PR #81, opened 2016-01-16
Remove unused #define's related to md32_common.h

frewsxcv:
`HASH_CBLOCK` and `DATA_ORDER_IS_BIG_ENDIAN` were previously required by
'md32_common.h' as a part of a generic 32-bit "collector", but that has
been removed, so these are no longer used.

### DIFF
--- a/crypto/digest/md32_common.h
+++ b/crypto/digest/md32_common.h
@@ -57,15 +57,6 @@ extern "C" {
 #endif
 
 
-#if !defined(DATA_ORDER_IS_BIG_ENDIAN)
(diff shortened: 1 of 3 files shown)


briansmith:
I merged this in 20f63fcd5dc1d73b42e369523000914d3c59a2d9 after updating the commit message slightly. Thanks for the patch.

Also, this change was just deleting code, so it doesn't matter. But, please state for the record that you agree to license your contributions to each file under the same terms given at the top of each file you changed. Note that each file is often licensed under different terms.

frewsxcv:
Sounds good!

Issue #71, opened 2016-01-05
Replace `HOST_c2l` in md32_common.h with byteswap function


briansmith:
`HOST_c2l` modifies its argument, which isn't obvious when reading the code. Also, all the users are of this form:

```
HOST_c2l(data, l);
X(n) = l;
```

This can be simplified to:
```
X(n) = HOST_c2l(data);
data += 4;
```.

**Comment by frewsxcv** [comment]
Can I take this?

**Comment by frewsxcv** [comment]
sort of relevant: https://boringssl-review.googlesource.com/#/c/6652/

**Comment by briansmith** [comment]
Yes, go ahead. For some reason I can't assign the bug to you. Check out the definition of BSWAP4 in crypto/modes/internal.h. That should probably be moved to crypto/internal.h and reused everywhere.

**Comment by frewsxcv** [comment]
> For some reason I can't assign the bug to you.

Unfortunately: "You can only create assignments for yourself, collaborators on personal projects, or members of your organization with read permissions on the repository."

https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/

**Comment by frewsxcv** [comment]
How does `BSWAP4` relate to `HOST_c2l`?

**Comment by briansmith** [comment]
> How does BSWAP4 relate to HOST_c2l?

Here is one of the definitions of `HOST_c2l`:

```
#define HOST_c2l(c, l)                       \
  (void)({                                   \
    uint32_t r = *((const uint32_t *)(c));   \
    __asm__("bswapl %0" : "=r"(r) : "0"(r)); \
    (c) += 4;                                \
    (l) = r;                                 \
  })
```

**Comment by frewsxcv** [comment]
@briansmith Did you mean for the second code block in your original post to be:

```c
X(n) = BSWAP4(data);
data += 4;
```

briansmith:
Something like that. Note that these functions need to be defined as no-ops for big-endian systems, and bswap on little-endian systems. At the time I wrote the initial ones, I didn't think about HOST_c2l being the same as bswap on little-endian platforms. The Rust functions are called `to_be` so renaming these functions `to_be_u32` & `from_be_u32` for 32-bit and `to_be_u64` & `from_be_u64` for 64-bit. (Note: The 64-bit versions are in sha512.c.)

This might be something best split up into multiple patches.

Also, instead of using `#define`, please define them as `static inline` functions.

briansmith:
[...]

PR #82, opened 2016-01-17
Replace usages of `HOST_c2l` macro with `from_be_u32` function

frewsxcv:
https://github.com/briansmith/ring/issues/71


frewsxcv:
@briansmith This is my first time using C since university, so please forgive me if I'm doing anything blatantly wrong :stuck_out_tongue:

frewsxcv:
I'm out right now and will address the issues you brought up later. Thanks for all the suggestions! Much appreciated :)

briansmith:
>> If we run into performance problems,
> Did you mean to finish this thought?

If we run into performance problems, we can explore using the built-in intrinsic functions, but I'd like `_byteswap_uint64`/`_byteswap_ulong` and the GCC/clang equivalents, but I'd like to avoid all inline assembly if at all possible, and I'd like to avoid `_byteswap_uint64`/`_byteswap_ulong` and friends if posssible.

frewsxcv:
Intentionally made a new commit so you could see the changes easier. I figure a `git squash` will happen later.

Thanks again for walking me through some of this, I really appreciate it!

Let me know how the latest changes look

frewsxcv:
Addressed your comments.

frewsxcv:
Ran crypto-bench on this: https://gist.github.com/frewsxcv/5acf2a5a6f4e112a9421

briansmith:
When working on stuff related to this, I realized another issue: It will eventually be better to have two versions of `from_be_u32`: One that takes a `const uint8_t *` like the current one does, and one that takes a `uint32_t` (not a pointer). The name `from_be_u32` makes more sense for the non-pointer variant. Because of that, I'm going to  rename `from_be_u32` to `from_be_u32_ptr`.

frewsxcv:
>Because of that, I'm going to rename from_be_u32 to from_be_u32_ptr.

SGTM :+1:

frewsxcv:
Merged in as https://github.com/briansmith/ring/commit/16af088fbc2099d3b03e04b8086278cdf03104e1

briansmith:
See the discussion in #71 about the follow-up to address potentially-undefined behavior and performance regressions.

briansmith:
Thanks again!

PR #87, opened 2016-01-21
Replace usages of `PULL64` macro with `from_be_u64`.

frewsxcv:


frewsxcv:
Addressed your comments

briansmith:
Thanks! Fixed in abb5cbebd969b6482a320d3c49eabd8c27ad8c70. I changed the commit message to be clearer (IMO) and squashed all the changes. Also note that the convention in BoringSSL & *ring* is to end the first line of the commit message with a period.

PR #88, opened 2016-01-21
Remove MSVC conditional checks for now-unsupported versions

frewsxcv:
According to @briansmith, Ring doesn't support any version lower than
1700, which makes these version checks obsolete.


frewsxcv:
I'll do that in a follow-up PR

briansmith:
Thanks again! Merged as 09f9f013e156d9e50917ca9fd31507a02304813e.

PR #89, opened 2016-01-22
Replace usages of `ROTR` macro with `rotate_right_u64`.

frewsxcv:

### DIFF
--- a/crypto/internal.h
+++ b/crypto/internal.h
@@ -519,12 +519,12 @@ static inline uint64_t from_be_u64(const uint64_t data) {
 #endif
 }
 
-/* rotate_right_u64 returns the |data| with its bits rotated |n| bits to the
+/* rotate_right_u64 returns the value |x| with its bits rotated |n| bits to the
    right */
-static inline uint64_t rotate_right_u64(uint64_t data, int n) {
+static inline uint64_t rotate_right_u64(uint64_t x, int n) {
     assert(n > 0);
     assert(n < 64);
-    return (data >> n) | (data << (64 - n));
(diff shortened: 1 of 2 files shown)


briansmith:
There's no performance difference on MSVC, so no intrinsics seem necessary, at least for that compiler.

frewsxcv:
Fairly big performance regression with GCC

https://gist.github.com/frewsxcv/4459dd9e7cdc78f5edb2

frewsxcv:
Actually, I think that's Clang, not GCC. For whatever reason, Apple aliases `gcc` to `clang`

briansmith:
Older verisons of clang haad some trouble optimizing things like this. What version of clang are you using?

frewsxcv:
Similar results for gcc-4.9: https://gist.github.com/frewsxcv/4459dd9e7cdc78f5edb2

> What version of clang are you using?

```
coreyf@frewbook-pro ~/D/c/ring (master) [130]> clang --version
Apple LLVM version 7.0.2 (clang-700.1.81)
Target: x86_64-apple-darwin15.2.0
Thread model: posix
```

frewsxcv:
Just realized I was running master without `OPENSSL_NO_ASM`. Here are number with `OPENSSL_NO_ASM=1` for all runs: https://gist.github.com/frewsxcv/502ef3886c71532d8a5d

briansmith:
Thanks! Committed as 13060ab6f1c52989804d7c9b3f1904c4cdf13ede. Note that I reduced the indentation of the function to two spaces (it was four spaces) and renamed `data` to `x`.

PR #90, opened 2016-01-22
Avoid 'data' name when referring to value (not a pointer)

frewsxcv:
As suggested:

https://github.com/briansmith/ring/pull/89#commitcomment-15602293


briansmith:
Thanks! Landed as 161165ff0dcf62129dee1cd763e64ee478c9e3a4.

Issue #69, opened 2016-01-05
Remove `#define asm __asm__` everywhere


briansmith:
crypto/bn/div.c
crypto/bn/asm/x86_64-gcc.c
crypto/modes/internal.h

briansmith:
Fixed in #69 with commit c882c2c3ec837f14267991122dc20e6aba5700bc.

PR #92, opened 2016-01-22
Remove '#define asm __asm__' everywhere

frewsxcv:
Closes https://github.com/briansmith/ring/issues/69

### DIFF
--- a/crypto/bn/asm/x86_64-gcc.c
+++ b/crypto/bn/asm/x86_64-gcc.c
@@ -70,30 +70,40 @@
 #define mul_add(r, a, word, carry)                                     \
   do {                                                                 \
     register BN_ULONG high, low;                                       \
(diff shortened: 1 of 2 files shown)


frewsxcv:
Addressed your comments

briansmith:
I merged the crypto/bn part in 2ed6c664fe255453965044f83871e1e0860a346e.

It was my mistake to ask you to reformat the patch. Once that happened, I couldn't review it in any useful way. So, I went back and un-reformatted it and then reformatted it according to how I think the BoringSSL people prefer to have it.

I dropped the changes to crypto/modes/internal.h because all the inline asm in that file is going to get deleted anyway in your other pull request.

briansmith:
Also, thanks!

frewsxcv:
:+1:

PR #97, opened 2016-01-24
A couple AES related cleanup commits

frewsxcv:


frewsxcv:
> Did you confuse crypto/aes/internal.h with include/openssl/aes.h? 

Yes, that was the case. Removed the commit.

briansmith:
Landed in cd06d58632d5af515c67e2d64733eb251dc4af0d. Thanks!

Issue #84, opened 2016-01-20
Replace crypto/cipher/e_chacha20poly1305.c with Rust code


briansmith:
This code has already diverged from the BoringSSL code quite a bit and there's no much left of it in C. We should just port it to C.

briansmith:
> We should just port it to Rust.

To do this, the following functions need to be written in Rust:

* evp_aead_chacha20_poly1305_init,
* evp_aead_chacha20_poly1305_seal,
* evp_aead_chacha20_poly1305_open,
* evp_aead_chacha20_poly1305_old_seal,
* evp_aead_chacha20_poly1305_old_open,

Then crypto/cipher/e_chacha20poly1305.c can be removed.

briansmith:
I should also mention that I don't mean for crypto/chacha/* to be replaced with Rust code. Soon (eventually, at least) all platforms will have assembly language code (or equivalent), so it isn't worth rewriting the primitive ChaCha20 and Poly1305 functions in Rust.

frewsxcv:
I'm currently working on this.

frewsxcv:
@briansmith I finished `evp_aead_chacha20_poly1305_init` this morning and the tests seem to pass. I implemented `evp_aead_chacha20_poly1305_seal`, but the tests don't seem to pass right now. Is it possible I could get a very hasty review of my progress so far, to make sure I'm heading in the right direction?

https://github.com/briansmith/ring/compare/master...frewsxcv:echacha20poly1305-to-rust?expand=1

Thanks!

briansmith:
See https://github.com/briansmith/ring/commit/3a1570c5cb9bd0fe381e80444bccf3d5fc00c0dd to see where your bug is.

Name the chacha20poly1305 module aead_chacha20poly1305, instead of using the e_ prefix.

I suggest that you don't spend too much effort trying to make the Rust code be organized like the C code. If there's an improvement to the organization, feel free to make it. It might help to avoid thinking about the `_old` code and optimize the organization of the code as though it doesn't exist.

The goal of this, of course, is to minimize "unsafe" code. So, the only places that pointers should be used is in calls to the functions `CRYPTO_chacha_20`, `CRYPTO_poly1305_init`, `CRYPTO_poly1305_finish`, and `CRYPTO_poly1305_update` functions, converting slices to pointers using `as_ptr` and `as_mut_ptr`.

[...]

PR #106, opened 2016-02-09
Create AEAD sub-module, separate modules for algorithms.

frewsxcv:


frewsxcv:
This structure will allow us to keep the private `Algorithm` fields private even with the specific `Algorithm` definitions living in different modules.

frewsxcv:
Will combine this with my patch for #84

PR #107, opened 2016-02-11
Replace crypto/cipher/e_chacha20poly1305.c with Rust code

frewsxcv:


Ms2ger:
Any particular reason to use `src/aead/aead.rs` rather than `src/aead/mod.rs`?

briansmith:
Yes. In my editor, I can type "Ctrl+,aead.rs" and navigate directly to the file. But, it doesn't work if I have a bunch of files named "mod.rs." IMO, the Rust toolchain should support this pattern directly, but until it does...

frewsxcv:
Excluding the comment about asserts, all other comments have been addressed

briansmith:
Here are the assertions for `init`:

*  `assert(ctx_buf != NULL);` - There's no NULL pointer possible in the new version of `init` because they are written in Rust (+ `ring::polyfill`) without using `unsafe`, so the assertion is not necessary.
*  `assert(((uintptr_t)ctx_buf) % ctx_struct_alignment == 0);` Again, since the new `init` function is written in the safe subset of Rust (+ `ring::polyfill`), the assertion is irrelevant as we assume Rust is happy with the alignment.
* `assert(ctx_buf_len >= ctx_struct_size);` Similar to the above, the point of this check is to make sure that the buffer size constants in the Rust code are at least as large as the data that will be stored in the buffer. Rust's slice bounds checking takes care of this in the new code.
* `assert(key != NULL);` Again, there is no NULL pointer possible because of Rust.

And here are the assertions for `open` and `seal`:

1. `assert(ctx_buf != NULL);`
2. `assert(((uintptr_t)ctx_buf) % ctx_struct_alignment == 0);`
3. `assert(out != NULL);`
4. `assert(out_len != NULL);`
5. `assert(nonce != NULL);`
6. `assert(in != NULL || in_len == 0);`
7. `assert(aead_check_in_len(in_len));`
8. `assert(aead_check_alias(in, in_len, out));`
9. `assert(ad != NULL || ad_len == 0);`

What's your assessment of which assertions are necessary?

frewsxcv:
1. `assert(ctx_buf != NULL);`
 * Unnecessary: There's no NULL pointer possible in the new version of `init`
2. `assert(((uintptr_t)ctx_buf) % ctx_struct_alignment == 0);`
 * Unnecessary: Assume Rust is satisfied with alignment in the new version of `init`
3. `assert(out != NULL);`
 * Probably unnecessary: The caller is responsible for passing in a slice, and if they pass in a null slice, it will fail upon reading or writing. Though, the assertion would cause a fail sooner
4. `assert(out_len != NULL);`
 * Probably unnecessary: Bounds checking should result in a fail, but a `assert!(in_out.len() > 0)` check would cause a fail sooner
5. `assert(nonce != NULL);`
 * See `3.`
6. `assert(in != NULL || in_len == 0);`
 * Unnecessary: This suggests that a NULL pointer was allowed previously, as long as the `len` was 0. This shouldn't be possible with the new system since `in` and `out` are combined in a single mutable slice, and because of `3.` above, `in_out` can't be NULL.
7. `assert(aead_check_in_len(in_len));`
 * Necessary. `CRYPTO_chacha_20` hasn't changed, so we should still ensure we don't pass it more data than it can handle
8. `assert(aead_check_alias(in, in_len, out));`
 * Not necessary. `in` and `out` are combined into `in_out`. Though, it might be worth checking `in_prefix < in_out.len()`
9. `assert(ad != NULL || ad_len == 0);`
 * Maybe unnecessary. On the one hand, we don't ever explicitly loop over `ad` or deref it so Rust will never encounter bounds issues with it, but maybe we should explicitly check? This is the assertion I'm most uncertain about

briansmith:
> assert(out != NULL);
> Probably unnecessary: The caller is responsible for passing in a slice, and if they pass in a null slice, it will fail upon reading or writing. Though, the assertion would cause a fail sooner

There are no NULL slices, just empty slices. An empty slice is a specific instance of the more general problem that in_out may not be large enough.

> assert(out_len != NULL);
> Probably unnecessary: Bounds checking should result in a fail, but a assert!(in_out.len() > 0) check would cause a fail sooner

Note that this is checking `|out_len| == NULL`, not that `*out_len == 0`. So, this is a standard "we're not using pointers so we don't need to check for NULL pointers" situation.

> assert(aead_check_in_len(in_len));
> Necessary. CRYPTO_chacha_20 hasn't changed, so we should still ensure we don't pass it more data than it can handle

I agree. It shouldn't be an assertion, though, just a normal check.

> assert(in != NULL || in_len == 0);
> Unnecessary: This suggests that a NULL pointer was allowed previously, as long as the len was 0. This shouldn't be possible with the new system since in and out are combined in a single mutable slice, and because of 3. above, in_out can't be NULL.

> assert(ad != NULL || ad_len == 0);
> Maybe unnecessary. On the one hand, we don't ever explicitly loop over ad or deref it so Rust will never encounter bounds issues with it, but maybe we should explicitly check? This is the assertion I'm most uncertain about

These two are the same kind of situation.

frewsxcv:
Addressed your comments. You might need to cancel my appveyor/travis builds as I accidentally force pushed sometime inbetween.

briansmith:
Sorry I didn't communicate this in my reviews before: Usually I try to point out an issue once and then expect that other instances of the issue will get fixed at the same time. But, also, generally it's worth keeping in mind:

* Use the strongest possible type: for example, `&[u8; X]` instead of `&[u8]` when the buffer is required to be `X` bytes.
* If you have the choice between doing checked arithmetic and doing an explicit check for overflow (`usize::max_value()`, etc.), always prefer the checked arithmetic.
* If you use `as` to cast, try to find a way to avoid using `as` as (haha) much as (haha) possible.
* Prefer to use `use` for bringing modules into scope, not items within modules.\
* Use qualified names `use aead; x = aead::something(...)` instead of prefixed names `use aead::{aead_x}; x = aead_something()`.

frewsxcv:
> Sorry I didn't communicate this in my reviews before: Usually I try to point out an issue once and then expect that other instances of the issue will get fixed at the same time.

No, that's reasonable to expect from someone, sorry about that.

frewsxcv:
I think I addressed all your comments.

briansmith:
The tests failed. Check out https://github.com/briansmith/ring/tree/chacha20_poly1305_fixes.

Notes:

1. `in_len as u64` was taking the address of the `in_len` function and converting it into an integer.
2. `slice[n..].as_ptr()` is safer than `slice.as_ptr().offset(n as isize)`. In particular, think about what happens when `n > isize::max_value()` or when `n > slice.len()`.

Also, please rename gcm.rs to aes_gcm.rs.

briansmith:
More notes:

* We need to add an assertion to `CRYPTO_chacha20`: `assert(out <= in);`
* We need to add similar assertions to the AES-GCM code.
* We need to add tests for non-zero prefix length. The fact that the tests didn't catch the earlier bugs is due to the lack of tests.
* The `in_out` + prefix + suffix pattern is way too error-prone for what we're trying to achieve. To add insult to injury, it doesn't even cover all the interesting cases, such as when the output is supposed to overwrite `ad` and/or `nonce`, which actually happens in TLS 1.2 for in-place decryption. After this PR lands, I am going to redo the whole thing.

frewsxcv:
> in_len as u64 was taking the address of the in_len function and converting it into an integer.

Wow, I didn't even know this was possible.

> slice[n..].as_ptr() is safer than slice.as_ptr().offset(n as isize). In particular, think about what happens when n > isize::max_value() or when n > slice.len().

Definitely a good call. It also results in one fewer use of `as`

> Also, please rename gcm.rs to aes_gcm.rs.

Maybe we should rename the other file to `chacha20_poly1305.rs` for consistency?

All your other comments sound good to me.

briansmith:
Great! Merged in 279ca8e1b0e0f4eef93c0c5ab458671b095ec41f and 5b38a81a67ef6048931abfc28c00d131d8aa99a1. Thanks for all the effort you put into this!

briansmith:
Also, I should mention that I made a few adjustments to the commits as shared on IRC, which were purely stylistic. I also updated the STYLE.md style guide accordingly.

PR #110, opened 2016-02-13
Utilize polyfill utility instead of manual fill.

frewsxcv:


briansmith:
Landed as b95fd0986ac46dc8c5caf1bd85a2526c8b507958. Thanks!

PR #111, opened 2016-02-15
A couple minor cleanup commits for pbkdf2.

frewsxcv:


frewsxcv:
Just saw the conversation in #rust last night about https://github.com/frewsxcv/ring/commit/d9ce775a0450e55b766c6c3bdfe3e038aa18a621. Feel free to ignore that commit and just cherry-pick the other one.

briansmith:
Thanks! I cherry-picked f46015fa1225c21358d2de03e901c8a66d2ab3ec as  7912df51504f67279a1892a1d2a3be68f6b99529.

Regarding the other commit with `static`, I chatted with people on the #rust channel and they seemed to agree amongst themselves that that is a no-op. So, maybe it is better to just leave the code the way it is? I.e. maybe I was wrong in suggesting you to make a similar change in your other PR?

frewsxcv:
> maybe I was wrong in suggesting you to make a similar change in your other PR?

Previously, it was a `let` binding because

PR #116, opened 2016-02-21
Fix invalid syntax for inline code within doc comment.

frewsxcv:


briansmith:
Thanks! Landed as fcefa13b51cdb54ccf2882a45286c79e3b6ac338.

PR #117, opened 2016-02-22
Remove unnecessary lower bound when slicing in digest.

frewsxcv:
At this point in the function, `num_pending` should always be `0`.

### DIFF
--- a/src/digest.rs
+++ b/src/digest.rs
@@ -133,8 +133,7 @@ impl Context {
         }
         if num_to_save_for_later > 0 {
             polyfill::slice::fill_from_slice(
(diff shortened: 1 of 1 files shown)


briansmith:
I adjusted the commit message slightly and landed this as 908161a8246691220e6e48b57ad77146f880de77. Thanks!

briansmith:
I also added some notes about this to the style guide and made some similar adjustments to other parts of the code: See https://github.com/briansmith/ring/blob/master/STYLE.md#arrays-and-slices and commit ba126545ca2652682d1af60228d892e5d9b0f26a.

PR #122, opened 2016-02-27
Utilize 'unit struct' where appropriate.

frewsxcv:

### DIFF
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -200,17 +200,13 @@ ecdsa!(ECDSA_P384_SHA512, "P-384 (secp384r1)", ecc::EC_GROUP_P384, "SHA-512",
 
 
 /// EdDSA signatures.
-struct EdDSA {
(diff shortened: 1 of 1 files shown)


src/signature.rs
```diff
@@ -200,17 +200,13 @@ ecdsa!(ECDSA_P384_SHA512, "P-384 (secp384r1)", ecc::EC_GROUP_P384, "SHA-512",
 
 
 /// EdDSA signatures.
-struct EdDSA {
-    _unused: u8, // XXX: Stable Rust doesn't allow empty structs.
-}
+struct EdDSA;
```
frewsxcv: Alternatively, this could be written:

```rust
struct EdDSA { }
```

...which was not stable in 1.0 (https://github.com/rust-lang/rfcs/pull/218).

briansmith:
Thanks! I was pretty sure that this didn't work before but I guess I was confiused. Committed as 922406200805d1e386cb71c9cea8d093f3668e5f.

PR #124, opened 2016-02-28
Componentization of the 'signature' module.

frewsxcv:
Similar to 279ca8e1b0e0f4eef93c0c5ab458671b095ec41f.


src/signature/signature.rs
```diff
@@ -0,0 +1,123 @@
...
+//! Currently this module does not support digesting the message to be signed
+//! separately from the public key operation, as it is currently being
+//! optimized for Ed25519 and for the implementation of protocols that do not
+//! requiring signing large messages. An interface for efficiently supporting
+//! larger messages will be added later. Similarly, the signing interface is
+//! not available yet.
+
+mod ecdsa;
+mod ed_dsa;
+mod rsa_pkcs1;
```
frewsxcv: One option here is to add `#[cfg(not(feature = "no_heap"))]` to `rsa_pkcs1` and `ecdsa` instead of having a bunch of these `cfg` feature conditionals inside their respective modules.

src/signature/ecdsa.rs
```diff
@@ -0,0 +1,116 @@
...
+// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+// OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+#![allow(unsafe_code)]
+
+use {c, bssl};
+#[cfg(not(feature = "no_heap"))] use {digest, ecc};
+use input::Input;
+use super::{VerificationAlgorithm, VerificationAlgorithmImpl};
```
frewsxcv: Not sure if you'd prefer use `use signature` and namespace everything below.

frewsxcv:
@briansmith I made these changes because I feel `signature.rs` is already pretty cluttered with implementation specific details, and adding `RSA_PSS` is not going to make it any cleaner. I'm pretty satisfied with how minimal `signature.rs` after these changes; it's easy to see what the core functionality/interface of the module is.

briansmith:
"eddsa" should be kept as "eddsa" and not split into "ed_dsa". But, actually, I think that the EdDSA code can just as well stay in signature.rs for now. Similarly, unless/until we move a more substantial portion of the ECDSA logic to Rust, we might as well keep the ECDSA part as-is.

I don't think that RSA-PKCS#1 and RSA-PSS should separate from each other. They are exactly the same except the padding function is different. In the long run, a lot of the C code in rsa.c and rsa_impl.c will likely get replaced with Rust code, and most of that Rust code will be shared for PSS and PKCS#1 padding.

Note that RSA-PSS signatures and RSA-OEAP encryption use the same function: MGF1. Similarly, PKCS#1 encryption/decryption and PKCS#1 signatures use the same PKCS#1 1.5 padding. Thus, I don't think it makes sense to put the code for MGF1 in a submodule of `ring::signature`. Maybe a `ring::rsa` module that contains all the padding code (and eventually other shared code) makes more sense.

Like the previous PR of this nature for ChaCha20-Poly1305, I think that this commit should be part of a series of commits that make a substantial change (e.g. adding the PSS padding functions).

frewsxcv:
Alright, I'll just keep everything in `signature.rs` for now. Sounds good.

PR #125, opened 2016-02-28
Refactor aead

briansmith:
@frewsxcv Do you have time to look at this and verify that it makes sense? Basically, it's an attempt to generalize the work you did to Rustify the ChaCha20-Poly1305 code to also Rustify more of the AES-GCM code and then simplify the code further by factoring out common parts.

### DIFF
--- a/crypto/cipher/e_aes.c
+++ b/crypto/cipher/e_aes.c
@@ -235,27 +235,24 @@ int evp_aead_aes_gcm_seal(const void *ctx_buf, uint8_t *in_out,
   return 1;
 }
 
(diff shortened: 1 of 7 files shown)


briansmith:
Thanks, @Ms2ger. I had thought I'd tried that and had it fail, but it does indeed work. I updated the PR accordingly.

frewsxcv:
Everything here looks good to me, nice cleanup!

briansmith:
I found some minor nits and I corrected them and landed everything. Thanks for taking a look!

Issue #58, opened 2015-12-30
Replace use of crypto/rand with more direct use of OS RNG


briansmith:
In particular, consider `rand::os::OsRng`. If `rand::os::OsRng` is insufficient in some way then try to fix it. `rand::os::OsRng` is preferable because other Rust code is likely using it, and we can avoid duplicate code. Also, crypto/rand is pretty complicated and has a weird interface. To the extent that that complication is useful, we should implement it as a layer over `rand::os::OsRng`.

briansmith:
One problem is that `OsRng` implements `Rng` and there are a LOT of methods, in particular methods that have floating-point parameters and lots of methods that we'd never, ever use. It's not clear how the presence of the floating point types in the trait interface affects platforms where we want to avoid floating point stuff. It's not clear how good a job does in removing the extraneous methods. Perhaps we should propose a new, simpler, interface, that is just `rand::os::OsRng::fill_bytes`.

briansmith:
See also https://boringssl.googlesource.com/boringssl/+/dca63cfa754503d75c615f5c134d9334f54c132b.

briansmith:
See https://github.com/rust-lang/rust/issues/27703#issuecomment-178143920

briansmith:
I've started a new implementation of `ring::rand` where more of the logic is in Rust.

It is now possible to explicitly manage when the file handle for `/dev/urandom` is open by constructing/destroying `rand::SystemRandom` instances appropriately.

Most applications should only have a single implementation of rand::SystemRandom instantiated to avoid the waste of having multiple file descriptors to /dev/urandom open and to avoid the overhead of opening/closing /dev/urandom. That was effectively how BoringSSL works and how *ring* worked before these changes. However, I didn't implement that because, generally, we shouldn't be using /dev/urandom anyway. Instead, we should be using RDRAND (or equivalents on ARM and other platforms) + PRP (e.g. ChaCha20), or `getrandom` syscall (or the equivalent on other platforms), for performance reasons. The `/dev/urandom` implementation should only exist as a least-common-denominator fallback.

I also dropped the RDRAND+ChaCha20 implementation.

So, we should:
* [ ] Make `SystemRandom` use `getrandom` on Linux whenever possible.
* [ ] Add support for the equivalent of the `getrandom` on Mac OS X, BSD, etc. to SystemRandom (It is already done on Windows).
* [ ] Provide some way to avoid the fallback to /dev/urandom, for cases where we know that one of the better options will always be available.

[...]

PR #137, opened 2016-03-04
Use RDRAND intrinsics, enabling RDRAND on 32-bit x86 and Windows.

briansmith:


briansmith:
Closing this, as I've refactoring `ring::rand` significantly and this will need to be redone. I will update #58 with the current status of things.

PR #140, opened 2016-03-07
Use `<[T]>::clone_from_slice()`.

### LINKED ISSUE #136: Replace `ring::polyfill::fill_from_slice` with use of `<[T]>::clone_from_slice()`
See http://blog.rust-lang.org/2016/03/02/Rust-1.7.html. `<[T]>::clone_from_slice()` was stabilized in Rust 1.7, which is now the minimum supported version in *ring*.

Note that `clone_from_slice` may not actually get optimized as well as `fill_from_slice` does. If that is found to be the case, then we'll have to wait for https://github.com/rust-lang/rust/issues/31755 to be resolved.

pyfisch:
Replace all occurences of `ring::polyfill::fill_from_slice`.

Closes #136

### DIFF
--- a/src/aead/chacha20_poly1305.rs
+++ b/src/aead/chacha20_poly1305.rs
@@ -108,7 +108,7 @@ fn chacha20_poly1305_update_old(state: &mut [u8; POLY1305_STATE_LEN],
 
 /// Copies |key| into |ctx_buf|.
 pub fn init(ctx_buf: &mut [u8], key: &[u8]) -> Result<(), ()> {
-    polyfill::slice::fill_from_slice(&mut ctx_buf[..key.len()], key);
+    ctx_buf[..key.len()].clone_from_slice(key);
     Ok(())
 }
 
--- a/src/digest.rs
+++ b/src/digest.rs
@@ -93,10 +93,8 @@ impl Context {
     /// C analog: `EVP_DigestUpdate`
     pub fn update(&mut self, data: &[u8]) {
         if data.len() < self.algorithm.block_len - self.num_pending {
-            polyfill::slice::fill_from_slice(
-                &mut self.pending[self.num_pending..
-                                  (self.num_pending + data.len())],
-                data);
+            self.pending[self.num_pending..
+                (self.num_pending + data.len())].clone_from_slice(data);
             self.num_pending += data.len();
             return;
         }
@@ -104,9 +102,8 @@ impl Context {
         let mut remaining = data;
         if self.num_pending > 0 {
             let to_copy = self.algorithm.block_len - self.num_pending;
-            polyfill::slice::fill_from_slice(
-                &mut self.pending[self.num_pending..self.algorithm.block_len],
-                &data[..to_copy]);
+            self.pending[self.num_pending..self.algorithm.block_len]
+                .clone_from_slice(&data[..to_copy]);
 
             unsafe {
                 (self.algorithm.block_data_order)(self.state.as_mut_ptr(),
@@ -132,8 +129,7 @@ impl Context {
                                           .unwrap();
         }
         if num_to_save_for_later > 0 {
-            polyfill::slice::fill_from_slice(
-                &mut self.pending[..num_to_save_for_later],
+            self.pending[..num_to_save_for_later].clone_from_slice(
                 &remaining[(remaining.len() - num_to_save_for_later)..]);
             self.num_pending = num_to_save_for_later;
         }
--- a/src/polyfill.rs
+++ b/src/polyfill.rs
@@ -34,17 +34,6 @@ pub mod slice {
         }
     }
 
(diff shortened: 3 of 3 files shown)


briansmith:
What was the difference in performance?

pyfisch:
Not yet testet. Are there benchmarks for `ring`?

briansmith:
https://github.com/briansmith/crypto-bench.

You'll need to use a .cargo/config that contains a path override so that your local copy of *ring* is used instead of the one from GitHub when doing the "after" comparisons.

pyfisch:
The benchmarks show that `<[T]>::clone_from_slice` is equally fast or even faster than `ring::polyfill::fill_from_slice`. The benchmarks are available at https://gist.github.com/pyfisch/3e557d4ff53c311c655b

briansmith:
The thing is, *all* your results are faster, even ones that shouldn't be affected by this change like the AES-GCM changes. When I run the benchmarks with your changes on my machine, I don't get results consistent with that. We need to check out the generated assembly code and see what's going on.

briansmith:
I used the following code on play.rust-lang.org to verify that the old code and the new code seem to optimize to the same thing:

```
mod digest {
  pub const MAX_BLOCK_LEN: usize = 1024 / 8;
  pub const MAX_OUTPUT_LEN: usize = 512 / 8;
  pub const MAX_CHAINING_LEN: usize = MAX_OUTPUT_LEN;

  #[inline(always)]
  fn widen_u64(x: usize) -> u64 { x as u64 }
  
  // XXX: endian-specific.
  // XXX: Replace with `const fn` when `const fn` is stable:
  // https://github.com/rust-lang/rust/issues/24111
  macro_rules! u32x2 {
    ( $first:expr, $second:expr ) =>
    ( ((($second as u64) << 32) | ($first as u64)) )
  }

  pub struct Context {
    // We use u64 to try to ensure 64-bit alignment/padding.
    state: [u64; MAX_CHAINING_LEN / 8],

    block_len: usize,

    // Note that SHA-512 has a 128-bit input bit counter, but this
    // implementation only supports up to 2^64-1 input bits for all algorithms,
    // so a 64-bit counter is more than sufficient.
    completed_data_blocks: u64,

    // TODO: More explicitly force 64-bit alignment for |pending|.
    pending: [u8; MAX_BLOCK_LEN],
    num_pending: usize,
  }

  impl Context {
    pub fn new() -> Context {
        Context {
            state: [
              u32x2!(0x67452301u32, 0xefcdab89u32),
              u32x2!(0x98badcfeu32, 0x10325476u32),
              u32x2!(0xc3d2e1f0u32, 0u32),
              0, 0, 0, 0, 0,
            ],
            block_len: 512 / 8,
            completed_data_blocks: 0,
            pending: [0u8; MAX_BLOCK_LEN],
            num_pending: 0,
        }
    }
    
    /// Updates the digest with all the data in `data`. `update` may be called
    /// zero or more times until `finish` is called. It must not be called
    /// after `finish` has been called.
    ///
    /// C analog: `EVP_DigestUpdate`
    pub fn update(&mut self, data: &[u8]) {
        if data.len() < self.block_len - self.num_pending {
            println!("copying to pending");
            self.pending[self.num_pending..
                         (self.num_pending + data.len())].clone_from_slice(data);
            self.num_pending += data.len();
            return;
        }

        let mut remaining = data;
        if self.num_pending > 0 {
            let to_copy = self.block_len - self.num_pending;
            println!("something else");
            &mut self.pending[self.num_pending..self.block_len].clone_from_slice(&data[..to_copy]);
            self.completed_data_blocks =
                self.completed_data_blocks.checked_add(1).unwrap();

            remaining = &remaining[to_copy..];
            self.num_pending = 0;
        }

        let num_blocks = remaining.len() / self.block_len;
        let num_to_save_for_later = remaining.len() % self.block_len;
        if num_blocks > 0 {
            self.completed_data_blocks =
                self.completed_data_blocks.checked_add(widen_u64(num_blocks))
                                          .unwrap();
        }
        if num_to_save_for_later > 0 {
            println!("saving for later");
            self.pending[..num_to_save_for_later].clone_from_slice(&remaining[(remaining.len() - num_to_save_for_later)..]);
            self.num_pending = num_to_save_for_later;
        }
    }
  }
  
}

fn main() {
 let mut x = digest::Context::new();
 x.update(&[1,2,3,4,5,6,7,8,9,0,1,2,3,4,5,6]);
 x.update(&[1,2,3,4,5,6,7,8,9,0,1,2,3,4,5,6]);
 x.update(&[1,2,3,4,5,6,7,8,9,0,1,2,3,4,5,6]);
}
```

briansmith:
Thanks! Landed in 7cdaeeb.

Ms2ger:
Actually landed in 43af354271857d9c034862de94e95a8b4f8f28df.

PR #142, opened 2016-03-08
Add AEAD tests that test all prefix length from 0-4096 bytes.

### LINKED ISSUE #127: Expand crypto primitive tests to more diverse inputs
1. Expand the `in_prefix_len` part of the AEAD tests to test a wide variety of prefixes, e.g. 0-4096 bytes.
2. Expand the tests of various primitives to test every input length 0-16K or so.

This might make the tests slower. We may need to add a “slow_tests” feature to enable these slow tests.

pyfisch:
The new tests are optional. To run them use `cargo test -- --ignored`.

Partially fixes #127.

### DIFF
--- a/src/aead/aead.rs
+++ b/src/aead/aead.rs
@@ -295,9 +295,9 @@ mod tests {
                   "crypto/cipher/test/aes_128_gcm_tests.txt");
     }
 
+    #[cfg(not(debug_assertions))]
     #[test]
-    #[ignore]
-    pub fn test_aes_gcm_128_slow() {
+    pub fn test_aes_gcm_128_in_prefix_len_extra() {
         test_aead_slow(&aead::AES_128_GCM,
                   "crypto/cipher/test/aes_128_gcm_tests.txt");
     }
@@ -308,9 +308,9 @@ mod tests {
                   "crypto/cipher/test/aes_256_gcm_tests.txt");
     }
 
+    #[cfg(not(debug_assertions))]
     #[test]
-    #[ignore]
-    pub fn test_aes_gcm_256_slow() {
+    pub fn test_aes_gcm_256_in_prefix_len_extra() {
         test_aead_slow(&aead::AES_256_GCM,
                   "crypto/cipher/test/aes_256_gcm_tests.txt");
     }
@@ -321,9 +321,9 @@ mod tests {
                   "crypto/cipher/test/chacha20_poly1305_tests.txt");
     }
 
+    #[cfg(not(debug_assertions))]
     #[test]
-    #[ignore]
-    pub fn test_chacha20_poly1305_slow() {
+    pub fn test_chacha20_poly1305_in_prefix_len_extra() {
         test_aead_slow(&aead::CHACHA20_POLY1305,
                   "crypto/cipher/test/chacha20_poly1305_tests.txt");
     }
@@ -334,9 +334,9 @@ mod tests {
                   "crypto/cipher/test/chacha20_poly1305_old_tests.txt");
     }
 
+    #[cfg(not(debug_assertions))]
     #[test]
-    #[ignore]
-    pub fn test_chacha20_poly1305_old_slow() {
+    pub fn test_chacha20_poly1305_old_in_prefix_len_extra() {
         test_aead_slow(&aead::CHACHA20_POLY1305_OLD,
                   "crypto/cipher/test/chacha20_poly1305_old_tests.txt");
     }
@@ -459,6 +459,7 @@ mod tests {
         });
     }
 
+    #[cfg(not(debug_assertions))]
     fn test_aead_slow(aead_alg: &'static aead::Algorithm, file_path: &str) {
         file_test::run(file_path, |section, test_case| {
             const PREFIX: [u8; 4096] = [123; 4096];
--- a/src/aead/aead.rs
+++ b/src/aead/aead.rs
@@ -286,7 +286,6 @@ fn check_per_nonce_max_bytes(in_out_len: usize) -> Result<(), ()> {
 
 #[cfg(test)]
 mod tests {
-    use std::string::String;
     use super::super::{aead, file_test};
     use std::vec::Vec;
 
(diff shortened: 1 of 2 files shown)


pyfisch:
Is this a "real" build failure? For me it does not look like one.

Are there any other problems remaining in this pull request?

(I ask because this was mentioned in #156)

src/aead/aead.rs
```diff
@@ -396,38 +425,29 @@ mod tests {
...
+            let plaintext = test_case.consume_bytes("IN");
+            let ad = test_case.consume_bytes("AD");
+            let mut ct = test_case.consume_bytes("CT");
+            let tag = test_case.consume_bytes("TAG");
+            let error = test_case.consume_optional_string("FAILS");
+
+            ct.extend(tag);
+
+            for in_prefix_len in 0..4096 {
+                let in_prefix = &vec![123; in_prefix_len][..];
```
briansmith: Perhaps the test would run faster if, instead of using `vec!`, we did something like `static PREFIX: [u8; 4096] = [123; 4096]` (outside the loop) and then `let in_prefix = PREFIX[..in_prefix_len]` or similar?

src/aead/aead.rs
```diff
@@ -613,4 +633,46 @@ mod tests {
...
+                       key_bytes: &[u8],
+                       nonce: &[u8],
+                       plaintext: &Vec<u8>,
+                       ad: &[u8],
+                       ct: &[u8],
+                       error: &Option<String>,
+                       in_prefix: &[u8]) {
+        let max_overhead_len = aead_alg.max_overhead_len();
+        let mut s_in_out = plaintext.clone();
+        for _ in 0..max_overhead_len {
```
briansmith: Similarly, maybe we need to optimize this so the tests run faster. Maybe the loop can be replaced with `s_in_out.resize(plantext.len() + max_overhead_len)` or similar?

src/aead/aead.rs
```diff
@@ -613,4 +633,46 @@ mod tests {
...
+            s_in_out.push(0);
+        }
+        let s_key = aead::SealingKey::new(aead_alg, key_bytes).unwrap();
+        let s_result = aead::seal_in_place(&s_key, nonce,
+                                           &mut s_in_out[..],
+                                           max_overhead_len, &ad);
```
briansmith: All of the steps above only need to be done once per test vector in `test_aead_slow`, not once per `in_prefix_len`, right? Refactoring the test to pull out the invariant steps should speed this up a lot.

src/aead/aead.rs
```diff
@@ -613,4 +633,46 @@ mod tests {
...
+
+        let mut o_in_out = Vec::from(in_prefix);
+        o_in_out.extend_from_slice(&ct[..]);
```
briansmith: Here's an idea:

1. Construct a Vec with `4096 + ct.len()` capacity.
2. For each `in_prefix_len`, reset the vec to zero length, then append `in_prefix_len` bytes to it, then append `ct` to it.

This way, only one memory allocation will be done per test vector, instead of one memory allocation per in_prefix_len per test vector.

src/aead/aead.rs
```diff
@@ -613,4 +633,46 @@ mod tests {
...
+        let o_key = aead::OpeningKey::new(aead_alg, key_bytes).unwrap();
```
briansmith: The OpeningKey only needs to be constructed once per test vector, not once per test vector * in_prefix_len.

src/aead/aead.rs
pyfisch: Should this really be a `static`? It can also be a `const` and Rust documentation favors `const` over `static` since it allows more optimizations.

src/aead/aead.rs
pyfisch: Fixed.

src/aead/aead.rs
pyfisch: Fixed.

src/aead/aead.rs
pyfisch: Done.

pyfisch:
Adressed all suggestions by @briansmith and rebased the pull request.

pyfisch:
The clang builds fail: `mk/travis.sh: line 58: clang-3.8: command not found`.

briansmith:
First, thanks for making these tests.

@pyfisch The clang-3.8 issue is due to failure to download clang, I believe. Don't worry about it.

I ran these tests using `cargo test --release -- --ignored` and actually they are NOT that slow; they take less than 2 seconds on my laptop, even when it isn't in high-performance mode, and even on Windows where the assembly-language optimizations are not enabled! They are only slow in the debug builds. Therefore, I think we should just always run them.

Note in particular that we don't do `cargo test --ignored` in Appveyor and Travis CI, so these tests don't get run during CI, which seems unfortunate. If we just always enable them then it will work fine.

I am looking into making the I-U-F tests faster now.

briansmith:
As of 117cf35fb1d8e72eee2abf72defc8aeda53b25ab the i-u-f tests are no longer super-slow, at least in release builds. So, I think we should drop the i-u-f part of the patch and I think that these tests should follow that same "only run when not in debug mode" pattern as the i-u-f tests. I hate that pattern but it is less error-prone than --ignored as everybody should know to run the tests in `--release` before deploying, but it is easy to overlook the need to run the tests with `--ignored`

src/aead/aead.rs
```diff
@@ -296,23 +297,51 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
+    pub fn test_aes_gcm_128_slow() {
```
briansmith: Instead of `_slow`, let's use a suffix like `_in_prefix_len_extra` or another self-documenting name.

src/aead/aead.rs
```diff
@@ -296,23 +297,51 @@ mod tests {
...
+        test_aead_slow(&aead::AES_128_GCM,
```
briansmith: Similarly, use a more descriptive suffix for this.

("_slow" may be descriptive, but it doesn't hint at the reason we wrote these tests.)

briansmith:
The SHA-{256,384,512} tests should be faster still as of 314747c2cff8f076d4c9547f6852a89357df3341 if you have a Haswell+ CPU.

pyfisch:
I've added more descriptive test names. Now all tests are run in release mode.

Should I squash all commits into a single one?

briansmith:
First, I'm sorry this took so long for me to review. I didn't notice the email I got when you updated the PR.

Secondly, thanks for cleaning up my wastefulness with the existing tests for this.

Thirdly, I landed this in 43daba8c8740f4a8d00dfb93c369d418d9b9385f, but with significant simplifications. In particular, the new way of choosing whether to test the more comprehensive set of prefixes or the old set of prefixes allows us to avoid duplicating some cases when your more comprehensive set is chosen. Also, less code is duplicated this way. Please let me know if you think this is problematic.

In the future, please include a statement along the lines of "I agree to license my contributions to each file under the terms given at the top of each file I changed." in future commit messages. This way, it is clear to everybody that everybody understands and agrees to the licensing stuff. (I hate this but I guess I should be careful about it.)

Sorry again about the delays, and thanks again for doing all this work!

Issue #99, opened 2016-01-25
Simplify RSA signing code


briansmith:
Note that some of these refactorings are only valid because *ring* doesn't use the `RSA` struct for verifying signatures, only for signing and decryption.

* [x] Drop support for signing or decrypting with a key that only has `n` and `d` or only `n`, `d`, and `e`. Instead, we'll require that any such keys be converted to the full PKCS#8 form before they can be used in *ring*.
* [x] Remove `rsa_get_public_exp` in blinding.c. Instead, just fail if `rsa->e` is `NULL`.
* [x] Remove all uses of `BN_with_flags(..., BN_FLG_CONSTTIME)` in the RSA code. Instead, set the `BN_FLG_CONSTTIME` flag on `rsa->n` and other members as required, at the time they are assigned.
* [x] At the time an `RSA` is constructed (using `RSA_generate` or reading in a PKCS#8 file), compute `rsa->mont_n`. Then, remove the lazy initialization of that member. In particular, do this without regard to the `RSA_FLAG_CACHE_PUBLIC` flag, which should be removed. Note that at these times of construction, there is no need for locking, so there's no need to use `BN_MONT_CTX_set_locked`; `BN_MONT_CTX_set` can be used instead. Hopefully this will allow us to remove `BN_MONT_CTX_set_locked` completely.
* [x] Replace `BN_mod`/`BN_div` operations with Montgomery reductions. (in progress by @briansmith)
* [x] Drop the attempt to recover from CRT faults when verifying the just-signed value (pending BoringSSL review).
* [ ] Improve/simplify the RSA blinding code: #144 
* [x] Require p > q and simplify the code accordingly. See https://crypto.stackexchange.com/a/22661. (In progress by @briansmith)
* [ ] Add tests for the corner cases mentioned in https://crypto.stackexchange.com/a/22661.

All of these should be done in separate commits.

frewsxcv:
>  Remove all uses of BN_with_flags(..., BN_FLG_CONSTTIME) in the RSA code. Instead, set the BN_FLG_CONSTTIME flag on rsa->n and other members as required, at the time they are assigned.

For reference:

```c
void BN_with_flags(BIGNUM *out, const BIGNUM *in, int flags) {
  memcpy(out, in, sizeof(BIGNUM));
  out->flags &= ~BN_FLG_MALLOCED;
  out->flags |= BN_FLG_STATIC_DATA | flags;
}
```

Correct me if I'm wrong, but this function:

[...]

PR #154, opened 2016-03-19
Replace usages of `BN_with_flags` with flag assignment.

frewsxcv:
Instead of using `BN_with_flags` (which relies on an extra allocated
`BIGNUM`), we'll make sure that all `BIGNUM`s we're working with have
the `BN_FLG_CONSTTIME` flag assigned at the time of creation.

Part of https://github.com/briansmith/ring/issues/99.

### DIFF
--- a/crypto/bn/gcd.c
+++ b/crypto/bn/gcd.c
@@ -417,8 +417,6 @@ BIGNUM *BN_mod_inverse(BIGNUM *out, const BIGNUM *a, const BIGNUM *n,
 BIGNUM *BN_mod_inverse_no_branch(BIGNUM *out, int *out_no_inverse,
                                  const BIGNUM *a, const BIGNUM *n, BN_CTX *ctx) {
   BIGNUM *A, *B, *X, *Y, *M, *D, *T, *R = NULL;
-  BIGNUM local_A, local_B;
-  BIGNUM *pA, *pB;
   BIGNUM *ret = NULL;
   int sign;
 
@@ -452,12 +450,8 @@ BIGNUM *BN_mod_inverse_no_branch(BIGNUM *out, int *out_no_inverse,
   A->neg = 0;
 
   if (B->neg || (BN_ucmp(B, A) >= 0)) {
-    /* Turn BN_FLG_CONSTTIME flag on, so that when BN_div is invoked,
-     * BN_div_no_branch will be called eventually.
-     */
-    pB = &local_B;
-    BN_with_flags(pB, B, BN_FLG_CONSTTIME);
-    if (!BN_nnmod(B, pB, A, ctx)) {
+    BN_set_flags(B, BN_FLG_CONSTTIME);
+    if (!BN_nnmod(B, B, A, ctx)) {
       goto err;
     }
   }
@@ -478,14 +472,10 @@ BIGNUM *BN_mod_inverse_no_branch(BIGNUM *out, int *out_no_inverse,
      *      sign*Y*a  ==  A   (mod |n|)
      */
 
-    /* Turn BN_FLG_CONSTTIME flag on, so that when BN_div is invoked,
-     * BN_div_no_branch will be called eventually.
-     */
-    pA = &local_A;
-    BN_with_flags(pA, A, BN_FLG_CONSTTIME);
+    BN_set_flags(A, BN_FLG_CONSTTIME);
 
     /* (D, M) := (A/B, A%B) ... */
-    if (!BN_div(D, M, pA, B, ctx)) {
+    if (!BN_div(D, M, A, B, ctx)) {
       goto err;
     }
 
--- a/crypto/rsa/rsa.c
+++ b/crypto/rsa/rsa.c
@@ -120,24 +120,34 @@ int rsa_new_end(RSA *rsa, BN_CTX *ctx) {
   /* Make sure BN_mod_inverse in Montgomery intialization uses the
    * BN_FLG_CONSTTIME flag. */
 
+  assert(BN_get_flags(rsa->d, BN_FLG_CONSTTIME));
+
+  if (rsa->dmp1) {
+    assert(BN_get_flags(rsa->dmp1, BN_FLG_CONSTTIME));
+  }
+
+  if (rsa->dmq1) {
+    assert(BN_get_flags(rsa->dmq1, BN_FLG_CONSTTIME));
+  }
+
+  if (rsa->iqmp) {
+    assert(BN_get_flags(rsa->iqmp, BN_FLG_CONSTTIME));
+  }
+
   if (rsa->p) {
-    BIGNUM local_p;
-    BN_init(&local_p);
-    BN_with_flags(&local_p, rsa->p, BN_FLG_CONSTTIME);
+    assert(BN_get_flags(rsa->p, BN_FLG_CONSTTIME));
     rsa->mont_p = BN_MONT_CTX_new();
     if (rsa->mont_p == NULL ||
-        !BN_MONT_CTX_set(rsa->mont_p, &local_p, ctx)) {
+        !BN_MONT_CTX_set(rsa->mont_p, rsa->p, ctx)) {
       return 0;
     }
   }
 
   if (rsa->q) {
-    BIGNUM local_q;
-    BN_init(&local_q);
-    BN_with_flags(&local_q, rsa->q, BN_FLG_CONSTTIME);
+    assert(BN_get_flags(rsa->q, BN_FLG_CONSTTIME));
     rsa->mont_q = BN_MONT_CTX_new();
     if (rsa->mont_q == NULL ||
-        !BN_MONT_CTX_set(rsa->mont_q, &local_q, ctx)) {
+        !BN_MONT_CTX_set(rsa->mont_q, rsa->q, ctx)) {
       return 0;
     }
   }
--- a/crypto/rsa/rsa_asn1.c
+++ b/crypto/rsa/rsa_asn1.c
@@ -67,12 +67,13 @@
 #include "internal.h"
 
 
-static int parse_integer(CBS *cbs, BIGNUM **out) {
+static int parse_integer_consttime(CBS *cbs, BIGNUM **out) {
   assert(*out == NULL);
   *out = BN_new();
   if (*out == NULL) {
     return 0;
   }
+  BN_set_flags(*out, BN_FLG_CONSTTIME);
   return BN_parse_asn1_unsigned(cbs, *out);
 }
 
@@ -172,14 +173,14 @@ RSA *RSA_parse_private_key(CBS *cbs) {
     goto err;
   }
 
-  if (!parse_integer(&child, &ret->n) ||
-      !parse_integer(&child, &ret->e) ||
-      !parse_integer(&child, &ret->d) ||
-      !parse_integer(&child, &ret->p) ||
-      !parse_integer(&child, &ret->q) ||
-      !parse_integer(&child, &ret->dmp1) ||
-      !parse_integer(&child, &ret->dmq1) ||
-      !parse_integer(&child, &ret->iqmp)) {
+  if (!parse_integer_consttime(&child, &ret->n) ||
+      !parse_integer_consttime(&child, &ret->e) ||
+      !parse_integer_consttime(&child, &ret->d) ||
+      !parse_integer_consttime(&child, &ret->p) ||
+      !parse_integer_consttime(&child, &ret->q) ||
+      !parse_integer_consttime(&child, &ret->dmp1) ||
+      !parse_integer_consttime(&child, &ret->dmq1) ||
+      !parse_integer_consttime(&child, &ret->iqmp)) {
     goto err;
   }
 
--- a/crypto/rsa/rsa_asn1.c
+++ b/crypto/rsa/rsa_asn1.c
@@ -67,13 +67,13 @@
 #include "internal.h"
 
 
-static int parse_integer_consttime(CBS *cbs, BIGNUM **out) {
+static int parse_integer_with_flags(CBS *cbs, BIGNUM **out, int flags) {
   assert(*out == NULL);
   *out = BN_new();
   if (*out == NULL) {
     return 0;
   }
-  BN_set_flags(*out, BN_FLG_CONSTTIME);
+  BN_set_flags(*out, flags);
   return BN_parse_asn1_unsigned(cbs, *out);
 }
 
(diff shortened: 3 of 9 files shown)


frewsxcv:
Going to rebase to address conflicts. Hold on.

frewsxcv:
Rebased.

crypto/bn/gcd.c
```diff
@@ -438,6 +436,15 @@ static BIGNUM *BN_mod_inverse_no_branch(BIGNUM *out, int *out_no_inverse,
...
   T = BN_CTX_get(ctx);
+
+  BN_set_flags(A, BN_FLG_CONSTTIME);
+  BN_set_flags(B, BN_FLG_CONSTTIME);
+  BN_set_flags(X, BN_FLG_CONSTTIME);
+  BN_set_flags(D, BN_FLG_CONSTTIME);
+  BN_set_flags(M, BN_FLG_CONSTTIME);
+  BN_set_flags(Y, BN_FLG_CONSTTIME);
+  BN_set_flags(T, BN_FLG_CONSTTIME);
+
```
briansmith: Please only set the CONSTTIME flag on the BIGNUMs that require it.

crypto/bn/gcd.c
```diff
@@ -458,12 +465,15 @@ static BIGNUM *BN_mod_inverse_no_branch(BIGNUM *out, int *out_no_inverse,
...
      * BN_div_no_branch will be called eventually.
      */
-    pB = &local_B;
-    BN_with_flags(pB, B, BN_FLG_CONSTTIME);
-    if (!BN_nnmod(B, pB, A, ctx)) {
+    assert(BN_get_flags(B, BN_FLG_CONSTTIME) != 0);
+    if (BN_get_flags(B, BN_FLG_CONSTTIME) == 0) {
+      OPENSSL_PUT_ERROR(BN, 100000);  // FIXME: what error code?
+      goto err;
+    }
```
briansmith: I suggest just moving the `BN_set_flags` call for `B` here and avoiding this pattern. The above pattern should only be used for `BIGNUM`s that were passed in as parameters to a function or structure members.

crypto/bn/gcd.c
```diff
@@ -484,14 +494,17 @@ static BIGNUM *BN_mod_inverse_no_branch(BIGNUM *out, int *out_no_inverse,
...
+    /* Ensure BN_FLG_CONSTTIME flag is on, so that when BN_div is invoked,
      * BN_div_no_branch will be called eventually.
      */
-    pA = &local_A;
-    BN_with_flags(pA, A, BN_FLG_CONSTTIME);
+    assert(BN_get_flags(A, BN_FLG_CONSTTIME) != 0);
+    if (BN_get_flags(A, BN_FLG_CONSTTIME) == 0) {
+      OPENSSL_PUT_ERROR(BN, 100000);  // FIXME: what error code?
+      goto err;
+    }
```
briansmith: Ditto here, since A is a local variable you can just move the `BN_set_flags` here.

crypto/rsa/blinding.c
```diff
@@ -317,16 +316,12 @@ BN_BLINDING *rsa_setup_blinding(const RSA *rsa, BN_CTX *ctx) {
...
-  BIGNUM *n;
-
   if (rsa->e == NULL) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_NO_PUBLIC_EXPONENT);
     return NULL;
   }
 
-  n = &local_n;
-  BN_with_flags(n, rsa->n, BN_FLG_CONSTTIME);
+  assert(BN_get_flags(rsa->n, BN_FLG_CONSTTIME) != 0);
```
briansmith: Let's use a consistent style: either we omit the `!= 0` always or we include it always. My thought is that we should just omit it always.

crypto/rsa/rsa.c
```diff
@@ -121,23 +121,17 @@ int rsa_new_end(RSA *rsa, BN_CTX *ctx) {
    * BN_FLG_CONSTTIME flag. */
```
briansmith: This comment should be replaced with an assertion that the relevant variables have the CONSTTIME flag set.. But, see my other comment below.

crypto/rsa/rsa_impl.c
```diff
@@ -644,11 +635,10 @@ static int mod_exp(BIGNUM *r0, const BIGNUM *I, RSA *rsa, BN_CTX *ctx) {
     goto err;
   }
 
-  /* Turn BN_FLG_CONSTTIME flag on before division operation */
-  pr1 = &local_r1;
-  BN_with_flags(pr1, r1, BN_FLG_CONSTTIME);
+  /* Ensure BN_FLG_CONSTTIME flag is on before division operation */
```
briansmith: All of the comment of this form are not needed any more because they are replaced by the assertions, so please remove all of them.

crypto/rsa/rsa_impl.c
```diff
@@ -644,11 +635,10 @@ static int mod_exp(BIGNUM *r0, const BIGNUM *I, RSA *rsa, BN_CTX *ctx) {
...
+  assert(BN_get_flags(r1, BN_FLG_CONSTTIME));
```
briansmith: r1 is a local variable and BN_set_flags(r1, BN_FLG_CONSTTIME) is never set. So, won't this assertion always fail? And if so, does that mean this code isn't being run during the tests at all?

frewsxcv: > does that mean this code isn't being run during the tests at all?

It looks like this code is being run in the tests. I put an `assert(0)` here and things seem to fail normally.

If I do this:

```c
  BN_CTX_start(ctx);
  r1 = BN_CTX_get(ctx);
  m1 = BN_CTX_get(ctx);
  vrfy = BN_CTX_get(ctx);
  assert(BN_get_flags(r1, BN_FLG_CONSTTIME));
  assert(BN_get_flags(m1, BN_FLG_CONSTTIME));
  assert(BN_get_flags(vrfy, BN_FLG_CONSTTIME));
```

No assertion fail, which makes me think that we are reusing old entries in the `BN_CTX_POOL` that already have the flags set. Maybe we should reset the flags in `BN_CTX_get`?

frewsxcv:
@briansmith Would it be excessive to offer constant time functions for the `BN` operations? For example, offer `BN_mod_consttime` which does the `assert` for the flag and then calls `BN_mod`

crypto/rsa/rsa.c
frewsxcv: I'm a little confused what you want me to do for this. Which variables are you referring to here?

crypto/bn/gcd.c
frewsxcv: https://github.com/briansmith/ring/pull/154/commits/3270b641a62890fcd75869ea7ed229aa34e02257

crypto/bn/gcd.c
frewsxcv: https://github.com/briansmith/ring/pull/154/commits/3270b641a62890fcd75869ea7ed229aa34e02257

crypto/rsa/blinding.c
frewsxcv: https://github.com/briansmith/ring/pull/154/commits/7fd66c5f03a01598244f0440405dab4d7c0854f5

crypto/rsa/rsa_impl.c
frewsxcv: https://github.com/briansmith/ring/pull/154/commits/da6b6ac630e79587927112cb56a6bf655f0eb7a7

crypto/bn/gcd.c
frewsxcv: https://github.com/briansmith/ring/pull/154/commits/af81e2d434e4b0913084e81dfb71a2e36a4d90bd

crypto/rsa/blinding.c
```diff
@@ -317,16 +316,12 @@ BN_BLINDING *rsa_setup_blinding(const RSA *rsa, BN_CTX *ctx) {
...
-  BIGNUM *n;
-
   if (rsa->e == NULL) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_NO_PUBLIC_EXPONENT);
     return NULL;
   }
 
-  n = &local_n;
-  BN_with_flags(n, rsa->n, BN_FLG_CONSTTIME);
+  assert(BN_get_flags(rsa->n, BN_FLG_CONSTTIME));
```
briansmith: :sigh: It looks like when I refactored this code last week, I didn't remove enough.; notice that `n` and `local_n` are not used. So, just drop the assertion here.

crypto/rsa/rsa.c
briansmith: `assert(BN_get_flags(rsa->p, BN_FLG_CONSTTIME));` etc. for all BIGNUM variables in the RSA structure.

crypto/rsa/rsa_impl.c
```diff
@@ -526,6 +526,8 @@ static int rsa_private_transform(RSA *rsa, uint8_t *out, const uint8_t *in,
     goto err;
   }
 
+  BN_set_flags(f, BN_FLG_CONSTTIME);
+
```
briansmith: Why?

frewsxcv: For [this line](https://github.com/briansmith/ring/blob/b2123c8e303a0b2204cbd97cef06bd7a60f6c4b2/crypto/rsa/rsa_impl.c#L552). Should I move it to immediately before?

crypto/rsa/blinding.c
frewsxcv: https://github.com/briansmith/ring/pull/155

crypto/rsa/rsa_impl.c
briansmith: OK, so IIRC it is because you added an `assert(BN_get_flags(I, BN_FLG_CONSTTIME));`, right? And, without this `BN_set_flags`, that fails, right?

Check this out:

```
#define BN_mod(rem, numerator, divisor, ctx) \
  BN_div(NULL, (rem), (numerator), (divisor), (ctx))
```
In mod_exp:
```
BN_with_flags(c, I, BN_FLG_CONSTTIME);
BN_mod(r1, c, rsa->q, ctx));
```
the `BN_mod` expands to:
```
BN_div(NULL, r, c, rsa->q, (ctx))
```

So, `c` is `BN_div`'s `num` parameter and `rsa->q` is `BN_div`'s `divisor` parameter.

Let's look at `BN_div`:
```
int BN_div(BIGNUM *dv, BIGNUM *rm, const BIGNUM *num, const BIGNUM *divisor,
           BN_CTX *ctx) {
  [...]
  if ((num->flags & BN_FLG_CONSTTIME) != 0 ||
      (divisor->flags & BN_FLG_CONSTTIME) != 0) {
    no_branch = 1;
  }
```

So, instead of setting the BN_FLG_CONSTTIME flag here, you can `assert(BN_get_flags(rsa->q, BN_FLG_CONSTTIME))` instead of asserting the same for `I`. Then you won't need this.

crypto/rsa/rsa_impl.c
briansmith: > No assertion fail, which makes me think that we are reusing old entries in the BN_CTX_POOL that already have the flags set. Maybe we should reset the flags in BN_CTX_get?

Sorry, I didn't see this note. I actually *just* submitted a change to do that to BoringSSL for unrelated reasons: https://boringssl-review.googlesource.com/#/c/7597/.

crypto/rsa/rsa_asn1.c
```diff
@@ -67,12 +67,13 @@
 #include "internal.h"
 
 
-static int parse_integer(CBS *cbs, BIGNUM **out) {
+static int parse_integer_consttime(CBS *cbs, BIGNUM **out) {
```
briansmith: I guess that I misled you here. This should probably be named parse_integer_with_flags and it should take an `int flags` argument because...

crypto/rsa/rsa_asn1.c
```diff
@@ -172,14 +173,14 @@ RSA *RSA_parse_private_key(CBS *cbs) {
...
-  if (!parse_integer(&child, &ret->n) ||
-      !parse_integer(&child, &ret->e) ||
-      !parse_integer(&child, &ret->d) ||
-      !parse_integer(&child, &ret->p) ||
-      !parse_integer(&child, &ret->q) ||
-      !parse_integer(&child, &ret->dmp1) ||
-      !parse_integer(&child, &ret->dmq1) ||
-      !parse_integer(&child, &ret->iqmp)) {
+  if (!parse_integer_consttime(&child, &ret->n) ||
+      !parse_integer_consttime(&child, &ret->e) ||
```
briansmith: ...as discussed previously (on IRC, IIRC), we don't want `n` and `e` to have the consttime flag.

crypto/rsa/rsa_impl.c
```diff
@@ -646,11 +635,9 @@ static int mod_exp(BIGNUM *r0, const BIGNUM *I, RSA *rsa, BN_CTX *ctx) {
     goto err;
   }
 
-  /* Turn BN_FLG_CONSTTIME flag on before division operation */
-  pr1 = &local_r1;
-  BN_with_flags(pr1, r1, BN_FLG_CONSTTIME);
+  assert(BN_get_flags(r1, BN_FLG_CONSTTIME));
```
briansmith: NIT: It doesn't matter much, but might as well assert `rsa->p` has the consttime flag, instead of assertint that `r1` has it, to be consistent with the above assertions.

briansmith:
> @briansmith Would it be excessive to offer constant time functions for the BN operations? For example, offer BN_mod_consttime which does the assert for the flag and then calls BN_mod.

As discussed on IRC, I would love to get rid of the `BN_FLG_CONSTTIME` flag and instead just call different functions depending on whether we want the consttime behavior or not. If you want to submit another commit that does that, go ahead. But, I recommend not doing it for `BN_mod` and `BN_nnmod` because we're going to get rid of all of the consttime calls to those functions anyway, because they can't be constant time.

Anyway, this is looking awesome.

frewsxcv:
>  If you want to submit another commit that does that, go ahead. 

I think I'd rather do that in a follow-up PR for now

All other comments have been addressed

briansmith:
Thanks! I made the following tweaks and pushed this as e37d03107df5e08af99ffc2b85e8d5c047306b73.

```
-  BN_set_flags(r0, BN_FLG_CONSTTIME);
-
   bitsp = (bits + 1) / 2;
   bitsq = bits - bitsp;
 
@@ -827,6 +825,7 @@ RSA *RSA_generate(int bits, uint32_t e, BN_GENCB *cb) {
   if (!BN_mul(r0, r1, r2, ctx)) {
     goto err; /* (p-1)(q-1) */
   }
+  BN_set_flags(r0, BN_FLG_CONSTTIME);
   if (!BN_mod_inverse(rsa->d, rsa->e, r0, ctx)) {
     goto err; /* d */
   }
diff --git a/crypto/rsa/rsa_test.cc b/crypto/rsa/rsa_test.cc
index dd9ddb5..add39ab 100644
--- a/crypto/rsa/rsa_test.cc
+++ b/crypto/rsa/rsa_test.cc
@@ -359,8 +359,8 @@ static bool TestOnlyDGiven() {
   ScopedBN_CTX ctx(BN_CTX_new());
   // XXX TODO: |rsa_new_end| should fail because the complete key isn't provided.
   if (!key ||
-      !BN_hex2bn_with_flags(&key->n, kN, BN_FLG_CONSTTIME) ||
-      !BN_hex2bn_with_flags(&key->e, kE, BN_FLG_CONSTTIME) ||
+      !BN_hex2bn(&key->n, kN) ||
+      !BN_hex2bn(&key->e, kE) ||
       !BN_hex2bn_with_flags(&key->d, kD, BN_FLG_CONSTTIME) ||
       !rsa_new_end(key.get(), ctx.get()) ||
       RSA_size(key.get()) > sizeof(buf)) {
@@ -403,7 +403,7 @@ static bool TestOnlyDGiven() {
   // XXX TODO: |rsa_new_end| should fail because the complete key isn't provided.
   ScopedRSA key2(rsa_new_begin());
   if (!key2 ||
-      !BN_hex2bn_with_flags(&key2->n, kN, BN_FLG_CONSTTIME) ||
+      !BN_hex2bn(&key2->n, kN) ||
       !BN_hex2bn_with_flags(&key2->d, kD, BN_FLG_CONSTTIME) ||
       !rsa_new_end(key2.get(), ctx.get())) {
     return false;
```

PR #156, opened 2016-03-24
Rewrite SHA-1 in Rust.

### LINKED ISSUE #60: Replace C implementation of SHA-1 with a small Rust implementation
At this point, SHA-1 is purely a legacy algorithm and it's more important to minimize its code size than to maximize its performance. The assembly language implementations have already been removed, so the BoringSSL C fallback implementation is always used.

When this is done, the documentation for `ring::digest::SHA1` and `ring::pbkdf2::
HMAC_SHA1` needs to be updated to note that these are optimized for size, not performance.

There are several open source, public domain, and CC0-licensed SHA-1 Rust implementations. Ideally we'd just integrate the smallest one. But, we should make sure that we use a public domain and/or CC0-licensed one and/or that the author is OK with us changing the license to the ISC license that the rest of the Rust code in *ring* uses.

SimonSapin:
Fix #60.

I found a number of SHA-1 implementations in Rust but none of them were in the public domain or under CC0 or ISC, so I wrote another one based on reading FIPS 180-4.

As discussed in #60 this does not optimize for speed but keeps the source code short and simple (for example without manual loop unrolling) as a first approximation of optimizing for machine code size.

### DIFF
--- a/src/digest/digest.rs
+++ b/src/digest/digest.rs
@@ -29,8 +29,6 @@
 use core;
 use super::{c, polyfill};
 
-pub use self::sha1::SHA1;
-
 // XXX: endian-specific.
 // XXX: Replace with `const fn` when `const fn` is stable:
 // https://github.com/rust-lang/rust/issues/24111
@@ -114,7 +112,7 @@ impl Context {
                 &data[..to_copy]);
 
             unsafe {
-                (self.algorithm.block_data_order)(self.state.as_mut_ptr(),
+                (self.algorithm.block_data_order)(&mut self.state,
                                                   self.pending.as_ptr(), 1);
             }
             self.completed_data_blocks =
@@ -128,7 +126,7 @@ impl Context {
         let num_to_save_for_later = remaining.len() % self.algorithm.block_len;
         if num_blocks > 0 {
             unsafe {
-                (self.algorithm.block_data_order)(self.state.as_mut_ptr(),
+                (self.algorithm.block_data_order)(&mut self.state,
                                                   remaining.as_ptr(),
                                                   num_blocks);
             }
@@ -161,7 +159,7 @@ impl Context {
             polyfill::slice::fill(
                 &mut self.pending[padding_pos..self.algorithm.block_len], 0);
             unsafe {
-                (self.algorithm.block_data_order)(self.state.as_mut_ptr(),
+                (self.algorithm.block_data_order)(&mut self.state,
                                                   self.pending.as_ptr(), 1);
             }
             // We don't increase |self.completed_data_blocks| because the
@@ -185,7 +183,7 @@ impl Context {
             completed_data_bits /= 0x100;
         }
         unsafe {
-            (self.algorithm.block_data_order)(self.state.as_mut_ptr(),
+            (self.algorithm.block_data_order)(&mut self.state,
                                               self.pending.as_ptr(), 1);
         }
 
@@ -283,7 +281,7 @@ pub struct Algorithm {
     /// The length of the length in the padding.
     len_len: usize,
 
-    block_data_order: unsafe extern fn(state: *mut u64, data: *const u8,
+    block_data_order: unsafe extern fn(state: &mut [u64; MAX_CHAINING_LEN / 8], data: *const u8,
                                        num: c::size_t),
     format_output: fn (input: &[u64; MAX_CHAINING_LEN / 8]) ->
                        [u64; MAX_OUTPUT_LEN / 8],
@@ -295,6 +293,26 @@ pub struct Algorithm {
     /// references to the same algorithm.
     pub nid: c::int,
 }
+
+/// SHA-1 as specified in [FIPS
+/// 180-4](http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).
+///
+/// SHA-1 is deprecated.
+pub static SHA1: Algorithm = Algorithm {
+    output_len: 160 / 8,
+    chaining_len: sha1::CHAINING_LEN,
+    block_len: sha1::BLOCK_LEN,
+    len_len: 64 / 8,
+    block_data_order: sha1::block_data_order,
+    format_output: sha256_format_output,
+    initial_state: [
+        u32x2!(0x67452301u32, 0xefcdab89u32),
+        u32x2!(0x98badcfeu32, 0x10325476u32),
+        u32x2!(0xc3d2e1f0u32, 0u32),
+        0, 0, 0, 0, 0,
+    ],
+    nid: 64, // NID_sha1
+};
 
 /// SHA-256 as specified in [FIPS
 /// 180-4](http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).
@@ -435,8 +453,8 @@ pub extern fn SHA512_4(out: *mut u8, out_len: c::size_t,
 }
 
 extern {
-    fn sha256_block_data_order(state: *mut u64, data: *const u8, num: c::size_t);
-    fn sha512_block_data_order(state: *mut u64, data: *const u8, num: c::size_t);
+    fn sha256_block_data_order(state: &mut [u64; MAX_CHAINING_LEN / 8], data: *const u8, num: c::size_t);
+    fn sha512_block_data_order(state: &mut [u64; MAX_CHAINING_LEN / 8], data: *const u8, num: c::size_t);
 }
 
 #[cfg(test)]
--- a/src/digest/sha1.rs
+++ b/src/digest/sha1.rs
@@ -17,31 +17,11 @@ use core;
 use core::num::Wrapping;
 use c;
 use polyfill;
-use super::{Algorithm, sha256_format_output, MAX_CHAINING_LEN};
+use super::MAX_CHAINING_LEN;
 
-/// SHA-1 as specified in [FIPS
-/// 180-4](http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).
-///
-/// SHA-1 is deprecated in *ring*.
-/// The main purpose in retaining it is to support legacy protocols and OCSP,
-/// none of which need a fast SHA-1 implementation.
-/// This implementation therfore favors size and simplicity over speed.
-/// Unlike SHA-256, SHA-384, and SHA-512, there is no assembly language implementation.
-pub static SHA1: Algorithm = Algorithm {
-    output_len: 160 / 8,
-    chaining_len: 160 / 8,
-    block_len: 512 / 8,
-    len_len: 64 / 8,
-    block_data_order: sha1_block_data_order,
-    format_output: sha256_format_output,
-    initial_state: [
-        u32x2!(0x67452301u32, 0xefcdab89u32),
-        u32x2!(0x98badcfeu32, 0x10325476u32),
-        u32x2!(0xc3d2e1f0u32, 0u32),
-        0, 0, 0, 0, 0,
-    ],
-    nid: 64, // NID_sha1
-};
+pub const BLOCK_LEN: usize = 512 / 8;
+pub const CHAINING_LEN: usize = 160 / 8;
+const CHAINING_WORDS: usize = CHAINING_LEN / 4;
 
 type W32 = Wrapping<u32>;
 
@@ -49,19 +29,31 @@ type W32 = Wrapping<u32>;
 #[inline] fn parity(x: W32, y: W32, z: W32) -> W32 { x ^ y ^ z }
 #[inline] fn maj(x: W32, y: W32, z: W32) -> W32 { (x & y) | (x & z) | (y & z) }
 
-unsafe extern fn sha1_block_data_order(state: *mut u64, mut data: *const u8,
-                                       num: c::size_t) {
-    let state = core::slice::from_raw_parts_mut(state, MAX_CHAINING_LEN / 8);
+/// The main purpose in retaining this is to support legacy protocols and OCSP,
+/// none of which need a fast SHA-1 implementation.
+/// This implementation therefore favors size and simplicity over speed.
+/// Unlike SHA-256, SHA-384, and SHA-512,
+/// there is no assembly language implementation.
+pub unsafe extern fn block_data_order(state: &mut [u64; MAX_CHAINING_LEN / 8],
+                                      data: *const u8,
+                                      num: c::size_t) {
+    let data = data as *const [u8; BLOCK_LEN];
+    let blocks = core::slice::from_raw_parts(data, num);
+    block_data_order_safe(state, blocks)
+}
+
+fn block_data_order_safe(state: &mut [u64; MAX_CHAINING_LEN / 8],
+                         blocks: &[[u8; BLOCK_LEN]]) {
     let state = polyfill::slice::u64_as_u32_mut(state);
     let state = polyfill::slice::as_wrapping_mut(state);
+    let state = &mut state[..CHAINING_WORDS];
+    let state = slice_as_array_ref_mut!(state, CHAINING_WORDS).unwrap();
 
     let mut w: [W32; 80] = [Wrapping(0); 80];
-    for _ in 0..num {
-        let block = core::slice::from_raw_parts(data, 512 / 8);
-        data = data.offset(512 / 8);
-
+    for block in blocks {
         for t in 0..16 {
-            w[t] = Wrapping(polyfill::slice::u32_from_be_u8_at(block, t * 4))
+            let word = slice_as_array_ref!(&block[t * 4..][..4], 4).unwrap();
+            w[t] = Wrapping(polyfill::slice::u32_from_be_u8(word))
         }
         for t in 16..80 {
             let wt = w[t - 3] ^ w[t - 8] ^ w[t - 14] ^ w[t - 16];
--- a/src/polyfill.rs
+++ b/src/polyfill.rs
@@ -38,11 +38,11 @@ pub mod slice {
     use core;
 
     #[inline(always)]
-    pub fn u32_from_be_u8_at(buffer: &[u8], at: usize) -> u32 {
-        u32::from(buffer[at]) << 24 |
-        u32::from(buffer[at + 1]) << 16 |
-        u32::from(buffer[at + 2]) << 8 |
-        u32::from(buffer[at + 3])
+    pub fn u32_from_be_u8(buffer: &[u8; 4]) -> u32 {
+        u32::from(buffer[0]) << 24 |
+        u32::from(buffer[1]) << 16 |
+        u32::from(buffer[2]) << 8 |
+        u32::from(buffer[3])
     }
 
     // https://github.com/rust-lang/rust/issues/27750


SimonSapin:
I was expecting this to be slower, but I was surprised by how much. `cargo test` before this PR took ~3.5 minutes on my machine (all but 15 seconds spent in `pkbdf2_tests`). It now takes almost 24 minutes.

I suspect this particular test case:

```
Hash = SHA1
P = "password"
S = "salt"
c = 16777216
DK = eefe3d61cd4da4e4e9945b3d6ba2158c2634e984
```

Maybe it’s OK to test fewer than 16 millions iterations? Removing a couple digits from `c` brings the total time under 17 seconds.

Enabling optimizations for tests also helps a lot. A comment in `Cargo.toml` suggests this is intended, but it isn’t actually done.

Curiously, optimizations seem to make tests other than `pbkdf2` *slower*…

At this point I have no idea what’s going on and it’s way past time for me to got to bed.

briansmith:
Simon, for benchmarking use https://github.com/briansmith/crypto-bench#how-to-run-all-the-benchmarks-for-a-specific-crypto-library. It would be great if you could report before/after numbers for one platform (please state which platform it is) for the various input sizes that crypto-bench supports.

I agree we need to fix the slow tests. @pyfisch has submitted one change in that direction and I've been waffling about how to resolve it. But, now's the time to do so.

src/digest.rs
```diff
@@ -448,8 +447,66 @@ pub extern fn SHA512_4(out: *mut u8, out_len: c::size_t,
     polyfill::slice::fill_from_slice(out, digest);
 }
 
+unsafe extern fn sha1_block_data_order(state: *mut u64, data: *const u8, num: c::size_t) {
+    sha1_process_blocks(
+        &mut *(state as *mut _),
```
briansmith: In general we don't use `as` in *ring* as we consider `as` to be too unsafe in general. We have `polyfill::slice::u64_as_u32` for this kind of conversion.

src/digest.rs
```diff
@@ -448,8 +447,66 @@ pub extern fn SHA512_4(out: *mut u8, out_len: c::size_t,
...
+        core::slice::from_raw_parts(data as *const _, num)
```
briansmith: This cast is not valid because there's no guarantee that `data` is aligned enough for accessing as `u32`. This is why the C code uses `from_be_u32_ptr`.

SimonSapin:
crypto-bench on Archlinux x86_64 with an Intel i7-5600U, rustc 1.9.0-nightly (0dcc413e4 2016-03-22).

ring master @ b12faf1fdb800fbdbf59ed1eee7b8758674f2627

```
test digest::sha1::_1000      ... bench:       2,147 ns/iter (+/- 109) = 465 MB/s
test digest::sha1::_16        ... bench:         182 ns/iter (+/- 28) = 87 MB/s
test digest::sha1::_2000      ... bench:       4,229 ns/iter (+/- 189) = 472 MB/s
test digest::sha1::_256       ... bench:         698 ns/iter (+/- 41) = 366 MB/s
test digest::sha1::_8192      ... bench:      16,872 ns/iter (+/- 1,541) = 485 MB/s
test digest::sha1::block_len  ... bench:         319 ns/iter (+/- 281) = 200 MB/s
test pbkdf2::hmac_sha1        ... bench:  39,665,282 ns/iter (+/- 7,941,013)
```

This PR @ 97749e2c12d6f6f1bd09d9ba0a2a3b601b568098

```
test digest::sha1::_1000      ... bench:       4,169 ns/iter (+/- 202) = 239 MB/s
test digest::sha1::_16        ... bench:         306 ns/iter (+/- 125) = 52 MB/s
test digest::sha1::_2000      ... bench:       8,224 ns/iter (+/- 669) = 243 MB/s
test digest::sha1::_256       ... bench:       1,360 ns/iter (+/- 68) = 188 MB/s
test digest::sha1::_8192      ... bench:      33,014 ns/iter (+/- 1,405) = 248 MB/s
test digest::sha1::block_len  ... bench:         563 ns/iter (+/- 32) = 113 MB/s
test pbkdf2::hmac_sha1        ... bench:  63,681,196 ns/iter (+/- 15,704,245)
```

So 1.6~2x slower. Not as dramatic as the ~8x I observed with unoptimized tests.

I’ll look separately at keeping `cargo test` times reasonable.

src/digest.rs
```diff
@@ -471,6 +528,22 @@ mod tests {
...
+    fn test_sha1_sizes() {
+        use core::mem::size_of;
+        use super::{Sha1State, Sha1Block};
+
+        assert_eq!(digest::SHA1.block_len, size_of::<Sha1Block>());
+        assert_eq!(digest::SHA1.chaining_len, size_of::<Sha1State>());
+    }
+
+    #[test]
+    fn test_single_sha1() {
```
briansmith: This test duplicates this test vector https://github.com/briansmith/ring/blob/3bbc5ebc09f328aa9c521dc14b476c7394051951/src/digest_tests.txt#L3 and so it should be removed. (In general, prefer to put test vectors into data files.)

src/digest.rs
```diff
@@ -448,8 +447,66 @@ pub extern fn SHA512_4(out: *mut u8, out_len: c::size_t,
     polyfill::slice::fill_from_slice(out, digest);
 }
 
+unsafe extern fn sha1_block_data_order(state: *mut u64, data: *const u8, num: c::size_t) {
```
briansmith: Please add some documentation about the speed-for-size/simplicity trade-off we are making here. The main purpose in retaining SHA-1 is to support legacy protocols and OCSP, none of which need a fast SHA-1 implementation. Also, please add a note that, unlike SHA-256, SHA-384, and SHA-512, there are no primary assembly language implementations.

Finally, could you please put this in a separate file, following the example of `ring::aead::chacha20_poly1305`. For the license header, please use the ISC license header used in this file, but make sure the Copyright line of the file contains the copyright holder's name.

briansmith:
We should just remove the `ring::pbkdf2::HMAC_SHA1`. It doesn't make sense—in fact, it is kind of harmful—to support a slow implementation of PBKDF2. Feel free to do so (perhaps as a separate commit), or I'll do it soon.

Also, thanks!

src/digest.rs
```diff
@@ -448,8 +447,66 @@ pub extern fn SHA512_4(out: *mut u8, out_len: c::size_t,
...
+            w[t] = (w[t - 3] ^ w[t - 8] ^ w[t - 14] ^ w[t - 16]).rotate_left(1)
+        }
+
+        let mut a = state[0];
+        let mut b = state[1];
+        let mut c = state[2];
+        let mut d = state[3];
+        let mut e = state[4];
+
+
```
briansmith: Remove one of these blank lines.

src/digest.rs
briansmith: See https://github.com/briansmith/ring/commit/279ca8e1b0e0f4eef93c0c5ab458671b095ec41f, which demonstrates how to add a (private) ring::digest::sha1 submodule.

src/digest.rs
```diff
@@ -448,8 +447,66 @@ pub extern fn SHA512_4(out: *mut u8, out_len: c::size_t,
...
+        for t in 0..80 {
+            let (k, f) = match t {
+                0...19 => (0x5a827999, ch(b, c, d)),
+                20...39 => (0x6ed9eba1, parity(b, c, d)),
+                40...59 => (0x8f1bbcdc, maj(b, c, d)),
+                60...79 => (0xca62c1d6, parity(b, c, d)),
+                _ => unreachable!()
+            };
+            let tt = a.rotate_left(5)
+                .wrapping_add(f)
```
briansmith: Nit: Is this wrapped correctly? It looks pretty horrific to me. It seems like it would be better to align the dots.

Also, wouldn't it be better to use `Wrapping<u32>` instead, since all the math here is wrapping? It seems like it would be less error -prone to use `Wrapping<u32>` and also easier to read.

SimonSapin: I did that at first but `Wrapping<u32>` doesn’t have `rotate_left`. On further thought I could use something like `Wrapping(a.0.rotate_left(5))`. Do you think that’s better ?

src/digest.rs
```diff
@@ -279,8 +279,7 @@ pub struct Algorithm {
     /// The length of the length in the padding.
     len_len: usize,
 
-    block_data_order: unsafe extern fn(state: *mut u64, data: *const u8,
-                                       num: c::size_t),
+    block_data_order: unsafe extern fn(state: *mut u64, data: *const u8, num: c::size_t),
```
briansmith: Please undo this change. (If you are using rustfmt, you need to set it to use an ~80 character limit.

src/digest.rs
briansmith: Sorry this isn't in the documentation: Please limit line length to 80 characters (punctuation can spill over by a character or two).

src/digest.rs
briansmith: I guess we should file a bug against the Rust stdlib about the missing `rotate_left` for `Wrapping`. We can define our own polyfill in `ring::polyfill` in the meantime; something like this:

```
// `core::num::Wrapping` doesn't support `rotate_left`. There is no usable
// trait for `rotate_left`, so this polyfill just hard-codes u32.
// https://github.com/rust-lang/rust/issues/XXXX
fn wrapping_rotate_left_u32(x: core::num::Wrapping<u32>, n: u32)
                            -> core::num::Wrapping<u32> {
   let core::num::Wrapping(x) = x;
   core::num::Wrapping(x.rotate_left(n)) 
}
```

This would allow us to use Wrapping<u32> for the time of every variable we do math on here, I think. It's ugly, but hopefully the ugliness is only temporary.

SimonSapin:
I’m removing `ring::pbkdf2::HMAC_SHA1` in a separate commit of this PR. This "fixes" `cargo test` being slow (now under 12 seconds for me) since all other PBKDF2 test cases with large iteration numbers are commented out.

src/digest.rs
SimonSapin: Removed. (I’d included this to be able to trace through a single digest for debugging, but that bug is solved now.)

SimonSapin:
Alright, I think I’ve addressed all comments. I’ve also change the signature of `block_data_order` (see second commit) and managed to add `#![deny(unsafe_code)]` to `sha1.rs`.

I’ve added this to `polyfill::slice`, for the state buffer:

```
pub fn as_wrapping_mut<'a, T>(src: &'a mut [T]) -> &'a mut [core::num::Wrapping<T>]
```

I don’t like it, but I think it’s correct. It can be removed if we go back to `wrapping_add` instead of `Wrapping`.

SimonSapin:
This is now 1.8~2.3x slower than the manually-unrolled C implementation.

src/digest/mod.rs
```diff
@@ -109,10 +109,8 @@ impl Context {
                 &mut self.pending[self.num_pending..self.algorithm.block_len],
                 &data[..to_copy]);
 
-            unsafe {
-                (self.algorithm.block_data_order)(self.state.as_mut_ptr(),
-                                                  self.pending.as_ptr(), 1);
-            }
+            (self.algorithm.block_data_order)(&mut self.state,
```
briansmith: The premise of this change is wrong. On ARM/AAarch64/x86/x86-64, block_data_order for SHA-{256,385,512} is implemented in assembly language and probably won't ever get replaced with Rust code. The C implementations might get replaced with Rust code, but they are mostly just placeholders for implementations that don't have assembly language implementations. SHA-1 is the exception because it is a deprecated algorithm and so the assembly language code was deleted.

Accordingly, the Rust implementations of these functions will be the exceptional case and so I don't think we should add overhead to the common case where the assembly language implementations are being used.

Note: On Windows, the assembly language implementations are not enabled yet, because I need to find a way to avoid depending on Perl on Windows, and I haven't gotten around to it yet.

briansmith:
When moving the digest modulue to a directory, please follow the guidelines at https://github.com/briansmith/ring/blob/master/STYLE.md#submodules-and-file-naming by naming the file "digest.rs" instead of "mod.rs" so that the project doesn't get filled with a bunch of files named "mod.rs".

src/digest/sha1.rs
```diff
@@ -0,0 +1,102 @@
...
+    let mut w: [W32; 80] = [Wrapping(0); 80];
+    for block in data.chunks(64) {
+        let block = slice_as_array_ref!(block, 64).unwrap();
+
+        for t in 0..16 {
+            w[t] = Wrapping(
+                (block[t * 4] as u32) << 24 |
+                (block[t * 4 + 1] as u32) << 16 |
+                (block[t * 4 + 2] as u32) << 8 |
+                (block[t * 4 + 3] as u32));
```
briansmith: 1. Is this change to deal with unaligned inputs the cause of the worse performance? Or was it the conversion to `Wrapping`? Or a combination?

If the comversion to `Wrapping` caused the worse performance then we should go back to the way it was before. (And, in that case we should file a bug on the Rust stdlib.)

2. Please extract this out to a function named `from_be_u32_ptr` or similar.

3. Please use `u32::from(x)` instead of `x as u32` to follow the guidelines at https://github.com/briansmith/ring/blob/master/STYLE.md#casting-as-and-conversions.

briansmith: 1. Is this change to deal with unaligned inputs the cause of the worse performance? Or was it the conversion to `Wrapping`? Or a combination?

If the comversion to `Wrapping` caused the worse performance then we should go back to the way it was before. (And, in that case we should file a bug on the Rust stdlib.)

2. Please extract this out to a function named `from_be_u32_ptr` or similar.

3. Please use `u32::from(x)` instead of `x as u32` to follow the guidelines at https://github.com/briansmith/ring/blob/master/STYLE.md#casting-as-and-conversions.

src/digest/mod.rs
SimonSapin: Ok. I’ll revert this.

src/digest/sha1.rs
SimonSapin: `Wrapping` vs `wrapping_add` doesn’t seem to affect crypto-bench. Casting `data` to `&[[u32; 16]]` does.

SimonSapin:
Updated.

briansmith:
I merged the first two commits: 2087d5a710334e3e6d65c87f4964232fa7dec29a and 6b339d4cab6a7fea7b5eac46bc747ce3a2611e9e.

I have some new comments on the main patch. Sorry for the delays and nit-picking here. This project is part of an experiment I'm doing in trying to use type systems and static analysis to improve the safety of crypto and so I'm kind of learning as I go along with this review. :)

src/digest/digest.rs
```diff
@@ -292,24 +296,6 @@ pub struct Algorithm {
     pub nid: c::int,
 }
 
-/// SHA-1 as specified in [FIPS
```
briansmith: I worry that if we move this to the submodule, any time we need to change all of the `Algorithm` specifications, we risk forgetting to change SHA-1's. So, I'd rather leave this here.

src/digest/sha1.rs
```diff
@@ -0,0 +1,100 @@
...
+
+#[inline] fn ch(x: W32, y: W32, z: W32) -> W32 { (x & y) | (!x & z) }
+#[inline] fn parity(x: W32, y: W32, z: W32) -> W32 { x ^ y ^ z }
+#[inline] fn maj(x: W32, y: W32, z: W32) -> W32 { (x & y) | (x & z) | (y & z) }
+
+unsafe extern fn sha1_block_data_order(state: *mut u64, mut data: *const u8,
+                                       num: c::size_t) {
+    let state = core::slice::from_raw_parts_mut(state, MAX_CHAINING_LEN / 8);
+    let state = polyfill::slice::u64_as_u32_mut(state);
+    let state = polyfill::slice::as_wrapping_mut(state);
```
briansmith: I would prefer to split this into two functions: One `unsafe` function like you have here, and then another non-`unsafe` function that does the computation. That way, it is clear that the implementation is safe in the Rust sense. I believe this is what you were trying to accomplish with one of your earlier refactorings. I think this way, we get to have our cake (safe SHA-1) and eat it too (no extra overhead when calling the assembly-language implementations).

briansmith: We should add:
`let state = slice_as_array_ref_mut(state, SHA1_CHAINING_WORDS);`

This way, type of state will be `[u32; SHA1_CHAINING_WORDS]`. Besides being a stronger type, this should also help the compiler reduce bounds checking.

src/digest/sha1.rs
```diff
@@ -0,0 +1,100 @@
...
+
+type W32 = Wrapping<u32>;
+
+#[inline] fn ch(x: W32, y: W32, z: W32) -> W32 { (x & y) | (!x & z) }
+#[inline] fn parity(x: W32, y: W32, z: W32) -> W32 { x ^ y ^ z }
+#[inline] fn maj(x: W32, y: W32, z: W32) -> W32 { (x & y) | (x & z) | (y & z) }
+
+unsafe extern fn sha1_block_data_order(state: *mut u64, mut data: *const u8,
+                                       num: c::size_t) {
+    let state = core::slice::from_raw_parts_mut(state, MAX_CHAINING_LEN / 8);
```
briansmith: I suggest that you just use `state.len()` here because we'll truncate state to the right length below and there' s no reason to do the truncation twice.

src/digest/sha1.rs
```diff
@@ -0,0 +1,100 @@
...
+    let state = polyfill::slice::u64_as_u32_mut(state);
+    let state = polyfill::slice::as_wrapping_mut(state);
+
+    let mut w: [W32; 80] = [Wrapping(0); 80];
+    for _ in 0..num {
+        let block = core::slice::from_raw_parts(data, 512 / 8);
+        data = data.offset(512 / 8);
```
briansmith: Once this becomes a safe function, we'll have to use safe alternatives to this.

src/digest/sha1.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        for t in 0..16 {
+            w[t] = Wrapping(polyfill::slice::u32_from_be_u8_at(block, t * 4))
```
briansmith: The type of the input to this function should be `&[u8; 4]` instead of `&[u8]`. That way, it will be guaranteed not to panic.

That means that, in order to call it, we'd need to use  `slice_as_array_ref!` before calling it. Hopefully the compiler will optimize out the bounds check in `slice_as_array_ref!`.

src/polyfill.rs
```diff
@@ -17,14 +17,34 @@
...
 
+use core;
+
 #[inline(always)]
 pub fn u64_from_usize(x: usize) -> u64 {
     x as u64
 }
 
+/// `core::num::Wrapping` doesn't support `rotate_left`.
+/// There is no usable trait for `rotate_left`, so this polyfill just hard-codes u32.
```
briansmith: This line is too long.

src/polyfill.rs
```diff
@@ -17,14 +17,34 @@
...
+                            -> core::num::Wrapping<u32> {
+    core::num::Wrapping(x.0.rotate_left(n))
+}
+
+
 pub mod slice {
     use core;
 
+    #[inline(always)]
+    pub fn u32_from_be_u8_at(buffer: &[u8], at: usize) -> u32 {
```
briansmith: As indicated above, the parameter should be `&[u8; 4]`

src/digest/sha1.rs
briansmith: Please use a named constant for the value `512 / 8`, e.g. `SHA1_BLOCK_LEN`.

Also, we should use `slice_as_array_ref!` to make the type of block `&[u8; SHA1_BLOCK_LEN]`.

briansmith:
BTW, this is looking awesome. I'm very excited about this.

src/digest/sha1.rs
```diff
@@ -0,0 +1,100 @@
...
+#[inline] fn ch(x: W32, y: W32, z: W32) -> W32 { (x & y) | (!x & z) }
+#[inline] fn parity(x: W32, y: W32, z: W32) -> W32 { x ^ y ^ z }
+#[inline] fn maj(x: W32, y: W32, z: W32) -> W32 { (x & y) | (x & z) | (y & z) }
+
+unsafe extern fn sha1_block_data_order(state: *mut u64, mut data: *const u8,
+                                       num: c::size_t) {
+    let state = core::slice::from_raw_parts_mut(state, MAX_CHAINING_LEN / 8);
+    let state = polyfill::slice::u64_as_u32_mut(state);
+    let state = polyfill::slice::as_wrapping_mut(state);
+
```
briansmith: We should have a `debug_assert_eq!(data.len() % SHA1_BLOCK_LEN, 0)`

briansmith: Actually, maybe there is some way, given `data: &[u8]`, to create a `&[[u8; SHA1_BLOCK_LEN]]` (i.e. a slice of 512-byte byte arrays)? This would be the strongest possible type and should eliminate almost all the bounds checking.

src/digest/sha1.rs
```diff
@@ -0,0 +1,100 @@
...
+use core;
+use core::num::Wrapping;
+use c;
+use polyfill;
+use super::{Algorithm, sha256_format_output, MAX_CHAINING_LEN};
+
+/// SHA-1 as specified in [FIPS
+/// 180-4](http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).
+///
+/// SHA-1 is deprecated in *ring*.
```
briansmith: Please remove "in *ring*".

src/digest/sha1.rs
```diff
@@ -0,0 +1,100 @@
...
+/// The main purpose in retaining it is to support legacy protocols and OCSP,
+/// none of which need a fast SHA-1 implementation.
+/// This implementation therfore favors size and simplicity over speed.
```
briansmith: s/therfore/therefore/

src/digest/sha1.rs
```diff
@@ -0,0 +1,100 @@
...
+/// Unlike SHA-256, SHA-384, and SHA-512, there is no assembly language implementation.
```
briansmith: Please move the text "The main purpose...assembly language implementation." to be directly above the code implementing the algorithm. In particular, it doesn't make sense for this text to appear in the rustdoc for the module. Also, people are more likely to be wondering about this design choice when reading that code than when reading the definition of the `SHA1` static.

SimonSapin:
> I merged the first two commits: 2087d5a and 6b339d4.

Looks like you got all 3. (Accidentally?) f645693e2afa893371843a600214fce4d8c687a3 is reachable from current master.

src/digest/sha1.rs
SimonSapin: `state` at that point is `*mut u64` and doesn’t have a `.len()` method. But this line can be removed entirely if we change the type to `&mut [u64; MAX_CHAINING_LEN / 8]` and change `self.state.as_mut_ptr()` in calls to `&mut self.state`

It seems to works out (tests still pass) despite the mismatch between the C signatures of `sha256_block_data_order` / `sha512_block_data_order` and their Rust `extern fn` declarations. The former already had a mismatch anyway (`uint32_t *` vs `*mut u64`).

briansmith:
:sigh: apparently I don't know how to use git. Please still make the suggested changes as a separate commit.

src/digest/sha1.rs
briansmith: Yes, that sounds like the right way to do it.

src/digest/sha1.rs
SimonSapin: Everything’s possible with `as` on raw pointers or `transmute` :) In this case I’d recommend:

```rust
let data = core::slice::from_raw_parts(data as *const [u8; SHA1_BLOCK_LEN], num);
```

Yes casting raw pointers with `as` is totally unsafe, but we have to deal with unsafety regardless if inputs are raw pointers. And this specific cast + `from_raw_parts` matches exactly the "contract" of how `block_data_order` functions are called.

src/digest/sha1.rs
SimonSapin: I had that at some point and removed it when I saw it didn’t affect `crypto-bench` results, but I don’t mind adding it back.

src/polyfill.rs
SimonSapin: I’m used to projects where the limit is 100 or 120, and the only help my editor gives with this is a vertical bar whose position is a global setting (not per-project).

I can add to the repo a subset of [Servo’s `tidy.py`](https://github.com/servo/servo/blob/0826a5b45455cdbecf484136b1bab39bcc0b15ff/python/tidy.py#L120-L125) to help with this.

SimonSapin: https://github.com/briansmith/ring/pull/158

SimonSapin:
Rebased and updated.

SimonSapin:
So, master is now c5fc35ff04ca95fef064cbd615c4db95cb2b03b0 which has the exact content of this PR’s bd18e47. Did you mean to merge this PR?

briansmith:
> So, master is now c5fc35f which has the exact content of this PR’s bd18e47. Did you mean to merge this PR?

Yes, this time I actually intended to land the change! :)

Thanks for addressing the comments. If you are interested, there are some optimizations that can be done that don't make the code (much) more complicated:

1. The `w` array has 80 elements, but only 16 elements are used at a time. Most implementations optimize this by indexing a 16-element array.

2. At https://en.wikipedia.org/wiki/SHA-1#Examples_and_pseudocode, "Instead of the formulation from the original FIPS PUB 180-1 shown..." shows some alternate math that can be used.

3. See https://software.intel.com/en-us/articles/improving-the-performance-of-the-secure-hash-algorithm-1. I don't know if there's any hope of rustc auto-vectorizing this code, but if there is, then you can help it vectorize by using the alternate formulation of SHA-1 described.

SimonSapin:
Alright, let’s close this as merged. I’ll try with what you suggest and send another PR if I see improvements in `crypto-bench`.

For the `w` array, FIPS 180-4 suggested this but presented it as a trading speed (additional modulo takes time) for memory (256 fewer bytes) so I didn’t bother. I’ll try it.

briansmith:
> For the w array, FIPS 180-4 suggested this but presented it as a trading speed (additional modulo takes time) for memory (256 fewer bytes) so I didn’t bother. I’ll try it.

I think that FIPS 180-4 text was written with an understanding of CPUs that is out of date. IIUC, the modulo 16 can be done with bitwise and so it's bascally free. If you have 80 elements then it seems impossible to vectorize and store the state in registers. If you have 16 elements then there's a chance, I think.

Anyway, I am OK with the current state. I'd love to see what is needed to get the Rust compiler to auto-vectorize this code but the benefit of that is more in learning how we'd need to code other stuff (e.g. SHA-256) than to make this not-needed-to-be-fast function faster.

PR #157, opened 2016-03-25
Remove code that supports AES-192.

### LINKED ISSUE #112: Remove all code that supports AES-192
See https://github.com/briansmith/ring/blob/c882c2c3ec837f14267991122dc20e6aba5700bc/crypto/aes/aes.c#L549-L551

and

https://github.com/briansmith/ring/blob/c882c2c3ec837f14267991122dc20e6aba5700bc/crypto/aes/aes.c#L549-L551

Also, I guess the assembly language code must have some bits that support 192-bit AES.

We don't need any of this since we don't (and don't plan to) expose AES-192.

Ms2ger:
Fixes #112.

### DIFF
--- a/crypto/aes/aes.c
+++ b/crypto/aes/aes.c
@@ -546,9 +546,6 @@ int AES_set_encrypt_key(const uint8_t *key, unsigned bits, AES_KEY *aeskey) {
     case 128:
       aeskey->rounds = 10;
       break;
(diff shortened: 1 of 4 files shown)


briansmith:
Awesome! Landed in fc99d3070ecce98d6487410a74ed53e6b4d47ebf.

PR #158, opened 2016-03-27
Keep lines under 90 chars and add a script to check them.

SimonSapin:
There is an exception for comment lines containing only an URL. More code changes would be needed to reach 80. The script is largely copied from Servo: https://github.com/servo/servo/blob/master/python/tidy.py

### DIFF
--- a/util/tidy.py
+++ b/util/tidy.py
@@ -1,0 +1,151 @@
+# Copyright 2013 The Servo Project Developers.
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import fnmatch
+import itertools
+import os
+import re
+
+URL_COMMENT = re.compile("^ *///? https?:\S+$")
+MAX_LINE_LENGTH = 90
+
+
+# File patterns to include in the tidy check.
+file_patterns_to_check = ["*.rs", "*.cpp", "*.c",
+                          "*.h", "Cargo.lock", "*.py",
+                          "*.toml"]
+
+# File patterns that are ignored for all tidy and lint checks.
+file_patterns_to_ignore = [
+    "*.#*",
+    "*.pyc",
+]
+
+# Files that are ignored for all tidy and lint checks.
+ignored_files = [
+    # Hidden files
+    os.path.join(".", "."),
+]
+
+# Directories that are ignored for the non-WPT tidy check.
+ignored_dirs = [
+    # Imported from BoringSSL
+    os.path.join(".", "crypto"),
+    os.path.join(".", "include", "openssl"),
+    # Hidden directories
+    os.path.join(".", "."),
+]
+
+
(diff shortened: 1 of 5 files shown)


frewsxcv:
This should maybe get checked by Travis?

SimonSapin:
Good point. Added another commit.

util/tidy.py
```diff
@@ -0,0 +1,152 @@
+# Copyright 2013 The Servo Project Developers.
```
briansmith: Sorry to be annoying, but I'm not sure I should accept this without knowing who the actual copyright holders are.

util/tidy.py
```diff
@@ -0,0 +1,152 @@
...
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
```
briansmith: I definitely am not interested in dual licensing with Apache 2.0. It would be better to just paste in the MIT license here, instead of this statement that references those two licenses.

util/tidy.py
SimonSapin: Output of `git shortlog -sn python/tidy.py` in the Servo repository is below. @bors-servo is a bot that only makes merge commits. The original file does lots more stuff, so probably not everyone listed has touched the parts kept in this PR. But if copyright/licensing is a problem, maybe it’d be easier to rewrite this script from scratch?

```
    21  Corey Farwell
    17  bors-servo
    11  Ms2ger
     5  Josh Matthews
     4  Simon Sapin
     3  Bogdan Cuza
     3  Brandon Fairchild
     3  Daan Sprenkels
     3  Dhananjay Nakrani
     3  Duncan Keall
     3  Manish Goregaokar
     3  Michael Howell
     3  Ravi Shankar
     3  wilmoz
     2  Anthony Ramine
     2  Bobby Holley
     2  Gilles Leblanc
     2  Guillaume Gomez
     2  Jack Moffitt
     2  João Oliveira
     2  Matt Brubeck
     2  PythonNut
     2  jmr0
     1  Attila Dusnoki
     1  Duncan
     1  Greg Guthe
     1  Hugo Thiessard
     1  Jaydeep
     1  Johann Tuffe
     1  Martin Robinson
     1  Mathieu Rheaume
     1  Mike Fairhurst
     1  OneKorg
     1  Paul Padier
     1  Pierre Chevalier
     1  Prabhjyot Singh Sodhi
     1  Rahul Sharma
     1  Simon Martin
     1  Till Schneidereit
     1  UK992
     1  erneyja

briansmith:
OK, I guess the next step here is for somebody (probably me) to go and clean up the code and/or modify the script so that the script can actually check what is intended to be checked--that the code is 80 columns wide (or approximately so, as described above). I will look into it when I have time.

briansmith:
It's very unlikely I will have time for this. If somebody else wants to pick it up and carry it across the finish line, then fine. Otherwise, we can just close it.

Note: I am OK with the license in the script as long as it stays separate from everyhting else.

briansmith:
I guess we're probably more likely to rely on rustfmt to do this, at least eventually. In any case, I don't see me making progress on this course of action and it's just making me sad to see it sitting around. Thanks for submitting this though. At the very least, it alerted me to the need to have something like this.

Issue #31, opened 2015-10-18
Remove all casts that use the `as` operator


briansmith:
Rust doesn't have implicit widening, so casts that are lossy (not necessarily safe) look just like casts that are 100% safe. Instead of using the `as` operator for widening casts, we should use (and create, if necessary) functions that safely do widening conversions, and use those. Then the use of `as` would be limited to only those functions.

This depends on #4.

Ms2ger:
`std::convert::From` is implemented for widening conversions in Rust nightly, so `let x: u8 = ...; let y: u16 = x.into();` should now work.

briansmith:
Thanks, Ms2ger. I'm looking forward to taking advantage of that.

See https://github.com/rust-lang/rfcs/pull/1218 and https://github.com/rust-lang/rust/pull/28921.

I left a comment https://github.com/rust-lang/rust/pull/28921#issuecomment-149427316 wherein I ask for additional cast-free conversions.

As part of fixing this issue, or in a follow-up, we should add a lint (static analysis) that verifies there are no uses of `as` in the code.

briansmith:
Fixed, at least as far as the current state of the code goes, in https://github.com/briansmith/ring/commit/377f6118593168d2252f9ed448b00e23a8170a56 and https://github.com/briansmith/ring/commit/a8276b2b9a27dcc5c8953955c9f6d0e33e7230a2. One line of test code from https://github.com/briansmith/ring/commit/377f6118593168d2252f9ed448b00e23a8170a56 casts `u16` to `usize`, but that will hopefully be removed if/when my Rust PR is accepted: https://github.com/rust-lang/rust/pull/29220.

briansmith:
This will become an issue against as we'll be casting `ring::der::OID as u8` for use in ``ring::input::Reader::peek``. :(

briansmith:
For the ones where implicit conversion doesn't work yet (e.g. conversions to/from `isize`), `into()` and `from()` might work:

```
fn main() {
    let x: u32 = 10;
    let y: u64 = x.into();
    let z: u64 = u64::from(x);
}
```

(from https://internals.rust-lang.org/t/the-problem-with-array-slice-vector-indexes/3129)

briansmith:
In sha1.rs there is now this:

```
let data = data as *const [u8; BLOCK_LEN];
```

We should make a function and put it in ring::polyfill to avoid using `as` directly in the code. It actually needs to be a macro until generics over integers are supported in Rust.

[...]

PR #159, opened 2016-03-28
Extract SHA-1’s cast with `as` into a dedicated function.

SimonSapin:
See https://github.com/briansmith/ring/issues/31#issuecomment-202311233

### DIFF
--- a/src/polyfill.rs
+++ b/src/polyfill.rs
@@ -146,3 +146,18 @@ macro_rules! slice_as_array_ref_mut {
         }
     }
 }
+
+/// Returns a slice of arrays starting at the given pointer.
+macro_rules! ptr_as_array_slice {
+    ($ptr:expr, $block_len:expr, $block_num:expr) => {
+        {
+            unsafe fn ptr_as_array_slice<'a, T>(ptr: *const T,
+                                                block_num: usize)
+                                                -> &'a [[T; $block_len]] {
+                let ptr = ptr as *const [T; $block_len];
+                core::slice::from_raw_parts(ptr, block_num)
+            }
+            ptr_as_array_slice($ptr, $block_num)
+        }
+    }
+}
--- a/src/digest/sha1.rs
+++ b/src/digest/sha1.rs
@@ -37,9 +37,7 @@ type W32 = Wrapping<u32>;
 pub unsafe extern fn block_data_order(state: &mut [u64; MAX_CHAINING_LEN / 8],
                                       data: *const u8,
                                       num: c::size_t) {
-    let data = data as *const [u8; BLOCK_LEN];
-    let blocks = core::slice::from_raw_parts(data, num);
-    block_data_order_safe(state, blocks)
+    block_data_order_safe(state, ptr_as_array_slice!(data, BLOCK_LEN, num))
 }
 
 fn block_data_order_safe(state: &mut [u64; MAX_CHAINING_LEN / 8],


src/polyfill.rs
```diff
@@ -146,3 +146,18 @@ macro_rules! slice_as_array_ref_mut {
         }
     }
 }
+
+/// Returns a slice of arrays starting at the given pointer.
+macro_rules! ptr_as_array_slice {
+    ($ptr:expr, $block_len:expr, $block_num:expr) => {
```
briansmith: It's probably helpful to check out my latest comment in the issue to understand the design goals for this submodule. In particular, it doesn't make sense for a feature of this submodule to take a pointer as a parameter because pointers are `unsafe` and the goal of this submodule is to define a purely-not-`unsafe` API implemented on top of `unsafe`. And, also, the goal is to do so in a way where the feature could reasonably be expected to make it into the Rust language or stdlib.

I actually started trying to do this but I had to switch to working on non-*ring* stuff before I finished it. Here's my in-progress attempt:
```
/// Chunks `$slice` into `$len`-length parts, returning (`chunks`, `leftover`),
/// where `chunks` is a subslice of `$slice` with elements of type `[T; $len]`
/// and `leftover` is a subslice of leftover elements.
macro_rules! slice_as_array_ref_chunks {
    ($slice:expr, $chunk_len:expr) => {
        {
            #[allow(unsafe_code)]
            fn slice_as_array_ref_chunks<'a, T>(slice: &'a [T])
                    -> (&'a [[T; $chunk_len]], &'a [T]) {
                let (to_chunk, leftover) = match slice.len() % $chunk_len {
                    n => slice.split_at(slice.len() - ($chunk_len - n))
                };
                let to_chunk = unsafe {
                    to_chunk.as_ptr() as *const [T; $chunk_len]
                };
                let chunks =
                    core::slice::from_raw_parts(to_chunk,
                                                to_chunk.len() / $chunk_len);
                (chunks, leftover)
            }
            slice_as_array_ref_chunks($slice)
        }
    }
}
```

SimonSapin: As long as `sha1::block_data_order` takes `data: *const u8, num: c::size_t`, calling this new API would require an unsafe `slice::from_raw_parts` call:

```
let data = core::slice::from_raw_parts(data, num * BLOCK_LEN);
let (blocks, rest) = slice_as_array_ref_chunks(data, BLOCK_LEN);
debug_assert!(rest.is_empty());
```

> do so in a way where the feature could reasonably be expected to make it into the Rust language or stdlib.

In my opinion this would be very tough sell for `slice_as_array_ref_chunks`. It’s a very narrow use case. Who knows how many slightly different functions like this would be needed to cover every use of `as` ever?

briansmith:
```
let data = core::slice::from_raw_parts(data, num * BLOCK_LEN);
```

Yes, that has to go into the `unsafe` function in the sha1 code because of this unusual aspect of *ring* where we're implementing a normally-performance-critical API in Rust that is almost exclusively implemented with assembly language code. Maybe in the future we can change things, e.g. changing the calling convention used to call the assembly language code, so that it is possible to pass the data in safe (non-pointer) types. I just don't know how to do it yet.

```
let (blocks, rest) = slice_as_array_ref_chunks(data, BLOCK_LEN);
debug_assert!(rest.is_empty());
```
These parts should go in the safe function.

As for this being a tough sell: In low-level crypto code, we are almost always breaking slices into fixed-sized chunks. So far in *ring* we don't implement low-level crypto in Rust, but I guess that will change and there will be lots of uses of it, at least within *ring*.

Anyway, this *ring* issue isn't the place to debate changing Rust or its stdlib. The goal of `ring::polyfill` is to gather data to inform those discussions, but they will happen elsewhere.

src/polyfill.rs
briansmith: Sorry to take so long to get back to you.

The purpose of the `array_ref` stuff is to basically work around the lack of generics over integer values. I don't expect the Rust lang/lib teams to ever implement `slice_as_array_ref_chunks` but I do expect people can agree that it is useful to have some kind of functionality like that.

Anyway, one goal of the polyfill module to to expose only an interface that is "safe" (doesn't require `unsafe`), so we should try to find some solution that achieves that goal. I am definitely open to alternatives to what I suggested above.

briansmith:
Hey, it seems like nothing's happening with this. Also, I've had to add some `as` uses in other places too. So, let's just close this and we'll deal with them all in the issue about getting rid of `as`.

PR #160, opened 2016-03-28
"Alternate method" for SHA-1

SimonSapin:
… with 16 elements in `w` instead of 80. See https://github.com/briansmith/ring/pull/156#issuecomment-202306263

`crypto-bench` results are slightly *worse* than before.

### DIFF
--- a/src/digest/sha1.rs
+++ b/src/digest/sha1.rs
@@ -49,38 +49,46 @@ fn block_data_order_safe(state: &mut [u64; MAX_CHAINING_LEN / 8],
     let state = &mut state[..CHAINING_WORDS];
     let state = slice_as_array_ref_mut!(state, CHAINING_WORDS).unwrap();
 
-    let mut w: [W32; 80] = [Wrapping(0); 80];
+    let mut w: [W32; 16] = [Wrapping(0); 16];
     for block in blocks {
-        for t in 0..16 {
-            let word = slice_as_array_ref!(&block[t * 4..][..4], 4).unwrap();
-            w[t] = Wrapping(polyfill::slice::u32_from_be_u8(word))
-        }
-        for t in 16..80 {
-            let wt = w[t - 3] ^ w[t - 8] ^ w[t - 14] ^ w[t - 16];
-            w[t] = polyfill::wrapping_rotate_left_u32(wt, 1);
-        }
-
         let mut a = state[0];
         let mut b = state[1];
         let mut c = state[2];
         let mut d = state[3];
         let mut e = state[4];
 
-        for t in 0..80 {
+        macro_rules! round {
+            ($k: expr, $f: expr, $w: expr) => {
+                {
+                    let tt = polyfill::wrapping_rotate_left_u32(a, 5) + $f + e +
+                             Wrapping($k) + $w;
+                    e = d;
+                    d = c;
+                    c = polyfill::wrapping_rotate_left_u32(b, 30);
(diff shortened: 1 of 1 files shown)


SimonSapin:
Example runs on x86_64 Linux. Before:

```
test digest::sha1::_1000      ... bench:       4,718 ns/iter (+/- 267) = 211 MB/s
test digest::sha1::_16        ... bench:         333 ns/iter (+/- 55) = 48 MB/s
test digest::sha1::_2000      ... bench:       9,290 ns/iter (+/- 814) = 215 MB/s
test digest::sha1::_256       ... bench:       1,499 ns/iter (+/- 89) = 170 MB/s
test digest::sha1::_8192      ... bench:      35,219 ns/iter (+/- 3,312) = 232 MB/s
test digest::sha1::block_len  ... bench:         593 ns/iter (+/- 49) = 107 MB/s
```

After:
```
test digest::sha1::_1000      ... bench:       5,275 ns/iter (+/- 354) = 189 MB/s
test digest::sha1::_16        ... bench:         362 ns/iter (+/- 42) = 44 MB/s
test digest::sha1::_2000      ... bench:      10,454 ns/iter (+/- 963) = 191 MB/s
test digest::sha1::_256       ... bench:       1,692 ns/iter (+/- 188) = 151 MB/s
test digest::sha1::_8192      ... bench:      42,058 ns/iter (+/- 2,882) = 194 MB/s
test digest::sha1::block_len  ... bench:         695 ns/iter (+/- 84) = 92 MB/s
```

src/digest/sha1.rs
```diff
@@ -49,38 +49,46 @@ fn block_data_order_safe(state: &mut [u64; MAX_CHAINING_LEN / 8],
...
-        }
-
         let mut a = state[0];
         let mut b = state[1];
         let mut c = state[2];
         let mut d = state[3];
         let mut e = state[4];
 
-        for t in 0..80 {
+        macro_rules! round {
```
briansmith: Does this really need to be a macro? It seems like it could be an inline function, albeit one that takes a lot of arguments. I'd rather avoid macros unless they are necessary for some purpose, in which case we should document why we use a macro instead of a function.

src/digest/sha1.rs
```diff
@@ -49,38 +49,46 @@ fn block_data_order_safe(state: &mut [u64; MAX_CHAINING_LEN / 8],
...
+            ($k: expr, $f: expr, $w: expr) => {
+                {
+                    let tt = polyfill::wrapping_rotate_left_u32(a, 5) + $f + e +
+                             Wrapping($k) + $w;
```
briansmith: It may be worth asking the rustc people for help. In particular, do they think there's any hope in the compiler auto-vectoring this code? If not, what would need to be changed for it to be auto-vectorizable? If auto-vectorization is impossible, is there a platform-neutral explicit vectorization we should use?

I suspect that in order for this to be vectorized, you'd need to target a platform that has PSHUFB or equivalent, and you may need to rewrite the rotations and shifts.

It would definitely be interesting if it can be done in a way that is a performance win. But, I don't htink it makes sense to make things more complicated and *also* make them slower.

SimonSapin:
> I don't htink it makes sense to make things more complicated and also make them slower.

Yes, that’s what I thought as well. I made this PR to show what it could look like, but I don’t think it should be merged.

There may be more to do here, but I’m not planning to pursue it myself.

Issue #72, opened 2016-01-05
Remove `RSA_decrypt` and related functions


briansmith:
See the comment in the code:

```
 * Passing |RSA_PKCS1_PADDING| into this function is deprecated and insecure. If
 * implementing a protocol using RSAES-PKCS1-V1_5, use |RSA_NO_PADDING| and then
 * check padding in constant-time combined with a swap to a random session key
 * or other mitigation. See "Chosen Ciphertext Attacks Against Protocols Based
 * on the RSA Encryption Standard PKCS #1", Daniel Bleichenbacher, Advances in
 * Cryptology (Crypto '98). */
```

briansmith:
This was done!

PR #165, opened 2016-04-14
Remove RSA_decrypt.

Ms2ger:
CC #72.


Ms2ger:
Not sure what related functions you meant; happy to do others as well.

include/openssl/rsa.h
```diff
@@ -311,7 +311,6 @@ struct rsa_st {
 #define RSA_R_DATA_TOO_SMALL_FOR_KEY_SIZE 117
 #define RSA_R_DIGEST_TOO_BIG_FOR_RSA_KEY 118
 #define RSA_R_D_E_NOT_CONGRUENT_TO_1 119
-#define RSA_R_EMPTY_PUBLIC_KEY 120
```
briansmith: I suggest we just drop this commit. These error codes aren't used in *ring*. instead, they exist only to make it easier to upstream code to BoringSSL. (Technically, they could all be removed in *ring* because `OPENSSL_PUT_ERROR` is a no-op in *ring*.)

briansmith:
Thanks! I landed the first two commits as e92cc0afab04fd059ab05ea956cd500b0eb8a0b9 and 02f39ead8a8178dd80d3720702120ef531476593. Let's just drop the third commit.

I don't even remember what else is "related" other than the padding function you also removed.

PR #166, opened 2016-04-14
Add clang 3.8 to linux builds

pietro:


briansmith:
Thanks for the patch!

This landed in ba0003a896397cb8dd833323e2cb6f77baf830d8.

I think you intended for the title of the PR to be the first line of the commit message, but instead the commit message was "Added to travis by travis-ci/apt-package-whitelist#561." which doesn't quite make sense. I fixed this before pushing the patch.

In the future, please include a statement such as "I agree to license my contributions to each file under the terms given at the top of each file I changed." in your commit message.

Thanks again!

PR #171, opened 2016-05-05
Build and run tests for ARM.

pietro:


briansmith:
Thanks. I committed two of your commits in this PR as f59aa7d7a9077e9200fd24d7d9549baad14cb644 and 043dfef73fda000c03c03f38f993179b0f4b8678. I also added my own 63b10be8e3005bd3745e969a54f2859697f01c84 to reduce some duplicated logic.

I didn't merge the travis_wait changes because, based on your comments, they don't solve the problem of the tests timing out. Let me know if I misunderstood.

Thanks a ton for doing this work!

pietro:
They actually do, but the tests that around 30 minutes. So using 30 minutes sometimes worked and sometimes didn't. That's why I ended up setting the limit to 60 minutes which is higher than the total build time allowed by travis.

PR #174, opened 2016-05-12
Test on armhf and use ubuntu's gcc packages.

pietro:
And use travis_wait on slow arm tests.


mk/travis.sh
```diff
@@ -21,30 +21,22 @@ printenv
 
 case $TARGET_X in
 aarch64-unknown-linux-gnu)
-  DL_TARGET=aarch64-linux-gnu
-  DL_DIGEST=b9137008744d9009877f662dbac7481d673cdcb1798e727e325a37c98a0f63da
+  sudo apt-get install --no-install-recommends g++-aarch64-linux-gnu gcc-aarch64-linux-gnu libc6-dev-arm64-cross -y
```
briansmith: Do we need to use `sudo apt-get install` to install these, or can we add them as package dependencies the normal way in update_travis_yml.py?

If these packages can only be installed using `sudo` then please add a note to that effect so people can understand why we're doing this a different way. Otherwise, let's switch to the declarative way.

mk/travis.sh
```diff
@@ -21,30 +21,22 @@ printenv
...
-  sudo apt-get install binfmt-support qemu-user-binfmt -y
-
-  DL_ROOT=https://releases.linaro.org/components/toolchain/binaries/
-  DL_RELEASE=5.1-2015.08
-  DL_BASENAME=gcc-linaro-$DL_RELEASE-x86_64_$DL_TARGET
-  wget $DL_ROOT/$DL_RELEASE/$DL_TARGET/$DL_BASENAME.tar.xz
-  echo "$DL_DIGEST  $DL_BASENAME.tar.xz" | sha256sum -c
-  tar xf $DL_BASENAME.tar.xz
-  export PATH=$PWD/$DL_BASENAME/bin:$PATH
+  sudo apt-get install --no-install-recommends binfmt-support qemu-user-binfmt -y
```
briansmith: ditto.

mk/travis.sh
```diff
@@ -21,30 +21,22 @@ printenv
...
-  DL_TARGET=aarch64-linux-gnu
-  DL_DIGEST=b9137008744d9009877f662dbac7481d673cdcb1798e727e325a37c98a0f63da
+  sudo apt-get install --no-install-recommends g++-aarch64-linux-gnu gcc-aarch64-linux-gnu libc6-dev-arm64-cross -y
+  export QEMU_LD_PREFIX=/usr/aarch64-linux-gnu
   ;;
-arm-unknown-linux-gnueabi)
-  DL_TARGET=arm-linux-gnueabi
-  DL_DIGEST=1c11a944d3e515405e01effc129f3bbf24effb300effa10bf486c9119378ccd7
+arm-unknown-linux-gnueabihf)
+  sudo apt-get install --no-install-recommends g++-arm-linux-gnueabihf gcc-arm-linux-gnueabihf libc6-dev-armhf-cross -y
```
briansmith: ditto.

briansmith:
Please add this to the end of your commit message:
```
```
Thanks!:

mk/travis.sh
pietro: >Do we need to use sudo apt-get install to install these, or can we add them as package dependencies the normal way in update_travis_yml.py?

These packages are whitelisted by travis and can be declared on .travis.yml. I can move them there if you prefer. As for update_travis_yml.py the arm/aarch64 are not defined there. Should I take a stab at defining them there?

briansmith: I suggest that we just move this to .travis.yml for now.

I would love to get a seperate PR for doing the ARM/Aarch64 stuff in update-travis-yml.py, though! I only hard-coded them in .travis.yml because I didn't have time to do it properly.

mk/travis.sh
pietro: These packages need sudo because we're using my backport that is not whitelisted on travis. I requested the whitelist on travis-ci/apt-source-whitelist#271. I'll add the comment for now and update it to use .travis.yml if/when the whitelist gets approved.

briansmith:
Great! This landed as 37892c52583220c4de8c802c02d3ce5a5ee4b330. Note that I redid the commit message a bit. Awesome contribution!

PR #175, opened 2016-05-12
Correctly set TARGET variables when building for Android.

pietro:
With this change I was able to compile ring using the Android SDK/NDK. I was able to build and run the tests using the docker container rust libc uses to test on android. Un fortunately the tests segfault like some of the other arm builds. I'll turn my android experimentation into a script and do a PR to add android to travis later.

### DIFF
--- a/mk/top_of_makefile.mk
+++ b/mk/top_of_makefile.mk
@@ -14,8 +14,12 @@
 
 .DEFAULT_GOAL := all
 
-# $(TARGET) must be of the form <arch>-<vendor>-<sys>-<abi>, except <abi> can
-# omitted on Mac OS X (Darwin).
+# $(TARGET) must be of the form <arch>[<sub>]-<vendor>-<sys>-<abi>.
+# Due to how Rust names platforms the following exceptions are allowed:
+# <abi> can be omitted for Mac OS X (Darwin) and iOS.
+# <vendor> can be omitted for the Android ABIs (android and androideabi).
+# The list of Rust supported platforms is at:
+# https://forge.rust-lang.org/platform-support.html
 TARGET_WORDS = $(subst -, ,$(TARGET))
 TARGET_ARCH_BASE = $(word 1,$(TARGET_WORDS))
 TARGET_ARCH_NORMAL = \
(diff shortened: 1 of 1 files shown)


pietro:
Well, back to reading about make conditionals. Also a similar change is needed to build on ios. I'll try to get my hands on a mac to test if setting the ABI to "macho" works.

mk/top_of_makefile.mk
```diff
@@ -27,10 +27,12 @@ TARGET_VENDOR = $(word 2,$(TARGET_WORDS))
 TARGET_SYS = $(word 3,$(TARGET_WORDS))
 TARGET_ABI = $(word 4,$(TARGET_WORDS))
 
-# Cargo doesn't pass the ABI as part of TARGET on Mac OS X.
+# Cargo doesn't pass the ABI as part of TARGET on Mac OS X or Android.
```
briansmith: If I understand your patch correctly, for Android it omits the *vendor* part, not the ABI. So, this comment should be updated.

mk/top_of_makefile.mk
```diff
@@ -27,10 +27,12 @@ TARGET_VENDOR = $(word 2,$(TARGET_WORDS))
...
 ifeq ($(TARGET_ABI),)
 ifeq ($(findstring apple-darwin,$(TARGET_VENDOR)-$(TARGET_SYS)),apple-darwin)
 TARGET_ABI = macho
+ifeq ($(findstring linux-androideabi,$(TARGET_VENDOR)-$(TARGET_SYS)),linux-androideabi)
+TARGET_ABI = androideabi
```
briansmith: We should set `TARGET_SYS` to `linux`. Maybe `TARGET_VENDOR` to `unknown`?

mk/top_of_makefile.mk
```diff
@@ -39,14 +41,16 @@ endef
 $(error TARGET must be of the form \
         <arch>[<sub>]-<vendor>-<sys>-<abi>.$(NEWLINE)\
 \
-\       Exceptions: <abi> defaults to "macho" on Mac OS X.\
-\
+        Exceptions: <abi> defaults to "macho" on Mac OS X \
+        and to "androideabi" on Android. $(NEWLINE)\
```
briansmith: Again, `<vendor>` defaults to `unknown` on Android, and the ABI is explicitly given.

briansmith:
Please add this to your commit messages:
```
<blank line>
```

Thanks!

mk/top_of_makefile.mk
pietro: Yeah, you're right and it's not exactly cargo, but how the rust team named the platforms it builds for. I'm working on the wording to explain this.

briansmith:
Awesome! This landed as d73194733209ffcc3704b79d71deee5375d6662b. Note that I wrapped the last line of the commit message at 72 characters.

Thanks!

PR #180, opened 2016-05-23
Improve and expand ECDH

briansmith:


briansmith:
I updated most of these changes and landed them in the master branch.

PR #182, opened 2016-05-23
Fix crash in ECDH

ctz:
We need to supply a `RAND` to `EC_KEY_generate_key_ex` now. There's a smoke test to exercise things.


briansmith:
I believe I just obviated the need for this with the new ECDH implementation. Please confirm.

briansmith:
Actually, no need to confirm, because I wrote a slew of tests for this. Thanks for the PR though! Note that the API changed slightly today so you may need to update your stuff a bit.

ctz:
Yup, that worked for me!

Issue #62, opened 2016-01-05
Replace C implementation of SHA-512 (and SHA-384) with a fast Rust implementation


briansmith:
The minimum requirement is that the Rust implementation is approximately the same speed as the BoringSSL C (OPENSSL_NO_ASM) implementation and that the code is clear and that the licensing is clear.

I noticed that the rust-crypto implementation of SHA-2 (SHA-256, SHA-384, and SHA-512) is actually faster than the BoringSSL C implementation. But, it is using the MIT/Apache dual license. I contacted the author of the optimized version about relicensing as ISC via email, but I never heard back. Maybe there are other alternatives.

Note that we only need the block function, and SHA-384 and SHA-512 use the same block function.

briansmith:
Also note that we already have an extensive test suite, so no new tests are needed.

briansmith:
It might be better to leave the C implementations of the transform functions and instead use formal methods to prove the C bits correct.

SimonSapin:
> The minimum requirement is that the Rust implementation is approximately the same speed as the BoringSSL C (OPENSSL_NO_ASM) implementation

In order to do this, how can one make sure `OPENSSL_NO_ASM` is enabled? Should it be exposed as a [Cargo feature](http://doc.crates.io/manifest.html#the-features-section)? (Or rather, since Cargo features are supposed to be additive, have an `asm` default feature and use `OPENSSL_NO_ASM` when `asm` is *not* enabled.)

briansmith:
This isn't the kind of thing that should be controlled by a cargo feature, except maybe temporarily on Windows since Windows has toolchain issues with the asm code.

The main problem with doing this work is that it makes little practical difference since the Rust code would almost never be run, as every major platform (32-bit ARM, AAarch64, x86, x86-64) has assembly language code for this. Also, OpenSSL upstream has asm code for more platforms. So, what's the point? How can we really evaluate the performance? Perhaps we should just code these like the SHA-1 implementation--prefer clarity over performance. Then, maybe we can experiment with using some interesting technique like using macros to unroll the code without changing its overall form (i.e. without a bunch of copy-pasta that manual unrolling entails). Or, maybe we should try to use the Rust code as a case study for the Rust compiler's optimizer? I think any of these options are reasonable.

My perspective on this is that, technically, there's not much benefit to doing this, but politically it makes people feel good about *ring* because it makes *ring* seem less like a Rust wrapper around C code and more like a Rust library that has assembly language optimizations.

[...]

PR #187, opened 2016-05-25
Implement sha256

gsoltis:
block_data_order algorithm ported from https://en.wikipedia.org/wiki/SHA-2

Let me know what you think! Aiming at https://github.com/briansmith/ring/issues/62

Is there a benchmark test already? Or does one need to be written?

### DIFF
--- a/src/digest/digest.rs
+++ b/src/digest/digest.rs
@@ -38,6 +38,7 @@ macro_rules! u32x2 {
 }
 
 mod sha1;
+mod sha256;
 
 /// A context for multi-step (Init-Update-Finish) digest calculations.
 ///
@@ -320,10 +321,10 @@ pub static SHA1: Algorithm = Algorithm {
 /// 180-4](http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).
 pub static SHA256: Algorithm = Algorithm {
     output_len: 256 / 8,
-    chaining_len: 256 / 8,
-    block_len: 512 / 8,
+    chaining_len: sha256::CHAINING_LEN,
+    block_len: sha256::BLOCK_LEN,
     len_len: 64 / 8,
-    block_data_order: sha256_block_data_order,
+    block_data_order: sha256::block_data_order,
     format_output: sha256_format_output,
     initial_state: [
         u32x2!(0x6a09e667u32, 0xbb67ae85u32),
@@ -455,7 +456,6 @@ pub extern fn SHA512_4(out: *mut u8, out_len: c::size_t,
 }
 
 extern {
-    fn sha256_block_data_order(state: &mut [u64; MAX_CHAINING_LEN / 8], data: *const u8, num: c::size_t);
     fn sha512_block_data_order(state: &mut [u64; MAX_CHAINING_LEN / 8], data: *const u8, num: c::size_t);
 }
 
--- a/src/digest/sha256.rs
+++ b/src/digest/sha256.rs
@@ -1,0 +1,126 @@
+use c;
+use core;
+use core::num::Wrapping;
+use polyfill;
+use super::MAX_CHAINING_LEN;
+
+pub const CHAINING_LEN: usize = 256 / 8;
+pub const BLOCK_LEN: usize = 512 / 8;
+const CHAINING_WORDS: usize = CHAINING_LEN / 4;
+
+type W32 = Wrapping<u32>;
+
+pub unsafe extern fn block_data_order(state: &mut [u64; MAX_CHAINING_LEN / 8],
+                                      data: *const u8,
+                                      num: c::size_t) {
+    let data = data as *const [u8; BLOCK_LEN];
+    let blocks = core::slice::from_raw_parts(data, num);
+    block_data_order_safe(state, blocks);
+}
+
+const K: [W32; 64] =
+    [Wrapping(0x428a2f98), Wrapping(0x71374491), Wrapping(0xb5c0fbcf), Wrapping(0xe9b5dba5),
+     Wrapping(0x3956c25b), Wrapping(0x59f111f1), Wrapping(0x923f82a4), Wrapping(0xab1c5ed5),
+     Wrapping(0xd807aa98), Wrapping(0x12835b01), Wrapping(0x243185be), Wrapping(0x550c7dc3),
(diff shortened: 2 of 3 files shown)


briansmith:
Thanks!

The benchmarks are in https://github.com/briansmith/crypto-bench. I suggest comparing your implementation to, in particular, the one in rust-crypto. IMO, if unless/until we have a clear plan for getting faster-than-rust-crypto (which is pretty fast) implementations, we should consider just using rust-crypto for the rare cases where we can't use the asm code.

Let me know the benchmark result. I'll find some time to look at the code soon.

gsoltis:
It looks like rust-crypto is about 2x faster than my implementation. So are you thinking to just declare a dependency on that library (and I guess remove the existing sha1 implementation)? Or just leave the asm implementation (which, if I understand, is generated by the perl scripts from crypto/sha/asm) ?

briansmith:
> It looks like rust-crypto is about 2x faster than my implementation. So are you thinking to just declare a dependency on that library (and I guess remove the existing sha1 implementation)? Or just leave the asm implementation (which, if I understand, is generated by the perl scripts from crypto/sha/asm) ?

Our goal is to always use the best implementation for each target.

There are four ASM implementations of each digest function: x86, x86-64, 32-bit ARM, and AAarch64. Actually, there are more, because within each implementation there are build-time (e.g. x86 has a SSE2 built-time switch) and/or runtime switches (based on CPU feature detection).

It might be the case that in some cases, for some target, the rust-crypto implementation is faster than the ASM implementation for that target. In that case, we should just delete the ASM code for that target and use rust-crypto instead. That has to be determined on a target-by-target basis.

There are some targets where OpenSSL has asm implementations but we don't have the asm implementation in *ring*. The primary case is MIPS. If the OpenSSL MIPS asm code is faster than rust-crypto (on MIPS), then we should import in the MIPS asm code from OpenSSL.

There are some targets for which we don't have asm code at all, e.g. WebAssembly. For these, *ring* currently just doesn't built. These targets are the ones where having a fast Rust implementation is important (OTOH, how important are these targets, really?).

As far as how to integrate a fast Rust implementation goes, for the cases where we don't already have an ASM implementation faster than what we can do in Rust, we have various options:
1. We can add a new implementation in *ring*, like you proposed here.
2. We can copy/paste the `block_data_order` function (which probably has a different name) from another Rust implementation, respecting its license. (In practice, this would mean choosing an ISC- or MIT- or BSD- licensed library to share the code with.)
3. We can add a dependency on another crypto library (only for targets where we don't implement everything internally.)

The difficulty with #3 is that usually crypto libraries don't expose the `block_data_order` type function that would be ideal for *ring*. Thus, #1 seems like the best option to me. However, #1 is a lot of work, because it is pretty hard to make a new implementation faster than every other Rust implementation. Thus, #2 might be the most pragmatic choice.

What do you think?

briansmith:
BTW, if you're interested in making your SHA-256 implementation 2x as fast, I'd be down to helping you with that too. Just know it will probably take some serious effort.

gsoltis:
Option 2 does sound the most pragmatic. 

I should mention, my own goals are to learn Rust, gain experience with native-code toolchains, and get more hands-on experience with the ssl protocol. So, I'm not super-confidant that I'll be able to get a 2x speedup out of my implementation, but I'll probably take a stab at doing at least some optimization, at least for the experience if nothing else. 

If there are other more straightforward issues you think would be good to tackle, I'd be happy to spend time on those as well, especially if the solution here is to use something that exists already. Otherwise, as I said, I'll probably at least take a first pass at trying to optimize my implementation a little.

gsoltis:
Quick update: I tried some limited manual unrolling in `block_data_order_safe`, from 64 iterations to 32, and that seemed to make it roughly 10% faster. So, this seems like a worthwhile avenue to pursue, and as you pointed out, seems to be what other implementations do.

gsoltis:
Unrolled all 64 rounds, in chunks of 4 via macro. A little less readable now, but significantly faster. Still not quite as fast as rust-crypto. I'll probably look at unrolling the `w` calculation next.

gsoltis:
Looks like the build issues are a problem w/ nightly? stable is passing...

briansmith:
Awesome!

If you have time, could you please split your change into two:
1. Replacing `Wrapping` with direct use of the `wrapping_xxx()` functions.
2. Your unrolling.

Now that your state `w` shouldn't need 64 entries. Almost all high-performance implementations use a 16-element `w`. This smaller state then makes register allocation, and sometimes auto-vectorization, easier.

I see that you defined your macro inside the function, but that isn't how macros work in Rust, at least as far as the namespacing goes. I recommend that you put the macro at the top level of the file.

gsoltis:
Reworked the commits a little, and unrolled `w`, using 16 slots, as suggested. I haven't cleaned up the macro placement yet. But, the benchmark is a lot closer now! The exact numbers vary a bit, but on my ~6 year old macbook pro, I get `rust-crypto` being somewhere around 10-15% faster. Much better than 2x faster.

briansmith:
Awesome. Being within 10-15% faster is quite reasonable. I'm pretty burned out today after working on a bunch of other stuff, but I'll look at this as soon as I get a chance.

samscott89:
I've got another optimisation which puts us within about 3% of the rust-crypto (on my machine at least). A simple trick to optimise the Sigma functions. I'll PR to this branch.

I get 213 MB/s -> 229 MB/s versus 235 MB/s for rust-crypto (on sha256::_1000).

Incidentally, if rust-crypto made the same change they would see a 20MB/s speedup...

gsoltis:
Ok, I've merged @samscott89's PR, which does make a difference. 

I looked at moving the macros out to the top level, but it appears that since they use local variables in the function, I need to modify them to take the identifiers for those variables. I'll add a commit to show you what I mean. If you have a preference on style, we can take that commit or not.

gsoltis:
re: last commit, note having to pass in `$block`, whereas before, it wasn't required. I don't have a strong opinion on the style either way, but if we go with moving the macros out, I'll have to rework the one remaining one.

briansmith:
Please see my latest comments in #199. Should I be reviewing the commits in this PR, or the ones in #199?

briansmith:
I am actually kind of unsure about what happened with this PR in relation to #199 but it seems like we all decided to work in #199. Sorry about any confusion! I am as confused as anybody, I think.

Regardless, thank you very much for the PR!

Issue #61, opened 2016-01-05
Replace C implementation of SHA-256 with a fast Rust implementation


briansmith:
The minimum requirement is that the Rust implementation is approximately the same speed as the BoringSSL C (OPENSSL_NO_ASM) implementation and that the code is clear and that the licensing is clear.

I noticed that the rust-crypto implementation of SHA-2 (SHA-256, SHA-384, and SHA-512) is actually faster than the BoringSSL C implementation. But, it is using the MIT/Apache dual license. I contacted the author of the optimized version about relicensing as ISC via email, but I never heard back. Maybe there are other alternatives.

briansmith:
Note that we only need the block function, as the Initialize-Update-Finalize code is already in Rust. Also note that we already have an extensive test suite, so no new tests are needed

briansmith:
It might be better to leave the C implementations of the transform functions and instead use formal methods to prove the C bits correct.

briansmith:
Work on this seems to be far along. Note that the C implementation is long gone.

briansmith:
We already removed the C implementation, though we didn't replace it, so I'm closing this. #199 will remain open while we investigate what to do with the pure Rust implementations.

PR #199, opened 2016-05-27
Implement SHA2 in Rust.

samscott89:
Code is simple adaptation of the pre-existing SHA1 code.
Algorithm implemented as specified in [FIPS 180-4].

This addresses #61 and #62.

### DIFF
--- a/src/digest/sha2.rs
+++ b/src/digest/sha2.rs
@@ -1,0 +1,231 @@
+// Copyright 2015-2016 Brian Smith.
+// Copyright 2016 Simon Sapin.
+// Copyright 2016 Sam Scott.
+//
+// Permission to use, copy, modify, and/or distribute this software for any
+// purpose with or without fee is hereby granted, provided that the above
+// copyright notice and this permission notice appear in all copies.
+//
+// THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHORS DISCLAIM ALL WARRANTIES
+// WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR
+// ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+// ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+// OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+use c;
+use core::num::Wrapping;
+use polyfill;
+
+use super::MAX_CHAINING_LEN;
+
+// SHA-256: 512-bit blocks.
+const BLOCK_LEN_256: usize = 512 / 8;
+
+// SHA-384, SHA-512: 1024-bit blocks.
+const BLOCK_LEN_512: usize = 1024 / 8;
+
+// SHA-256: state is 256 bits.
+pub const CHAINING_LEN_256: usize = 256 / 8;
+// SHA-384, SHA-512: state is 512 bits.
+pub const CHAINING_LEN_512: usize = 512 / 8;
+
+// Length of the state as number of words.
+const CHAINING_WORDS_256: usize = CHAINING_LEN_256 / 4;
+const CHAINING_WORDS_512: usize = CHAINING_LEN_512 / 8;
+
+type W32 = Wrapping<u32>;
+type W64 = Wrapping<u64>;
+
+macro_rules! ch {
+    ($x:expr, $y:expr, $z:expr) => (($x & $y) ^ (!$x & $z))
+}
+macro_rules! parity {
+    ($x:expr, $y:expr, $z:expr) => ($x ^ $y ^ $z)
+}
+macro_rules! maj {
+    ($x:expr, $y:expr, $z:expr) => (($x & $y) ^ ($x & $z) ^ ($y & $z))
+}
+macro_rules! rotr {
+    ($x:expr, $n:expr) => (Wrapping($x.0.rotate_right($n)))
+}
+
+// SHA256 sigma functions
+#[inline]
+fn big_s0_256(x: W32) -> W32   { rotr!(rotr!(rotr!(x, 9) ^ x, 11) ^ x, 2) }
+
+#[inline]
+fn big_s1_256(x: W32) -> W32   { rotr!(rotr!(rotr!(x, 14) ^ x, 5) ^ x, 6) }
+
+#[inline]
+fn small_s0_256(x: W32) -> W32 { rotr!((rotr!(x, 11) ^ x), 7)  ^ (x >> 3) }
+
+#[inline]
+fn small_s1_256(x: W32) -> W32 { rotr!((rotr!(x, 2)  ^ x), 17) ^ (x >> 10) }
+
+// SHA512 sigma functions
+#[inline]
+fn big_s0_512(x: W64) -> W64   { rotr!(rotr!(rotr!(x, 5) ^ x, 6) ^ x, 28) }
+
+#[inline]
+fn big_s1_512(x: W64) -> W64   { rotr!(rotr!(rotr!(x, 23) ^ x, 4) ^ x, 14) }
+
+#[inline]
+fn small_s0_512(x: W64) -> W64 { rotr!((rotr!(x, 7) ^ x), 1)  ^ (x >> 7) }
+
+#[inline]
+fn small_s1_512(x: W64) -> W64 { rotr!((rotr!(x, 42) ^ x), 19)  ^ (x >> 6) }
+
+pub fn block_data_order_256(state: &mut [u64; MAX_CHAINING_LEN / 8],
+                            data: &[u8],
+                            num: c::size_t) {
+
+    // Converts state to array of Wrapping<u32> values.
+    let state = polyfill::slice::u64_as_u32_mut(state);
+    let state = polyfill::slice::as_wrapping_mut(state);
+    let state = &mut state[..CHAINING_WORDS_256];
+    let state = slice_as_array_ref_mut!(state, CHAINING_WORDS_256).unwrap();
+
+    // Message schedule
+    let mut w: [W32; 64] = [Wrapping(0); 64];
+    for i in 0..num {
+        let block = &data[i * BLOCK_LEN_256..][..BLOCK_LEN_256];
+        for t in 0..16 {
+            let word = slice_as_array_ref!(&block[t * 4..][..4], 4).unwrap();
+            w[t] = Wrapping(polyfill::slice::u32_from_be_u8(word))
+        }
+        for t in 16..64 {
+            w[t] = small_s1_256(w[t - 2])  + w[t - 7]
+                 + small_s0_256(w[t - 15]) + w[t - 16];
+        }
+        let mut a = state[0];
+        let mut b = state[1];
+        let mut c = state[2];
+        let mut d = state[3];
+        let mut e = state[4];
+        let mut f = state[5];
+        let mut g = state[6];
+        let mut h = state[7];
+
+        for t in 0..64 {
+            let t1 = h + big_s1_256(e) + ch!(e, f, g)
+                       + Wrapping(K_256[t]) + w[t];
+
+            let t2 = big_s0_256(a) + maj!(a,b,c);
+            h = g;
+            g = f;
+            f = e;
+            e = d + t1;
+            d = c;
+            c = b;
+            b = a;
+            a = t1 + t2;
+        }
+        state[0] = a + state[0];
+        state[1] = b + state[1];
+        state[2] = c + state[2];
+        state[3] = d + state[3];
+        state[4] = e + state[4];
+        state[5] = f + state[5];
+        state[6] = g + state[6];
+        state[7] = h + state[7];
+    }
+}
+
+pub fn block_data_order_512(state: &mut [u64; MAX_CHAINING_LEN / 8],
+                            data: &[u8],
+                            num: c::size_t) {
+
+    // Converts state to array of Wrapping<u64> values.
+    let state = polyfill::slice::as_wrapping_mut(state);
+    let state = &mut state[..CHAINING_WORDS_512];
+    let state = slice_as_array_ref_mut!(state, CHAINING_WORDS_512).unwrap();
+
+    // Message schedule
+    let mut w: [W64; 80] = [Wrapping(0); 80];
+    for i in 0..num {
+        let block = &data[i * BLOCK_LEN_512..][..BLOCK_LEN_512];
+        for t in 0..16 {
+            let word = slice_as_array_ref!(&block[t * 8..][..8], 8).unwrap();
+            w[t] = Wrapping(polyfill::slice::u64_from_be_u8(word))
+        }
+        for t in 16..80 {
+            w[t] = small_s1_512(w[t - 2]) + w[t - 7]
+                 + small_s0_512(w[t - 15]) + w[t - 16];
+        }
+        let mut a = state[0];
+        let mut b = state[1];
+        let mut c = state[2];
+        let mut d = state[3];
+        let mut e = state[4];
+        let mut f = state[5];
+        let mut g = state[6];
+        let mut h = state[7];
+
+        for t in 0..80 {
+            let t1 = h + big_s1_512(e) + ch!(e, f, g) + Wrapping(K_512[t])
+                       + w[t];
+
+            let t2 = big_s0_512(a) + maj!(a,b,c);
+            h = g;
+            g = f;
+            f = e;
+            e = d + t1;
+            d = c;
+            c = b;
+            b = a;
+            a = t1 + t2;
+        }
+        state[0] = a + state[0];
+        state[1] = b + state[1];
+        state[2] = c + state[2];
+        state[3] = d + state[3];
+        state[4] = e + state[4];
+        state[5] = f + state[5];
+        state[6] = g + state[6];
+        state[7] = h + state[7];
+    }
+}
+
+// SHA256 constants K
+const K_256: [u32; 64] = [0x428a2f98, 0x71374491, 0xb5c0fbcf, 0xe9b5dba5,
+    0x3956c25b, 0x59f111f1, 0x923f82a4, 0xab1c5ed5, 0xd807aa98, 0x12835b01,
+    0x243185be, 0x550c7dc3, 0x72be5d74, 0x80deb1fe, 0x9bdc06a7, 0xc19bf174,
+    0xe49b69c1, 0xefbe4786, 0x0fc19dc6, 0x240ca1cc, 0x2de92c6f, 0x4a7484aa,
+    0x5cb0a9dc, 0x76f988da, 0x983e5152, 0xa831c66d, 0xb00327c8, 0xbf597fc7,
+    0xc6e00bf3, 0xd5a79147, 0x06ca6351, 0x14292967, 0x27b70a85, 0x2e1b2138,
+    0x4d2c6dfc, 0x53380d13, 0x650a7354, 0x766a0abb, 0x81c2c92e, 0x92722c85,
+    0xa2bfe8a1, 0xa81a664b, 0xc24b8b70, 0xc76c51a3, 0xd192e819, 0xd6990624,
+    0xf40e3585, 0x106aa070, 0x19a4c116, 0x1e376c08, 0x2748774c, 0x34b0bcb5,
+    0x391c0cb3, 0x4ed8aa4a, 0x5b9cca4f, 0x682e6ff3, 0x748f82ee, 0x78a5636f,
+    0x84c87814, 0x8cc70208, 0x90befffa, 0xa4506ceb, 0xbef9a3f7, 0xc67178f2];
+
+// SHA512 constants K
+const K_512: [u64; 80] = [0x428a2f98d728ae22, 0x7137449123ef65cd,
+      0xb5c0fbcfec4d3b2f, 0xe9b5dba58189dbbc, 0x3956c25bf348b538,
+      0x59f111f1b605d019, 0x923f82a4af194f9b, 0xab1c5ed5da6d8118,
+      0xd807aa98a3030242, 0x12835b0145706fbe, 0x243185be4ee4b28c,
+      0x550c7dc3d5ffb4e2, 0x72be5d74f27b896f, 0x80deb1fe3b1696b1,
+      0x9bdc06a725c71235, 0xc19bf174cf692694, 0xe49b69c19ef14ad2,
+      0xefbe4786384f25e3, 0x0fc19dc68b8cd5b5, 0x240ca1cc77ac9c65,
+      0x2de92c6f592b0275, 0x4a7484aa6ea6e483, 0x5cb0a9dcbd41fbd4,
+      0x76f988da831153b5, 0x983e5152ee66dfab, 0xa831c66d2db43210,
+      0xb00327c898fb213f, 0xbf597fc7beef0ee4, 0xc6e00bf33da88fc2,
+      0xd5a79147930aa725, 0x06ca6351e003826f, 0x142929670a0e6e70,
+      0x27b70a8546d22ffc, 0x2e1b21385c26c926, 0x4d2c6dfc5ac42aed,
+      0x53380d139d95b3df, 0x650a73548baf63de, 0x766a0abb3c77b2a8,
+      0x81c2c92e47edaee6, 0x92722c851482353b, 0xa2bfe8a14cf10364,
+      0xa81a664bbc423001, 0xc24b8b70d0f89791, 0xc76c51a30654be30,
+      0xd192e819d6ef5218, 0xd69906245565a910, 0xf40e35855771202a,
+      0x106aa07032bbd1b8, 0x19a4c116b8d2d0c8, 0x1e376c085141ab53,
+      0x2748774cdf8eeb99, 0x34b0bcb5e19b48a8, 0x391c0cb3c5c95a63,
+      0x4ed8aa4ae3418acb, 0x5b9cca4f7763e373, 0x682e6ff3d6b2b8a3,
+      0x748f82ee5defb2fc, 0x78a5636f43172f60, 0x84c87814a1f0ab72,
+      0x8cc702081a6439ec, 0x90befffa23631e28, 0xa4506cebde82bde9,
+      0xbef9a3f7b2c67915, 0xc67178f2e372532b, 0xca273eceea26619c,
+      0xd186b8c721c0c207, 0xeada7dd6cde0eb1e, 0xf57d4f7fee6ed178,
+      0x06f067aa72176fba, 0x0a637dc5a2c898a6, 0x113f9804bef90dae,
+      0x1b710b35131c471b, 0x28db77f523047d84, 0x32caab7b40c72493,
+      0x3c9ebe0a15c9bebc, 0x431d67c49c100d4c, 0x4cc5d4becb3e42b6,
+      0x597f299cfc657e2a, 0x5fcb6fab3ad6faec, 0x6c44198c4a475817];
--- a/src/digest/digest.rs
+++ b/src/digest/digest.rs
@@ -39,6 +39,9 @@ macro_rules! u32x2 {
 
 mod sha1;
 
+#[cfg(feature="no_asm")]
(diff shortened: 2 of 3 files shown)


samscott89:
I had a go at writing some simple code for the SHA2 family. I know the original issues specified *fast* implementations, so these might not cut it. Roughly 2x slower than previous version and bit slower than rust-crypto.

Timings on x86_64:
ring before:
```
test digest::sha256::_1000               ... bench:       2,245 ns/iter (+/- 43) = 445 MB/s  
test digest::sha256::_16                 ... bench:         187 ns/iter (+/- 6) = 85 MB/s  
test digest::sha256::_2000               ... bench:       4,433 ns/iter (+/- 83) = 451 MB/s  
test digest::sha256::_256                ... bench:         739 ns/iter (+/- 12) = 346 MB/s  
test digest::sha256::_8192               ... bench:      17,679 ns/iter (+/- 11,560) = 463 MB/s  
test digest::sha256::block_len           ... bench:         346 ns/iter (+/- 23) = 184 MB/s  
test digest::sha384::_1000               ... bench:       1,616 ns/iter (+/- 26) = 618 MB/s  
test digest::sha384::_16                 ... bench:         269 ns/iter (+/- 5) = 59 MB/s  
test digest::sha384::_2000               ... bench:       3,123 ns/iter (+/- 58) = 640 MB/s  
test digest::sha384::_256                ... bench:         661 ns/iter (+/- 12) = 387 MB/s  
test digest::sha384::_8192               ... bench:      12,203 ns/iter (+/- 8,610) = 671 MB/s  
test digest::sha384::block_len           ... bench:         498 ns/iter (+/- 7) = 257 MB/s  
test digest::sha512::_1000               ... bench:       1,614 ns/iter (+/- 188) = 619 MB/s  
test digest::sha512::_16                 ... bench:         266 ns/iter (+/- 13) = 60 MB/s  
test digest::sha512::_2000               ... bench:       3,109 ns/iter (+/- 35) = 643 MB/s  
test digest::sha512::_256                ... bench:         658 ns/iter (+/- 16) = 389 MB/s  
test digest::sha512::_8192               ... bench:      12,214 ns/iter (+/- 201) = 670 MB/s  
test digest::sha512::block_len           ... bench:         498 ns/iter (+/- 12) = 257 MB/s  
  
```
Ring after:
```
test digest::sha256::_1000               ... bench:       5,221 ns/iter (+/- 184) = 191 MB/s  
test digest::sha256::_16                 ... bench:         359 ns/iter (+/- 16) = 44 MB/s  
test digest::sha256::_2000               ... bench:      10,409 ns/iter (+/- 440) = 192 MB/s  
test digest::sha256::_256                ... bench:       1,668 ns/iter (+/- 36) = 153 MB/s  
test digest::sha256::_8192               ... bench:      41,795 ns/iter (+/- 1,400) = 196 MB/s  
test digest::sha256::block_len           ... bench:         691 ns/iter (+/- 22) = 92 MB/s  
test digest::sha384::_1000               ... bench:       3,365 ns/iter (+/- 96) = 297 MB/s  
test digest::sha384::_16                 ... bench:         455 ns/iter (+/- 27) = 35 MB/s  
test digest::sha384::_2000               ... bench:       7,014 ns/iter (+/- 675) = 285 MB/s  
test digest::sha384::_256                ... bench:       1,357 ns/iter (+/- 105) = 188 MB/s  
test digest::sha384::_8192               ... bench:      27,777 ns/iter (+/- 1,394) = 294 MB/s  
test digest::sha384::block_len           ... bench:         883 ns/iter (+/- 29) = 144 MB/s  
test digest::sha512::_1000               ... bench:       3,403 ns/iter (+/- 205) = 293 MB/s  
test digest::sha512::_16                 ... bench:         455 ns/iter (+/- 50) = 35 MB/s  
test digest::sha512::_2000               ... bench:       6,771 ns/iter (+/- 937) = 295 MB/s  
test digest::sha512::_256                ... bench:       1,305 ns/iter (+/- 1,058) = 196 MB/s  
test digest::sha512::_8192               ... bench:      27,661 ns/iter (+/- 1,363) = 296 MB/s  
test digest::sha512::block_len           ... bench:         880 ns/iter (+/- 45) = 145 MB/s  
```

rust-crypto:
```
test digest::sha256::_1000               ... bench:       4,270 ns/iter (+/- 660) = 234 MB/s  
test digest::sha256::_16                 ... bench:         294 ns/iter (+/- 10) = 54 MB/s  
test digest::sha256::_2000               ... bench:       8,632 ns/iter (+/- 898) = 231 MB/s  
test digest::sha256::_256                ... bench:       1,379 ns/iter (+/- 88) = 185 MB/s  
test digest::sha256::_8192               ... bench:      34,263 ns/iter (+/- 2,441) = 239 MB/s  
test digest::sha256::block_len           ... bench:         562 ns/iter (+/- 18) = 113 MB/s  
test digest::sha384::_1000               ... bench:       2,789 ns/iter (+/- 139) = 358 MB/s  
test digest::sha384::_16                 ... bench:         367 ns/iter (+/- 14) = 43 MB/s  
test digest::sha384::_2000               ... bench:       5,547 ns/iter (+/- 243) = 360 MB/s  
test digest::sha384::_256                ... bench:       1,057 ns/iter (+/- 32) = 242 MB/s  
test digest::sha384::_8192               ... bench:      32,851 ns/iter (+/- 10,469) = 249 MB/s  
test digest::sha384::block_len           ... bench:       1,048 ns/iter (+/- 268) = 122 MB/s  
test digest::sha512::_1000               ... bench:       4,090 ns/iter (+/- 99) = 244 MB/s  
test digest::sha512::_16                 ... bench:         380 ns/iter (+/- 175) = 42 MB/s  
test digest::sha512::_2000               ... bench:       5,568 ns/iter (+/- 298) = 359 MB/s  
test digest::sha512::_256                ... bench:       1,067 ns/iter (+/- 23) = 239 MB/s  
test digest::sha512::_8192               ... bench:      22,429 ns/iter (+/- 902) = 365 MB/s  
test digest::sha512::block_len           ... bench:         724 ns/iter (+/- 52) = 176 MB/s  
 ```

If you have any suggestions for optimisations, I'm happy to give them a go.

**Comment by samscott89** [comment]
Haha, #187  wasn't there a few days ago when I first looked for something to solve 😄 

Well, I was looking for something to get my hands dirty anyway.

File: src/digest/sha2.rs [hunk 1]
```diff
@@ -0,0 +1,248 @@
...
+
+// SHA256 functions
+#[inline] fn big_s0_256(x: W32) -> W32   { x.rotr(2)  ^ x.rotr(13) ^ x.rotr(22) }
+#[inline] fn big_s1_256(x: W32) -> W32   { x.rotr(6)  ^ x.rotr(11) ^ x.rotr(25) }
+#[inline] fn small_s0_256(x: W32) -> W32 { x.rotr(7)  ^ x.rotr(18) ^ (x >> 3) }
+#[inline] fn small_s1_256(x: W32) -> W32 { x.rotr(17) ^ x.rotr(19) ^ (x >> 10) }
+
+// SHA512 functions
+#[inline] fn big_s0_512(x: W64) -> W64   { x.rotr(28)  ^ x.rotr(34) ^ x.rotr(39) }
+#[inline] fn big_s1_512(x: W64) -> W64   { x.rotr(14)  ^ x.rotr(18) ^ x.rotr(41) }
```
briansmith: NIT: your alignment is messed up here.

src/digest/sha2.rs
```diff
@@ -0,0 +1,248 @@
...
+        state[4] = e + state[4];
+        state[5] = f + state[5];
+        state[6] = g + state[6];
+        state[7] = h + state[7];
+    }
+}
+
+pub unsafe extern fn sha512_block_data_order(state: &mut [u64; MAX_CHAINING_LEN / 8],
+                                      data: *const u8,
+                                      num: c::size_t) {
```
briansmith: NIT: alignment is off.

src/digest/sha2.rs
```diff
@@ -0,0 +1,248 @@
...
+    let data = data as *const [u8; BLOCK_LEN_512];
+    let blocks = core::slice::from_raw_parts(data, num);
+    sha512_block_data_order_safe(state, blocks)
+}
+
+fn sha512_block_data_order_safe(state: &mut [u64; MAX_CHAINING_LEN / 8],
+                         blocks: &[[u8; BLOCK_LEN_512]]) {
```
briansmith: ditto

src/digest/sha2.rs
```diff
@@ -0,0 +1,248 @@
...
+
+use c;
+use core;
+use core::num::Wrapping;
+use polyfill;
+use std::ops::{Add,BitAnd,BitXor,Shr,Not};
+
+use super::MAX_CHAINING_LEN;
+
+// SHA 224/256: 512-bit blocks
```
briansmith: NIT: Don't mention 224, since we don't implement it.
NIT: Use "SHA-" instead of "SHA " in these comments (also below).

src/digest/sha2.rs
```diff
@@ -0,0 +1,248 @@
...
+    #[inline]
+    fn rotr(self, n: u32) -> Self;
+}
+
+impl Word for W32 {
+    #[inline]
+    fn rotr(self, n: u32) -> W32 {
+        Wrapping(self.0.rotate_right(n))
+    }
+}
```
briansmith: I guess if we do it this way, we should update the SHA-1 code to follow the same pattern, right?

briansmith:
Thanks for submitting this!

The duplication of effort with #187 is unfortunate. Maybe you and @gsoltis can work together to determine a path forward?

Like I said in #187, since the rust-crypto code is faster, I think we need to choose between (a) adding more commits that optimize the code further (either the code in #187 or the code here), or (b) simply use the rust-crypto implementation of the block function, if we don't think we can implement something that is as efficient as rust-crypto.

Also, unless/until the pure-Rust implementation is as fast as the assembly-language implementations in *ring*, we'll continue to use the assembly-language implementations on whatever platforms we have faster assembly language implementations for. (Ideally, we'd figure out a way to do a pure Rust implementation that is faster than the ASM code, but at least on x86-64 it looks to be not even close. Maybe on ARM the story is different.) This means, in particular, that the final PR needs to include a commit that uses ![cfg()] to choose between using an asm implementation or the pure-Rust implementation based on what's available for that platform. (If we can't benchmark on every platform yet, then we can just assume that the asm implementation is faster, when an asm implementation is available, as the OpenSSL team has spent a lot of effort optimizing the asm code on many different platforms.)

gsoltis:
I did a little bit of profiling of my implementation, and it looks like the wrapping bits are showing up the most. I'm going to look at reducing usage of Wrapping to see if that makes a difference. But, failing that, I think the conditional include of the ASM implementation or rust-crypto is probably that way to go.

gsoltis:
I did a little more profiling, and it seems like `Wrapping` is not itself the issue. Or rather, it costs about the same as `.wrapping_add`. Nearly all the time is spent in `.wrapping_add` and `.rotate_right` for `core::num`. My guess is that any optimizations will have to center around doing fewer of those calls, or doing them more efficiently.

briansmith:
I think it is probably worth studying what other implementations do, regarding reducing the state size (I think) and regarding unrolling. Most implementations look much different than this because they are partially or often fully unrolled, which allows for some really important operations. Also, reformulating the structure of the implementation might enable some auto-vectorization to happen.

You're not going to reduce the number of additions or rotations by any significant amount, because they are fundamental to the algorithm.

samscott89:
Will have a look at implementing these suggestions to improve performance. And, ultimately, add the cfg flags where necessary.

samscott89:
Played around with some optimisations. The commits above made the following changes:
```
test digest::sha256::_1000
-----------------------

Before:                    191 MB/s
Alternate Sigma functions: 200 MB/s
Remove Wrapping:           194 MB/s
Loop unrolling:            218 MB/s
Add Wrapping:              213 MB/s

asm no extensions:         307 MB/s
asm sse3:                  386 MB/s
asm avx:                   406 MB/s
asm shaext:                462 MB/s
```

Interestingly, removing `Wrapping` initially made performance worse, but after unrolling the loops a little it is better to remove it.

At the bottom I've shown the times for the included openssl/boringssl asm when disabling different extensions. I assume rust has limited access to those instructions, so is the best we could hope to achieve the top line?

samscott89:
With this last push, performance for SHA-256 is now equivalent (on my machine) to rust-crypto. Whereas the SHA-512 performance is actually better (by just over 5%).

However, as I mentioned in the other PR comments, if rust-crypto were to use the optimised Sigma variants, they would again be ahead.

briansmith:
> With this last push, performance for SHA-256 is now equivalent (on my machine) to rust-crypto. Whereas the SHA-512 performance is actually better (by just over 5%).
> 
> However, as I mentioned in the other PR comments, if rust-crypto were to use the optimised Sigma variants, they would again be ahead.

I really appreciate the benchmarking done here and the evidence-based optimization work to get this into awesome shape.

I'm a little bit unsure what's going on with this PR and the other PR, #187. Did we decide to do everything in this PR (#199)?

It seems like we'll still need to add the conditional logic to use the assembly-language implementations on the platforms that we have them for, right?

The main goal I had when I filed the issue for adding a Rust SHA-2 implementation was to provide a fallback for platforms where we don't have asm implementations. Later, I got the idea we might just use rust-crypto for everything, so I investigated that path a bit. That path seems attractive because it would eliminate some duplication of effort. But, rust-crypto does a lot of stuff we intentionally avoid doing in *ring*, and also its license isn't as nice as the ISC-style license we're using in *ring*. Also, I contacted the rust-crypto author about splitting out the "core" of rust-crypto into a library we might share, but understandably that's kind of inconvenient for everybody too.

Anyway, investigating those things in the background was why I was slow to get back to this. Now it seems like we should go ahead and just add our own SHA-2, Poly1305, ChaCha20, etc. implementations.

The remaining thing I'm uncertain about is this: How are we going to test this? On Travis CI we have four architectures: x86, x86-64, ARM, and Aarch64. And, we have asm implementations for every one. We could pick one platform--probably x86-64, and modify travis.sh so that it rebuilds *ring* on that platform with the ASM implementations disabled, and then re-tests. Or, we could try adding MIPS to Travis CI via qemu like we did for ARM. Since we don't have any asm code for MIPS, that port would then serve as the platform for testing this. I would prefer the latter (adding MIPS), but I'm also fine with the former (adjusting the travis script for x86-64 to also test this.)

samscott89:
The reason I pushed some changes to the other PR branch, was because there was an incredibly simple optimisation trick which improved performance significantly, so I thought it would be interesting to see if it had the same effect for the other implementation.

I'll continue pushing things along in this branch and if @gsoltis is okay with it we'll just keep things here. 

Personally I'm a big fan of the approach and philosophy you have in this project, so I think it is definitely worth writing new implementations. I would also argue there is value in having pure rust implementations of these algorithms for auditing purposes.  For a ~50% performance penalty, some applications may consider it worthwhile to use the simpler code.

That being said, I'll have a play with qemu because I'm interested to see whether the optimisations I made were only effective because it allowed the compiler to leverage some of the more modern instructions which might not be available to other architectures anyway. In which case the naive code might be preferable?

samscott89:
I've had a go at getting a mips build (mipsel technically) up and running. Seems *mostly* straightforward, with the usual hiccups. 

The main difficulty was the large amount of code which assumes there is some form of C/asm external functions.

I've created a `no_asm` feature in [this branch](https://github.com/samscott89/ring/tree/mips_build), along with what appear to be the required changes. For the sake of getting it to compile, I've left the function bodies for `[cfg(feature = "no_asm")]` with `unimplemented!()` to verify the rest of the library will compile/test successfully. 

Given that, you can see that this travis build works as here: https://travis-ci.org/samscott89/ring/builds/145933431

Unfortunately, the large number of failing tests results in the Travis build failing by exceeding the maximum log size. This could either be temporarily hidden for the mips build, or would hopefully not be an issue any more after this PR is merged, with passing SHA tests (the lack of a SHA2 implementation produces a lot of the output).

Should I merge recent changes, clean that branch up and submit it as a new PR, or should I just roll it all into this PR?

briansmith:
> I've created a no_asm feature in this branch, along with what appear to be the required changes. For the sake of getting it to compile, I've left the function bodies for [cfg(feature = "no_asm")] with unimplemented!() to verify the rest of the library will compile/test successfully.

I guess it would be better to create a separate PR that changes lib.rs so that all the submodules that depend on asm code are `#[cfg(not(feature = "no_asm"))] pub mod <module name>;`. That is, just skip all the submodules for MIPS. Then we can get the build/dest infrastructure working.

Then, we can come back and rebase this PR on top of that one, and then enable the `digest` submodule using the Rust code.

Does that seem like a good plan?

briansmith:
Also, good work!

samscott89:
Coming back to what I said in a previous comment:

>That being said, I'll have a play with qemu because I'm interested to see whether the optimisations I made were only effective because it allowed the compiler to leverage some of the more modern instructions which might not be available to other architectures anyway. In which case the naive code might be preferable?

Using the work in PR #256 I've had a look at the numbers. These should be taken with a pinch of salt, since it's using emulated binaries  run inside an emulator... But in any case:

1. The first (naive) implementation of SHA2 (currently commit  [396cdbb](https://github.com/briansmith/ring/pull/199/commits/396cdbb33f3f98a84870d4cd29ac13314e7717cf):
```
test digest::sha256::_1000     ... bench:      15,905 ns/iter (+/- 380) = 62 MB/s
test digest::sha256::_16       ... bench:       1,953 ns/iter (+/- 35) = 8 MB/s
test digest::sha256::_2000     ... bench:      30,620 ns/iter (+/- 761) = 65 MB/s
test digest::sha256::_256      ... bench:       5,761 ns/iter (+/- 157) = 44 MB/s
test digest::sha256::_8192     ... bench:     119,075 ns/iter (+/- 3,035) = 68 MB/s
test digest::sha256::block_len ... bench:       2,991 ns/iter (+/- 93) = 21 MB/s
```

2. The optimised variant using unrolling etc:
```
test digest::sha256::_1000     ... bench:      14,515 ns/iter (+/- 228) = 68 MB/s
test digest::sha256::_16       ... bench:       1,836 ns/iter (+/- 31) = 8 MB/s
test digest::sha256::_2000     ... bench:      27,712 ns/iter (+/- 488) = 72 MB/s
test digest::sha256::_256      ... bench:       5,279 ns/iter (+/- 136) = 48 MB/s
test digest::sha256::_8192     ... bench:     108,315 ns/iter (+/- 2,641) = 75 MB/s
test digest::sha256::block_len ... bench:       2,768 ns/iter (+/- 58) = 23 MB/s
```

3. rust-crypto
```
test digest::sha256::_1000     ... bench:      16,096 ns/iter (+/- 305) = 62 MB/s
test digest::sha256::_16       ... bench:       1,670 ns/iter (+/- 60) = 9 MB/s
test digest::sha256::_2000     ... bench:      31,360 ns/iter (+/- 730) = 63 MB/s
test digest::sha256::_256      ... bench:       5,546 ns/iter (+/- 115) = 46 MB/s
test digest::sha256::_8192     ... bench:     124,101 ns/iter (+/- 1,901) = 66 MB/s
test digest::sha256::block_len ... bench:       2,682 ns/iter (+/- 75) = 23 MB/s
```

So even the naive implementation has pretty much equivalent performance to the rust-crypto one, and not far off the optimised variant. Suggesting that perhaps we could stick with a straightforward, more legible implementation?

briansmith:
> So even the naive implementation has pretty much equivalent performance to the rust-crypto one, and not far off the optimised variant. Suggesting that perhaps we could stick with a straightforward, more legible implementation?

In that case, I suggest that we keep the fully-unrolled version in a branch somewhere so we don't lose it, and commit the smaller version. But, we should still take the optimizations in "Use alternative Sigma{0,1} functions," right?

In particular, I guess the non-asm code is likely to be used for unusual targets like WebAssembly. And, I think unrolling and anything that increases code size or complexity is likely to be a net loss for WebAssembly and probably other targets.

If you agree, let me know, and I'll do the review on the basis that we're aiming to commit (only) everything up to "Use alternative Sigma{0,1} functions."

Thanks for doing these measurements! Really cool work here.

samscott89:
Ok, that should do it. I have tests passing on my machine and on qemu when I merge #256 followed by this.

briansmith:
OK, I had a serious look at this. Before we dive deep into it, do you think any of these transformations are a good idea?: https://github.com/briansmith/ring/tree/sha2

samscott89:
My first thought was that they all looked to be sensible modifications. However, according to the benchmarks, splitting the loop is a good idea, but the refactoring of the message schedule to only use 16 values hurts performance. Perhaps the extra arithmetic needed for the indices isn't worth it. 

Splitting the loop (commits  https://github.com/briansmith/ring/commit/08d0c4557a1b11acfe9634a66353ec44596546b7 ... https://github.com/briansmith/ring/commit/73f282f9c3b45139d63fd9ad7d496164756f7479) gives you a ~5% speedup but shrinking the message schedule (commit https://github.com/briansmith/ring/commit/2f18b5ba0f4fc83f2e3590e6983e1cec688a3dd1) ends up losing a full 10% (from the base case, i.e. over 15% overall).

One minor thing: I get that Cargo features are supposed to be additive so the syntax of `asm` makes more sense over `no_asm`. However, I would expect that if multiple packages used *ring* and one specified not to use asm code, then the overall preference would be to not use asm?

briansmith:
> but the refactoring of the message schedule to only use 16 values hurts performance. Perhaps the extra arithmetic needed for the indices isn't worth it.

Most likely, it's inhibiting vectorization. I guess we won't do that then. Thanks for measuring it.

briansmith:
> One minor thing: I get that Cargo features are supposed to be additive so the syntax of asm makes more sense over no_asm. However, I would expect that if multiple packages used ring and one specified not to use asm code, then the overall preference would be to not use asm?

Let's say there's a program P that uses crate C that uses *ring*, and the crate. Let's say that C's dependency on *ring* is just a normal dependency using the defaults. If we have "no_asm" then P to can depend on *ring* with the "no_asm" feature and then the build will work as long as C doesn't use any features that are disabled by "no_asm". If we have "asm" instead then would there be a way for "P" to get the same effect? I don't think so, but I'm not sure. So, this seems like evidence in favor of "no_asm".

On the other hand, with "asm" crates could, in theory, advertise that they work in the no-asm mode in their Cargo.toml, but with "no_asm" there's no such way. This seems like a minor benefit of "asm" though.

On the other other hand, let's say that we have "no_asm" and then C depends on *ring* with "no_asm" enabled. Then P and any other crates that P uses won't be able to use any of the features that require the asm optimizations, even if they would work! Thus, if we go with the "no_asm" design we would have to document that it shouldn't be used by libraries, only by the top-level program.

It seems like it will be common for library crates to ignore the asm/no_asm issue and so we should do whatever lets the top-level program override the default choice. If there's a way to do that overriding with "asm" then I think that's best; otherwise, I think "no_asm" is best.

samscott89:
> It seems like it will be common for library crates to ignore the asm/no_asm issue and so we should do whatever lets the top-level program override the default choice. If there's a way to do that overriding with "asm" then I think that's best; otherwise, I think "no_asm" is best.

Agreed. 

One particularly ugly option would be use "asm" as a default feature and add a further "no_asm_override" feature or something which when activated would take precedence over the "asm" feature. This would require `#[cfg(all(feature = "asm," not(feature = "no_asm_override")))]` 

At which point, the question would be why is there an asm option in the first place.

Another option would be to just change the syntax slightly. So as opposed to "no_asm", it could be "use_rust_code" .

briansmith:
Please check out https://github.com/briansmith/ring/tree/sha2-3.

I dropped my proposed no_asm -> asm change. I also dropped all of my proposed optimizations, including the splitting. It isn't clear to me why the splitting would make things so much faster, unless it is purely due to saving some load-stores or allowing better reordering of operations. In any case, if the splitting is really faster, we can do it later.

I did some experimenting with trying to reduce the duplicate code, without decreasing performance. Could you check which changes in the sha2-3 branch decrease performance? You mentioned earlier that using `Wrapping` is somehow actually faster than using `wrapping_add`. Is Could you double-check that? There should be no reason for it. Using `Wrapping` makes it harder to create trait-based functions that work over both `u32` and `u64`, so I'd like to switch (back) to `wrapping_add` if it doesn't hurt performance.

I suggested adding a dependency on the num-traits crate so that we can write generic functions. I prefer to write generic functions rather than copy-paste if we can do so without hurting performance. I'm not that excited on depending on an external crate just for this; maybe in the future we can create a trait similar (and simpler than) `PrimInt` in `ring::polyfill` so we can remove this dependency.

You said that using the precomputed tables is significantly faster, right? If so, we might consider using such a precomputed table for SHA-1 too, as the slow SHA-1 is actually a huge slowdown when it comes to running the test suite.

Anyway, I think this is getting close to landing.

samscott89:
I think the use of traits kills performance. Commit https://github.com/briansmith/ring/commit/644d870f0cd99f394e990672a97bd3310f15ccc3 slows everything down to barely 1MB/s.

Previously I've tested inline functions vs macros and there didn't seem to be any difference, so I can only assume it's the trait object which is causing the slowdown. Replacing the trait functions with macros in the head of your branch recovers performance again as I've done here: https://github.com/samscott89/ring/commit/cb8752fe81e2076211520b793b93c06236e4cf3a.

> You said that using the precomputed tables is significantly faster, right? If so, we might consider using such a precomputed table for SHA-1 too, as the slow SHA-1 is actually a huge slowdown when it comes to running the test suite.

I don't think that was me. Could simply remove the SHA-1 tests from the Travis set, since SHA-1 is mostly deprecated in *ring* anyway?

djc:
> I think the use of traits kills performance. Commit 644d870 slows everything down to barely 1MB/s.

@samscott89 maybe you can file a rustc bug about this. It seems like the compiler should be able to monomorphize the traits for this case to improve the performance. (I'd do it myself, but you'll probably be able to more accurately describe the problem and answer any questions they might have.)

samscott89:
>  maybe you can file a rustc bug about this. It seems like the compiler should be able to monomorphize the traits for this case to improve the performance. (I'd do it myself, but you'll probably be able to more accurately describe the problem and answer any questions they might have.)

It's a bit bizarre, I haven't been able to reproduce it with a minimal example. Checking the assembly code the problem is fairly obvious: the `PrimInt::rotate_right` method isn't inlined which adds a few cycles of overhead for calling the function and moving values around. Whereas in the macroed version there is just a single `ror` instruction used.

However, as I said before I can't reproduce this with a minimal example.

In any case, perhaps we should just leave a comment explaining why macros are used instead of traits over generics?

SimonSapin:
Does adding `#[inline]` to the method definition help?

samscott89:
Adding `#[inline]` to the definition in num_traits here: https://github.com/rust-num/num/blob/master/traits/src/int.rs#L302 indeed works.

samscott89:
Is there any way of forcing `#[inline]` for an external function?

Ms2ger:
I don't believe so, so I filed https://github.com/rust-num/num/issues/218 about it.

samscott89:
Cool, thanks!

briansmith:
@samscott89 There are many things we can do:

1. If we use the `block_data_order` function I suggested, then I think we can make these functions nested functions of the resultant block_data_order. This would be about the same as your "leave them as macros" approach but it seems cleaner to me. Something like this:

```
macro_rules! block_data_order {
[...]
{
+        // Same as the same-named function in `ring::digest::sha1`.
+        #[inline(always)]
+        fn ch(x: $Word, y: $Word, z: $Word) -> $Word {
+            (x & y) ^ (!x & z)
+        }
+
         let state = &mut $state[..$SHA.chaining_words];
```

2. I'm not very excited about depending on rust-num just for this anyway. I can spend some time to create something like `PrimInt` and then we could base this on the new thing.

I think #1 is good enough. I would eventually like to get rid of all the macros but it seems we need language improvements to the type system before it would be possible.

briansmith:
OK, I feel pretty strongly now that we should avoid the num-traits crate at least until we have more significant need for it. I don't want to duplicate that effort but also I don't want to add that dependency for such a small thing. Please let me know if you can update the PR to use the nested function approach inside the `block_data_order!` macro. Otherwise, I can take a stab at it next week.

Regardless, this has drug on for a long time (mostly my fault, it seems) so LMK what I can do to help finish this off.

samscott89:
> Please let me know if you can update the PR to use the nested function approach inside the block_data_order! macro. Otherwise, I can take a stab at it next week.

Will do.

samscott89:
Something like this: https://github.com/samscott89/ring/commit/d3711b44fc575dfce9f453b77e1df819133639fa ?

briansmith:
> Something like this: samscott89@d3711b4 ?

Yes, that seems fine. Will look at it again after the travis queue clears a bit.

briansmith:
See https://github.com/briansmith/ring/pull/256#issuecomment-262668392. This is very good work but it is in conflict with some other things we're trying to do with respect to increasing confidence in the code quality, so we're going to park it for now.

briansmith:
PR 863 has a new implementation of SHA-2. Please take a look. Thanks again for the effort here.

PR #211, opened 2016-06-05
Ed25519 signing

djc:


include/openssl/curve25519.h
```diff
@@ -39,9 +39,9 @@ OPENSSL_EXPORT void ED25519_keypair(uint8_t out_public_key[32],
 /* ED25519_sign sets |out_sig| to be a signature of |message_len| bytes from
  * |message| using |private_key|. It returns one on success or zero on
  * error. */
```
briansmith: Remove the "It returns one on success.." sentence.

src/ec/eddsa.rs
```diff
@@ -18,9 +18,76 @@
...
+
+
+impl<'a> Ed25519KeyPair {
+    /// Generate a new key pair from randomness.
+    pub fn generate(rng: &SecureRandom) -> Ed25519KeyPair {
+        let mut rand = RAND::new(rng);
+        let mut pair = Ed25519KeyPair { private_public: [0; 64] };
+        let mut public_key = [0u8; 32];
+        unsafe {
+            ED25519_keypair(public_key.as_mut_ptr(),
```
briansmith: Sorry, the implementation and return type of `ED25519_keypair` is wrong. It isn't checking the return value of `RAND_bytes`. It needs to return zero when `RAND_bytes` returns zero. That means that this `generate` function needs to use `bssl::map_result` and return a `Result`.

src/ec/eddsa.rs
```diff
@@ -18,9 +18,76 @@
...
 use {bssl, c, signature, signature_impl};
 use input::Input;
+use rand::{RAND, SecureRandom};
 
 struct EdDSA;
 
+/// An Ed25519 key pair.
+//
+// This can be used for signing. Allows generation of a new key pair through
+// the `generate()` method, or using an existing key pair with `from_bytes()`.
```
briansmith: These should all be `///` (triple slashes).

src/ec/eddsa.rs
```diff
@@ -18,9 +18,76 @@
...
+pub struct Ed25519KeyPair {
+    private_public: [u8; 64],
+}
+
+
+impl<'a> Ed25519KeyPair {
+    /// Generate a new key pair from randomness.
+    pub fn generate(rng: &SecureRandom) -> Ed25519KeyPair {
```
briansmith: I suggest you just drop "from randomness."

src/ec/eddsa.rs
```diff
@@ -18,9 +18,76 @@
...
 
+/// An Ed25519 key pair.
+//
+// This can be used for signing. Allows generation of a new key pair through
+// the `generate()` method, or using an existing key pair with `from_bytes()`.
+pub struct Ed25519KeyPair {
+    private_public: [u8; 64],
+}
+
+
```
briansmith: Drop one of these blank lines.

src/ec/eddsa.rs
```diff
@@ -18,9 +18,76 @@
...
+                            &mut rand);
+        };
+        debug_assert_eq!(pair.private_public[32..], public_key[..]);
+        pair
+    }
+
+    /// Copies key data from the given slices to create a new key pair.
+    pub fn from_bytes(private_key: &[u8; 32], public_key: &[u8; 32]) -> Ed25519KeyPair {
+        let mut pair = Ed25519KeyPair { private_public: [0; 64] };
+        {
```
briansmith: I suggest you use `let (private_key, public_key) = pair.private_public.split_at(32)` to simplify this function.

src/ec/eddsa.rs
```diff
@@ -18,9 +18,76 @@
...
+        unsafe {
+            ED25519_keypair(public_key.as_mut_ptr(),
+                            pair.private_public.as_mut_ptr(),
+                            &mut rand);
+        };
+        debug_assert_eq!(pair.private_public[32..], public_key[..]);
+        pair
+    }
+
+    /// Copies key data from the given slices to create a new key pair.
```
briansmith: Mention that the private key and public key are little-endian.

src/ec/eddsa.rs
```diff
@@ -18,9 +18,76 @@
...
+        {
+            let pair_pub_key_bytes = &mut pair.private_public[32..];
+            for i in 0..pair_pub_key_bytes.len() {
+                pair_pub_key_bytes[i] = public_key[i];
+            }
+        }
+        pair
+    }
+
+    /// Access private key data.
```
briansmith: I suggest "Returns a reference to the little-endian-encoded private key bytes."

src/ec/eddsa.rs
```diff
@@ -18,9 +18,76 @@
...
+    pub fn private_key_bytes(&'a self) -> &'a [u8] {
```
briansmith: Shouldn't the return type be `&'a [u8; 32]`?

src/ec/eddsa.rs
```diff
@@ -18,9 +18,76 @@
...
+        &self.private_public[..32]
+    }
+
+    /// Access public key data.
+    pub fn public_key_bytes(&'a self) -> &'a [u8] {
```
briansmith: Similar comments apply to this.

src/ec/eddsa.rs
```diff
@@ -18,9 +18,76 @@
...
+        &self.private_public[32..]
+    }
+
+    /// Sign a given message, returning the signature.
+    pub fn sign(&self, msg: &[u8]) -> signature::Signature {
+        let mut signature = signature::Signature { value: [0; 64] };
+        unsafe {
+            ED25519_sign(signature.value.as_mut_ptr(),
```
briansmith: Wrap the lines at 80 characters, instead of the one-argument-per-line.

src/signature.rs
```diff
@@ -66,6 +67,15 @@ pub use rsa::{
     RSA_PKCS1_3072_8192_SHA384_VERIFY,
 };
 
+/// A signature holder struct.
+//
+// The `Signature` struct is a just a wrapper around a `u8` array. It is
+// only used for signing operations, not for verification operations.
+pub struct Signature {
+    /// The array holding the signature data.
```
briansmith: Drop this comment.

src/signature.rs
```diff
@@ -66,6 +67,15 @@ pub use rsa::{
     RSA_PKCS1_3072_8192_SHA384_VERIFY,
 };
 
+/// A signature holder struct.
+//
+// The `Signature` struct is a just a wrapper around a `u8` array. It is
+// only used for signing operations, not for verification operations.
```
briansmith: I suggest simply "/// A public-key signature."

src/signature.rs
briansmith: Please add a comment explaining why `value` is public and also mark this:
```
#[doc(hidden)]
```

Also, please add this function:
```
/// Returns the signature's encoded value.
pub fn as_slice(&self) -> &'a [u8] { &self.value[..] }
```

briansmith:
This looks awesome.

If you have time, please (perhaps in a separate commit) add an example code fragment (a.k.a. doctest) that explains how to use the API, similar to the examples for the other APIs.

Thanks!

src/ec/eddsa.rs
briansmith: It would be great if we remove the `out_public_key` argument of `ED25519_keypair` completely, in a follow-up commit.

briansmith:
Oh, I forgot one important thing: Please acknowledge that you are the copyright holder of your patch (i.e. it isn't owned by an employer or similar) and that you agree to license it according to the license used in the files you modified. See the suggested wording of the commit message at https://github.com/briansmith/ring#contributing. Thanks again!

src/ec/eddsa.rs
djc: Hah, that's what I had, but then changed it after looking at `rustfmt` suggestions. I've changed it back.

src/ec/eddsa.rs
djc: It doesn't look like I can get that to work...

djc:
I started on doctests, but was a bit hesitant about how to do it nicely. This was my attempt:

```diff
diff --git a/src/signature.rs b/src/signature.rs
index 381f14a..4bcb12e 100644
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -136,6 +136,22 @@ pub struct VerificationAlgorithm {
 /// }
 /// ```
 ///
+/// ## Signing with Ed25519 private key
+///
+/// ```
+/// use ring::input::Input;
+/// use ring::signature::{Ed25519KeyPair, Signature};
+/// use ring::polyfill::slice_as_array_ref;
+///
+/// fn sign_ed25519(private_key: Input, public_key: Input, msg: Input)
+///                 -> Result<Signature, ()> {
+///     let pub_key = slice_as_array_ref!(public_key.as_slice_less_safe(), 32);
+///     let priv_key = slice_as_array_ref!(private_key.as_slice_less_safe(), 32);
+///     let key_pair = try!(Ed25519KeyPair::from_bytes(priv_key, pub_key));
+///     key_pair.sign(msg.as_slice_less_safe())
+/// }
+/// ```
+///
 /// ## Verify an Ed25519 signature
 ///
 /// ```
```

I started out with the arguments as `Input` to match the verification example, below, but that makes the code pretty ugly, what with all the `slice_as_array_ref!()` calls and such. Should it just be normal slices everywhere?

Other than that, I think all of your comments have been addressed except for the one about returning an array reference from the key pair accessor methods. Also, I have not rustified ED25519_keypair yet, since that function looks a bit scary. Do we just replicate types like `ge_p3` and `extern` declare functions like `x25519_ge_scalarmult_base`?

Anyway, personally I'd like to see if we can land this either with or without example docs, then we can work from there. I'm thinking the `Ed25519KeyPair` API might be better served by having slices and then checking the lengths inside rather than projecting array references outwards. Until `slice` gains an `as_array_ref()` method natively, I guess...

briansmith:
> I started out with the arguments as Input to match the verification example, below, but that makes the code pretty ugly, what with all the slice_as_array_ref!() calls and such. Should it just be normal slices everywhere?

The idea of `Input` is that it is used for (all) *untrusted* inputs. But, we don't (shouldn't!) sign stuff that is untrusted! So, that means it makes sense for the inputs to the `sign` function to be slices.

Note that `slice_as_array_ref!` isn't part of the *ring* API. If the examples are too ugly without `slice_as_array_ref!` then either we should improve the API to make it more usable or we should make `slice_as_array_ref!` part of the API.



> I have not rustified ED25519_keypair yet, since that function looks a bit scary. Do we just replicate types like ge_p3 and extern declare functions like x25519_ge_scalarmult_base?

Here's what I would do:

```c
void GFp_X25519_scalarmult_base(uint8_t out[32], const uint8_t in[32]) {
  ge_p3 A;
  x25519_ge_scalarmult_base(&A, in);
  ge_p3_tobytes(out, &A);
}
```

Then call this `GFp_X25519_scalarmult_base` function from Rust. This way you can avoid Rustifying the `ge_p3` struct.

> with or without example docs.

We can land it without the example. If you'd rather work on RSA then I'd be happy to take over writing the Ed25519 examples and other documentation.

djc:
I've pushed an API simplification in d224153, does that look okay? Could squash it into the previous commit if you prefer, though it's a clean improvement as it is IMO.

Also added a very basic example.

include/openssl/rand.h
```diff
@@ -24,7 +24,8 @@ extern "C" {
 
 /* Random number generation. */
 
-/* RAND_bytes writes |len| bytes of random data to |buf| and returns one. */
+/* RAND_bytes writes |len| bytes of random data to |buf| and returns one on
```
briansmith: s/ and/. It/

The convention in the code is to have "It returns one..." as a separate sentence. Let's be consistent with that.

crypto/curve25519/curve25519.c
```diff
@@ -4639,6 +4638,7 @@ int ED25519_keypair(uint8_t out_public_key[32], uint8_t out_private_key[64],
 
   ge_p3 A;
   x25519_ge_scalarmult_base(&A, az);
+  uint8_t out_public_key[32];
   ge_p3_tobytes(out_public_key, &A);
```
briansmith: Please just do `ge_p3_tobytes(&out_private_key[32], &A)` and remove `out_public_key` and `memmove` below. In particular, `out_` is for output parameters, and `out_public_key` isn't an output parameter so it shouldn't be named like that.

src/ec/eddsa.rs
```diff
@@ -18,9 +18,67 @@
...
+    pub fn private_key_bytes(&'a self) -> &'a [u8] {
+        &self.private_public[..32]
+    }
+
+    /// Returns a reference to the little-endian-encoded public key bytes.
+    pub fn public_key_bytes(&'a self) -> &'a [u8] {
+        &self.private_public[32..]
+    }
+
+    /// Sign a given message, returning the signature.
```
briansmith: s/ a given/the message `msg`/

src/ec/eddsa.rs
```diff
@@ -61,8 +105,11 @@ impl signature_impl::VerificationAlgorithmImpl for EdDSA {
 }
 
 
+#[allow(improper_ctypes)]
 extern {
-    #[cfg(test)]
+    fn ED25519_keypair(out_private_key: *mut u8/*[64]*/,
+                       rng: *mut RAND) -> c::int;
```
briansmith: Please use two `extern` blocks, one with `allow(improper_ctypes)` and one without. Only `E25519_keypair` should be marked `allow(improper_ctypes)`

src/ec/eddsa.rs
```diff
@@ -90,9 +137,11 @@ mod tests {
             let msg = test_case.consume_bytes("MESSAGE");
             let expected_sig = test_case.consume_bytes("SIG");
 
-            let mut actual_sig = [0u8; 64];
-            assert!(ed25519_sign(&private_key, &msg, &mut actual_sig).is_ok());
-            assert_eq!(&expected_sig[..], &actual_sig[..]);
+            let priv_key_sized = slice_as_array_ref!(&private_key[..32], 32).unwrap();
+            let pub_key_sized = slice_as_array_ref!(&public_key, 32).unwrap();
```
briansmith: NIT: drop the "_sized" suffixes.

src/signature.rs
```diff
@@ -66,6 +67,23 @@ pub use rsa::{
...
 };
 
+/// A public key signature.
+pub struct Signature {
+    value: [u8; 64],
+}
+
+impl<'a> Signature {
+    /// Initialize signature from slice of bytes.
+    pub fn new(signature_bytes: [u8; 64]) -> Signature {
```
briansmith: Please mark this as `#[doc(hidden)]` with a comment explaining why it has to be public. (It doesn't make sense to have `new` be part of the public API, but we need to construct `Signature`s from other modules within *ring* so it has to be public until Rust's module system is extended to support what we want.)

src/ec/eddsa.rs
```diff
@@ -43,8 +43,14 @@ impl<'a> Ed25519KeyPair {
 
     /// Copies key data from the given slices to create a new key pair.
     /// The arguments are interpreted as little-endian-encoded key bytes.
-    pub fn from_bytes(private_key: &[u8; 32], public_key: &[u8; 32]) -> Ed25519KeyPair {
+    pub fn from_bytes(private_key: &[u8], public_key: &[u8])
+                      -> Result<Ed25519KeyPair, ()> {
         let mut pair = Ed25519KeyPair { private_public: [0; 64] };
```
briansmith: Please put the construction of `pair` underneath the input validation (to minimize its scope and to show more clearly that nothing bad happens with invalid inputs).

src/ec/eddsa.rs
```diff
@@ -43,8 +43,14 @@ impl<'a> Ed25519KeyPair {
...
+        if private_key.len() != 32 {
+            return Err(());
+        } else if public_key.len() != 32 {
```
briansmith: No `else`. The early return above makes it unnecessary.

briansmith:
Because I'm making some disruptive changes, I'm going to make the small updates I suggested and check in your changes, so they don't bitrot.

src/ec/eddsa.rs
```diff
@@ -90,9 +137,11 @@ mod tests {
             let msg = test_case.consume_bytes("MESSAGE");
             let expected_sig = test_case.consume_bytes("SIG");
 
-            let mut actual_sig = [0u8; 64];
-            assert!(ed25519_sign(&private_key, &msg, &mut actual_sig).is_ok());
-            assert_eq!(&expected_sig[..], &actual_sig[..]);
+            let priv_key_sized = slice_as_array_ref!(&private_key[..32], 32).unwrap();
+            let pub_key_sized = slice_as_array_ref!(&public_key, 32).unwrap();
+            let key_pair = Ed25519KeyPair::from_bytes(&priv_key_sized, &pub_key_sized);
```
briansmith: This line was too long; I fixed it.

frewsxcv:
This looks great, nice work @djc! :)

djc:
Thanks for merging, and for all the hand-holding along the way!

briansmith:
Thanks a ton for doing this! As you probably noticed, I cleaned up the nits that I mentioned in my comments, plus a few more. Also, I added the verification that the public and private key are in sync, along with a test for that. Finally, I wrote a new example that matches the new style of module-level example documentation that I'm working on adding to every module. Thanks for contributing your example though! In other circumstances it would have worked fine.

I recommend that you look at the diff between your commits and the commits that landed, to see if there are any problems and to see what nits I addressed. That will make your next contribution much easier, I think.

Thanks again! Awesome work!

PR #218, opened 2016-06-18
Implement RSA PKCS1 v1.5 signing (fixes #208)

### LINKED ISSUE #208: Add RSA PKCS#1 1.5 signing
There's a bunch of code to do the low-level RSA operation in `GFp_rsa_private_transform` in `crypto/rsa/rsa_impl.c` that should be used.

We need a new implementation of PKCS#1 1.5 padding (in Rust, unless there's a reason it can't be done in Rust for some reason).

We can (should, unless we're going to somehow implement blinding in Rust) use the blinding code in crypto/rsa/blinding.c. But, we need to write new Rust code that manages the pool of `BN_BLINDING`s in some kind of thread-safe way. I guess this largely depends on the threading model chosen for the signing API.

NIST publishes test vectors for this, and testing using those test vectors are the minimum acceptable testing. Note that NIST's test vectors are in a format that's not really usable by *ring*'s testing framework. The comments in `src/rsa_pkcs1_verify_tests.txt` explain what I did to reformat NIST's RSA PKCS#1 1.5 *verification* test vectors for *ring*; a similar thing needs to be done for the *signing* test vectors.

This is complicated by the fact that there's no code for deserializing RSA private keys or for serializing RSA private keys. Adding the deserialization/serialization is probably a big task on its own so it might be worth splitting into a separate bug.

See also #205 for any general requirements on signing APIs.

djc:
I think it's ready for an official first review. I can write a doctest/example and polish the docs tomorrow.


crypto/rsa/rsa.c
```diff
@@ -76,8 +76,6 @@ RSA *rsa_new_begin(void) {
   }
 
   memset(rsa, 0, sizeof(RSA));
```
briansmith: Either keep the blank line, or remove the blank line above the `memset`.

briansmith:
The first line of each commit message should end with a period (it should be a sentence, or sentence-ish).

Every commit should have the "I agree..." at the end.

crypto/rsa/rsa.c
```diff
@@ -103,8 +101,12 @@ int rsa_new_end(RSA *rsa, BN_CTX *ctx) {
   assert(rsa->iqmp != NULL);
   assert(BN_get_flags(rsa->iqmp, BN_FLG_CONSTTIME));
 
-  int ret = 0;
+  BN_CTX *ctx = BN_CTX_new();
+  if (ctx == NULL) {
+    return 0;
+  }
 
+  int ret = 0;
```
briansmith: Blank line after the declaration of `ret`.

src/rsa.rs
```diff
@@ -17,7 +17,9 @@
 /// RSA PKCS#1 1.5 signatures.
 
 use {bssl, c, der, digest, signature, signature_impl};
+use std::boxed::Box;
 use untrusted;
+use std::ptr;
```
briansmith: Just `use::std` and the qualify the uses as `std::boxed::` and `std::ptr::`.

src/rsa.rs
```diff
@@ -163,8 +165,107 @@ fn parse_public_key<'a>(input: untrusted::Input<'a>) ->
...
 }
 
+#[allow(non_camel_case_types)]
+enum BIGNUM {}
+#[allow(non_camel_case_types)]
+enum BN_MONT_CTX {}
+
+#[allow(non_camel_case_types)]
+#[repr(C)]
+struct RSA {
```
briansmith: Add a comment that says that this needs to be kept in sync with `rsa_st` in rsa.h. Add a similar comment where `rsa_st` is defined in rsa.h.

src/rsa.rs
```diff
@@ -163,8 +165,107 @@ fn parse_public_key<'a>(input: untrusted::Input<'a>) ->
...
+
+const BN_FLG_CONSTTIME: c::int = 4;
+
+fn build_bn(input: untrusted::Input) -> *mut BIGNUM {
+    unsafe {
+        let bytes = input.as_slice_less_safe();
+        let res = BN_bin2bn(bytes.as_ptr(), bytes.len(), ptr::null_mut());
+        BN_set_flags(res, BN_FLG_CONSTTIME);
+        res
+    }
```
briansmith: See https://github.com/briansmith/ring/blob/master/STYLE.md#unsafe, in particular the part about minimizing the scope of the `unsafe`. In particular, `input.as_slice_less.safe()` shouldn't be inside the `unsafe`.

Do you need to use `BN_set_flags` here? Won't `rsa_new_end` set it on everything as needed?

Let's consider the error handling strategy. BN_bin2bn might return a null pointer when it fails. It seems you want to return the NULL pointer. Then, if any of the `BN_bin2bn` calls fail, we'll construct an `RSA` that has at least one NULL pointer member. Then, we are expecting `rsa_new_end` to check for all these NULL pointers. But, that seems error prone, and in fact `rsa_new_end` doesn't check for the NULL pointers (it only `asserts` they aren't NULL). Further, this seems prone to leak the `BIGNUM`, if we were to return early before constructing the `RSA`.

Maybe it would make more sense for this function to return `Result<PositiveInteger, ()>`, where `struct PositiveInteger { value: *mut BIGNUM }`, where PositiveInteger works exactly like `Box` except that it uses `BN_free()` instead of `free()`? Then, when we construct the `RSA`, we'd do it like `Box::new(RSA { n: n.into_raw(), e: e.into_raw(), ... })`.

Note that `build_bn` (perhaps renamed `PositiveInteger::new`) could also be the thing that calls `der::positive_integer`.

src/rsa.rs
```diff
@@ -163,8 +165,107 @@ fn parse_public_key<'a>(input: untrusted::Input<'a>) ->
...
+
+impl RSAKeyPair {
+    /// Reads key data from the given DER-encoded input, which must contain
+    /// the full private key data (n, e, d, p, q, dmp1, dmq1, iqmp).
+    ///
+    /// This is intended for use by code that deserializes key pairs.
+    pub fn from_der(input: untrusted::Input) -> Result<RSAKeyPair, ()> {
+        input.read_all((), |input| {
+            der::nested(input, der::Tag::Sequence, (), |input| {
+                let _ = try!(der::small_nonnegative_integer(input)); // version
```
briansmith: You need to check that the version is zero and fail if it isn't zero. Please add four new test vectors for this: One where the version field is missing, one where version is 1, and one where the version is -1 (encoded as value 0xff), and one where the version is 2. (Please create these test vectors by taking an existing good test vector and modifying it.)

src/rsa.rs
```diff
@@ -163,8 +165,107 @@ fn parse_public_key<'a>(input: untrusted::Input<'a>) ->
...
+                    rsa_new_end(rsa.as_mut())
+                }));
+                Ok(RSAKeyPair { rsa: rsa })
+            })
+        })
+    }
+}
+
+impl Drop for RSAKeyPair {
+    fn drop(&mut self) {
```
briansmith: `RSA_free` should be removed given that this replaces it.

src/rsa.rs
```diff
@@ -163,8 +165,107 @@ fn parse_public_key<'a>(input: untrusted::Input<'a>) ->
...
+    rsa: Box<RSA>,
+}
+
+const BN_FLG_CONSTTIME: c::int = 4;
+
+fn build_bn(input: untrusted::Input) -> *mut BIGNUM {
+    unsafe {
+        let bytes = input.as_slice_less_safe();
+        let res = BN_bin2bn(bytes.as_ptr(), bytes.len(), ptr::null_mut());
+        BN_set_flags(res, BN_FLG_CONSTTIME);
```
briansmith: I was wrong in my other comment: We do need to call `BN_set_flags` to set the `BN_FLG_CONSTTIME` flag here. BUT, not for `n` and `e`, because those aren't private values.

util/generate-rsa-signing-tests.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python2
+# Copyright 2016 Dirkjan Ochtman
```
briansmith: Please add the license block.

util/generate-rsa-signing-tests.py
```diff
@@ -0,0 +1,81 @@
...
+        hex_sig = ''.join('{:02x}'.format(ord(c)) for c in sig)
+        assert hex_sig == case['S']
+
+        # Serialize the private key in DER format
+        der = key.private_bytes(serialization.Encoding.DER,
+                                serialization.PrivateFormat.TraditionalOpenSSL,
+                                serialization.NoEncryption())
+        hex_der = ''.join('{:02x}'.format(ord(c)) for c in der)
+
+        # Print the test case data in the format used by ring test files
```
briansmith: Please spell *ring* as `*ring*`.

src/rsa.rs
```diff
@@ -165,10 +167,19 @@ fn parse_public_key<'a>(input: untrusted::Input<'a>) ->
     })
 }
 
+static PREFIXES: [(&'static digest::Algorithm, &'static [u8]); 4] = [
+    (&digest::SHA1, b"\x30\x21\x30\x09\x06\x05\x2b\x0e\x03\x02\x1a\x05\x00\x04\x14"),
```
briansmith: These prefixes are already defined immediately above. See SHA512_PKCS1_DIGESTINFO_PREFIX and friends.

src/rsa.rs
```diff
@@ -244,6 +255,49 @@ impl RSAKeyPair {
     pub fn public_modulus_len(&self) -> usize {
         unsafe { RSA_size(self.rsa.as_ref()) }
     }
+
+    /// Sign the given message, hashed with the given digest algorithm.
```
briansmith: Please document the function by mentioning what each parameter does. (See the documentation for other public API functions at https://briansmith.org/rustdoc/ring/.)

src/rsa.rs
```diff
@@ -244,6 +255,49 @@ impl RSAKeyPair {
...
+    pub fn sign(&self, digest_alg: &'static digest::Algorithm, msg: &[u8], signature: &mut [u8]) -> Result<(), ()> {
```
briansmith: Hmm...are we sure this is the order the parameters should be in? I'm actually not sure. In particular, should we put the `signature` parameter before or after `msg`?

src/rsa.rs
```diff
@@ -244,6 +255,49 @@ impl RSAKeyPair {
...
+        if signature.len() != self.public_modulus_len() {
+            return Err(());
+        }
+
+        let found = PREFIXES.into_iter().find(|pair| pair.0.nid == digest_alg.nid);
+        if found.is_none() {
+            return Err(());
+        }
```
briansmith: There's a problem with doing it this way: Since `PREFIXES` references every digest algorithm, that means that all the code for every digest algorithm has to be linked into the program, even if only a subset are used.

Instead, I suggest that we rename `RSA_PKCS1` to `PKCS1Padding`, make it public, and then have `sign` take `padding: &PKCS1Padding` instead of `digest_alg` as a parameter.

src/rsa.rs
```diff
@@ -268,11 +322,17 @@ impl Drop for RSAKeyPair {
 }
 
 
+#[allow(improper_ctypes)]
```
briansmith: Only `GFp_rsa_private_transform` should be annotated `allow(improper_ctypes)`.

src/rsa.rs
```diff
@@ -288,9 +348,10 @@ extern {
 
 #[cfg(test)]
 mod tests {
-    use {der, file_test, signature};
+    use {der, digest, file_test, signature};
     use untrusted;
     use super::*;
+    use std::vec::Vec;
```
briansmith: just `use std`.

src/rsa.rs
```diff
@@ -338,4 +399,39 @@ mod tests {
...
+            let key_pair = RSAKeyPair::from_der(private_key).unwrap();
+
+            let msg = test_case.consume_bytes("Msg");
+            let expected = test_case.consume_bytes("Sig");
+            let mut actual: Vec<u8> = Vec::with_capacity(key_pair.public_modulus_len());
+            for _ in 0..actual.capacity() {
+                actual.push(0);
+            }
+            try!(key_pair.sign(alg, &msg, actual.as_mut_slice()));
+            assert_eq!(actual.as_slice(), &expected[..]);
```
briansmith: In another commit, we should add a check that the signature verifies with the public key. Note that currently we don't have the function to get the public key for the private key! We need to add that.

src/rsa.rs
```diff
@@ -244,6 +255,49 @@ impl RSAKeyPair {
...
+        pad_bytes.push(0);
+        pad_bytes.push(1);
+        for _ in 0..padding_len {
+            pad_bytes.push(255);
+        }
+        pad_bytes.push(0);
+        pad_bytes.extend(prefix);
+        pad_bytes.extend(hashed.as_ref());
+
+        let rng = rand::SystemRandom::new();
```
briansmith: `sign` should take a `rng: &SecureRandom` parameter, instead. In general, we never use `SystemRandom::new()`, but require the library user to do that.

src/rsa.rs
```diff
@@ -244,6 +255,49 @@ impl RSAKeyPair {
...
+        let mut rand = RAND::new(&rng);
+        try!(bssl::map_result(unsafe {
+            let blinding = BN_BLINDING_new();
```
briansmith: This is OK for the first attempt. But, this is very inefficient. Besides the memory allocation that `BN_BLINDING_new()` does, the first time you *use* a `BN_BLINDING` is much more expensive than the subsequent times. So, later, we should store `blinding` in the `RSAPrivateKeyPair` struct and reuse it (using a mutex).

src/rsa.rs
```diff
@@ -244,6 +255,49 @@ impl RSAKeyPair {
...
+        if msg_len < digest_len + 11 {
+            return Err(());
+        }
+
+        let padding_len = msg_len - digest_len - 3;
+        let mut pad_bytes: Vec<u8> = Vec::with_capacity(msg_len);
+        pad_bytes.push(0);
+        pad_bytes.push(1);
+        for _ in 0..padding_len {
+            pad_bytes.push(255);
```
briansmith: NIT: please s/255/0xff`.

src/rsa.rs
```diff
@@ -244,6 +255,49 @@ impl RSAKeyPair {
...
+        let msg_len = self.public_modulus_len();
+        let hashed = digest::digest(digest_alg, msg);
+        let prefix = found.unwrap().1;
+        let digest_len = prefix.len() + hashed.as_ref().len();
+        if msg_len < digest_len + 11 {
+            return Err(());
+        }
+
+        let padding_len = msg_len - digest_len - 3;
+        let mut pad_bytes: Vec<u8> = Vec::with_capacity(msg_len);
```
briansmith: I think we should do a follow up where we do all of this in-place in `signature`, avoiding the need to allocate this memory at all.

src/rsa.rs
```diff
@@ -163,8 +165,107 @@ fn parse_public_key<'a>(input: untrusted::Input<'a>) ->
...
+        let bytes = input.as_slice_less_safe();
+        let res = BN_bin2bn(bytes.as_ptr(), bytes.len(), ptr::null_mut());
+        BN_set_flags(res, BN_FLG_CONSTTIME);
+        res
+    }
+}
+
+impl RSAKeyPair {
+    /// Reads key data from the given DER-encoded input, which must contain
+    /// the full private key data (n, e, d, p, q, dmp1, dmq1, iqmp).
```
briansmith: Please base the documentation of this function on the documentation for other, similar things. For example, see https://briansmith.org/rustdoc/ring/signature/static.ECDSA_P384_SHA256_VERIFY.html. Notice that it references the specification, in this case [RFC3447 Appendix A.1.2](https://tools.ietf.org/html/rfc3447#appendix-A.1.2). Instead of mentioning (`n`, `e`, `d`, ...) individually, just reference the `RSAPrivateKey` structure.

Please mention that only two-prime keys (version 0) are supported; multiprime keys (version 1) are not yet supported.

src/rsa.rs
djc: It's still used, in the error handling for `RSA_generate()`. So it should be removed as part of adding `RSAKeyPair` generation, I think.

src/rsa.rs
djc: So there will be a `private: bool` argument on whatever replaces `build_bn()`?

briansmith: I would just have a `flags: c::int` parameter. Then you can pass `BN_FLG_CONSTTIME` for the ones that need it.

src/rsa.rs
djc: To me, source before destination makes more sense.

src/rsa.rs
```diff
@@ -163,8 +164,133 @@ fn parse_public_key<'a>(input: untrusted::Input<'a>) ->
...
 }
 
+#[allow(non_camel_case_types)]
+enum BIGNUM {}
+#[allow(non_camel_case_types)]
+enum BN_MONT_CTX {}
+
+#[allow(non_camel_case_types)]
+#[repr(C)]
+/// Needs to be kept in sync with `struct rsa_st` (in `include/openssl/rsa.h`).
```
briansmith: NIT: I think the comment should go above the `#[...]` stuff, unless there's some Rust style guide that says the opposite.

util/generate-rsa-signing-tests.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # Read private key components
+        n = int(case['n'], 16)
+        e = int(case['e'], 16)
+        d = int(case['d'], 16)
+
+        # Recover prime factors and CRT numbers
+        p, q = rsa.rsa_recover_prime_factors(n, e, d)
+        # cryptography returns p, q with p < q by default
+        # OpenSSL requires p > q, so swap them here
```
briansmith: s/OpenSSL/*ring*/

util/generate-rsa-signing-tests.py
```diff
@@ -0,0 +1,94 @@
...
+            if name.strip() == 'S':
+                cases.append(cur)
+                cur = copy.copy(cur)
+    return cases
+
+def main(fn):
+    for case in parse(fn):
+
+        if case['SHAAlg'] == 'SHA224':
+            # SHA224 not supported in *ring* for now
```
briansmith: NIT: Please end comments with periods. (We follow the BoringSSL rule where, unless constrained for space, we make comments full sentences.) Here and elsewhere below.

src/rsa.rs
```diff
@@ -86,9 +92,29 @@ impl signature_impl::VerificationAlgorithmImpl for RSA_PKCS1 {
     }
 }
 
+macro_rules! rsa_pkcs1_padding {
+    ( $PADDING_ALGORITHM:ident, $digest_alg_name:expr,
+      $digest_alg:expr, $digestinfo_prefix:expr ) => {
+
+        #[doc="PKCS#1 1.5 padding with the "]
+        #[doc=$digest_alg_name]
+        #[doc=" digest algorithm."]
```
briansmith: We need to have a simlar "not available in "no_heap" mode comment.

Also, it seems better to just fold this macro into rsa_pkcs1. I doubt that we'll ever have a mismatch between the padding types we support for signing and the types we support for verification.

src/rsa.rs
```diff
@@ -100,29 +126,23 @@ macro_rules! rsa_pkcs1 {
...
                 min_bits: $min_bits,
-                digestinfo_prefix: $digestinfo_prefix,
             }
         };
     }
 }
 
-rsa_pkcs1!(RSA_PKCS1_2048_8192_SHA1_VERIFY, 2048, "2048", "SHA-1",
-           &digest::SHA1, &SHA1_PKCS1_DIGESTINFO_PREFIX);
+rsa_pkcs1!(RSA_PKCS1_2048_8192_SHA1_VERIFY, 2048, "2048", "SHA-1", RSA_PKCS1_SHA1);
```
briansmith: These are all things we should figure out in follow-up patches:

Should we rename these to have the pattern `{padding}`_{min}_{max}_VERIFY'? Now it seems weird for the key size to be in the middle.

And/or, if we start all padding names with `PADDING_` and all verify algorithm names with `VERIFY_` then they'll sort separately.

We are not enforcing minimum/maximum key size for signing like we are for verification. IMO, we should have a way of enforcing the same 2048-8192 and 3072-8192 ranges for signing.

src/rsa.rs
briansmith: Never mind the second part. I see that the `RSA_PKCS1_3072_8192_SHA384_VERIFY` would screw this up.

src/rsa.rs
```diff
@@ -16,7 +16,8 @@
 
 /// RSA PKCS#1 1.5 signatures.
 
-use {bssl, c, der, digest, signature, signature_impl};
+use {bssl, c, der, digest, rand, signature, signature_impl};
+use rand::RAND;
```
briansmith: No `use rand::RAND`.

src/rsa.rs
```diff
@@ -27,6 +28,30 @@ pub struct RSAPKCS1Padding {
     digestinfo_prefix: &'static [u8],
 }
 
+impl RSAPKCS1Padding {
+    fn pad(&self, msg: &[u8], out: &mut [u8]) -> Result<(), ()> {
+        let digest_len = self.digestinfo_prefix.len() + self.digest_alg.output_len;
+        if out.len() < digest_len + 11 {
```
briansmith: Please document where 11 comes from with a comment. Also, please write `digest_len + 11` as (`1 + 1 + 8 + digest_len`) to make things even more clear.

src/rsa.rs
```diff
@@ -27,6 +28,30 @@ pub struct RSAPKCS1Padding {
...
+        let pad_len = out.len() - digest_len - 3;
+        out[0] = 0;
+        out[1] = 1;
+        for i in 0..pad_len {
+            out[2 + i] = 0xff;
+        }
+        out[2 + pad_len] = 0;
+
+        let mid = 3 + pad_len + self.digestinfo_prefix.len();
+        let (digest_prefix, digest_dst) = out.split_at_mut(mid);
```
briansmith: I suggest you write this as
```rust
let (digest_prefix, digest_value) =
    out[(3 + pad_len)..].split_at_mut(self.digestinfo_prefix.len())];`
```

Then you can avoid the `[pad_len + 3]` below, which reduces the amount of redundancy in the code.

src/rsa.rs
```diff
@@ -289,6 +316,31 @@ impl RSAKeyPair {
     pub fn public_modulus_len(&self) -> usize {
         unsafe { RSA_size(self.rsa.as_ref()) }
     }
+
+    /// Pad the message `msg` with the padding algorithm `padding_alg` and write
+    /// the resulting signature into `signature`. The random number generattor `rng`
+    /// is used for blinding (to help make signing constant time).
```
briansmith: The `(to make signing constant time)` is incorrect; just drop that parenthetical. (The blinding is actually used because the rest of RSA isn't 100% constant time, though it tries hard to be, and so the blinding randomizes the timing.)

src/rsa.rs
```diff
@@ -289,6 +316,31 @@ impl RSAKeyPair {
...
+    ///
+    /// `signature` must be exactly the length of the `public_modulus_len()`.
+    pub fn sign(&self, padding_alg: &'static RSAPKCS1Padding, rng: &rand::SecureRandom,
+                msg: &[u8], signature: &mut [u8]) -> Result<(), ()> {
+        if signature.len() != self.public_modulus_len() {
+            return Err(());
+        }
+
+        let msg_len = self.public_modulus_len();
```
briansmith: I think it doesn't make sense to have this `msg_len` variable. Maybe this is an artifact of the previous iteration? If the goal is to avoid calling `public_modulus_len()` multiple times, then `msg_len` should be used in the first line of the function too. But, I think it just makes more sense to call `public_modulus_len()` below too.

src/rsa.rs
```diff
@@ -289,6 +316,31 @@ impl RSAKeyPair {
...
+        try!(padding_alg.pad(msg, signature));
+
+        let mut rand = RAND::new(rng);
+        try!(bssl::map_result(unsafe {
```
briansmith: You don't need the `try!(....); Ok(())`. map_result returns a `Result<(), ()>`, which is what you need.

src/rsa.rs
```diff
@@ -289,6 +316,31 @@ impl RSAKeyPair {
...
+            let blinding = BN_BLINDING_new();
+            let ret = GFp_rsa_private_transform(self.rsa.as_ref(), signature.as_mut_ptr(),
+                                                msg_len, blinding, &mut rand);
```
briansmith: `msg_len` comes from `public_modulus_len()`. Although we verified that that is the same as `signature.len()`, it would still be clearer to use `signature.len()` here. In general, when we pass a slice to a C function, we do so by `x.as_mut_ptr(), x.len()`.

src/rsa.rs
briansmith: Actually, this is dead code, at least as far as our test suite goes, right?

I think instead, we should have `rsa_verify_end()` reject keys where `n < 2048`, and then change this check to a `debug_assert!()` because we reject small keys.

Consider: a SHA-512 hash is 64 bytes. The digest_info prefix is 19 bytes. 64 + 19 + 11 = 94. 94 * 8 bits per byte =  752 bits.

Otherwise, we need to make a test vector with a <752 bit key for testing this.

src/rsa.rs
```diff
@@ -443,4 +444,39 @@ mod tests {
...
+                &RSA_PKCS1_SHA384
+            } else if digest_name == "SHA512" {
+                &RSA_PKCS1_SHA512
+            } else {
+                panic!("Unsupported digest: {}", digest_name);
+            };
+
+            let private_key = test_case.consume_bytes("Key");
+            let private_key = try!(untrusted::Input::new(&private_key));
+            let key_pair = RSAKeyPair::from_der(private_key).unwrap();
```
briansmith: Use `try!` here.

briansmith:
I pushed the first 5 commits, plus https://github.com/briansmith/ring/commit/1ac871ea0d3a0fef9fdafc233897ae0737d1ebaf.

briansmith:
### Formatting

Please check out https://github.com/djc/ring/compare/208-rsa-signing...briansmith:208-rsa-signing (or https://github.com/briansmith/ring/commits/208-rsa-signing, if you don't like the GitHub comparison view), where I suggest some formatting and wording tweaks. It looks like a big diff but most of it is:

1. Move the implementation details of `RSAKeyPair` to be below `RSAKeyPair`.
2. Wrap code at 80 characters. (We need to figure out how to compare rustfmt to format *ring* code in a reasonable way, so that this isn't such a manual process.)
3. Tweak some comments.
4. Remove some unneeded annotations about snake case naming. Rust accepts ALLCAPS type names as long as they don't contain underscores `LIKE_THIS`.

### Public Key refactoring
The idea is exactly right: we need a way to extract the public key from the private key, and we need to verify that the public key we extract is valid for the private key by verifying that we can verify a signature using the public key that was signed with the private key. But this new design is actually going in the opposite direction from what I've been trying to do with *ring*. In particular, I don't want the have `RSAPublicKey` objects if we can avoid them. I'd much rather we return the ASN.1-DER-encoded bytes of the public key and avoid `RSAPublicKey` at all. This would be consistent with what we did with Ed25519KeyPair's `public_key_bytes`, for example.

I do realize, though, that it's a lot more work to do it the way I prefer because we don't have any infrastructure to serialize ASN.1 DER encoded stuff. Thinking about it more, the main time we need the public key is right after we've generated a new key pair, so that we can put it in a certificate or some other certificate-like message. And, we're not doing RSA keygen in this series. So, maybe we should move this part of the series over to a different branch that we can refactor during the implementation of the RSA keygen API? WDYT?

### What's left to do for MVP?

* Verify that this builds and the test suite runs (without the RSA tests) in `no_heap` mode.

* Verify that when the rustdoc is generated, everything newly-added to the rustdoc properly references the relevant specifications and that everything new notes that it isn't available in the no_heap mode. In particular, the documentation for the `RSAPKCS1Padding`, at least, needs to be improved in these respects. (The spec is https://tools.ietf.org/html/rfc3447#section-8.2.)

* Create an example in the `ring::signature` module documentation for deserializing an RSA private key, signing a message with it, and verifying the message, similar to the Ed25519 example at https://briansmith.org/rustdoc/ring/signature/index.html#signing-and-verifying-with-ed25519. Since we don't have the keygen or public key formatting, I guess the example can just hand-wave those aspects and assume that we already have the public key bytes and that the key was previously serialized.

* I thought on IRC we agreed to just name `RSAPKCS1Padding` as `RSAPadding` on the assumption that we'll be able to (somehow) use the same type for RSA-PSS if/when it is implemented? I think we should do that before this lands to save some churn. (If you agree, it can be done in a separate commit.)

* I feel like we need more testing regarding different lengths of keys, as I alluded to in the line-by-line review comments. Particularly, we should test what happens with very small keys, or, preferably, prohibit small keys. And, particularly, we should prohibit keys that are larger than what we have tests for; i.e. either we should limit the private key public modulus size to 3072 (which is what we test up to), or we should add more tests up to the maximum. In particular, the memory management in `BN_mod_exp_consttime` is very complicated and there is potential for stack overflows and other badness. I may need to merge in some more OpenSSL changes. See https://www.mail-archive.com/openssl-dev@openssl.org/msg44759.html

6. On 32-bit ARM, the tests are now timing out on Travis CI. So, we need to figure out a way to prevent that.

7. We need to document the `p > q` restriction and also add a test for this. Also, we should improve the behavior when `p < q` so that the failure isn't "BN_mod_inverse failed", which is confusing and terrible. I will work on this.

djc:
(Nothing new, just rebased on master.)

djc:
**Formatting**: I think your rules make sense. I have to say that I find keeping to the 80-column limit much more onerous in Rust than in Python; I'm never this lax in Python. Also, I was under the impression that *ring* isn't totally consistent about 80 chars everywhere, so I had a bit of broken window theory going on. What do you want me to do with your reformatting commits? Squash into my own, or keep them in the series as is?

**Public key refactoring**: I disagree quite a bit with your point of view, I rather like the direction I was moving the code in. I like having public key object that I can pass around and that's a thing that the type system can deal with. I also like not having the overhead of deserializing the public key for every verification.

**Other MVP bits**: Forgot about the renaming of the padding struct, will still do that. Realistically, it would seem to me that we need support for 4096-bit keys, but not anything beyond that. I can cook up some test vectors with a freshly generated key and the existing NIST messages/digest parametrization, does that work for you? Will document the `p > q` restriction. I'd like to leave the work of dealing with the ARM timeout to you somehow... I have no clue what a good solution for that might be.

In general: I'm not writing webpki/TLS code, and I'm not familiar with the kinds of constraints it involves. Let's say that I'm lazy crypto-layperson programming, and I just want to get some stuff done with the high-level primitives that cryptographers tell me are good or that some standard tells me I need -- my primary motivation right now is OpenID Connect, and the RSA-PKCS1-SHA256 signing and verification it requires. So I need signing, which means deserializing a key pair that I generated using the openssl command-line tool, and verification, which means getting a public key thing from JWS data, which are base64-encoded strings representing `n` and `e`. For the latter, I'd like to be able to add `RSAPublicKey.from_bytes()` or `from_slice()` or maybe `from_input()` and have it Just Work.

Having to use constants like `RSA_PKCS1_2048_8192_SHA256_VERIFY` are just off-putting to me; I have these public key parameters, and now I want to verify this thing. In particular, the documentation does a bad job (well, not really a job at all) of explaining what the `2048_8192` part is good for (although I now understand it, and it makes sense to me). Also, to me, the `RSA_PKCS1_SHA256` part should be orthogonal to sign vs verify, and the key length limitations are more closely coupled to the key then the verification. So in my mind, the ideal API would have an `RSAPublicKey` struct with constructors `from_der()` and `from_bytes()` where the constructors can take min_bits and max_bits arguments. The `verify()` function can then be a wrapper around instantiating the `RSAPublicKey` with the size contraints and the given `RSAPadding`  as they are now, if you still want to have it that way.

briansmith:
> Formatting: I think your rules make sense. I have to say that I find keeping to the 80-column limit much more onerous in Rust than in Python; I'm never this lax in Python. Also, I was under the impression that ring isn't totally consistent about 80 chars everywhere, so I had a bit of broken window theory going on.

It is true that *ring* sometimes goes over 80 characters because I've been following BoringSSL where sometimes they'd prefer some punctuation on the 81st column than to wrap something. I know that the current situation isn't ideal. I need to get rustfmt working acceptably on *ring* and then it will become a non-issue.

> What do you want me to do with your reformatting commits? Squash into my own, or keep them in the series as is?

Please review them and make sure they make sense, and squash them into your changes. (I wrote them because me telling you how I want things reformatted would have just been a huge timesink for both of us, and it's faster for me to communicate it by patch.)

> Public key refactoring: I disagree quite a bit with your point of view, I rather like the direction I was moving the code in. I like having public key object that I can pass around and that's a thing that the type system can deal with. I also like not having the overhead of deserializing the public key for every verification.

Let's move the discussion of the public key stuff to another forum (perhaps another issue in the issue tracker.) Otherwise, it will dominate this discussion. I think your points are valid concerns and I want to address them properly. The main thing we need to decide is whether it's OK to split the public key parts of this PR into a separate PR. WDYT?

> Other MVP bits: Forgot about the renaming of the padding struct, will still do that. Realistically, it would seem to me that we need support for 4096-bit keys, but not anything beyond that. I can cook up some test vectors with a freshly generated key and the existing NIST messages/digest parametrization, does that work for you?

Yes.

> I'd like to leave the work of dealing with the ARM timeout to you somehow... I have no clue what a good solution for that might be.

Yes, I'll take care of it.

src/rsa.rs
djc: Are you sure about the first part? Seems to me like nothing about the PKCS1 padding, per se, requires the heap...

src/rsa.rs
djc: From a high-level perspective, it's totally unclear why `sign()` requires an RNG argument, so I think there needs to be a comment to explain it. I can make it about randomizing the timing, instead.

src/rsa.rs
briansmith: Good point. `RSAKeyPair` is only available in the `use_heap` mode and `RSAPadding` is only useful for `RSAKeyPair.sign`. So, technically it doens't use the heap but there's no point in exposing it in the `no_heap` config. OTOH, no harm, no foul; I'm not sure it's worth extra effort to hide it either.

(BTW, "no_heap" got replace with "use_heap" yesterday.)

briansmith:
Please review the changes I made at https://github.com/briansmith/ring/tree/208-rsa-signing-2. In some cases, I just tweaked your wording and/or formatting and amended your commit. In other cases, I made a more substantial change in a separate commit. This time, let's leave those separate commits separate instead of squashing them into yours. This will help me keep track of what I overlooked during this review, for improving future reviews.

Regarding the 4096-bit test vectors: Sorry, I misunderstood what you said on IRC. We don't need that many test vectors with a 4096-bit key. We only need one, or maybe one per digest algorithm. Also, when the message doesn't matter, let's use "" as the message to digest.

Please add test a vector for a 2040 bit key so that the failure condition for too-small keys is covered. Also, please add a test vector for a 4104 (4096 + 8) bit key, so that the failure condition for the too-large key is covered. These will cover https://coveralls.io/builds/6677166/source?filename=crypto%2Frsa%2Frsa.c#L186 and the new code I added.

Please add a test that triggers this: https://coveralls.io/builds/6677166/source?filename=src%2Frsa.rs#L287, by passing in a too-small `signature` slice.

Please add a test vector that triggers this: https://coveralls.io/builds/6677166/source?filename=src%2Frsa.rs#L372. The easiest way to do this would be to use a valid public key, but remove the second half of it, so that parsing fails before we construct the `RSA`.

Please add a test vector that triggers this: https://coveralls.io/builds/6677166/source?filename=src%2Frsa.rs#L480, or drop the SHA-1 support for now (e.g. by making the SHA1 `RSAPadding` private).

Regarding the example doctest:
* Please name the private key file something like "src/signature_rsa_example_private_key.der".
* The private key file is missing! You forgot to "git add" it.
* The example's title says that the example shows how to sign *and verify* using RSA PKCS#1, but the verification step isn't included. Please add the verification step. You can use `include_bytes!` again. (I believe openssl will generate the `RSAPublicKey` if you use `-pubout` instead of `-out`.)

Meanwhile, I will work on improving the other parts, including at least some of the stuff I mention in https://github.com/briansmith/ring/issues/225.

briansmith:
Belatedly closing this. It all landed! Again, thanks a ton!

PR #226, opened 2016-06-22
Add `signature::primitive::verify_rsa`

djc:


briansmith:
From the old IRC conversation, my understanding is that there are two reasons we need such a function:

1. Application need to verify an RSA signatures for public keys that were given to them in non-ASN.1-encoded form.
2. When we generate an RSA key pair, we need a way to extract the public key from the key pair.

For #1, I suggest that instead we just add a new function `ring::signature::primitive::verify_rsa_signature(padding: &RSAPadding, msg: untrusted::Input, sig: untrusted::Input, /*public_key*/(n, e): (untrusted::Input, untrusted::Input)) -> Result<(), ()>` for the verification. Then the application won't have to construct the DER-encoded key (which you don't like anyway, IIRC) and we can avoid doing memory allocation during signature verification.

For #2, I suggest that when we create a public key, we should create and store the ASN.1-encoded public key in a `Vec` owned by the `RSAKeyPair`. Then `RSAKeyPair` can have a `public_key_bytes` function like `Ed25519KeyPair`. This would use the code that you submitted in this PR, but it would be internal to `RSAKeyPair`.

WDYT?

djc:
Sounds good. Please review my implementation of #1.

briansmith:
It looks like you forgot to `git add` the test file?

djc:
Yeah... I've fixed it up now.

src/rsa.rs
```diff
@@ -116,6 +116,25 @@ fn verify(padding: &RSAPadding, msg: untrusted::Input,
...
+    /// Takes an `RSAPadding` static to determine the padding algorithm used,
+    /// a message, a signature (which should match the length of public
+    /// `n`), the parts of the public key (in big-endian byte order). The key
+    /// length must be between 2048 and 8192 bits.
+    pub fn verify(padding: &super::RSAPadding, msg: untrusted::Input,
+                  signature: untrusted::Input,
+                  (n, e): (untrusted::Input, untrusted::Input))
+                  -> Result<(), ()> {
+        let n = n.as_slice_less_safe();
+        let e = e.as_slice_less_safe();
```
briansmith: I think instead we should change `parse_public_key` so that it returns `untrusted::Input`s and then have `super::verify` take `(n, e): (untrusted::Input, untrusted::Input)`. This removes the duplication of the `as_slice_less_safe` and fits the pattern I'm trying to promote wherein we avoid using slices for untrusted inputs when possible.

djc:
Ah, sorry if I implemented the API wrong. I'm not yet well-versed in the intricacies of Rust's module system and how it interacts with public APIs. Will try to rejigger so that it exposes `ring::signature::primitive::verify_rsa()`, with an implementation in `ring::rsa`.

briansmith:
Suggestion: do `cargo doc -p ring --open` to preview what your API changes end up looking like. (`--open` doesn't always work for me.)

djc:
This should be better, let me know what you think!

src/rsa.rs
```diff
@@ -119,6 +119,27 @@ fn verify(padding: &RSAPadding, msg: untrusted::Input,
...
+    ///
+    /// Takes an `RSAPadding` static to determine the padding algorithm used,
+    /// a message, a signature (whose length should match the length of public
+    /// exponent, `n`) and a tuple containing the parts of the public key (in
+    /// big-endian byte order). The key length must be between 2048 and 8192 bits.
+    pub fn verify(padding: &super::RSAPadding, msg: untrusted::Input,
+                  signature: untrusted::Input,
+                  (n, e): (untrusted::Input, untrusted::Input))
+                  -> Result<(), ()> {
+        super::verify(padding, msg, signature, (n, e), 2048)
```
briansmith: Please see my comments in the other PR where you renamed `RSA_PKCS1_2048_8192_SHA*_VERIFY` to `RSA_PKCS1_SHA*_2048_8192`. I had landed that change, and then I saw this change and decided to revert that change because this helped me see that we're going in the wrong direction.

As discussed previously, our goal is to eventually develop an API that treats ECDSA, RSA, etc. uniformly or at least as similarly as possible. In ECDSA, the only parameters you have are curve and digest algorithm. In RSA, you have min bits, max bits, digest algorithm, exponent, and padding scheme. In *ring*, I've chosen to use the following mapping:

| ECDSA                               | RSA                                  | EdDSA         |
|-------------------------------------|--------------------------------------|---------------|
| Algorithm = ECDSA | Algorithm = (RSA, Padding Scheme) | Algorithm = EdDSA |
| Curve                       | modulus min bits, modulus max bits, exponent min bits, exponent max bits)               | Curve |
| Digest Alg.                         | Digest Alg.                          | implicit      |
| Encoding (ASN.1, PKCS#11/WebCrypto) | implicit                             | implicit      |

Note that the exponent range is fixed at (2 bits, 33 bits). But, I'm thinking of changing that so that for the 3072-8192 variant (Suite B), at least, the minimum exponent length will be 16 bits.

So, anyway, I'm a little unsure how important it is to be explicit about the allowed exponent range, but at least this new primitive API should allow the caller to be explicit about the modulus range--at least the minimum modulus length.

Going back to your other PR, the issue you had was that `RSAPadding`s are named `RSA_<padding_scheme>_<digest_alg>` but RSA `SignatureAlgorithm`s are named similarly, but they put <min_modulus_bits>_<max_modulus_bits> in between. And, that affects this API because it takes `RSAPadding` as a parameter. But, maybe this API shouldn't take `RSAPadding`.

Ideally, this API should take `RSA_PKCS1_2048_8192_SHA256`, etc. as parameters instead, like `verify()` does, but with a type that isn't `VerificationAlgorithm` but instead something like `RSAParameters` or something. After all, this API is exactly like `verify` except that it takes its public key argument pre-parsed.

But, I'm not sure how to do that without making `VerificationAlgorithm` a trait, which I don't want to do for the reasons I explained before. If I have to choose between the main `verify` API forcing people to `use` traits or the `primitive` API forcing people to `use` traits, I'd prefer to put that inconvenience on the `primitive` API.

Maybe it is possible to change the type of `RSA_PKCS1_2048_8192_SHA256` et al. from `&'static VerificationAlgorithm` to `&'static <VerificationAlgorithm + RSAParameters>` and make `RSAParameters` a trait? I've never tried to do that before so i'm not sure.

Or, if that's all too complicated, then we can punt on trying to create a really uniform API and just, at least, make the minimum key size a parameter of this new function. The main thing I want to avoid doing is making the key size implicit like it is in this patch, because one of the most important design goals for *ring* is to avoid such implicit security-critical parameters.

src/rsa.rs
```diff
@@ -605,4 +605,39 @@ mod tests {
...
+            let rng = rand::SystemRandom::new();
+
+            const PRIVATE_KEY_DER: &'static [u8] =
+                include_bytes!("signature_rsa_example_private_key.der");
+            let key_bytes_der = untrusted::Input::from(PRIVATE_KEY_DER);
+            let key_pair = RSAKeyPair::from_der(key_bytes_der).unwrap();
+
+            // The output buffer is one byte too short.
+            let mut signature = vec![0; key_pair.public_modulus_len()];
+            try!(key_pair.sign(&RSA_PKCS1_SHA256, &rng, MESSAGE,
```
briansmith: Let's not call `sign` here. We have the static test vectors in the data files to avoid this pattern where we use the same code to calculate the test vector that we are then going to test. In particular, RSA verification and RSA signing share a lot of code in the modular exponentiation step (especially on 32-bit platforms) and if there is a bug in modular multiplication or modular exponentiation then it is too likely that the bug would go unnoticed with this style, since the computation of the signature and the computation of the verification of the signature might miscalculate the same way.

Plus, we already have significant performance problems with the tests on ARM where we are right (sometimes past) the timeout that Travis CI imposes, so as the set of test vectors for this grows, calling `sign` for each one would eventually be too wasteful.

The bulk of the test vectors for RSA signature verification came from NIST already in this form (separate `n` and `e` attributes) and I had to DER-encode them. We can't afford to have all the test vectors from NIST go through both this function and the normal `verify` test. But, I would like to see some assurance that all the edge cases tested through both APIs.  Maybe the best way to do that, while minimizing redundant work in the CI, is to document very clearly in `VerificationAlgorithmImpl::verify` that the test coverage for this new API depends on `VerificationAlgorithmImpl::verify` being implemented in terms of it. e.g. a comment like this immediately above the call to `verify_rsa`:

```
// The bulk of test coverage for "signature::primitive::verify_rsa"
// depends on it being called through this function. If we change
// that, then the test coverage for `verify_rsa needs to be
// reconsidered.

File: src/rsa_primitive_verify_tests.txt [hunk 4]
```diff
@@ -0,0 +1,5 @@
+# Big-endian public key parts for signature_rsa_example_public_key.der,
+# encoded in hex.
+
+n = CEA80475324C1DC8347827818DA58BAC069D3419C614A6EA1AC6A3B510DCD72CC516954905E9FEF908D45E13006ADF27D467A7D83C111D1A5DF15EF293771AEFB920032A5BB989F8E4F5E1B05093D3F130F984C07A772A3683F4DC6FB28A96815B32123CCDD13954F19D5B8B24A103E771A34C328755C65ED64E1924FFD04D30B2142CC262F6E0048FEF6DBC652F21479EA1C4B1D66D28F4D46EF7185E390CBFA2E02380582F3188BB94EBBF05D31487A09AFF01FCBB4CD4BFD1F0A833B38C11813C84360BB53C7D4481031C40BAD8713BB6B835CB08098ED15BA31EE4BA728A8C8E10F7294E1B4163B7AEE57277BFD881A6F9D43E02C6925AA3A043FB7FB78D
```
briansmith: The most useful things to test are the misuse-resistance of the API. The reason I initially designed the API to do the parsing itself is to minimize the changes that somebody could screw up; e.g. by swapping `(e, n)` for `(n, e)`. So, the tests I really want to see answer the question of "what happens if `e` and `n` are swapped. It turns out, actually, that the maximum value for `e` is smaller than the minimum value for `n`. We should have test vectors that test this: A test vector with the maximum value of `e` and the minimum value of `n` that passes, a test vector for that value of `e` - 1 and the same `n` that fails, a test tvector for the original value of `e` with original value of `n` 0 - 1 that fails. This will demonstrate what the minimum of `n` and maximum of `e` is. Then, another test vector should swap `(n, e)` as `(e, n)` with the minimum/maximum values of n/e to show that there's no way that swapped parameters would ever cause a problem.

Also, I want to see test vectors for leading zeros and for values of `n` and `e` that have the highest bit set. This will demonstrate that the API doesn't enforce the ASN.1 semantics for leading zeros. So, for example, we should show that `n = 00<whatever>` is equivalent to `n = <whatever>` and that `n = FF<whatever>` is equivalent to `00FF<whatever>`. Or, if we want to require that there's no leading zeros, then that needs to be documented and also tested.

src/rsa.rs
```diff
@@ -60,61 +60,67 @@ struct RSAVerificationAlgorithm {
...
 
 
-impl signature_impl::VerificationAlgorithmImpl for RSAVerificationAlgorithm {
-    fn verify(&self, public_key: untrusted::Input, msg: untrusted::Input,
-              signature: untrusted::Input) -> Result<(), ()> {
-        const MAX_BITS: usize = 8192;
-
-        let (n, e) = try!(parse_public_key(public_key));
-        let signature = signature.as_slice_less_safe();
+fn verify(padding: &RSAPadding, msg: untrusted::Input,
```
briansmith: I would rather just call this `verify_rsa` so it is easier to find. Most people will be looking for it under that name. Also, we need to add the end-user documentation for this function.

src/rsa.rs
```diff
@@ -60,61 +60,67 @@ struct RSAVerificationAlgorithm {
...
+        Ok(())
+    })
+}
+
+
+impl signature_impl::VerificationAlgorithmImpl for RSAVerificationAlgorithm {
+    fn verify(&self, public_key: untrusted::Input, msg: untrusted::Input,
+              signature: untrusted::Input) -> Result<(), ()> {
+        let (n, e) = try!(parse_public_key(public_key));
+        verify(self.padding_alg, msg, signature, (n, e), self.min_bits)
```
briansmith: Please see my comment about adding documentation about the test coverage of `verify_rsa` being dependent on it being called here.

Also, please call it as `signature::primitive::verify_rsa` to make that aspect crystal-clear. (Go ahead and squash all your changes to make that possible, if needed.)

src/rsa.rs
djc: So we just have `RSAParameters` that carries a reference to an `RSAPadding`, and then a `SignatureVerificationAlgorithm` that carries a reference to an `RSAParameters`? For now, `RSAParameters` can also carry `min_bits`, and then you could add more stuff later.

briansmith: > So we just have RSAParameters that carries a reference to an RSAPadding, and then a SignatureVerificationAlgorithm that carries a reference to an RSAParameters? For now, RSAParameters can also carry min_bits, and then you could add more stuff later.

That seems like one reasonable way of doing it. TBH, it's a little hard for me to tell w/o seeing some code for it.

src/rsa_primitive_verify_tests.txt
djc: So I'm back to working on this PR now, and I think I have the other comments covered, but the testing here is still hard for me to wrap my head around. Can you lay out the test matrix in a table or list a bit more, please? In particular, how do I create keys that have the maximum or minimum value for `n` (which is the public exponent, right?)? Isn't the size of `n` now covered by the `min_bits` property of `RSAParameters`?

briansmith:
```
e = 2**34 - 1 (maximum value of e).
n = <modulus generated from openssl that is slightly smaller than the smallest accepted, e.g. 2040 bits>.

e = 2**34 + 1 (maximum value of e + 2)
n = <modulus generated from openssl that is the smallest accepted, 2048 bits>.

e = 2**34 - 1 (maximum value of e)
n = <modulus generated from openssl that is the smallest accepted, 2048 bits>.

# Above, we showed what the minimum size of `n` is and what the
# maximum value of `e` is. This shows that if we (accidentally) swap
# the minimum value of n and e when verifying a signature, the
# `signature::primitive::rsa_verify` will return an error. In other
# words, there's no way the validation could succeed for any case
# where the user accidentally swapped the values of `n` and `e`
# when calling `rsa_verify`, due to the maximum/minimum
# checks on `e` and `n`.
e = <n from the previous test vector>
n = <e from the previous test vector>.
```

briansmith:
@samscott89 Your work on PSS might benefit from looking at this.

djc:
Pushed the latest version for reference.

briansmith:
Rebased: https://github.com/briansmith/ring/commits/public-key-as-der

src/rsa.rs
```diff
@@ -68,61 +68,68 @@ pub struct RSAParameters {
...
-        })
+        Ok(())
+    })
+}
+
+
+impl signature::VerificationAlgorithm for RSAParameters {
+    fn verify(&self, public_key: untrusted::Input, msg: untrusted::Input,
+              signature: untrusted::Input) -> Result<(), ()> {
+        verify(self, msg, signature, public_key)
```
briansmith: The arguments to the new `verify` function should be in the same order as the arguments to the existing `verify` functions: "public_key, msg, signature".

src/rsa.rs
```diff
@@ -127,6 +127,30 @@ fn verify(params: &RSAParameters, msg: untrusted::Input,
...
+    /// bytes for the public key components as `untrusted::Input` arguments.
+    ///
+    /// Takes an `RSAParameters` static to determine the parameters used,
+    /// a message, a signature (whose length should match the length of public
+    /// exponent, `n`) and a tuple containing the parts of the public key (in
+    /// big-endian byte order). The key length must be between 2048 and 8192 bits.
+    // Note that test coverage for this function thoroughly depends on test
+    // coverage for the `super::verify()` implementation. If we change that,
+    // test coverage for `verify_rsa()` will need to be reconsidered.
+    pub fn verify_rsa(params: &super::RSAParameters, msg: untrusted::Input,
```
briansmith: Ditto here: The public key should be before the `msg` argument, to be consistent with the other `verify` functions.

src/rsa.rs
```diff
@@ -127,6 +127,30 @@ fn verify(params: &RSAParameters, msg: untrusted::Input,
...
+    /// Lower-level verification function for RSA signatures.
+    ///
+    /// If you do not have the public key available as DER-encoded
+    /// PKCS#1-formatted ASN.1, this function can be used to pass in the raw
+    /// bytes for the public key components as `untrusted::Input` arguments.
+    ///
+    /// Takes an `RSAParameters` static to determine the parameters used,
+    /// a message, a signature (whose length should match the length of public
+    /// exponent, `n`) and a tuple containing the parts of the public key (in
+    /// big-endian byte order). The key length must be between 2048 and 8192 bits.
```
briansmith: This comment will need to be adjusted for the new argument order.

The "The key length must be between..." comment can be dropped, because that's part of `RSAParameters` now.

src/rsa.rs
```diff
@@ -127,6 +127,30 @@ fn verify(params: &RSAParameters, msg: untrusted::Input,
...
+    // Note that test coverage for this function thoroughly depends on test
```
briansmith: Blank line before this note, s/Note that/The/, and rewrap.

src/signature.rs
```diff
@@ -184,6 +184,12 @@ pub use rsa::{
     RSA_PKCS1_3072_8192_SHA384,
 };
 
+/// Lower-level verification primitives. Usage of the `signature::verify()`
+/// function is preferred.
```
briansmith: I suggest: "Lower-level verification primitives. Usage of `ring::signature::verify()` is preferred when the public key and signature is encoded in standard formats, as it also handles the parsing."

src/rsa.rs
```diff
@@ -127,6 +127,30 @@ fn verify(params: &RSAParameters, msg: untrusted::Input,
 }
 
 
+pub mod primitive {
+    use untrusted;
+    /// Lower-level verification function for RSA signatures.
+    ///
+    /// If you do not have the public key available as DER-encoded
```
briansmith: I suggest something like "When the public key is in DER-encoded PKCS#1 ASN.1 format, it is recommended to use `ring::signature::verify()` with `ring::signature::RSA_PKCS1_*`, because then `ring::signature::verify()` will handle the parsing."

Then, s/If you do not have the public key available as DER-encoded PKCS#1-formatted ASN.1,/Otherwise/

src/rsa.rs
briansmith: Update the commit message to refer to `ring::signature::primitive::verify_rsa` instead of `rsa::primitive::verify()`.

briansmith:
@djc, Please see https://github.com/briansmith/ring/commits/public-key-as-der. There are several modifications I made:

* Reorder the code in rsa.rs a bit to fit the ordering I try to use elsewhere in *ring*.
* Clarify some of the comments.
* Merge `rsa::primitive::verify_rsa` into `rsa::verify`. I think that they were separate because you were expecting originally to expose the function as `ring::rsa::primitive::verify`, and then things got a bit confused when we changed that.
* Clarify the documentation.

Please let me know if I FUBAR'd anything. Otherwise, I'll push this this weekend.

Thanks for your work, and patience, on this!

djc:
@briansmith your changes look great to me! I still have an open item to look into the extra tests you proposed. I haven't gotten to it yet, but also haven't forgotten. Hope to get to it this weekend, but would be happy if you already push these.

briansmith:
> @briansmith your changes look great to me! I still have an open item to look into the extra tests you proposed. I haven't gotten to it yet, but also haven't forgotten. Hope to get to it this weekend, but would be happy if you already push these.

Let me know if you need some help with the tests. I thought I was going to be able to do them today but I probably won't be able to as something has come up.

Also, we need to document what happens with leading zeros and leading high bits and stuff in `n` and `e`. In particular, we need to make it clear that we don't use ASN.1 semantics when parsing a leading high bit and that either leading zeros are rejected or accepted. Each of these cases should be tested too.

djc:
@briansmith I've been having resistance to starting work on these tests, probably because it's unfamiliar terrain and it feels low-value despite knowing that it really isn't. I'll need to give it a real try before I ask for help, but will ask if I don't get anywhere once I actually do.

BTW, I was planning to look at reviewing the organization of the `rsa` module after this, because it's definitely gotten a bit messy. I suppose that after #262, it will clear up a bit already, and I'll take a look at #263 after the refactoring parts of the former land.

(Did you see my questions about review practice and funding in IRC?)

djc:
@briansmith if you're happy with them, maybe it makes sense to land the refactoring commits here already, to prevent further churn with all the other refactoring going on? I'm also happy to rebase on the `RSASigningState` series.

briansmith:
> @briansmith if you're happy with them, maybe it makes sense to land the refactoring commits here already, to prevent further churn with all the other refactoring going on? I'm also happy to rebase on the RSASigningState series.

I'll merge in the refactoring changes and `RSASigningState` ASAP after finishing the more urgent refactoring of the C internals. No need for you to do anything at this point w.r.t. that refactoring. It seems increasingly unlikely I'm going to get time to write the additional tests and documentation for the new API soon, so that's what I really need help with.

djc:
> It seems increasingly unlikely I'm going to get time to write the additional tests and documentation for the new API soon, so that's what I really need help with.

Yeah, that seems sane, you have more important stuff to work on. I'll try to make progress.

src/rsa_primitive_verify_tests.txt
djc: Is it possible you made an off-by-one error here in suggesting putting `e` around 2 ** 34? According to [AGL](https://www.imperialviolet.org/2012/03/16/rsae.html), it seems like that maximum `e` should be 2 ** 32 + 1, which has a length of 33 bits. As such, it seems that `GFp_rsa_check_modulus_and_exponent()`should check that `e` is not more than 33 bits, which is actually what it does, with `e_bits > kMaxExponentBits`. However, 2 ** 34 - 1 has a length of 34 bits. I've therefore used 2 ** 33 +/- 1 in 1f450f9. If you think that's okay, I'll look into the zero padding next.

djc: 2c12b61 now has a test for zero-padding, though not for the `00FF` variety. I guess that's a separate case because of ASN.1 handling? Do you have a suggestion how I can get a key that has `n` starting with `FF`? I've also attempted a fix for the issue with the `drop_with_repr_extern` with nightly Rust, in 2099dc6. If it works, you could maybe cherry-pick it to master.

src/lib.rs
```diff
@@ -90,6 +89,8 @@
     while_true,
 )]
 
+#![cfg_attr(not(feature = "nightly"), deny(drop_with_repr_extern))]
```
briansmith: I didn't see this until I'd already written a different patch. This is smarter than what I did. I don't think we need to do this though, because in 6 weeks we'd have to remove it. Thanks though!

src/rsa/rsa_primitive_verify_tests.txt
```diff
@@ -1,7 +1,32 @@
 # Big-endian public key parts for signature_rsa_example_public_key.der,
 # encoded in hex.
 
-Msg = 68656c6c6f2c20776f726c64
-Sig = 048efbc9eb5f7a6f55f6d7b9f7e6c3ce58e2db226562ca905e7f972e8f43b6969b0ad878e0d6b290c5bbf2c05410a1efc9de051d91e5faa537e454306f5f526c828379fe28a17e50c8bd4e7c834479da482305a78e198c988a177b9263cea27a2a99c0da98e03b0cc8d880eccdeba7c16dd07f78d980739753690953d1b63106145a80059ed38f52100a9a8d2c7c5371d91b70ce5b7b36d6b97ebef8798d09c01e5b6cb8a6a7fd1a4100d3527327b7d23f8a26187985d8702f8951346ea4a7253e87f765ef587a728021bff37be55d1a8639809e3453ea5a2da482bfedeae18579b51037cfecff5bece21d8c82ee6fa8eb0f43c43c3a23a983c3a2eea4e7d2dc
 n = CEA80475324C1DC8347827818DA58BAC069D3419C614A6EA1AC6A3B510DCD72CC516954905E9FEF908D45E13006ADF27D467A7D83C111D1A5DF15EF293771AEFB920032A5BB989F8E4F5E1B05093D3F130F984C07A772A3683F4DC6FB28A96815B32123CCDD13954F19D5B8B24A103E771A34C328755C65ED64E1924FFD04D30B2142CC262F6E0048FEF6DBC652F21479EA1C4B1D66D28F4D46EF7185E390CBFA2E02380582F3188BB94EBBF05D31487A09AFF01FCBB4CD4BFD1F0A833B38C11813C84360BB53C7D4481031C40BAD8713BB6B835CB08098ED15BA31EE4BA728A8C8E10F7294E1B4163B7AEE57277BFD881A6F9D43E02C6925AA3A043FB7FB78D
```
briansmith: Please add a comment that labels each test: e.g. "Maximum value of `e`,", `"Minimal value of `n`", "minimal `n` and maximal `e` swapped".

src/rsa/rsa_primitive_verify_tests.txt
```diff
@@ -7,6 +7,12 @@ Msg = 68656c6c6f2c20776f726c64
 Sig = 048efbc9eb5f7a6f55f6d7b9f7e6c3ce58e2db226562ca905e7f972e8f43b6969b0ad878e0d6b290c5bbf2c05410a1efc9de051d91e5faa537e454306f5f526c828379fe28a17e50c8bd4e7c834479da482305a78e198c988a177b9263cea27a2a99c0da98e03b0cc8d880eccdeba7c16dd07f78d980739753690953d1b63106145a80059ed38f52100a9a8d2c7c5371d91b70ce5b7b36d6b97ebef8798d09c01e5b6cb8a6a7fd1a4100d3527327b7d23f8a26187985d8702f8951346ea4a7253e87f765ef587a728021bff37be55d1a8639809e3453ea5a2da482bfedeae18579b51037cfecff5bece21d8c82ee6fa8eb0f43c43c3a23a983c3a2eea4e7d2dc
 Result = Pass
 
+n = 00CEA80475324C1DC8347827818DA58BAC069D3419C614A6EA1AC6A3B510DCD72CC516954905E9FEF908D45E13006ADF27D467A7D83C111D1A5DF15EF293771AEFB920032A5BB989F8E4F5E1B05093D3F130F984C07A772A3683F4DC6FB28A96815B32123CCDD13954F19D5B8B24A103E771A34C328755C65ED64E1924FFD04D30B2142CC262F6E0048FEF6DBC652F21479EA1C4B1D66D28F4D46EF7185E390CBFA2E02380582F3188BB94EBBF05D31487A09AFF01FCBB4CD4BFD1F0A833B38C11813C84360BB53C7D4481031C40BAD8713BB6B835CB08098ED15BA31EE4BA728A8C8E10F7294E1B4163B7AEE57277BFD881A6F9D43E02C6925AA3A043FB7FB78D
```
briansmith: I expected to see a test like "Same test as above, but with `n` is prefixed with a leading zero".

Also, we should have a "Same test as above, but with `n` has a trailing zero appended" which should fail.

src/rsa/rsa_primitive_verify_tests.txt
```diff
@@ -7,6 +7,12 @@ Msg = 68656c6c6f2c20776f726c64
...
+e = 260445
+Msg = 68656c6c6f2c20776f726c64
+Sig = 048efbc9eb5f7a6f55f6d7b9f7e6c3ce58e2db226562ca905e7f972e8f43b6969b0ad878e0d6b290c5bbf2c05410a1efc9de051d91e5faa537e454306f5f526c828379fe28a17e50c8bd4e7c834479da482305a78e198c988a177b9263cea27a2a99c0da98e03b0cc8d880eccdeba7c16dd07f78d980739753690953d1b63106145a80059ed38f52100a9a8d2c7c5371d91b70ce5b7b36d6b97ebef8798d09c01e5b6cb8a6a7fd1a4100d3527327b7d23f8a26187985d8702f8951346ea4a7253e87f765ef587a728021bff37be55d1a8639809e3453ea5a2da482bfedeae18579b51037cfecff5bece21d8c82ee6fa8eb0f43c43c3a23a983c3a2eea4e7d2dc
+Result = Pass
```
briansmith: What is the justification for this passing? Is there some protocol like JWS or similar that requires support for leading zeros? I had expected that we should have this test fail.

Also, we should have similar cases for leading zeros in `e`.

Finally, we should document the behavior of leading zeros in the doc comments for `verify_rsa`.

briansmith:
I updated https://github.com/briansmith/ring/commits/public-key-as-der with your patches rebased on top of the previous version and on top of the current master.

src/rsa_primitive_verify_tests.txt
briansmith: On Sat, Aug 27, 2016 at 3:07 AM, Dirkjan Ochtman <notifications@github.com>
wrote:

> Is it possible you made an off-by-one error here in suggesting putting e
> around 2 ** 34? According to AGL
> <https://www.imperialviolet.org/2012/03/16/rsae.html>, it seems like that
> maximum e should be 2 ** 32 + 1, which has a length of 33 bits. As such,
> it seems that GFp_rsa_check_modulus_and_exponent()should check that e is
> not more than 33 bits, which is actually what it does, with e_bits >
> kMaxExponentBits. However, 2 ** 34 - 1 has a length of 34 bits. I've
> therefore used 2 ** 33 +/- 1 in 1f450f9
> <https://github.com/briansmith/ring/commit/1f450f9fbe2f69becfe31a14d96b734a9aafa6a8>.
> If you think that's okay, I'll look into the zero padding next.
>
Yes, do what I mean, not what I say! :) The important thing is to test that
bounds check.

Cheers,
Brian
-- 
https://briansmith.org/

djc:
The latest should address all your comments, and has your comment extension folded in. I hope we're converging!

djc:
So my change to `BN_bin2bn()` in d77a179 appears to have broken RSA signing, where I guess there can be leading zero coming out of the parsing of DER-encoded ASN.1. I tried to remedy it with something like this, but it doesn't seem to work. Any suggestions?

```diff
diff --git a/src/rsa/rsa.rs b/src/rsa/rsa.rs
index f540c3b..7d2870e 100644
--- a/src/rsa/rsa.rs
+++ b/src/rsa/rsa.rs
@@ -126,8 +126,14 @@ impl PositiveInteger {
     fn from_der(input: &mut untrusted::Reader)
                 -> Result<PositiveInteger, error::Unspecified> {
         let bytes = try!(der::positive_integer(input)).as_slice_less_safe();
+        let raw_bytes = if bytes[0] == 0 {
+                &bytes[1..]
+            } else {
+                &bytes[..]
+            };
         let res = unsafe {
-            GFp_BN_bin2bn(bytes.as_ptr(), bytes.len(), core::ptr::null_mut())
+            GFp_BN_bin2bn(raw_bytes.as_ptr(), raw_bytes.len(),
+                          core::ptr::null_mut())
         };
         if res.is_null() {
             return Err(error::Unspecified);
```

briansmith:
> So my change to BN_bin2bn() in d77a179 appears to have broken RSA signing, where I guess there can be leading zeroes coming out of the parsing of DER-encoded ASN.1. I tried to remedy it with something like this, but it doesn't seem to work. Any suggestions?

`der::positive_integer` already strips leading zeros; if it doesn't, that's a bug that I'll fix ASAP. Your patch to `PositiveInteger::from_der` doesn't account for the case of `[0]`, i..e. in the case of the actual value zero, the result shouldn't be an empty slice.

But, also, the point of this PR is to completely bypass the need for DER decoding in the first place, right? I think instead we need to add the leading zero detection within `verify_rsa`. Maybe we should have a (Rust?) function that acts like a wrapper around `BN_bin2bn` that checks that there isn't a leading zero before calling `BN_bin2bn`? Something like the following (untested; I'm traveling):

```
impl PositiveInteger {
    fn from_be_bytes(input: untrusted::Input) -> Result<PositiveInteger, error::Unspecified> {
        if input.len() > 0 {
            try!(input.read_all(error::Unspecified, |input| {
                if try!(input.read_byte()) == 0 {
                }
            });
        }
        let res = GFp_BN_bin2bn(input.as_slice_less_safe().as_ptr(), input.len(),
                                core::ptr::null_mut());
       [...]
    }
}
```

Note that `PositiveInteger::from_der` could probably be then be implemented as a call to `der::positive_integer()` followed by a call to such a `from_be_bytes()` function.

djc:
Well, the problem doesn't show up for verification, only for signing, so I'm not sure it makes sense to add something to `verify_rsa()`? I can try out your `BN_bin2bn()` wrapper and see if it resolves the issue.

djc:
I followed through on your suggestion, in 18fdc13. The result seems sane, and passes the tests for me even with `rsa_signing` enabled. Note that this required moving some code in `rsa::rsa` out from behind the `rsa_signing` guard.

briansmith:
This looks good. I made some adjustments for the new `PositiveInteger` constructor. Please see the last commit at https://github.com/briansmith/ring/tree/public-key-as-der. WDYT?

djc:
It looks good to me.

So you want tests for both the low-level and higher-level APIs, even if the lower-level API is covered by tests for the higher-level one?

What's the merit of `Reader::peek()` over just indexing into the slice here?

Also, note that "indention" is an archaic word for "indentation".

When this is merged, would you mind bumping *ring* on crates.io?

briansmith:
> So you want tests for both the low-level and higher-level APIs, even if the lower-level API is covered by tests for the higher-level one?

I try to add tests where they are helpful, without getting carried away. In this case, I think there was a bug where `[]` was accepted as zero even though it shouldn't have been. It was easier to test that this way.

> What's the merit of Reader::peek() over just indexing into the slice here?

`Reader::peek()` cannot panic. In general, we avoid indexing into slices whenever possible in *ring*, by using `untrusted` instead.

> Also, note that "indention" is an archaic word for "indentation".

:) Too late, I guess. Thanks though.

> When this is merged, would you mind bumping ring on crates.io?

Sure!

briansmith:
https://crates.io/crates/ring/0.4.1
https://briansmith.org/rustdoc/ring/signature/primitive/index.html

5778edb 0.4.1 release.
771dbda Clarify encoding of `n` and `e` in `verify_rsa`.
c1ad0ba PositiveInteger::from_be_bytes: Add minimal tests; reject empty inputs.
4637a96 Add tests for zero-padding public key components with primitive verification.
15cd1d0 Test misuse-resistance for signature::primitive::verify_rsa().
0f5cfd5 Use rsa::PositiveInteger for public key parts in verification.
931800b Expand historical context of testing of `verify_rsa`.
f3452ef Expose rsa::verify as ring::signature::verify_rsa.
09484b1 Fix indention in src/rsa/rsa.rs.
80e96e0 Change rsa::parse_public_key to return components as Input.
9c85dfd Extract rsa::verify_rsa function from RSAParameters impl.

Thanks a lot for contributing this!

PR #227, opened 2016-06-23
Validate ECDSA and ECDH inputs using Rust code

briansmith:


briansmith:
This landed in master.

PR #230, opened 2016-06-26
Test if *ring* compiles without default features.

### LINKED ISSUE #222: Add `--no-default-features` configurations to CI
We need to verify that people don't break the builds when the functionality that uses the heap is disabled, in particular.

pyfisch:
Closes #222

### DIFF
--- a/mk/update-travis-yml.py
+++ b/mk/update-travis-yml.py
@@ -61,7 +61,8 @@ compilers = {
 
 modes = [
     "DEBUG",
-    "RELWITHDEBINFO"
+    "RELWITHDEBINFO",
+    "NODEFAULTFEATURES",
 ]
 
 # Mac OS X is first because we don't want to have to wait until all the Linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -147,6 +147,11 @@ matrix:
       os: osx
       osx_image: xcode7.2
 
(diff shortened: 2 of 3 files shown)


mk/update-travis-yml.py
```diff
@@ -61,7 +61,8 @@
 
 modes = [
     "DEBUG",
-    "RELWITHDEBINFO"
+    "RELWITHDEBINFO",
+    "NODEFAULTFEATURES"
```
djc: Nit: add a trailing comma here, so that next time you don't have to change this line if you're adding more.

pyfisch:
fixed nit, added copyright statement to commit message.

pyfisch:
@briansmith running `cargo build --no-default-features` does not assure that the tests pass without `use_heap`.

briansmith:
> @briansmith running cargo build --no-default-features does not assure that the tests pass without use_heap.

Fair enough. I took the "compiles" in the title of your PR too literally.

At any rate, I do think that this is a good thing to do. We just need to find some way to limit the amount of resources this consumes. I suggest that, for example, we limit this to one version of GCC per platform and and Linux only. I don't think the no-heap mode is useful for Mac OS X, iPhone, or Android. That would give us {ARM, Aarch64, x86, x86-64} * {1 version of GCC} * {debug, release} = 8 new configurations to test, instead of 20 new configurations.

The issue is that Travis CI already takes *hours* to finish, because we only get 5 concurrent builds and we have a large number already.

PR #231, opened 2016-06-29
Rename RSA signature verification algorithm constants.

djc:
As discussed before.

Question: do we rename ECDSA/EDDSA-related constants to remove the VERIFY as well?


briansmith:
If we're going to take the `_VERIFY` off the end of the RSA ones then we should also do the same for ED25519 and ECDSA.

The unfortunate thing about this change, IIUC, is that the output of `cargo doc` will now interleave the `RSAPadding` and `VerificationAlgorithm` statics.

djc:
I was thinking about this some more last night. How about we remove the RSAPadding statics from the public API, anyway? I think the full verification algorithm statics that additionally reference a minimum key size can be used for signing just fine. This makes signing and verification APIs more symmetric, which I think is a good thing.

In general, I feel that we should try harder to make signing (and possibly generation) more generic across algorithms. Signing takes an algorithm and appropriate keypair (or the key pair can carry the algorithm), a message slice and a mutable signature destination slice. Verification takes a public key (in the form of DER-encoded Input or, in the case of rsa::primitive, two Inputs), a message Input and a signature Input. Generate takes no arguments but returns a key pair. For RSA, we'd put the RNG signing param in the key pair instantiation rather than the sign() function. The key pair can also have a generic signature_length() method (which would replace public_modulus_len() in the RSA case).

I forget exactly how you articulated your concerns with having traits or the RSAPublicKey (did it have to do with heap usage?), and either way, this could probably implemented in different ways anyway. WDYT?

briansmith:
> I think the full verification algorithm statics that additionally reference a minimum key size can be used for signing just fine. This makes signing and verification APIs more symmetric, which I think is a good thing.

In general, I do very much prefer that symmetry. But, I am skeptical that it will be reasonable to actually implement the RSA signing in the long term with an API that is symmetric with the RSA verification. In particular, I'm doubting we'll be able to support the same range of key sizes for signing. In fact, we already don't--we only support up to 4096.

> In general, I feel that we should try harder to make signing (and possibly generation) more generic across algorithms.

Yes. I agree. I expect that the signing and generation interface will change a lot. We had agreed to land the interfaces we did so that we could get something working and learn from things. I am very open to improvements and we have absolutely no backward compatibility constraints--we can refactor the API from commit to commit as we please.

> For RSA, we'd put the RNG signing param in the key pair instantiation rather than the sign() function. 

I think are more complex than this in real life. Consider a TLS server. It needs to sometimes sign with SHA-256 and sometimes SHA-384 and/or even SHA-1 depending on what the client's advertised capabilities are. Also, because the RSA keys are so big, and because the way one manages the `BN_BLINDING`s really matters a lot in a high-performance server, there ends up being a lot of complexity that I think isn't being accounted for yet. ECDSA and EdDSA are more easy to do like you suggest. I'm not sure if/how we can do RSA the same way yet.

> The key pair can also have a generic signature_length() method (which would replace public_modulus_len() in the RSA case).

Just FYI, in the ECDSA case, the typical ASN.1-encoded signatures are variable-sized: One cannot know the exact length of the signature until it is calculated. That's one of the reasons the `Ed25519KeyPair` returns a `Signature` object, to kind of hide that issue with the upcoming ECDSA signing integration.

> I forget exactly how you articulated your concerns with having traits or the RSAPublicKey (did it have to do with heap usage?), and either way, this could probably implemented in different ways anyway. WDYT?

In general, I want a more uniform interface. More important for me, though, is getting the functionality we need implemented (in particular, the blinding management, and ECDSA) so that everybody involved can see the complexity involved in trying to unify things into a single interface.

djc:
> In general, I do very much prefer that symmetry. But, I am skeptical that it will be reasonable to actually implement the RSA signing in the long term with an API that is symmetric with the RSA verification. In particular, I'm doubting we'll be able to support the same range of key sizes for signing. In fact, we already don't--we only support up to 4096.

In that case, can we live with the crappy ordering in the docs for now?

> Yes. I agree. I expect that the signing and generation interface will change a lot. We had agreed to land the interfaces we did so that we could get something working and learn from things. I am very open to improvements and we have absolutely no backward compatibility constraints--we can refactor the API from commit to commit as we please.

Even after we publish to crates.io? How do you want to manage versions/API changes after that? There's a clear trade-off in making *ring* more easily available and still iterating on large parts of the API surface.

Implementing more functionality first so we can figure out where the complexities actually are makes a lot of sense to me.

briansmith:
> In that case, can we live with the crappy ordering in the docs for now?

Yes, that's not blocking this.

djc:
Removed `_VERIFY` from all the signature algorithms.

briansmith:
I merged these changes, but then I reverted the part that changes the naming pattern of the `RSA_PKCS1` constants. Maybe we'll still do that renaming, but I want to explore another idea first. I reverted that part to avoid extra churn in `webpki`'s API. (Unfortunately, I only decided to do that after changing webpki's API, so I had to revert the change in webpki too.)

Removing the `_VERIFY` suffix seems like a good usability improvement and I kept that part. I hope we don't have to undo it in the future when we try to create a unified signing API.

Previously, `ECDSA_P256_SHA256` and `RSA_PKCS1_2048_8192_SHA256` followed a common pattern of {algorithm}_{strength}_{digest}. Now that's no longer the case, which disappoints me. I agree it was confusing to have `RSAPadding`s be named `RSA_{padding}_{digest}` but RSA `SignatureAlgorithm`s to be named `RSA_{strength}_{padding}_{digest}`. But, isn't the problem really with `RSAPadding`?

I think instead we should change `RSAPadding` to `RSAParameters` and add the key strength range to it, then rename the `RSAPadding` constants to `RSA_PKCS1_{strength}_{digest}`. This will also solve a problem I didn't notice previously with the new RSA signing API. See my commit in that PR.

briansmith:
Closing this. If we decide to revert my revert of the naming pattern changes, let's do it in a separate PR. Thanks!

PR #238, opened 2016-07-02
Make `RSAKeyPair` conditional on a new `rsa_key_pair` feature.

briansmith:
This is one of the changes necessary to drop `use_heap` from the
default feature set.


briansmith:
This landed.

Issue #243, opened 2016-07-09
parse_big_endian_value wrong for inputs not a multiple of the limb length


ctz:
I've noticed that sometimes ring rejects valid ECDSA signatures. I've tracked this down to a bug in `parse_big_endian_value`. For example, it parses `0x0e0d0c0b0a09080706050403020100` to `[0x0d0c0b0a09080706, 0x0e, 0, ...]`. The same bug causes a panic in other length inputs.

I'm preparing a PR at the minute with a fix, sample signatures, and some extra unit tests.

briansmith:
Thanks! For the testing, you should be able to just put the new test vectors into src/ec/suite_b/ecdsa_verify_tests.txt.

briansmith:
I see. I thought I had fixed this already but maybe I only fixed the case where the short value had no more than one limb?

Please confirm that you are the sole copyright holder on your contributions to *ring* and/or please identify who the copyright holder is. Then I can merge this. Thanks!

briansmith:
Thanks for the bug report and for the PR.

I am going to look into fixing this a different way. The code worked left-to-right instead of right-to-left because originally I used `untrusted::Input` to read the bytes. That old code would have avoided the panic. But, I changed the code to stop using `untrusted::Input` because, at the time, `Input::new()` had return type `Result<(), ()>`, and I didn't want to add an `unwrap()`. But now, `untrusted::Input::from()` is infallible so we should be able to go back to using `untrusted::Input` like I originally wanted.

ctz:
Yes, I'm the copyright holder. :+1:

briansmith:
The fix for this landed as 669d6ca869cf1a112f3eea0a893099be1f7062dd,
b609d69771820b92a2782de9b825947f388da2cf, and 2a4ce0b466d649b7c09a9b38a840372a7d060568,

Thanks for reporting the issue and for providing the patch! As I indicated above, I decided to use `untrusted::Input` for the parsing, so I couldn't switch to the right-to-left mechanism.

Please take a look at the above commits and close the issue if you think it is resolved. Thanks again!

ctz:
yup, wfm!

PR #244, opened 2016-07-09
Test and fix parse_big_endian_value

ctz:
See issue #243


briansmith:
Thanks for reporting the issue and for providing the patch! As I indicated above, I decided to use `untrusted::Input` for the parsing, so I couldn't switch to the right-to-left mechanism. If I hadn't switched to `untrusted::Input` then your patch would have been good.

briansmith:
Oh, I should add: I did commit the test changes, though I had to modify them in order to work on 32-bit targets.

PR #247, opened 2016-07-09
New P-384 implementation.

briansmith:
The primary goal of this implementation is to reduce the variance between the P-384 and P-256 implementations so that analysis of one is more likely to apply to the other, and so that we can handle everything (e.g. the special form of the point at infinity in nistz256) uniformly. The secondary goal is to remove usage of the heap, which means removing dependencies on `BIGNUM`, `EC_POINT`, etc. The tertiary goal is to improve the side channel protection, particularly protection against timing attacks. The final goal is to improve performance significantly.


briansmith:
I think I fixed all the nits. This landed.

PR #248, opened 2016-07-11
Clarify the bit shift in gfp_little_endian_bytes_from_scalar.

Ms2ger:
The bitshift-assignment in the middle of an expression is
unneccesarily confusing, and in any case is not necessary if BN_BYTES
is not 8.

### DIFF
--- a/crypto/ec/ecp_nistz.c
+++ b/crypto/ec/ecp_nistz.c
@@ -40,14 +40,14 @@ void gfp_little_endian_bytes_from_scalar(uint8_t str[], size_t str_len,
     str[i + 0] = d & 0xff;
     str[i + 1] = (d >> 8) & 0xff;
     str[i + 2] = (d >> 16) & 0xff;
-    str[i + 3] = (d >>= 24) & 0xff;
-    if (BN_BYTES == 8) {
-      d >>= 8;
-      str[i + 4] = d & 0xff;
-      str[i + 5] = (d >> 8) & 0xff;
-      str[i + 6] = (d >> 16) & 0xff;
-      str[i + 7] = (d >> 24) & 0xff;
-    }
+    str[i + 3] = (d >> 24) & 0xff;
+#if BN_BYTES == 8
+    d >>= 32;
+    str[i + 4] = d & 0xff;
+    str[i + 5] = (d >> 8) & 0xff;
+    str[i + 6] = (d >> 16) & 0xff;
+    str[i + 7] = (d >> 24) & 0xff;
+#endif
   }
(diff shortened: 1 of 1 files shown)


crypto/ec/ecp_nistz.c
```diff
@@ -40,14 +40,14 @@ void gfp_little_endian_bytes_from_scalar(uint8_t str[], size_t str_len,
...
-    str[i + 3] = (d >>= 24) & 0xff;
-    if (BN_BYTES == 8) {
-      d >>= 8;
-      str[i + 4] = d & 0xff;
-      str[i + 5] = (d >> 8) & 0xff;
-      str[i + 6] = (d >> 16) & 0xff;
-      str[i + 7] = (d >> 24) & 0xff;
-    }
+    str[i + 3] = (d >> 24) & 0xff;
+#if BN_BYTES == 8
```
briansmith: Hi. AGL decided that we should avoid `#if` in favor of using `if` in these situations. I don't know if it is in the official style guide or not. Basically, we only use the preprocessor the minimal amount.

Otherwise, I agree with the main part of your change.

Ms2ger: That was what I did first, but then you get a warning-as-error about shifting by the width of the integer. I could try to trick that warning if you prefer.

briansmith: I see.

Let's zoom out a bit. In the Rust code, we store private keys as arrays of bytes already. For P-256 and P-384, we store them in big-endian order, for Curve25519 we store them in little-endian order. For P-256 and P-384, we take the big-endian byte arrays and convert them to an array of `Limb`s (`Limb` == `GFp_Limb` == `BN_ULONG`). Then, this function converts them into a little-endian byte array. So, you see, it is circuitous.

In the p384 branch, I have new code that builds on this code. And, in the upcoming `ecdsa_verify` branch I will have more code to make it easier to switch the callers of this code to pass in byte arrays instead of `Limb` arrays. So, it doesn't look like this code will have a long life. Maybe we should just leave it as it is, and wait for this grossness to die a natural death?

(FWIW, this isn't new code, but instead code that was inherited from BoringSSL & OpenSSL.)

Ms2ger: That works for me.

briansmith:
OK, I'm going to close this. The discussion above showed why the code is the way it is; it's basically an unfortunate artifact of the BoringSSL coding style. I don't think it's worth changing just this little bit, especially because I expect soon we'll be able to get rid of this function completely.

PR #250, opened 2016-07-16
Implements err::Result

thejpster:
I've implemented err::Result and err::EmptyResult, leaving the Err case as (). To me, the code looks clearer now.


briansmith:
I think we don't want to qualify `Result` as `err::Result`  all over the code.

I'm still still uncertain this is an improvement. Now the user of the library has to look up the documentation for `EmptyResult` to see that it is `Result<(), ()>`. I remember having this problem with `std::io::Result` myself. `Result<(), ()>` is ugly but it doesn't have that problem.

Let's think about this in the context of your other suggestion, which is to report errors with some error codes, not just `()` to indicate that the operation failed. There are two approaches that I see in common use. One approach is to define a single big `Error` enum that contains every single error code that the library can generate. The other approach is to have multiple enums, e.g. `PrivateKeyError` for `RSAKeyPair::from_der` might have `SyntaxError` and `InconsistentParameters` and `PLessThanQ` variants, while `SignatureVerificationError` may have `MalformedPublicKey` and `SignatureNotValidForGivenKey` errors, or whatever. I guess the single enum pattern doesn't seem very appealing to me, but I guess the approach in this bug is a step towards the single big `Error` enum?

thejpster:
To your first point - I think I agree actually. You didn't have any declarations in the top level lib.rs, so I didn't want to put any there, but actually that's probably the right place. For example, std::io::Result is in std::io, not std::io::err. I'm happy to re-work this.

To your second point, this is Don't Repeat Yourself in action. To get a consistent error handling mechanism throughout any given library, I think it is best to define it once and use it everywhere. Someone might need to look it up the custom Result type the first time they see it but then it's effectively 'cached' and doesn't require any more thought. See https://doc.rust-lang.org/book/error-handling.html#the-result-type-alias-idiom. Mentally string matching one Result&lt;(), ()&gt; with another Result&lt;(), ()&gt; and knowing they could be treated the same would, I suggest, be more difficult. In general, I think copying stable API idioms from the Rust standard library is to be considered A Good Thing - they've probably thought about it more than I have.

To your third point, in my mind the two key points about having the errors harmonised are a) you can try!() in multiple levels in your call stack and the errors will bubble up (without loosing any details about the error and/or having to implement a bunch of Into traits) and b) it means the users of the library know they can call any of the API calls and get a consistent error they can return and/or handle. If the various API calls returned different types, consider how you would have to implement the sign_and_verify_rsa() doctest function would be in signature.rs.

I think this library is sufficiently small and the set of errors sufficiently small that placing them in a single enum would be fine. An alternative approach would be to return a &'static str - you couldn't programatically distinguish between a nonce that's too short and an output buffer that's too short, and you'd be tied to English langage responses (probably), but it would avoid pre-declaring the full set of errors codes in a single place.

briansmith:
Let's move the discussion that's not specific to this PR over to the issue. I'll reply over there.

src/err.rs
```diff
@@ -0,0 +1,26 @@
...
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+//! Implements types influenced by `std::io::Result` for indicating success or failure
+//! of a routine.
+
+use std::result;
+
+/// Used by routines that return a certain type if they pass, or
+/// wish to indicate failure without any sense of why they failed.
+pub type Result<T> = result::Result<T, ()>;
```
briansmith: Unfortunately, this seems to be working too poorly, looking at the rest of the PR. `::Result` vs `result::Result` and the caveat about `use ring::*` make this cure seem worse than the disease. So, I think we should just drop Result completely.

src/err.rs
```diff
@@ -0,0 +1,26 @@
...
+
+/// Used by routines that wish to indicate success or failure, with
+/// no associated value either way.
+pub type EmptyResult = Result<()>;
```
briansmith: The main goal is to make it possible to implement the `From` trait to convert a *ring* error into an application-specific error.

Instead of this, I think we need something like: `enum UnspecifiedError;` and then change the return types of functions from `Result<R, ()>` to `Result<R, UnspecifiedError>`. Then the application can implement `From<ring::UnspecifiedError>` or `From<Result<R, ring::UnspecifiedError>>` to get the conversion as described in https://doc.rust-lang.org/book/error-handling.html#composing-custom-error-types.

We could also implement `std::error::Error` for `UnspecifiedError` to get the "default" conversion as specified here: https://doc.rust-lang.org/book/error-handling.html#the-from-trait.

thejpster: Could you call the enum RingError and give it a single member called
Unspecified? It would give you options in the future for being more
specific with errors and without requiring a second breaking change to the
API.

I agree the result::Result/Result confusion is a bit of a mess. If you get
the API right first time (or at least this time) then aliasing the return
type in this way could be argued as unnecessary.

On 8 Aug 2016 02:04, "Brian Smith" <notifications@github.com> wrote:

In src/err.rs
<https://github.com/briansmith/ring/pull/250#discussion_r73813197>:

> +// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
> +// OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
> +// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
> +
> +//! Implements types influenced by `std::io::Result` for indicating success or failure
> +//! of a routine.
> +
> +use std::result;
> +
> +/// Used by routines that return a certain type if they pass, or
> +/// wish to indicate failure without any sense of why they failed.
> +pub type Result<T> = result::Result<T, ()>;
> +
> +/// Used by routines that wish to indicate success or failure, with
> +/// no associated value either way.
> +pub type EmptyResult = Result<()>;

The main goal is to make it possible to implement the From trait to convert
a *ring* error into an application-specific error.

Instead of this, I think we need something like: enum UnspecifiedError; and
then change the return types of functions from Result<R, ()> to Result<R,
UnspecifiedError>. Then the application can implement
From<ring::UnspecifiedError> or From<Result<R, ring::UnspecifiedError>> to
get the conversion as described in https://doc.rust-lang.org/
book/error-handling.html#composing-custom-error-types.

We could also implement std::error::Error for UnspecifiedError to get the
"default" conversion as specified here: https://doc.rust-lang.org/
book/error-handling.html#the-from-trait.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/briansmith/ring/pull/250/files/0b98364802124e49957ff6203bd45677fdc2fa80#r73813197>,
or mute the thread

briansmith: > Could you call the enum RingError and give it a single member called
> Unspecified? It would give you options in the future for being more
> specific with errors and without requiring a second breaking change to the
> API.
> 
> I agree the result::Result/Result confusion is a bit of a mess. If you get
> the API right first time (or at least this time) then aliasing the return
> type in this way could be argued as unnecessary.

I have an idea. Please see https://github.com/briansmith/ring/tree/error. I created `ring::error::Unspecified` which is a unit type. I also created `untrusted::EndOfInput` and a `From<EndOfInput>` implementation for `ring::error::Unspecified`. I also made `ring::error::Unspecified` implement `std::error::Error`.

Here's the idea. In the future, we might have some functions return more specific error information, maybe. If so, then we can have a trait:

```

mod error  {
    #[cfg(feature = "use_heap")]
    trait Error :  std::error::Error { }

    #[cfg(not(feature = "use_heap"))]
    trait Error { }

    impl Error for Unspecified { }
}

mod signature {
    #[derive(Clone, Copy, Debug, PartialEq)]
    enum Error {
        WeakPublicKey,
        InvalidSignatureForGivenKey,
        SyntacticallyInvalidSignature,
    }

    impl error::Error for VerificationError { }
}
```

Then users that want to abstract over all *ring* errors could do so using the `ring::error::Error` type. But, I actually doubt that such a type would be useful except when an application is being prototyped.

There's a specific reason that I don't want to use an `enum Error { Unspecified }` that might get extended later with more error codes. Some functions really can't safely return anything more than a binary result. `ring::constant_time::verify_slices_are_equal` is a good example. If we use an `Error` enum like you suggest, then the type system can't enforce that `verify_slices_are_equal` only returns a binary result and not a more nuanced (and unsafe) result. This is the case for quite a few other functions where side-channel differentiation of errors is a significant problem.

briansmith: OK, I think we're about as close to agreement as we're going to get on the basic issue that we shouldn't use `()` for the error type. I pushed to master a change that replaces `()` with `ring::error::Unspecified`. This is not exactly what you suggested but actually it is not too far off.

If you still want to try to clarify the implementation of the code by having some kind of aliases for `Result<(), error::Unspecified>` and/or `Err(error::Unspecified)` and/or `Ok(())`, I am open to discussing that. But, that should probably happen either in the issue or in another PR.

Thanks for pushing for these improvements!

PR #251, opened 2016-07-17
Really fix Mac build

tarcieri:
Without this change, I'm getting errors like:

    crypto/ec/gfp_p384.c:69:20: error: unused function 'elem_sqr_mul_mont' [-Werror,-Wunused-function]
    static inline void elem_sqr_mul_mont(Elem r, const Elem a, size_t squarings,
                       ^
                       1 error generated.


briansmith:
Thanks!

PR #252, opened 2016-07-18
Support pbkdf2 outputs larger than prf.digest_alg.output_len

tarcieri:
My main motivation for this is to implement scrypt on top of ring::pbkdf2, which requires support for this from the pbkdf2 implementation.

One of the test vectors already included in the pbkdf2 text file (the one with a non-ridiculous iteration count) now passes.

I'm curious if there's a better way to do the big endian conversion for the block counter.


src/pbkdf2.rs
```diff
@@ -150,9 +149,39 @@ pub fn derive(prf: &'static PRF, iterations: usize, salt: &[u8], secret: &[u8],
...
+        idx += 1;
+        let chunk_len = chunk.len();
+
+        if chunk_len == output_len {
+            derive_block(&secret, iterations, salt, idx, chunk);
+        } else {
+            let mut tmp = Vec::with_capacity(output_len);
+            for _ in 0..output_len {
+                tmp.push(0u8);
+            }
```
frewsxcv:
```
let tmp = vec![0; output_len];
```

https://doc.rust-lang.org/std/macro.vec!.html

src/pbkdf2.rs
```diff
@@ -105,6 +105,7 @@
 //!     assert!(db.verify_password("alice", "@74d7]404j|W}6u").is_ok());
 //! }
 
+use std::vec::Vec;
```
briansmith: This code can't use `Vec` because `Vec` requires a heap, which this code isn't supposed to require.

src/pbkdf2.rs
frewsxcv: Considering [this](https://github.com/briansmith/ring/pull/252/files#r71096506):

```rust
let tmp = [0; output_len];
```

src/pbkdf2.rs
```diff
@@ -150,9 +149,39 @@ pub fn derive(prf: &'static PRF, iterations: usize, salt: &[u8], secret: &[u8],
...
+
+        if chunk_len == output_len {
+            derive_block(&secret, iterations, salt, idx, chunk);
+        } else {
+            let mut tmp = Vec::with_capacity(output_len);
+            for _ in 0..output_len {
+                tmp.push(0u8);
+            }
+
+            derive_block(&secret, iterations, salt, idx, &mut tmp);
```
briansmith: Why can't you do this?:

```
derive_block(&secret, iterations, salt, idx, &mut chunk[(idx * output_len)..((idx * output_len) + output_len)];
```

src/pbkdf2.rs
```diff
@@ -150,9 +149,39 @@ pub fn derive(prf: &'static PRF, iterations: usize, salt: &[u8], secret: &[u8],
     // Clear |out|.
     polyfill::slice::fill(out, 0);
```
briansmith: It seems like this should go in `derive_block`?

src/pbkdf2.rs
```diff
@@ -150,9 +149,39 @@ pub fn derive(prf: &'static PRF, iterations: usize, salt: &[u8], secret: &[u8],
...
 
+    let mut idx: u32 = 0;
+
+    for chunk in out.chunks_mut(output_len) {
+        idx += 1;
```
briansmith: Add a comment explaining why this won't overflow.

src/pbkdf2.rs
briansmith: Actually, this is probably more efficient than putting it in `derive_block`.

src/polyfill.rs
```diff
@@ -45,6 +45,14 @@ pub mod slice {
         u32::from(buffer[3])
     }
 
+    #[inline(always)]
+    pub fn u8_as_be_u32(dest: &mut [u8], value: u32) {
```
briansmith: This should be named `be_u8_from_u32` to be consistent with `u32_from_be_u8`.

Instead of taking a `dest: &mut [u8]` parameter, you can return a `[u8; 4]` result.

src/pbkdf2.rs
tarcieri: What if `chunk_len != output_len`?

src/pbkdf2.rs
```diff
@@ -150,14 +148,42 @@ pub fn derive(prf: &'static PRF, iterations: usize, salt: &[u8], secret: &[u8],
...
+        let chunk_len = chunk.len();
+
+        if chunk_len == output_len {
+            derive_block(&secret, iterations, salt, idx, output_len, chunk);
```
briansmith: Why doesn't this work when `chunk_len != output_len`?

tarcieri: Because `derive_block` expects `output_len`-sized blocks, and if `chunk_len` is smaller than that it will cause an out-of-bounds write.

src/pbkdf2.rs
```diff
@@ -150,14 +148,39 @@ pub fn derive(prf: &'static PRF, iterations: usize, salt: &[u8], secret: &[u8],
...
+        }
+    }
+}
+
+fn derive_block(
+        secret: &hmac::SigningKey,
+        iterations: usize,
+        salt: &[u8],
+        idx: u32,
+        len: usize,
```
briansmith: I think that `len` is redundant with `out.len()`. We should just use `out.len()`.

tarcieri: Done

src/pbkdf2.rs
```diff
@@ -133,12 +133,14 @@ use {constant_time, digest, hmac, polyfill};
...
 ///
-/// `derive` panics if `out.len()` is larger than the output length of the
-/// digest function used by the PRF algorithm.
+/// `derive` panics if `out.len()` is larger than 2^32 - 1, per the
+/// PBKDF2 specification
 pub fn derive(prf: &'static PRF, iterations: usize, salt: &[u8], secret: &[u8],
               out: &mut [u8]) {
     assert!(iterations >= 1);
-    assert!(out.len() <= prf.digest_alg.output_len);
+    assert!(out.len() <= 0xFFFFFFFF);
```
briansmith: The spec says `0xffffffff * output_len, not just 0xffffffff`.

Note that this check is only necessary and only possible to do when usize is larger than 32-bits. Otherwise, `0xffffffff * output_len` itself would overflow.

It seems good to keep the limit at `0xffffffff` to avoid the messiness, but at least the doc comment is wrong and needs to be updated.

src/pbkdf2.rs
```diff
@@ -133,12 +133,14 @@ use {constant_time, digest, hmac, polyfill};
 ///
 /// `derive` panics if `iterations < 1`.
 ///
-/// `derive` panics if `out.len()` is larger than the output length of the
-/// digest function used by the PRF algorithm.
+/// `derive` panics if `out.len()` is larger than 2^32 - 1, per the
```
briansmith: Please use the Python notation `2**32`. As I noted below, the spec actually says something slightly different, so we need to change this somehow to not lie.

briansmith: Above, the existing documentation says:

> out.len()` must be no larger than the output length of the digest function used in the PRF algorithm. This limit is more strict than what the		 used in the PRF algorithm. This limit is more strict than what the
specification requires. As noted at https://github.com/ctz/fastpbkdf2,		 specification requires.

That part of the documentation also needs to be corrected.

tarcieri:
I think that's all the nits?

briansmith:
We need some test vectors that test big outputs, especially outputs with at least 256 blocks so that we can verify that the block counter gets formatted correctly. I would like to see test vectors with 256 and 255 blocks, using SHA-256 and using a very small iteration count for this.

Let's also add test vectors for the follow output lengths (with low iteration counts): 1, output_len - 1, output_len, output_len + 1, 2 * output_len - 1, 2 * output_len, 2 * output_len + 1.

Please let me some instructions for how to verify that the newly-created test vectors are correct using another implementation (preferably Python openssl's command line tools or something of that nature.)

tarcieri:
Mmmk, will look into that

briansmith:
It looks like Python's hashlib can be used to generate the new vectors. See `hashlib.pbkdf2_hmac`, https://docs.python.org/3/library/hashlib.html#key-derivation. Probably there's a Ruby way too.

tarcieri:
I've been using Ruby's OpenSSL wrapper. Hope that's ok :wink:

```ruby
digest = OpenSSL::Digest::SHA256.new
OpenSSL::PKCS5.pbkdf2_hmac("256blockboundarytest", "salt", 8, 32 * 256, digest).unpack("H*")
```

src/pbkdf2.rs
```diff
@@ -133,12 +130,13 @@ use {constant_time, digest, hmac, polyfill};
 ///
 /// `derive` panics if `iterations < 1`.
 ///
-/// `derive` panics if `out.len()` is larger than the output length of the
-/// digest function used by the PRF algorithm.
+/// `derive` panics if `out.len()` is larger than (2**32 - 1) * the PRF digest
+/// length, per the PBKDF2 specification
```
briansmith: NIT: add period to the end of this.

src/pbkdf2.rs
```diff
@@ -190,17 +203,34 @@ pub fn derive(prf: &'static PRF, iterations: usize, salt: &[u8], secret: &[u8],
...
+        idx = idx.checked_add(1).expect("derived key too long");
+
+        let derived_chunk = &mut derived_buf[..previously_derived_chunk.len()];
+        polyfill::slice::fill(derived_chunk, 0);
+
+        derive_block(&secret, iterations, salt, idx, derived_chunk);
+        let compares_equal = constant_time::verify_slices_are_equal(derived_chunk,
+                                                                    previously_derived_chunk);
+
+        if result.is_ok() {
```
briansmith: To avoid this branch, we need to have an internal version of `verify_slices_are_equal` that returns a `c::size_t` and then use `constant_time_select_size_t(old, new, old)` to combine all the results together. That means we have to expose `constant_time_select_size_t` to Rust, which we don't do yet.

Then, we need to map the `c::size_t` to a `Result`. But, this doesn't need to be constant time.

Let me know if you feel comfortable making these changes. If not, I'll find some time to do it soon.

tarcieri: I  would probably prefer you make them. That said I think in the pantheon of security risks this is probably a pretty low one.

src/pbkdf2.rs
tarcieri: Nit nitted, and I squashed while I was at it

briansmith:
Hi, please add the contributor's statement at https://github.com/briansmith/ring#contributing and also please let me know if you are the copyright holder, or your employer is (who?), or somebody else is (who?). Thanks.

tarcieri:
I added the contributor statement to the commit message.

I am the copyright holder.

briansmith:
Thanks Tony. I finally got time to deal with this. Please check my changes at https://github.com/briansmith/ring/commit/08e7de7fd591fd46887af242790e78d162c95db0.

Thanks!

tarcieri:
👍

briansmith:
Thank you very much for contributing this. Committed as a316606b402a0c13884f196e4cc46b36e8a8589c. My updates are f87e2e87ef18b784009d1224e56dc853d611806f. The updated documentation is live at https://briansmith.org/rustdoc/ring/pbkdf2/.

Thanks again!

PR #254, opened 2016-07-19
Replace BN_rand_range C implementation with Rust

### LINKED ISSUE #239: Replace the use of `BN_rand_range` in RSA blinding with Rust code
The Rust code logic in `ring::ec::suite_b::private_key::generate_private_key` should be factored out into something that both `ring::ec::suite_b::private_key` and RSA blinding can use.

With this, all of crypto/bn/random.c (`BN_rand` and `BN_rand_range`) can go away.

djudd:
Fixes #239

### DIFF
--- a/crypto/bn/bn_test.cc
+++ b/crypto/bn/bn_test.cc
@@ -572,8 +572,7 @@ static bool TestBN2BinPadded(RAND *rng) {
 
   // Test a random numbers at various byte lengths.
   for (size_t bytes = 128 - 7; bytes <= 128; bytes++) {
-    if (!BN_rand(n.get(), bytes * 8, 0 /* make sure top bit is 1 */,
-                 0 /* don't modify bottom bit */, rng)) {
+    if (!BN_rand(n.get(), bytes * 8, rng)) {
       return false;
     }
     if (BN_num_bytes(n.get()) != bytes ||
@@ -661,37 +660,15 @@ static bool TestRand(RAND *rng) {
     return false;
   }
 
-  // Test BN_rand accounts for degenerate cases with |top| and |bottom|
-  // parameters.
-  if (!BN_rand(bn.get(), 0, 0 /* top */, 0 /* bottom */, rng) ||
-      !BN_is_zero(bn.get())) {
-    fprintf(stderr, "BN_rand gave a bad result.\n");
-    return false;
-  }
-  if (!BN_rand(bn.get(), 0, 1 /* top */, 1 /* bottom */, rng) ||
+  // Test BN_rand accounts for degenerate cases
+  if (!BN_rand(bn.get(), 0, rng) ||
       !BN_is_zero(bn.get())) {
     fprintf(stderr, "BN_rand gave a bad result.\n");
     return false;
   }
 
-  if (!BN_rand(bn.get(), 1, 0 /* top */, 0 /* bottom */, rng) ||
+  if (!BN_rand(bn.get(), 1, rng) ||
       !BN_is_word(bn.get(), 1)) {
-    fprintf(stderr, "BN_rand gave a bad result.\n");
-    return false;
-  }
-  if (!BN_rand(bn.get(), 1, 1 /* top */, 0 /* bottom */, rng) ||
-      !BN_is_word(bn.get(), 1)) {
-    fprintf(stderr, "BN_rand gave a bad result.\n");
-    return false;
-  }
-  if (!BN_rand(bn.get(), 1, -1 /* top */, 1 /* bottom */, rng) ||
-      !BN_is_word(bn.get(), 1)) {
-    fprintf(stderr, "BN_rand gave a bad result.\n");
-    return false;
-  }
-
-  if (!BN_rand(bn.get(), 2, 1 /* top */, 0 /* bottom */, rng) ||
-      !BN_is_word(bn.get(), 3)) {
     fprintf(stderr, "BN_rand gave a bad result.\n");
     return false;
   }
@@ -822,7 +799,7 @@ static bool TestBadModulus(BN_CTX *ctx) {
 // TestExpModZero tests that 1**0 mod 1 == 0.
 static bool TestExpModZero(RAND *rng, BN_CTX *ctx) {
   ScopedBIGNUM zero(BN_new()), a(BN_new()), r(BN_new());
-  if (!zero || !a || !r || !BN_rand(a.get(), 1024, 0, 0, rng)) {
+  if (!zero || !a || !r || !BN_rand(a.get(), 1024, rng)) {
     return false;
   }
   BN_zero(zero.get());
--- a/crypto/bn/random.c
+++ b/crypto/bn/random.c
@@ -117,7 +117,6 @@
 
 #include "../internal.h"
 
-
 int BN_rand(BIGNUM *rnd, int bits, int top, int bottom, RAND *rng) {
   uint8_t *buf = NULL;
   int ret = 0, bit, bytes, mask;
@@ -177,61 +176,23 @@ err:
   return (ret);
 }
 
-int BN_rand_range_ex(BIGNUM *r, BN_ULONG min_inclusive,
-                     const BIGNUM *max_exclusive, RAND *rng) {
-  unsigned n;
-  unsigned count = 100;
+int BN_rand_range_ex(BIGNUM *r, const BIGNUM *max_exclusive, RAND *rng) {
 
-  if (BN_cmp_word(max_exclusive, min_inclusive) <= 0) {
+  if (BN_cmp_word(max_exclusive, 1) <= 0) {
     OPENSSL_PUT_ERROR(BN, BN_R_INVALID_RANGE);
     return 0;
   }
 
-  n = BN_num_bits(max_exclusive); /* n > 0 */
-
-  /* BN_is_bit_set(range, n - 1) always holds */
-  if (n == 1) {
-    BN_zero(r);
-    return 1;
+  if (bn_wexpand(r, max_exclusive->top) == NULL) {
+    return 0;
   }
 
-  do {
-    if (!--count) {
-      OPENSSL_PUT_ERROR(BN, BN_R_TOO_MANY_ITERATIONS);
-      return 0;
-    }
+  if (!GFp_rand_mod(r->d, max_exclusive->d, max_exclusive->top, rng)) {
+    return 0;
+  }
 
-    if (!BN_is_bit_set(max_exclusive, n - 2) &&
-        !BN_is_bit_set(max_exclusive, n - 3)) {
-      /* range = 100..._2, so 3*range (= 11..._2) is exactly one bit longer
-       * than range. This is a common scenario when generating a random value
-       * modulo an RSA public modulus, e.g. for RSA base blinding. */
-      if (!BN_rand(r, n + 1, -1 /* don't set most significant bits */,
-                   0 /* don't set least significant bits */, rng)) {
-        return 0;
-      }
-
-      /* If r < 3*range, use r := r MOD range (which is either r, r - range, or
-       * r - 2*range). Otherwise, iterate again. Since 3*range = 11..._2, each
-       * iteration succeeds with probability >= .75. */
-      if (BN_cmp(r, max_exclusive) >= 0) {
-        if (!BN_sub(r, r, max_exclusive)) {
-          return 0;
-        }
-        if (BN_cmp(r, max_exclusive) >= 0) {
-          if (!BN_sub(r, r, max_exclusive)) {
-            return 0;
-          }
-        }
-      }
-    } else {
-      /* range = 11..._2  or  range = 101..._2 */
-      if (!BN_rand(r, n, -1, 0, rng)) {
-        return 0;
-      }
-    }
-  } while (BN_cmp_word(r, min_inclusive) < 0 ||
-           BN_cmp(r, max_exclusive) >= 0);
+  r->top = max_exclusive->top;
+  bn_correct_top(r);
 
   return 1;
 }
--- a/crypto/bn/random.c
+++ b/crypto/bn/random.c
@@ -117,67 +117,8 @@
 
 #include "../internal.h"
 
-int BN_rand(BIGNUM *rnd, int bits, int top, int bottom, RAND *rng) {
-  uint8_t *buf = NULL;
-  int ret = 0, bit, bytes, mask;
-
-  if (rnd == NULL) {
-    return 0;
-  }
-
-  if (bits == 0) {
-    BN_zero(rnd);
-    return 1;
-  }
-
-  bytes = (bits + 7) / 8;
-  bit = (bits - 1) % 8;
-  mask = 0xff << (bit + 1);
-
-  buf = OPENSSL_malloc(bytes);
-  if (buf == NULL) {
-    OPENSSL_PUT_ERROR(BN, ERR_R_MALLOC_FAILURE);
-    goto err;
-  }
-
-  /* Make a random number and set the top and bottom bits. */
-  if (!RAND_bytes(rng, buf, bytes)) {
-    goto err;
-  }
-
-  if (top != -1) {
-    if (top && bits > 1) {
-      if (bit == 0) {
-        buf[0] = 1;
-        buf[1] |= 0x80;
-      } else {
-        buf[0] |= (3 << (bit - 1));
-      }
-    } else {
-      buf[0] |= (1 << bit);
-    }
-  }
-
-  buf[0] &= ~mask;
-
-  /* set bottom bit if requested */
-  if (bottom)  {
-    buf[bytes - 1] |= 1;
-  }
-
-  if (!BN_bin2bn(buf, bytes, rnd)) {
-    goto err;
-  }
-
-  ret = 1;
-
-err:
-  OPENSSL_free(buf);
-  return (ret);
-}
 
 int BN_rand_range_ex(BIGNUM *r, const BIGNUM *max_exclusive, RAND *rng) {
-
   if (BN_cmp_word(max_exclusive, 1) <= 0) {
     OPENSSL_PUT_ERROR(BN, BN_R_INVALID_RANGE);
     return 0;
--- a/crypto/test/bn_test_lib.c
+++ b/crypto/test/bn_test_lib.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 1995-1997 Eric Young (eay@cryptsoft.com)
+/* Copyright (C) 1995-1998 Eric Young (eay@cryptsoft.com)
  * All rights reserved.
  *
  * This package is an SSL implementation written
@@ -55,7 +55,7 @@
  * [including the GNU Public Licence.]
  */
 /* ====================================================================
- * Copyright (c) 1998-2006 The OpenSSL Project.  All rights reserved.
+ * Copyright (c) 1998-2005 The OpenSSL Project.  All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -104,65 +104,57 @@
  *
  * This product includes cryptographic software written by Eric Young
  * (eay@cryptsoft.com).  This product includes software written by Tim
- * Hudson (tjh@cryptsoft.com).
- *
- */
-/* ====================================================================
- * Copyright 2002 Sun Microsystems, Inc. ALL RIGHTS RESERVED.
- *
- * Portions of the attached software ("Contribution") are developed by
- * SUN MICROSYSTEMS, INC., and are contributed to the OpenSSL project.
- *
- * The Contribution is licensed pursuant to the Eric Young open source
- * license provided above.
- *
- * The binary polynomial arithmetic software is originally written by
- * Sheueling Chang Shantz and Douglas Stebila of Sun Microsystems
- * Laboratories. */
+ * Hudson (tjh@cryptsoft.com). */
 
-#ifndef OPENSSL_HEADER_BN_TEST_LIB_H
-#define OPENSSL_HEADER_BN_TEST_LIB_H
-
-#include <openssl/base.h>
 #include <openssl/bn.h>
+#include <openssl/err.h>
+#include <openssl/mem.h>
+#include <openssl/rand.h>
+#include <openssl/type_check.h>
 
-#if defined(__cplusplus)
-extern "C" {
-#endif
-
-/* BIGNUM functions that used to be in libcrypto, but which were removed
- * because no library code uses them anymore. They are retained here so that
- * bn_test.cc can use them to test functions that still remain in libcrypto. */
-
-/* Basic functions. */
-
-void BN_set_negative(BIGNUM *bn, int sign);
-
-/* Conversion functions. */
-
-/* BN_bn2bin serialises the absolute value of |in| to |out| as a big-endian
- * integer, which must have |BN_num_bytes| of space available. It returns the
- * number of bytes written. */
-OPENSSL_EXPORT size_t BN_bn2bin(const BIGNUM *in, uint8_t *out);
+#include "bn_test_lib.h"
+#include "../bn/internal.h"
 
 
-/* Exponentiation. */
+int BN_rand(BIGNUM *rnd, int bits, RAND *rng) {
+  uint8_t *buf = NULL;
+  int ret = 0, bit, bytes, mask;
 
-/* BN_exp sets |r| equal to |a|^{|p|}. It does so with a square-and-multiply
- * algorithm that leaks side-channel information. It returns one on success or
- * zero otherwise. */
-OPENSSL_EXPORT int BN_exp(BIGNUM *r, const BIGNUM *a, const BIGNUM *p,
-                          BN_CTX *ctx);
+  if (rnd == NULL) {
+    return 0;
+  }
 
+  if (bits == 0) {
+    BN_zero(rnd);
+    return 1;
+  }
 
-/* Random generation. */
+  bytes = (bits + 7) / 8;
+  bit = (bits - 1) % 8;
+  mask = 0xff << (bit + 1);
 
-/* BN_rand sets |rnd| to a random number of length |bits|. The
- * most-significant bit, if any, will always be set. */
-OPENSSL_EXPORT int BN_rand(BIGNUM *rnd, int bits, RAND *rng);
+  buf = (uint8_t*) OPENSSL_malloc(bytes);
+  if (buf == NULL) {
+    OPENSSL_PUT_ERROR(BN, ERR_R_MALLOC_FAILURE);
+    goto err;
+  }
 
-#if defined(__cplusplus)
+  // Make a random number.
+  if (!RAND_bytes(rng, buf, bytes)) {
+    goto err;
+  }
+
+  // Ensure top bit is always set, and higher bits are not.
+  buf[0] |= (1 << bit);
+  buf[0] &= ~mask;
+
+  if (!BN_bin2bn(buf, bytes, rnd)) {
+    goto err;
+  }
+
+  ret = 1;
+
+err:
+  OPENSSL_free(buf);
+  return (ret);
 }
-#endif
-
-#endif /* OPENSSL_HEADER_BN_TEST_LIB_H */
--- a/include/openssl/bn.h
+++ b/include/openssl/bn.h
@@ -394,7 +394,7 @@ OPENSSL_EXPORT int BN_mod_sub_quick(BIGNUM *r, const BIGNUM *a, const BIGNUM *b,
                                     const BIGNUM *m);
 
 
-/* Random and prime number generation. */
+/* Random generation. */
 
 /* BN_rand sets |rnd| to a random number of length |bits|. If |top| is zero, the
  * most-significant bit, if any, will be set. If |top| is one, the two most
@@ -408,11 +408,14 @@ OPENSSL_EXPORT int BN_mod_sub_quick(BIGNUM *r, const BIGNUM *a, const BIGNUM *b,
 OPENSSL_EXPORT int BN_rand(BIGNUM *rnd, int bits, int top, int bottom,
                            RAND *rng);
 
+extern int GFp_rand_mod(BN_ULONG *dest, const BN_ULONG *max_exclusive,
+                        size_t num_limbs, RAND *rand);
+
 /* BN_rand_range_ex sets |rnd| to a random value in
- * [min_inclusive..max_exclusive). It returns one on success and zero
+ * [1..max_exclusive). It returns one on success and zero
  * otherwise. */
-OPENSSL_EXPORT int BN_rand_range_ex(BIGNUM *r, BN_ULONG min_inclusive,
-                                    const BIGNUM *max_exclusive, RAND *rng);
+OPENSSL_EXPORT int BN_rand_range_ex(BIGNUM *r, const BIGNUM *max_exclusive,
+                                    RAND *rng);
 
 
 /* Number theory functions */
--- a/include/openssl/bn.h
+++ b/include/openssl/bn.h
@@ -396,17 +396,6 @@ OPENSSL_EXPORT int BN_mod_sub_quick(BIGNUM *r, const BIGNUM *a, const BIGNUM *b,
 
 /* Random generation. */
 
-/* BN_rand sets |rnd| to a random number of length |bits|. If |top| is zero, the
- * most-significant bit, if any, will be set. If |top| is one, the two most
- * significant bits, if any, will be set.
- *
- * If |top| is -1 then no extra action will be taken and |BN_num_bits(rnd)| may
- * not equal |bits| if the most significant bits randomly ended up as zeros.
- *
- * If |bottom| is non-zero, the least-significant bit, if any, will be set. The
- * function returns one on success or zero otherwise. */
-OPENSSL_EXPORT int BN_rand(BIGNUM *rnd, int bits, int top, int bottom,
-                           RAND *rng);
 
 extern int GFp_rand_mod(BN_ULONG *dest, const BN_ULONG *max_exclusive,
                         size_t num_limbs, RAND *rand);
--- a/src/ec/suite_b/ops/ops.rs
+++ b/src/ec/suite_b/ops/ops.rs
@@ -18,6 +18,8 @@ use {c, der, error};
 use core;
 use untrusted;
 
+pub use limb::*;
+
 /// Field elements. Field elements are always Montgomery-encoded and always
 /// fully reduced mod q; i.e. their range is [0, q).
 pub struct Elem {
@@ -98,12 +100,6 @@ impl Clone for Point {
 impl Copy for Point {
 }
 
-// XXX: Not correct for x32 ABIs.
-#[cfg(target_pointer_width = "64")] pub type Limb = u64;
-#[cfg(target_pointer_width = "32")] pub type Limb = u32;
-#[cfg(target_pointer_width = "64")] pub const LIMB_BITS: usize = 64;
-#[cfg(target_pointer_width = "32")] pub const LIMB_BITS: usize = 32;
-
 #[cfg(all(target_pointer_width = "32", target_endian = "little"))]
 macro_rules! limbs {
     ( $limb_b:expr, $limb_a:expr, $limb_9:expr, $limb_8:expr,
@@ -129,7 +125,6 @@ macro_rules! limbs {
     }
 }
 
-pub const LIMB_BYTES: usize = (LIMB_BITS + 7) / 8;
 pub const MAX_LIMBS: usize = (384 + (LIMB_BITS - 1)) / LIMB_BITS;
 
 static ONE: ElemDecoded = ElemDecoded {
@@ -215,11 +210,9 @@ impl CommonOps {
 
     pub fn elem_verify_is_not_zero(&self, a: &ElemUnreduced)
                                    -> Result<(), error::Unspecified> {
-        match unsafe {
-            GFp_constant_time_limbs_are_zero(a.limbs.as_ptr(), self.num_limbs)
-        } {
-            0 => Ok(()),
-            _ => Err(error::Unspecified),
+        match limbs_are_zero_constant_time(&a.limbs[..self.num_limbs]) {
+            false => Ok(()),
+            true => Err(error::Unspecified),
         }
     }
 
@@ -562,8 +555,6 @@ pub fn limbs_less_than_limbs(a: &[Limb], b: &[Limb]) -> bool {
 
 
 extern {
-    fn GFp_constant_time_limbs_are_zero(a: *const Limb, num_limbs: c::size_t)
-                                        -> Limb;
     fn GFp_constant_time_limbs_reduce_once(r: *mut Limb, m: *const Limb,
                                            num_limbs: c::size_t);
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -130,6 +130,13 @@ mod init;
 pub mod pbkdf2;
 pub mod rand;
 
+mod limb;
+
+// Really a private method; only has public visibility so that C compilation
+// can see it.
+#[doc(hidden)]
+pub use limb::GFp_rand_mod;
+
 #[cfg(feature = "use_heap")]
 #[path = "rsa/rsa.rs"]
 mod rsa;
--- a/src/limb.rs
+++ b/src/limb.rs
@@ -1,0 +1,208 @@
+use {rand, polyfill, c, core, error};
+
+// XXX: Not correct for x32 ABIs.
+#[cfg(target_pointer_width = "64")] pub type Limb = u64;
+#[cfg(target_pointer_width = "32")] pub type Limb = u32;
+#[cfg(target_pointer_width = "64")] pub const LIMB_BITS: usize = 64;
+#[cfg(target_pointer_width = "32")] pub const LIMB_BITS: usize = 32;
+
+pub const LIMB_BYTES: usize = (LIMB_BITS + 7) / 8;
+
+/// References a positive integer range `[1..max_exclusive)` where
+/// `max_exclusive` is a slice of limbs ordered least-significant-limb to
+/// most-significant-limb (and the limbs use the native endianness).
+pub struct Range<'a> {
+    pub max_exclusive: &'a [Limb],
+}
+
+impl <'a>Range<'a> {
+    pub fn from_max_exclusive(max_exclusive: &[Limb]) -> Range {
+        debug_assert!(!limbs_are_zero_constant_time(max_exclusive));
+        debug_assert!(max_exclusive[max_exclusive.len() - 1] > 0);
+
+        Range {
+            max_exclusive: max_exclusive
+        }
+    }
+
+    fn leading_zero_bits(&self) -> usize {
+        let most_significant_limb =
+            self.max_exclusive[self.max_exclusive.len() - 1];
+        most_significant_limb.leading_zeros() as usize
+    }
+
+    /// Are these little-endian limbs within the range?
+    ///
+    /// Checks in constant time.
+    pub fn are_limbs_within(&self, limbs: &[Limb]) -> bool {
+        assert_eq!(self.max_exclusive.len(), limbs.len());
+
+        let is_gt_zero = !limbs_are_zero_constant_time(limbs);
+        let is_lt_max = limbs_less_than_limbs_constant_time(limbs, self.max_exclusive);
+
+        is_lt_max && is_gt_zero
+    }
+
+    // Loosely based on NSA Guide Appendix B.2:
+    // "Key Pair Generation by Testing Candidates".
+    //
+    // TODO: DRY-up with `ec::suite_b::private_key::generate_private_key`?
+    //
+    /// Chooses a positive integer within the range and stores it into `dest`
+    /// encoded as limbs ordered least-significant-limb to most-significant-limb.
+    ///
+    /// This function is intended to be suitable for generating private keys.
+    pub fn sample_into_limbs(&self, dest: &mut [Limb], rng: &rand::SecureRandom)
+                             -> Result<(), error::Unspecified> {
+
+        assert_eq!(self.max_exclusive.len(), dest.len());
+
+        let bits_to_mask = self.leading_zero_bits();
+
+        // XXX: The value 100 was chosen to match OpenSSL due to uncertainty of
+        // what specific value would be better, but it seems bad to try 100 times.
+        for _ in 0..100 {
+            {
+                let mut dest_as_bytes = limbs_as_bytes_mut(dest);
+                try!(rng.fill(&mut dest_as_bytes));
+            }
+
+            if bits_to_mask > 0 {
+                let mask: Limb = (1 << (LIMB_BITS - bits_to_mask)) - 1;
+                dest[self.max_exclusive.len() - 1] &= mask;
+            }
+
+            if self.are_limbs_within(&dest) {
+                return Ok(());
+            }
+        }
+
+        Err(error::Unspecified)
+    }
+
+}
+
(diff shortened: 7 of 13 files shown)


src/limb.rs
```diff
@@ -0,0 +1,298 @@
...
+        Err(())
+    }
+
+}
+
+
+/// Initializes little-endian limbs from big-endian bytes.
+///
+/// Equivalent to `BN_bin2bn`.
+pub fn set_limbs_from_big_endian_bytes(limbs: &mut[Limb], bytes: &[u8]) {
```
djudd: I think this has to be constant-time, and AFAICT it is. Am I right? If so, should that be in the name and/or doc comment?

crypto/bn/bn_test.cc
```diff
@@ -171,6 +172,66 @@ static ScopedBIGNUM GetBIGNUM(FileTest *t, const char *attribute) {
   return ret;
 }
 
+// TODO Remove as dependency, per comment at top of file
```
briansmith: Please put this in bn_test_lib.c. (To determine which of bn_test_lib.c or bn_test_convert.c to move the code too, I diff a diff of the licenses at the top of random.c and those two functions. bn_test_lib.c has the same license so that's where it goes.)

src/lib.rs
```diff
@@ -133,6 +133,7 @@ pub mod test;
 
 #[cfg(test)]
 mod tests {
+    #[cfg(feature = "rsa_signing")]
```
briansmith: We still need to run the BIGNUM tests because RSA verification still uses BIGNUM (and ECC too, for a short while still.)

briansmith:
Thanks for submitting. I added a couple of comments about where to move the now-test-only code. Looking at the second commit, I can see that this will take me some time to review.

djudd:
Thanks for taking a look! I understand it will take some time to look over, it ended up being bigger than I was expecting.

Just rebased to add licensing comments to the commit messages.

src/lib.rs
djudd: A few of the tests depend on BN_rand_range, which in turn pulls in some other stuff. Options:
- Ship the actual BN_rand_range even without RSA signing;
- Move that subset of the BN tests to their own file and include it conditionally;
- Some other way to disable these tests depending on the Rust feature config.
Your preference / thoughts?

src/rsa.rs
```diff
@@ -475,6 +486,51 @@ extern {
...
+
+    let range = limb::Range {
+        limbs: core::slice::from_raw_parts((*range).d, (*range).top as usize),
+    };
+
+    if !range.valid().is_ok() {
+        return ERR;
+    }
+
+    let mut buf = &mut [0u8; limb::MAX_RANGE_BYTES][..range.bytes_count()];
```
djudd: In general I went with stack-allocated buffers that are as big as the maximum key we want to use, even where using the heap might be more natural. Tracking down those maximums took some inference, and it's not obvious they'll never change, so I'm not sure this is the right strategy.

src/lib.rs
briansmith: In BoringSSL, all of the use of `BN_rand_range` and most of the uses of `BN_range` have been removed from bn_test.cc. I will merge those changes soon, before I merge this PR. After that, and after #247, you should be able to update this PR to completely remove ``BN_rand_range`. We'll still need `BN_rand` for bn_test.cc, but we can do as you did in your first commit: just move it to bn_test_lib.c. Similarly, if there are other `BN_*` functions that would then only be used in tests, we can move them in a similar way; we just need to make sure we sort everything into files according to the license.

briansmith: Note also that eventually (hopefully soon) we will make ECC (#247) and RSA verification work without using `BIGNUM`. Then, RSA signing will be the only thing using `BIGNUM` and we can indeed skip `bn_test.cc` when the RSA signing feature is not enabled. But, we don't need to add that skipping in *this* PR.

src/rsa.rs
```diff
@@ -475,6 +486,51 @@ extern {
                                  rng: *mut rand::RAND) -> c::int;
 }
 
+#[cfg(feature = "rsa_signing")]
+#[allow(non_snake_case)]
+#[doc(hidden)]
+#[no_mangle]
+pub unsafe extern fn BN_rand_range(dest: *mut BIGNUM, range: *const BIGNUM,
+                                   rng: *mut rand::RAND) -> c::int {
```
briansmith: What do you think about this other strategy, which might be simpler.

1. In the C code, create the BIGNUM.
2. Use `bn_wexpand` to expand the BIGNUM's storage to the necessary number of words, `num_limbs`, which comes from the `top` of the modulus.
3. Call a Rust function `GFp_rand_mod(dest: *mut Limb, mod: *const Limb, num_limbs: c::size_t) -> c::int`, which returns zero on failure and non-zero on success.
4. Create a function that uses either `u64_as_u8_mut` or `u32_as_u8_mut` to view `dest` as an array of bytes and use SecureRandom.fill() to fill `dest` with random bytes. (Or, add functions to `SecureRandom for filling arrays of `u32` and `u64`.)
5. Use `GFP_constant_time_limbs_lt_limbs` to ensure that `dest` is less than `m`; Repeat until that is true.

Then, in a separate commit (maybe even a separate PR):

6. Change the ECC code to work a similar way as above, and then have it serialize the resulting `[Limb]` as big-endian bytes. (Instead of generating random bytes, then deserializing the bytes to `[Limb]`, then comparing.)

I think this alternate strategy might greatly reduce the amount of new code you need *for this PR*. And, also, it would avoid the need to worry about allocating very buffers on the stack. Then we'd have more time, at least, to work together on a strategy for dealing with the memory allocation.

I suspect that some and maybe even most of the code you wrote for this PR would then potentially be usable for some future refactoring of the RSA code.

WDYT?

src/limb.rs
briansmith: We already have a very similar function named `parse_big_endian_value`. I think we should unify these together. If the existing function isn't constant time, we can use this new definition.

See also the discussion in #248. Here's what we do now:

1. Generate an array of bytes.
2. Convert that array of bytes into limbs, viewing it as big-endian, to see if it is in range.
3. Store the big-endian array of bytes in `ec::PrivateKey`.
4. During ECDH (and eventually Signing), convert the big-endian bytes into an array of `Limb`s.
5. Then, convert the array of `Limb`s into a little-endian array of bytes.
6. Use that little-endian array of bytes to drive the scalar multiplication.

Meanwhile, for X25519/Ed25519 keys, we store the little-endian bytes in `ec::PrivateKey` and then use the little-endian bytes to drive the scalar multiplication. Much simpler.

So, I'd like to change the code for Suite B curves to work like the Curve25519 code. If/when we do that, we won't need this `set_limbs_from_big_endian_bytes` function for ECC, which seems good.

briansmith:
I added some more comments. Nothing definitive, but just my gut reaction. LMK what you think.

src/limb.rs
briansmith: > parse_big_endian_value. I think we should unify these together. 

Actually, I guess the point of this function to to be more clearly constant time, which is presumably why there were two functions to start with. So, it is OK if we don't unify them together. Again, see my other comment; I'm kind of hoping we can just avoid this function completely.

src/rsa.rs
djudd: Hmm. It seems like what you're suggesting a few differences from this PR, which could be taken separately:

(A) Instead of converting bytes to limbs each time we generate a number, we fill the limbs buffer directly. I like that: seems more efficient, requires one constant-time function fewer, and lets us avoid a few fishily-sized stack buffers. However, it's not immediately obvious to me that even so we'll be able to avoid writing new code to handle the case of range bounds which use only a few bits of the most significant limb, which motivated most of the code that I added in `limb.rs`. FWIW I suspect a lot of that code for counting leading zero bytes/bits etc could be shortened by somebody more familiar with bitwise APIs and tricks.

(B) We leave the definition of BN_rand_range in C, and call into a Rust function internally. I'm neutral on this; it seems the advantage is we have a smaller diff, and the disadvantage is we can't delete `random.c` yet. If it helps us make (A) work, though, I'm for it.

(C) We start by writing a new key-generation function, then merge them later, instead of refactoring first. My gut reaction is against that. What's the advantage?

djudd: Hmm. It seems like what you're suggesting a few differences from this PR, which could be taken separately:

(A) Instead of converting bytes to limbs each time we generate a number, we fill the limbs buffer directly. I like that: seems more efficient, requires one constant-time function fewer, and lets us avoid a few fishily-sized stack buffers. However, it's not immediately obvious to me that even so we'll be able to avoid writing new code to handle the case of range bounds which use only a few bits of the most significant limb, which motivated most of the code that I added in `limb.rs`. FWIW I suspect a lot of that code for counting leading zero bytes/bits etc could be shortened by somebody more familiar with bitwise APIs and tricks.

(B) We leave the definition of BN_rand_range in C, and call into a Rust function internally. I'm neutral on this; it seems the advantage is we have a smaller diff, and the disadvantage is we can't delete `random.c` yet. If it helps us make (A) work, though, I'm for it.

(C) We start by writing a new key-generation function, then merge them later, instead of refactoring first. My gut reaction is against that. What's the advantage?

briansmith: > (A) Instead of converting bytes to limbs each time we generate a number, we fill the limbs buffer directly.

Right.

> However, it's not immediately obvious to me that even so we'll be able to avoid writing new code to handle the case of range bounds which use only a few bits of the most significant limb, which motivated most of the code that I added in limb.rs.

If I understand the problem correctl : We could generate `num_limbs * BN_BYTES` of random bytes to fill the buffer. Then, in the highest limb, we need to mask `n` bits, where `n = (BN_num_bits(x) % BN_BITS2)` (or so). Then we do the comparison using GFp_constant_time_limbs_lt_limbs and try again if that fails. The issue is that it is a little bit unclear how to do the masking of the bits? Is that right?

Given n, `(1 << n) - 1` should give you the mask, I think. For example, we want to preserve 4 bits. `1 << 4 == 0b10000` and so `(1 << 4) - 1 == 0b1111`.

You can also look at the boringssl code at https://github.com/google/boringssl/blob/bfefc27c2bb4af62e09569e36b018d60da98a680/crypto/bn/random.c#L247 for some inspiration, in case my math above is off.

briansmith: > (B) We leave the definition of BN_rand_range in C, and call into a Rust function internally. I'm neutral on this; it seems the advantage is we have a smaller diff, and the disadvantage is we can't delete random.c yet. If it helps us make (A) work, though, I'm for it.

I think this is a good first step because it lets us review the change in parts.

Also, you can imagine that after this PR, we'll want to replace `BN_BLINDING` stuff with Rust code and then replace `GFp_private_key_transform` with Rust code. This is actually a step towards all of that. And, I think the question is, in between now and then, do we want to have `#[repr(C)] struct BIGNUM {...}` in the Rust code? I don't think we do. I think we should say "old code uses BIGNUM as it needs to, new code doesn't use BIGNUM at all" and in particular, leave all the stuff that relies on a `BIGNUM` internals in C. Then, later, we'll get rid of `BIGNUM` completely.

briansmith: > C) We start by writing a new key-generation function, then merge them later, instead of refactoring first. My gut reaction is against that. What's the advantage?

One of my concerns is that we can't use the same strategy of stack-allocated oversized buffers for RSA blinding that we use for ECC. We do it for ECC because the objects in ECC are all roughly the same size (+/- 16 bytes). In RSA, the difference between the smallest `BIGNUM` (1024 bits) and largest `BIGNUM` (8192 bits) is extreme. And, the largest `BIGNUM`s we have to support are very rarely used. So, if we make everything big enough to accomodate the large keys, then we'll end up wasting quite a bit of memory, because RSA private keys consist of many such values, and often there are many blindings in memory at once as well. So, I'm hoping to find a solution for this particular PR that doesn't require us to add heap allocations in the Rust code and which doesn't require the large static buffers.

Also, I'm hoping we can, at least, find some convenient way to break up the PR into smaller commits.

djudd:
@briansmith I pushed a revised version which follows the strategy you suggested. The first commit re-implements the guts of BN_rand_range in Rust, copying the code from EC key generation. The second commit switches EC key generation to use the code path created in the first commit.

FYI I'll be traveling & then back at my day job starting tomorrow so expect the pace of work & responses to slow down.

crypto/bn/random.c
```diff
@@ -118,120 +118,21 @@
...
-      if (!--count) {
-        OPENSSL_PUT_ERROR(BN, BN_R_TOO_MANY_ITERATIONS);
-        return 0;
-      }
-    } while (BN_cmp(r, range) >= 0);
+  if (!GFp_rand_mod(r->d, range->d, range->top, rng)) {
+    return 0;
   }
 
+  r->top = range->top;
```
briansmith: This needs to be a call to `bn_correct_top`. `BIGNUM`s always need to have the minimal value of `top` so that the top-level limb is never zero. (This destroys the constant-time property of `BIGNUM`, which is why we're getting rid of it. But certain other code depends on it, so whenever we have a `BIGNUM` we have to maintain that invariant.)

include/openssl/bn.h
```diff
@@ -423,19 +423,10 @@ OPENSSL_EXPORT int BN_mod_lshift1_quick(BIGNUM *r, const BIGNUM *a,
...
- *
- * If |top| is -1 then no extra action will be taken and |BN_num_bits(rnd)| may
- * not equal |bits| if the most significant bits randomly ended up as zeros.
- *
- * If |bottom| is non-zero, the least-significant bit, if any, will be set. The
- * function returns one on success or zero otherwise. */
-OPENSSL_EXPORT int BN_rand(BIGNUM *rnd, int bits, int top, int bottom,
-                           RAND *rng);
+extern int GFp_rand_mod(BN_ULONG* dest, const BN_ULONG* mod, size_t num_limbs,
+                        RAND* rand);
```
briansmith: s/* / */.

src/lib.rs
```diff
@@ -123,6 +123,9 @@ mod init;
 pub mod pbkdf2;
 
 pub mod rand;
+mod limb;
+pub use limb::GFp_rand_mod;
```
briansmith: I don't understand why this `pub use` is here. I think it should be removed. We actually don't want to expose this in the API.

src/limb.rs
```diff
@@ -0,0 +1,271 @@
...
+    if !range.is_ok() {
+        return ERR;
+    }
+
+    let mut dest = core::slice::from_raw_parts_mut(dest, num_limbs);
+
+    let result = range.unwrap().sample_into_limbs(&mut dest, (*rng).rng);
+    if result.is_err() {
+        ERR
+    } else {
```
briansmith: I would prefer to have this written as though it were a `try!()`:

```
if result.is_err() {
    return ERR;
}

SUCCESS
```

src/limb.rs
```diff
@@ -0,0 +1,271 @@
...
+            let bit = 1 << (LIMB_BITS - 1 - i);
+            if self.most_significant_limb() & bit > 0 {
+                return i;
+            }
+        }
+
+        unreachable!();
+    }
+
+    fn is_valid(&self) -> bool {
```
briansmith: This should just be inlined into the constructor. Also, all of these checks should be made `debug_assert!`s. This code will only be used internally and only when we know that the range meets all the conditions here.

src/limb.rs
```diff
@@ -0,0 +1,271 @@
...
+            {
+                let mut dest_as_bytes = limbs_as_bytes_mut(dest);
+                try!(rng.fill(&mut dest_as_bytes));
+            }
+
+            let bits_to_mask = self.leading_zero_bits();
+            if bits_to_mask > 0 {
+                let keep: Limb = (1 << (LIMB_BITS - bits_to_mask)) - 1;
+                dest[self.limbs.len() - 1] &= keep;
+            }
```
briansmith: For the ECC curves we currently support, `bits_to_mask` will always be zero and so the calculation `leading_zero_bits()` and the calculation of `keep` is unneeded. It is only useful for RSA.

So, I would prefer that the caller calculate `keep` (perhaps renamed to `top_limb_mask` or similar) and pass it to this function, instead of having this function calculate it. I know it's not as clean of an API, but OTOH it would keep the ECC code as simple as it currently is. I don't want there to be any additional (code size) cost for using the ECC code due to the RSA code, as much as possible. (One goal of this library is to shrink the size of the ECC code so that it will work on very small processors. But, we don't have such hopes for RSA signing.)

src/limb.rs
```diff
@@ -0,0 +1,271 @@
...
+fn limbs_as_bytes_mut<'a>(src: &'a mut [Limb]) -> &'a mut [u8] {
+    polyfill::slice::u32_as_u8_mut(src)
+}
+
+#[allow(unsafe_code)]
+pub fn limbs_less_than_limbs_constant_time(a: &[Limb], b: &[Limb]) -> bool {
+    assert_eq!(a.len(), b.len());
+    let result = unsafe {
+        GFp_constant_time_limbs_lt_limbs(a.as_ptr(),
+                                         b.as_ptr(),
```
briansmith: I guess this is the result of rustfmt or following rustfmt style. But, I use the BoringSSL style, which is to wrap to as few lines as possible. Please use the BoringSSL style. (If you know a way to write a rustfmt configuration file that does that, I'm happy to add one to this repo to make it easier for people..)

src/limb.rs
```diff
@@ -0,0 +1,271 @@
...
+#[cfg(target_pointer_width = "32")] pub const LIMB_BITS: usize = 32;
+
+/// Stores a positive integer range (e.g. that governing a private key) as
+/// little-endian limbs representing the top end of the range (exclusive).
+pub struct Range<'a> {
+    limbs: &'a [Limb],
+}
+
+impl <'a>Range<'a> {
+
```
briansmith: NIT: No blank line here.

src/limb.rs
```diff
@@ -0,0 +1,271 @@
...
+#[cfg(target_pointer_width = "64")] pub const LIMB_BITS: usize = 64;
+#[cfg(target_pointer_width = "32")] pub const LIMB_BITS: usize = 32;
+
+/// Stores a positive integer range (e.g. that governing a private key) as
+/// little-endian limbs representing the top end of the range (exclusive).
+pub struct Range<'a> {
+    limbs: &'a [Limb],
+}
+
+impl <'a>Range<'a> {
```
briansmith: NIT: space before "Range"

crypto/bn/random.c
```diff
@@ -118,120 +118,21 @@
...
-  }
-
-  ret = 1;
-
-err:
-  OPENSSL_free(buf);
-  return (ret);
-}
-
 int BN_rand_range(BIGNUM *r, const BIGNUM *range, RAND *rng) {
```
briansmith: Great. I'm going to work on the BoringSSL merge now so that the `BN_BLINDING` code is the only caller of `BN_rand_range`. Then (in a separate commit) we can inline the new body of this into the `BN_BLINDING` code and simplify it further.

src/lib.rs
djudd: My understanding was that this is necessary to make it visible to the C code. Without this, I get:

```error: function GFp_rand_mod is marked #[no_mangle], but not exported```

Is there another way around that?

File: src/limb.rs (hunk 10 above)
Comment (djudd) [inline src/limb.rs]: Ok. (The zero check was in BN_rand_range and I wasn't sure under what conditions that could be called.)

File: src/limb.rs (hunk 11 above)
Comment (djudd) [inline src/limb.rs]: That API seems more error prone to me. How do you feel about using `cfg(attr)` to make `leading_zero_bits` explicitly always return 0 in the non-RSA case instead? The compiler should be able to optimize this block away in that case, right?

**Comment by briansmith** [comment]
I have now done the BoringSSL merge so that `BN_BLINDING` is now the only users of `BN_rand_range`. Also, the only user of `BN_rand` is now `TestBN2BinPadded`, which means we're not far away from nuking it; for now we'll just have to make move it to bn_test_lib.c. Note that I deleted bn_test_lib.c during the merge as the contents were no longer being used, so when you merge your PR on top of the current master, you'll have to add it back (basically revert most of that commit).

It will be a day or two before I can look at this again.

File: src/lib.rs (hunk 8 above)
Comment (briansmith) [inline src/lib.rs]: In other areas of the code, we use this:

```
#[allow(non_snake_case)]
#[doc(hidden)]
#[no_mangle]
pub extern fn ...
```

File: src/limb.rs (hunk 11 above)
Comment (briansmith) [inline src/limb.rs]: Definitely I don't want to do the `cfg(attr)` thing. I think, at a minimum, we should pull the calculation of the mask out of the loop so that it is clearer that it doesn't change from one interation to another.

**Comment by briansmith** [comment]
I'm curious about whetehr the special case logic for the case where `range is 0b100...` might be really important for RSA blinding? Because, `n` is the product of two primes and it seems like values of `n` of the form `0b100` might be really common? Do you have any thoughts about that?

Sorry about the delay. I hope to get to this during the week.

**Comment by djudd** [comment]
As far as I can tell, the purpose of that logic is to increase the odds that any given iteration finds a value in the range. The comment says it makes the probability of succeeding on a given iteration >= 75%. With the current Rust logic, by contrast, the probability of success is >= 50% (since we know we succeed if the first bit is zero). I think that still makes the probability of failing for 100 iterations in a row sufficiently low (1 / 2^100), so it's a question of performance. 

If we think this code is performance critical, I could see a case for introducing that extra complication. It seems like that should come with benchmarks (since we're adding subtraction, no guarantee it's actually faster on average) and possibly a separate pull request, though.

BTW I don't see how that deleted logic avoids the bias towards small values that you mention in your comment on the NSA guide's alternative method.

File: src/lib.rs (hunk 8 above)
Comment (djudd) [inline src/lib.rs]: It already looks like this:

```Rust
#[allow(unsafe_code)]
#[allow(non_snake_case)]
#[doc(hidden)]
#[no_mangle]
pub unsafe extern fn GFp_rand_mod ...
```

But that's not enough, I get an error if I delete the `pub use`, I think because the `limb` module is private.

briansmith:
> BTW I don't see how that deleted logic avoids the bias towards small values that you mention in your comment on the NSA guide's alternative method.

Good points, especially this one.

crypto/test/bn_test_lib.c
```diff
@@ -116,7 +116,7 @@
 #include "../bn/internal.h"
 
 
-int BN_rand(BIGNUM *rnd, int bits, int top, int bottom, RAND *rng) {
+int BN_rand(BIGNUM *rnd, int bits, RAND *rng) {
```
briansmith: This looks good.

Another option, which I prefer, would be to rewrite `TestBN2BinPadded` to be based on bn_tests.txt test vectors instead of using `BN_rand` to generate random values. e.g. Change `TestBN2BinPadded` so that it prints out its inputs and outputs in bn_test.txt format, then save the output of one or two runs, and then redo the test logic to use the bn_test.txt test vectors.

Then we could delete `BN_rand` completely.

Not required, but just letting you know what I see as being the ideal scenerio.

src/limb.rs
```diff
@@ -8,8 +8,9 @@ use {rand, polyfill, c, core};
 
 pub const LIMB_BYTES: usize = (LIMB_BITS + 7) / 8;
 
-/// Stores a positive integer range (e.g. that governing a private key) as
-/// little-endian limbs representing the top end of the range (exclusive).
+/// Stores a positive integer range 0 < x < N (e.g. that governing a private
```
briansmith: I suggest `s/Stores/References` since the limbs aren't stored in the `Range`. Also, I suggest that you clarify the endianness comment a bit: the words are ordered least-significant-word to most-significant-word, and the bytes within the words are in the native endianness.

src/limb.rs
```diff
@@ -84,7 +87,6 @@ impl <'a>Range<'a> {
                 try!(rng.fill(&mut dest_as_bytes));
             }
 
-            let bits_to_mask = self.leading_zero_bits();
             if bits_to_mask > 0 {
                 let keep: Limb = (1 << (LIMB_BITS - bits_to_mask)) - 1;
```
briansmith: I would prefer to name `keep` as `mask` as `mask` seems clearer to me. Also, why did you pull bits_to_mask out of the loop, but not  the calculation of `keep`? I think both should be pulled out of the loop. It won't affect performance but it would improve clarity.

briansmith:
David, please fold the updates into the original changes. I'm going to re-review the whole thing and it would help me a lot with the review. Thanks.

src/lib.rs
briansmith: OK, thanks for correcting me. Please just add a comment where the `pub use` is and ensure that we have the `#[doc(hidden)]` right so that it doesn't show up in the docs.

briansmith:
> BTW I don't see how that deleted logic avoids the bias towards small values that you mention in your comment on the NSA guide's alternative method.

First, let's prove that 3x < 2<sup>n+1</sup>. If x is of the form 0b100... then x < 2<sup>n</sup> - 2<sup>n-2</sup> - 2<sup>n-3</sup>, because the largest n-bit value is less than 2<sup>n</sup> and we can subtract 2<sup>n-2</sup> and 2<sup>n-3</sup> since we know the 2nd- and 3rd- highest bits are zero. Then:

3x < 3×(2<sup>n</sup> - 2<sup>n-2</sup> - 2<sup>n-3</sup>)
3x < (2 + 1)×(2<sup>n</sup> - 2<sup>n-2</sup> - 2<sup>n-3</sup>)
3x < 2×(2<sup>n</sup> - 2<sup>n-2</sup> - 2<sup>n-3</sup>) + 2<sup>n</sup> - 2<sup>n-2</sup> - 2<sup>n-3</sup>
3x < 2×2<sup>n</sup> - 2×2<sup>n-2</sup> - 2×2<sup>n-3</sup> + 2<sup>n</sup> - 2<sup>n-2</sup> - 2<sup>n-3</sup>
3x < 2<sup>n+1</sup> - 2<sup>n-1</sup> - 2<sup>n-2</sup> + 2<sup>n</sup> - 2<sup>n-2</sup> - 2<sup>n-3</sup>
3x < 2<sup>n+1</sup> + 2<sup>n</sup> - 2<sup>n-1</sup> - 2<sup>n-2</sup> - 2<sup>n-2</sup> - 2<sup>n-3</sup>
3x < 2<sup>n+1</sup> + 2<sup>n</sup> - 2<sup>n-1</sup> - (2<sup>n-2</sup> - 2<sup>n-2</sup>) - 2<sup>n-3</sup>
 3x < 2<sup>n+1</sup> + 2<sup>n</sup> - 2<sup>n-1</sup> - 2×2<sup>n-2</sup> - 2<sup>n-3</sup>
3x < 2<sup>n+1</sup> + 2<sup>n</sup> - 2<sup>n-1</sup> - 2<sup>n-1</sup> - 2<sup>n-3</sup>
3x < 2<sup>n+1</sup> + 2<sup>n</sup> - (2<sup>n-1</sup> - 2<sup>n-1</sup>) - 2<sup>n-3</sup>
3x < 2<sup>n+1</sup> + 2<sup>n</sup> - 2×2<sup>n-1</sup> - 2<sup>n-3</sup>
3x < 2<sup>n+1</sup> + 2<sup>n</sup> - 2<sup>n</sup> - 2<sup>n-3</sup>
3x < 2<sup>n+1</sup> + (2<sup>n</sup> - 2<sup>n</sup>) - 2<sup>n-3</sup>
3x < 2<sup>n+1</sup> - 2<sup>n-3</sup>
3x < 2<sup>n+1</sup> - 2<sup>n-3</sup>

Clearly 2<sup>n+1</sup> - 2<sup>n-3</sup> < 2<sup>n+1</sup> so 3x < 2<sup>n+1</sup>

This means that if we generate a value in the range [0, 2<sup>n+1</sup>), which would fall into one of four sub-intervals:

 *    [0, max)          => return value.
 *    [max, 2×max)      => return value - max.
 *    [2×max, 3×max)    => return value - max - max.
 *    [3×max, 2<sup>n+1</sup>) => Generate a new random value and try again.

This avoids biasing the result towards small values, which is what reducing the random value (mod max) would do, while reducing the probability that a new random value will be needed.

djudd:
```This avoids biasing the result towards small values```

Ah, right, because despite the comment, we don't quite actually use `r := r MOD range`, since we don't use the interval `[3×max, 2n+1)`.

Do you think I should implement this logic in this PR?

**Comment by briansmith** [comment]
First, if you do implement it, here's the above in Rust comment form:

```
// `x` is of the form 0b100..... This means:
//
//    x < 2**n - 2**(n-2) - 2**(n-3).
//
// This means that `3*x < 2**(n+1)`. Proof:
//
//  3*x < 3*(2**n - 2**(n-2) - 2**(n-3))
//      < (2 + 1)*(2**n - 2**(n-2) - 2**(n-3))
//      < 2*(2**n - 2**(n-2) - 2**(n-3)) + 2**n - 2**(n-2) - 2**(n-3)
//      < 2**(n+1) - 2**(n-1) - 2**(n-2) + 2**n - 2**(n-2) - 2**(n-3)
//      < 2**(n+1) + 2**n - 2**(n-1) - 2**(n-2) - 2**(n-2) - 2**(n-3)
//      < 2**(n+1) + 2**n - 2**(n-1) - 2*(2**(n-2)) - 2**(n-3)
//      < 2**(n+1) + 2**n - 2**(n-1) - 2**(n-1) - 2**(n-3)
//      < 2**(n+1) + 2**n - 2*(2**(n-1)) - 2**(n-3)
//      < 2**(n+1) + 2**n - 2**n - 2**(n-3)
//      < 2**(n+1) - 2**(n-3)
//
//  Then clearly 2**(n+1) - 2**(n-3) < 2**(n+1) since n is positive.
//
// This means we can generate a value in the range [0, 2**(n+1)), which would
// fall into one of four sub-intervals:
//
//    [0, max)          => Return the value as-is.
//    [max, 2*max)      => Return `value - max`.
//    [2*max, 3*max)    => Return `value - max - max`.
//    [3*max, 2**(n+1)) => Generate a new random value and try again.
//
// This avoids biasing the result towards small values, which is what reducing
// the random value (mod max) would do, while reducing the probability that a
// new random value will be needed.
```

I don't want to get in the way of incremental progress. If you want to implement this logic, I do think it is better to have it than not have it. But, we could also add it later, I guess.

This brings up the issue again about whether it makes sense for the ECC-oriented random value generation and the RSA-oriented random value generation are actually similar enough to share code. Now I feel like the RSA logic is so different from the ECC logic, because of this and because of the extra masking that RSA requires, that I'd rather have them be separate. So, at least, don't spend too much effort integrating them. (Sorry to lead you down the wrong path in this respect.)

**Comment by djudd** [comment]
Been busier than expected and might not get to work on this for another couple of weeks, but still planning to come back to it.

**Comment by djudd** [comment]
@briansmith I was looking at your change which added `BN_rand_range_ex` and noticed that `BN_cmp_word`, which is now used for the min_inclusive comparison, is not constant-time. And then I noticed that we were already using `BN_cmp`, which is also not constant-time, so `BN_rand_range` doesn't seem ever to have been constant-time. But the `scalar_is_in_range` Rust function I replaced was careful to be constant-time. Is this just another case where the requirements are different - we don't care about constant-time comparisons in BN_rand_range because it's only used for blinding?

**Comment by briansmith** [comment]
> @briansmith I was looking at your change which added BN_rand_range_ex and noticed [...]

tl;dr: Because of `BN_mod_inverse_blinded` for (re)creating the RSA blinding, it does seem like the RSA-related uses of BN_rand_range_ex` don't need to be so careful about timing leaks. If it is straightforward to make the code in *ring* constant-time, then let's do so. If not, let's consider doing something simpler for RSA, as I did in BoringSSL.

I had been experimenting with the implementation of constant-time inversion for RSA blinding. Unfortunately, the performance of the constant-time implementation was pretty bad, and efforts to make it faster made it hard to read and still not very fast. Also, this particular effort (this PR here) demonstrated that trying to make all this stuff constant-time is creating complication, and also the more we learn about it the less similar the RSA blinding and ECC keygen seem. Thus, I decided to try to avoid the need for constant-timeness for the blinding inversion in the hope that it makes our work on RSA-blinding-related stuff easier. That's why I created `BN_mod_inverse_blinded` and `BN_rand_range_ex` to support it.

As for `BN_cmp_word`, that is a workaround for the way the OpenSSL/BoringSSL `BIGNUM` API works. it isn't constant time because it's current goal is to avoid creating a `BIGNUM` object for the comparison. In *ring*, I suspect that we could easily do these comparisons using our existing constant-time techniques, replacing `BN_cmp_word` with a constant-time thing. I would have liked to do the same for BoringSSL, but contributing the `GFp_constant_time_*` stuff to BoringSSL is too much work for now.

**Comment by djudd** [comment]
I've updated PR to avoid any significant changes to the EC code path for now. It doesn't include the special case `0b100...` logic yet, I'll add that in a separate commit at some point.

I don't think this is ready to merge, so feel free to wait on reviewing, but FYI this is what I'm thinking about (besides the `0b100` case):
* Should the `Range` struct and `GFp_rand_range_ex` be moved into a `blinding.rs`, or just into `rsa.rs`? (I'm leaning yes)
* How do we deal with the fact that `Range`'s APIs appear to support `min_inclusive` up to `2^64-1`, but won't actually be able to randomly sample for the highest values? (This was also true with the C implementation of `BN_rand_range_ex`; I'm leaning towards changing the API so `min_inclusive` is just a `u8`)

**Comment by briansmith** [comment]
> How do we deal with the fact that Range's APIs appear to support min_inclusive up to 2^64-1, but won't actually be able to randomly sample for the highest values? (This was also true with the C implementation of BN_rand_range_ex; I'm leaning towards changing the API so min_inclusive is just a u8)

First, it is inclusive, right, so in theory it could sample for the highest values, but it would result in a lot of loop iterations and we'd give up.

The C code in `BN_rand_range_ex` uses `BN_ULONG` for consistency with the other code and because it is likely that any constant-time stuff added to the BoringSSL `BIGNUM` library will be based on `BN_ULONG`.

In BoringSSL, the values {0, 1, 2} are used for `min_inclusive`. In *ring*, we only need {0, 1}, I believe. Also, the choice of zero or one is static. Therefore, we could do something like this:

```rust
#[cfg(target_pointer_width = "64")]
pub type Limb = u64;

#[cfg(target_pointer_width = "32")]
pub type Limb = u32;

#[cfg(target_pointer_width = "64")]
#[repr(u64)]
enum RangeMin { ZERO = 0, ONE = 1 }

#[cfg(target_pointer_width = "32")]
#[repr(u32)]
enum RangeMin { ZERO = 0, ONE = 1 }

fn main() {
  let foo = RangeMin::ZERO;
  let bar = RangeMin::ONE;
   
  assert_eq!(foo as Limb, 0);
  assert_eq!(bar as Limb, 1);
}
```

Unfortunately, `#[repr(Limb)]` doesn't work.

briansmith:
> Should the Range struct and GFp_rand_range_ex be moved into a blinding.rs, or just into rsa.rs? (I'm leaning yes)

I do think that they should live in the same place as the other RSA blinding code. Whether the RSA blinding code should be split out into blinding.rs or whatever, I am less sure about. More comments about splitting up rsa.rs at https://github.com/briansmith/ring/issues/263.

briansmith:
> In BoringSSL, the values {0, 1, 2} are used for min_inclusive. In ring, we only need {0, 1},

Actually, in *ring*, don't we just need the minimum to always be one? AFAICT, we don't even need it to be paraneterized.

crypto/bn/random.c
```diff
@@ -179,59 +179,22 @@ int BN_rand(BIGNUM *rnd, int bits, int top, int bottom, RAND *rng) {
 
 int BN_rand_range_ex(BIGNUM *r, BN_ULONG min_inclusive,
```
briansmith: You can remove the `min_inclusive` argument since it is always 1.

src/limb.rs
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+}
+
+#[allow(unsafe_code)]
+#[allow(non_snake_case)]
+#[doc(hidden)]
+#[no_mangle]
+pub unsafe extern fn GFp_rand_range_ex(dest: *mut Limb,
+                                       min_inclusive: Limb,
```
briansmith: Ditto, it seems `min_inclusive` is always 1, so it doesn't need to be a parameter.

src/limb.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+#[cfg(target_pointer_width = "32")]
+fn limbs_as_bytes_mut<'a>(src: &'a mut [Limb]) -> &'a mut [u8] {
+    polyfill::slice::u32_as_u8_mut(src)
+}
+
+// Assumes limbs are ordered least-significant to most-significant, and is not
+// constant time. TODO Is there an easy constant-time implementation?
+fn limbs_gte_limb(limbs: &[Limb], min_inclusive: Limb) -> bool {
```
briansmith: You can replace this with a call to `GFp_constant_time_limbs_are_zero` since the minimum is always 1.

djudd:
Ok, I've updated to remove support for explicit `min_inclusive` (back to the way it was in the beginning, in other words). Given the work you started in your other branch, my preference, I think, would be to merge as-is, and then go back and add the `0b100...` support and factor out the blinding code into a common module in separate PRs, so that I'm not chasing a changing master as I do so. But if you think the `0b100...` bit at least is a blocker to merge, which would be reasonable, I will do that first. Might be able to get to it tomorrow, depending on how long it takes.

briansmith:
> k, I've updated to remove support for explicit min_inclusive (back to the way it was in the beginning, in other words).

Great

> Given the work you started in your other branch, my preference, I think, would be to merge as-is, and then go back and add the 0b100... support and factor out the blinding code into a common module in separate PRs, so that I'm not chasing a changing master as I do so.

Great.

> But if you think the 0b100... bit at least is a blocker to merge, which would be reasonable, I will do that first. Might be able to get to it tomorrow, depending on how long it takes.

No, it absolutely isn't a blocker.

src/ec/suite_b/ops/ops.rs
```diff
@@ -215,11 +210,9 @@ impl CommonOps {
...
     pub fn elem_verify_is_not_zero(&self, a: &ElemUnreduced)
                                    -> Result<(), error::Unspecified> {
-        match unsafe {
-            GFp_constant_time_limbs_are_zero(a.limbs.as_ptr(), self.num_limbs)
-        } {
-            0 => Ok(()),
-            _ => Err(error::Unspecified),
+        match limbs_are_zero_constant_time(&a.limbs[..self.num_limbs]) {
+            false => Ok(()),
+            true => Err(error::Unspecified),
```
briansmith: we have a function `GFp_constant_time_limbs_are_zero` that returns `(Limb)-1` for TRUE and `0` for FALSE. `limbs_are_zero_constant_time` maps `(Limb)-1` => `true` and `0` => `false`. Then we map `true` => `Ok(())` and `false => `Err(error::Unspecified)`. This seems quite circuitous to me. I'll try to write a follow-up commit to this that creates a `BooleanMask` enum type that can be used to remove at least one layer of mapping here.

src/lib.rs
```diff
@@ -130,6 +130,13 @@ mod init;
 pub mod pbkdf2;
 pub mod rand;
 
+mod limb;
+
+// Really a private method; only has public visibility so that C compilation
+// can see it.
+#[doc(hidden)]
+pub use limb::GFp_rand_mod;
```
briansmith: I learned that the `public-from-private` or whatever error is generated by a lint, and we might be able to turn that lint off using `#[allow(...)]`. I'll file a follow-up for trying that.

src/limb.rs
```diff
@@ -0,0 +1,220 @@
...
+
+/// References a positive integer range `[1..max_exclusive)` where
+/// `max_exclusive` is a slice of limbs ordered least-significant-limb to
+/// most-significant-limb (and the limbs use the native endianness).
+pub struct Range<'a> {
+    pub max_exclusive: &'a [Limb],
+}
+
+impl <'a>Range<'a> {
+
```
briansmith: Let's get rid of these blank lines.

src/limb.rs
```diff
@@ -0,0 +1,220 @@
...
+    pub fn from_max_exclusive(max_exclusive: &[Limb]) -> Range {
+
```
briansmith: ...and this blank line.

src/ec/suite_b/ops/ops.rs
djudd: Ok. I thought the indirection was worth it to remove an unneeded use of `unsafe` and keep the unsafe implementation private to `limb.rs`

src/limb.rs
```diff
@@ -0,0 +1,220 @@
...
+        debug_assert!(!limbs_are_zero_constant_time(max_exclusive));
+        debug_assert!(max_exclusive[max_exclusive.len() - 1] > 0);
+
+        Range {
+            max_exclusive: max_exclusive
+        }
+    }
+
+    fn leading_zero_bits(&self) -> usize {
```
briansmith: I think this is duplicating these functions (which I just learned about accidentally):

https://doc.rust-lang.org/std/primitive.u32.html#method.leading_zeros
https://doc.rust-lang.org/std/primitive.u64.html#method.leading_zeros

src/limb.rs
```diff
@@ -0,0 +1,220 @@
...
+        // XXX: The value 100 was chosen to match OpenSSL due to uncertainty of
+        // what specific value would be better, but it seems bad to try 100 times.
+        for _ in 0..100 {
+            {
+                let mut dest_as_bytes = limbs_as_bytes_mut(dest);
+                try!(rng.fill(&mut dest_as_bytes));
+            }
+
+            if bits_to_mask > 0 {
+                let mask: Limb = (1 << (LIMB_BITS - bits_to_mask)) - 1;
```
briansmith: Thanks for factoring out the leading zero bits calculation.

I think we should probably follow this up by factoring out the calculation of mask out of the loop:

```
let mask = if bits_to_mask > 0 { 1 << (LIMB_BITS - bits_to_mask)) - 1 } else { LIMB_MAX };
```

Actually, if we look ahead a little bit to how we'll eventually rewrite `BN_mod_inverse_blinded`, we actually need to generate 2 random values in the same range, so it might make the most sense to store the mask as a field of `Range`. It won't make anything faster but it will be easier to see that the mask is purely function of the modulus, not a function of any of the input.

src/limb.rs
```diff
@@ -0,0 +1,220 @@
...
+                dest[self.max_exclusive.len() - 1] &= mask;
+            }
+
+            if !self.are_limbs_within(&dest) {
+                continue;
+            }
+
+            return Ok(());
```
briansmith: It seems like the various refactorings have left us with something that can be further simplified:

```
if self.are_limbs_within(&dest) { 
    return Ok(());
}
```.

File: src/ec/suite_b/ops/ops.rs (hunk 22 above)
Comment (briansmith) [inline src/ec/suite_b/ops/ops.rs]: No worries. I'll try to see if I can simplify it myself. I think there's other code that can benefit from such a `BooleanMask` type.

File: src/limb.rs (hunk 27 above)
Comment (briansmith) [inline src/limb.rs]: I investigated how we can improve this later a bit: https://github.com/briansmith/ring/issues/276.

File: crypto/bn/bn_test.cc [hunk 29]
```diff
@@ -655,50 +654,6 @@ static bool TestHex2BN() {
...
-    return false;
-  }
-
-  // Test BN_rand accounts for degenerate cases with |top| and |bottom|
-  // parameters.
-  if (!BN_rand(bn.get(), 0, 0 /* top */, 0 /* bottom */, rng) ||
-      !BN_is_zero(bn.get())) {
-    fprintf(stderr, "BN_rand gave a bad result.\n");
-    return false;
-  }
```
briansmith: We should keep the ones where top == 0 and bottom == 0 until we remove BN_rand.

briansmith:
Please see https://github.com/briansmith/ring/commits/oxidize-bn-rand-range. This attempts to address all my review comments plus a few more, plus factors out the masking of the top limb and adds tests for it.

Please add this to the top of the new Rust source file you added:

```
/// Copyright 2016 David Judd
/// Copyright 2016 Brian Smith
///
/// <normal license>
```

I didn't change the second commit; please add back the BN_rand tests for the (0 ,0) case. Also, please add the "I agree..." statement to the end of that commit's commit message.

Also, please verify that you are the sole copyright holder for your contribution.

Thanks!

src/limb.rs
djudd: Huh, I googled for something like these and didn't have any luck. I guess I should have just scanned the methods on u32/64 instead.

src/limb.rs
djudd: This logic will also have to change for the `0b100` special case, which I started poking at a little - we'll want to mask one fewer bit, basically. So I don't think we should worry about further optimization yet. (Also re that: I'm going to do some microbenchmarks to see if the added complexity is really a net gain before committing anything.)

djudd:
Addressed some of your review comments before seeing your link above, whoops. Shouldn't matter, but I'm not sure what you want me to do with the commits in your version of the branch - pull into mine, but leave as separate commits?

I'm the sole copyright holder for everything I've committed, with the exception of code copied and pasted from within ring and presumably copyrighted by you.

crypto/bn/bn_test.cc
djudd: It seems a little silly to me to test a test library that's going away soonish, but your call.

briansmith: I agree, it is unfortunate. But, the test is already there so it's just a matter of deleting less of it. The test for `BN_rand` partially establishes the validity of the `BN_bn2bin_padded` test. Ideally we'd rewrite the `BN_bn2bin_padded` test "soonish" but I kind of suspect it's likely to stay like it is for a while as it's "just a test" and so likely not very urgent to everybody. Especially since we've already remove `BN_rand_range` and `BN_rand` from the main lib, which is the main win.

src/limb.rs
briansmith: I will comment about it in #276. The issue is whether or not we need the operation to be constant-time with respect to `max_exclusive`. Right now, we don't. Maybe in the future we will. So, we don't need to worry about it until the future.

crypto/bn/bn_test.cc
djudd: Ok, fair enough.

briansmith:
@djudd I rebased my branch on top of your updated branch and folded some of the commits I added into them. I left some of my commits as separate commits. Could you please take a look at them and let me know if anything needs improving?: https://github.com/briansmith/ring/commits/oxidize-bn-rand-range

Thanks!

djudd:
I left a couple of comments but otherwise looks good. I was thinking about adapting the ECDH tests myself but didn't because on that iteration of the PR they were testing the same codepath. Now that these are separate codepaths, definitely good to have them!

djudd:
BTW, microbenchmark results suggest the 0b100 special case is worth keeping. Sampling to a bound of `[0, 0, 0, 0, 0, 0, 0, 16]` takes 7 microseconds with it, vs 10 without, on my Macbook. And that's with first-pass code and a constant-time subtraction. (Not the result I was rooting for :/).

briansmith:
Closed via:

9cf1e16 Replace scalar_is_in_range with direct use of `Range`.
1bc7c7f Adapt the ECDH key generation tests to test Range::sample_into_limbs.
36a91ba Expand `Range::are_limbs_within` tests.
bb52036 Simplify conversions from masks => bool => Err to masks => Err.
01fea0f Expand and clarify comments regarding origin of sample_into_limbs.
ee97b43 Factor out most-significant-limb masking and test it.
d25ceb6 Make RAND_bytes test-only.
f96d640 Make BN_rand test-only.
0a8efed Use `Limb::max_value()` instead of defining `MAX_LIMB`.
ef7a07c Describe limb ordering in just one place.
7cbe5d7 Replace BN_rand_range C implementation with call to Rust.

Thanks for your work on this!

djudd:
Thanks for your time and patience in helping me get it right!

briansmith:
Thanks again! Please submit your other PR, for dealing with the 0b100... case, when it is ready. It will probably be ~1 week before I'll be able to look at it because I'm working on clearing a backlog of stuff.

PR #255, opened 2016-07-19
Use enum to dispatch to signature verification algorithm handlers.

djc:

This seems the best solution that I can think of for the problems we identified in #226. The solution for the problem there is then that `verify_rsa()` can accept a `VerificationAlgorithm` and match on `RSA`, or return `Err(())` for other algorithms.

I'm not sure how to test if this new `VerificationAlgorithm` is marked as `Sync` by the compiler. However, it would seem that this whole setup is much more transparent for the compiler then it was with the trait. Also, `enum`s cannot be extended the way traits can be.

It might make sense to align the names of `EdDSA`, `ECDSAVerification` and `RSAVerificationAlgorithm` more, but I left that for follow-up, so you can review  this approach first.


djc:
So the other direction we could go with this, which I think I slightly prefer, is to expose `EdDSA`, `ECDSAVerification` and `RSAVerificationAlgorithm` as opaque types in the `signature` APIs. Then, we expose the parameter sets (like `RSA_PKCS1_2048_8192_SHA256`) as their original type, and require callers of generic API (like `signature::verify()`) to wrap (a reference to) them in the enum, exposing the enum variants in the process. We could provide optional `Into` trait implementations that help coercion of `&EdDSA` to `VerificationAlgorithm::ED25519` etc.

briansmith:
I've been swamped and haven't gotten to this yet.

My first instinct is that I feel like we've reached the maximum level of contortion to try to make things work the current way. Maybe we just need to accept that Rust requires the awkward extra `use`s for traits to work. Nobody seems to be doing things the way *ring* is, and why should *ring* work differently than other libs?

I'm curious what others think.

djc:
Actually, my instinct once I figured out this patch was that it seems to a better design for what you were trying to do in the first place. On the other hand, I do feel traits are probably the more idiomatic way to do this, even if they require the stupid extra `use`.

How do you propose getting more feedback? I could open a thread on `users.rust-lang.org` about enums vs traits and see what they say. Or maybe you can tweet about this PR and see what people think that way.

djc:
I posted a link to this PR in #rust-crypto, and @tarcieri seems to think this is an improvement. No responses from anyone else.

tarcieri:
I also said I could go either way ¯\_(ツ)_/¯

src/signature.rs
```diff
@@ -246,7 +241,14 @@ pub fn verify(alg: &VerificationAlgorithm, public_key: untrusted::Input,
               msg: untrusted::Input, signature: untrusted::Input)
               -> Result<(), ()> {
     init::init_once();
-    alg.implementation.verify(public_key, msg, signature)
+    match alg {
```
briansmith: One of the main goals is to make it so, for example, if you only use Ed25519, then the compiler/linker won't bring along ECDSA and RSA with it. This is why we never switch on enums like this, because it makes the dead code elimination in the compiler nearly impossible. So, unless we have clear evidence that the compiler consistently optimizes away the unused cases in this match, I'd definitely like to find a different solution to this.

briansmith:
It seems like the trait-based approach is going to win. Closing this. Let me know if I'm misunerstanding.

PR #258, opened 2016-07-25
WIP: Reuse `BN_BLINDING` multiple times in RSA signing

### LINKED ISSUE #234: Reuse `BN_BLINDING` multiple times in RSA signing
Right now, we create a new temporary `BN_BLINDING` value before signing and then throw it away right after signing, never using it again. Instead, we should make use of the reuse facilities already built into `BN_BLINDING` by saving a pointer to the `BN_BLINDING` in the `RSAKeyPair`. It should be saved in a `Mutex` so that multiple threads can access the blinding safely.

ranweiler:
Closes #234.

**Todo**:
- [x] Save a `Mutex`-wrapped pointer to `BN_BLINDING` in `RSAKeyPair`
- [x] Export `BN_BLINDING_COUNTER` as an extern `GFp_BN_BLINDING_COUNTER`
- [x] Create a test that uses the same `BN_BLINDING` at least `GFp_BN_BLINDING_COUNTER` +1 times, to exercise all the code paths in blinding.c
- [x]  Create a test that is similar to `test_agreement_suite_b_ecdh_generate`, where the `RAND` returns unusable values forever, to test the logic where the blinding gives up after a bunch of tries.
- [ ] Ensure `BN_BLINDING` coverage via coveralls report

**Questions**:
- ~~How do we want to handle a poisoned `Mutex`? Right now this work just propagates a panic with `unwrap()`.~~ 

We agreed that a poisoned mutex warrants a panic.

- ~~Right now the blinding reuse test is inadequate— it only checks that the `BN_BLINDING` counter is updated. Ideally we'd check that the actual `A` and `Ai` blinding values have been regenerated when the counter is reset. What's the "cheapest" way to do this, wrt minimizing extraneous FFI work in this PR?~~ 

We are accepting the committed `BN_BLINDING` sync burden in exchange for the level of assurance offered by checking the counter, and stopping there for now.

- ~~There is a function-local `extern` (for `GFp_BN_BLINDING_COUNTER`) and `use` (for `BN_BLINDING`) in `test_signature_rsa_pkcs1_sign_blinding_reuse`. We should make sure we're happy with that code style.~~

Reverted to module-level imports.

### DIFF
--- a/src/rsa.rs
+++ b/src/rsa.rs
@@ -430,9 +430,15 @@ impl Drop for PositiveInteger {
 #[cfg(feature = "rsa_signing")]
 enum BIGNUM {}
 
+/// Needs to be kept in sync with `struct bn_blinding_st` (in `crypto/rsa/blinding.c`).
 #[cfg(feature = "rsa_signing")]
 #[allow(non_camel_case_types)]
-enum BN_BLINDING {}
+#[repr(C)]
+struct BN_BLINDING {
+    a: *mut BIGNUM,
+    ai: *mut BIGNUM,
+    counter: u32,
+}
 
 #[cfg(feature = "rsa_signing")]
 #[allow(non_camel_case_types)]
--- a/src/rsa.rs
+++ b/src/rsa.rs
@@ -662,4 +662,30 @@ mod tests {
             assert_eq!(counter, (prev_counter + 1) % GFp_BN_BLINDING_COUNTER);
         }
     }
+
+    // In `crypto/rsa/blinding.c`, when `bn_blinding_create_param` fails
+    // to randomly generate an invertible blinding factor too many times
+    // in a loop, it returns an error. Check that we observe this.
+    #[cfg(feature = "rsa_signing")]
+    #[test]
+    fn test_signature_rsa_pkcs1_sign_blinding_creation_failure() {
+        const MESSAGE: &'static [u8] = b"hello, world";
+
+        // Stub RNG that is constantly 0. In `bn_blinding_create_param`,
+        // this causes the candidate blinding factors to always be 0,
+        // which has no inverse, so `BN_mod_inverse_no_branch` fails.
+        let rng = rand::test_util::FixedByteRandom { byte: 0x00 };
+
+        const PRIVATE_KEY_DER: &'static [u8] =
+            include_bytes!("signature_rsa_example_private_key.der");
+        let key_bytes_der = untrusted::Input::from(PRIVATE_KEY_DER);
+        let key_pair = RSAKeyPair::from_der(key_bytes_der).unwrap();
+
+        let mut signature = vec![0; key_pair.public_modulus_len()];
+
+        let result = key_pair.sign(&RSA_PKCS1_SHA256, &rng, MESSAGE,
+                                   &mut signature);
+
+        assert!(result.is_err());
+    }
 }
--- a/src/rsa.rs
+++ b/src/rsa.rs
@@ -227,6 +227,7 @@ fn parse_public_key<'a>(input: untrusted::Input<'a>) ->
 #[cfg(feature = "rsa_signing")]
 pub struct RSAKeyPair {
     rsa: std::boxed::Box<RSA>,
+    blinding: std::sync::Mutex<*mut BN_BLINDING>,
 }
 
 #[cfg(feature = "rsa_signing")]
@@ -295,7 +296,15 @@ impl RSAKeyPair {
                 try!(bssl::map_result(unsafe {
                     rsa_new_end(rsa.as_mut())
                 }));
-                Ok(RSAKeyPair { rsa: rsa })
+                let blinding = unsafe { BN_BLINDING_new() };
+                if blinding.is_null() {
+                    return Err(());
+                }
+
+                Ok(RSAKeyPair {
+                    rsa: rsa,
+                    blinding: std::sync::Mutex::new(blinding),
+                })
             })
         })
     }
@@ -365,6 +374,7 @@ impl Drop for RSAKeyPair {
             BN_MONT_CTX_free(self.rsa.mont_qq);
             BN_free(self.rsa.qmn_mont);
             BN_free(self.rsa.iqmp_mont);
+            BN_BLINDING_free(*(self.blinding.lock().unwrap()));
         }
     }
 }
--- a/src/rsa.rs
+++ b/src/rsa.rs
@@ -344,14 +344,12 @@ impl RSAKeyPair {
         try!(padding_alg.pad(msg, signature));
         let mut rand = rand::RAND::new(rng);
         bssl::map_result(unsafe {
-            let blinding = BN_BLINDING_new();
-            let ret =
-                GFp_rsa_private_transform(self.rsa.as_ref(),
-                                          signature.as_mut_ptr(),
-                                          signature.len(), blinding,
-                                          &mut rand);
-            BN_BLINDING_free(blinding);
-            ret
+            let blinding = *(self.blinding.lock().unwrap());
+
+            GFp_rsa_private_transform(self.rsa.as_ref(),
+                                      signature.as_mut_ptr(),
+                                      signature.len(), blinding,
+                                      &mut rand)
         })
     }
 }
@@ -505,6 +503,9 @@ mod tests {
     use super::*;
     use untrusted;
 
+    #[cfg(feature = "rsa_signing")]
+    extern { static GFp_BN_BLINDING_COUNTER: u32; }
+
     #[test]
     fn test_signature_rsa_pkcs1_verify() {
         test::from_file("src/rsa_pkcs1_verify_tests.txt", |section, test_case| {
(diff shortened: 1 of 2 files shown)


src/rsa.rs
```diff
@@ -295,7 +296,11 @@ impl RSAKeyPair {
                 try!(bssl::map_result(unsafe {
                     rsa_new_end(rsa.as_mut())
                 }));
-                Ok(RSAKeyPair { rsa: rsa })
+                let blinding = unsafe { BN_BLINDING_new() };
```
briansmith: `BN_BLINDING_new()` might return NULL. We need to check that.

src/rsa.rs
```diff
@@ -227,6 +227,7 @@ fn parse_public_key<'a>(input: untrusted::Input<'a>) ->
 #[cfg(feature = "rsa_signing")]
 pub struct RSAKeyPair {
     rsa: std::boxed::Box<RSA>,
+    blinding: std::sync::Mutex<*mut BN_BLINDING>,
```
briansmith: We have the `rsa` as a `Box`. Should we have the blinding as a `Mutex<Box>`?

crypto/rsa/internal.h
```diff
@@ -64,6 +64,7 @@
 extern "C" {
 #endif
 
+const uint32_t GFp_BN_BLINDING_COUNTER;
```
briansmith: We can drop this. We only put stuff in the headers if it is used by other C code in other files, not if it is used by only Rust code.

src/rsa.rs
```diff
@@ -432,7 +432,12 @@ enum BIGNUM {}
 
 #[cfg(feature = "rsa_signing")]
 #[allow(non_camel_case_types)]
-enum BN_BLINDING {}
+#[repr(C)]
+struct BN_BLINDING {
+    a: *mut BIGNUM,
+    ai: *mut BIGNUM,
+    counter: u32,
+}
```
briansmith: Usually we avoid doing this, so I'd like to know more about the motivation.
Which test do you expect will need to access these?

briansmith:
> How do we want to handle a poisoned Mutex? Right now this work just propagates a panic with unwrap().

The Mutex will only become poisoned if some code panics while it is holding the mutex, right? And, (1) we avoid holding the mutex for a long time, and (2) we avoid panicing as much as possible. So, this can basically never happen, right? If you agree, I am fine with panicing on a poisoned mutex.

briansmith:
> Right now the blinding reuse test is inadequate— it only checks that the BN_BLINDING counter is updated. Ideally we'd check that the actual A and Ai blinding values have been regenerated when the counter is reset. What's the "cheapest" way to do this, wrt minimizing extraneous FFI work in this PR?

Well, first, I'm not sure such a test is necessary. But, if we want to do it, we could create a C function `BN_BLINDING_check_consistency` that verifies that `a` * `ai` == 1 mod m and which returns the counter. Then (AFAICT) you wouldn't need the `#[repr(C)]` definition of the struct and you'd need one one FFI declaration for the `BN_BLINDING_check_consistency` function. WDYT?

briansmith:
> There is a function-local extern (for GFp_BN_BLINDING_COUNTER) and use (for BN_BLINDING) in test_signature_rsa_pkcs1_sign_blinding_reuse. We should make sure we're happy with that code style.

I usually don't do that. Instead I would do them at the module level in the `tests` submodule. But, I'm not sure I care much either way.

src/rsa.rs
briansmith: I see we already did this with `RSA` and I guess that is what allowed us to use `Box`? So, I guess I don't care strongly either way. But, we need to include the comments like this in both the Rust and C definitions of the struct:

/// Needs to be kept in sync with `struct rsa_st` (in `include/openssl/rsa.h`).

src/rsa.rs
briansmith: Never mind. I think the way you did it is fine.

ranweiler: Yeah, my assumption is that we'll always be passing `blinding` pointer by value when crossing the FFI boundary, so there's no need to heap allocate, unlike the `rsa` field.

crypto/rsa/internal.h
ranweiler: Cool, will keep that standard in mind.

src/rsa.rs
ranweiler: I would like to avoid this, if we can, especially since it is only for the sake of testing. Responding below re alternatives.

ranweiler:
> The Mutex will only become poisoned if some code panics while it is holding the mutex, right? And, (1) we avoid holding the mutex for a long time, and (2) we avoid panicing as much as possible. So, this can basically never happen, right? If you agree, I am fine with panicing on a poisoned mutex.

I agree with this.

ranweiler:
> Well, first, I'm not sure such a test is necessary. But, if we want to do it, we could create a C function BN_BLINDING_check_consistency that verifies that a * ai == 1 mod m and which returns the counter. Then (AFAICT) you wouldn't need the #[repr(C)] definition of the struct and you'd need one one FFI declaration for the BN_BLINDING_check_consistency function. WDYT?

I really like the idea of only touching the `BN_BLINDING` fields in C code, if possible. On the other hand, creating a new function in C (not used anywhere else) just for the sake of Rust testing feels like adding more overall complexity for less assurance. Unless we also test the `BN_BLINDING_check_consistency` function, we're now depending on its correctness, too.

Since with either approach we're relying on manual inspection for the correctness of `bn_blinding_update`, and just using the `counter` as a means of checking that its paths are being exercised, I'm inclined to leave the filled-in `BN_BLINDING` (+ comment). I don't expect the C `BN_BLINDING` struct to change, either, so it seems like a negligible synchronization burden for the Rust code.

ranweiler:
>> There is a function-local extern (for GFp_BN_BLINDING_COUNTER) and use (for BN_BLINDING) in test_signature_rsa_pkcs1_sign_blinding_reuse. We should make sure we're happy with that code style.
>
> I usually don't do that. Instead I would do them at the module level in the tests submodule. But, I'm not sure I care much either way.

I'm going to move them to the module level to be less surprising, as I don't have an actual reason to make them function local.

briansmith:
> Since with either approach we're relying on manual inspection for the correctness of bn_blinding_update, and just using the counter as a means of checking that its paths are being exercised, I'm inclined to leave the filled-in BN_BLINDING (+ comment). I don't expect the C BN_BLINDING struct to change, either, so it seems like a negligible synchronization burden for the Rust code.

That's fine with me. I think this is a good strategy because I think it will motivate somebody to convert `BN_BLINDING` into a "native" Rust structure, using `PositiveInteger`, etc., in the future.

I see that you made a bunch of fixes to the code. Could you please fold them into the original commits that they are relevant to? e.g. Fold "Check for `NULL` return value from `BN_BLINDING_new` " into the patch that adds the call to `BN_BLINDING_new`. That would make it easier to review.

Thanks!

ranweiler:
> I see that you made a bunch of fixes to the code. Could you please fold them into the original commits that they are relevant to?

Squashed and force pushed, continuing with remaining work.

src/rsa.rs
```diff
@@ -430,9 +430,15 @@ impl Drop for PositiveInteger {
...
 
+/// Needs to be kept in sync with `struct bn_blinding_st` (in `crypto/rsa/blinding.c`).
 #[cfg(feature = "rsa_signing")]
 #[allow(non_camel_case_types)]
-enum BN_BLINDING {}
+#[repr(C)]
+struct BN_BLINDING {
+    a: *mut BIGNUM,
+    ai: *mut BIGNUM,
+    counter: u32,
```
briansmith: Please change the C definition to use `uint32_t` instead of `unsigned` just out of an abundance of caution.

src/rsa.rs
```diff
@@ -295,7 +296,16 @@ impl RSAKeyPair {
                 try!(bssl::map_result(unsafe {
                     rsa_new_end(rsa.as_mut())
                 }));
-                Ok(RSAKeyPair { rsa: rsa })
+                let blinding = unsafe { BN_BLINDING_new() };
+                if blinding.is_null() {
+                    Err(())
+                }
+                else {
```
briansmith: `else` should go on the same line as the closing brace of the `if`.

briansmith: In *ring* we usually use the early-return-on-errors style consistently (if we don't, it's probably not intentional). So, I'd prefer to have it be `return Err(());` with no `else`.

src/rsa.rs
```diff
@@ -345,14 +345,12 @@ impl RSAKeyPair {
...
         bssl::map_result(unsafe {
-            let blinding = BN_BLINDING_new();
-            let ret =
-                GFp_rsa_private_transform(self.rsa.as_ref(),
-                                          signature.as_mut_ptr(),
-                                          signature.len(), blinding,
-                                          &mut rand);
-            BN_BLINDING_free(blinding);
-            ret
+            let blinding: *mut BN_BLINDING = *(self.blinding.lock().unwrap());
```
briansmith: NIT: drop the type signature.

src/rsa.rs
ranweiler: Ha, that was actually what I first wrote. I'll go with the early return, with the `} else` style noted in case it pops up again.

src/rsa.rs
ranweiler: Ah, it looks like those annotations are now vestigial! In some previous iteration they were necessary to deref through the `MutexGuard`. Removing throughout.

src/rsa.rs
```diff
@@ -628,4 +629,37 @@ mod tests {
...
+            include_bytes!("signature_rsa_example_private_key.der");
+        let key_bytes_der = untrusted::Input::from(PRIVATE_KEY_DER);
+        let key_pair = RSAKeyPair::from_der(key_bytes_der).unwrap();
+
+        let mut signature = vec![0; key_pair.public_modulus_len()];
+
+        for _ in 0 .. GFp_BN_BLINDING_COUNTER + 1 {
+            let prev_counter = unsafe {
+                let blinding = *(key_pair.blinding.lock().unwrap());
+                (*blinding).counter
```
ranweiler: Since the type annotations for `blinding` ended up being unnecessary, we _could_ make this a one-liner, as well as the similar line below. I tried it and found it less readable, but don't feel strongly about it.

ranweiler:
@briansmith, comments addressed and squashed into the relevant commits.

briansmith:
Great. Are you the copyright holder on this? If not, who is? And do you have permission to license it? (Same questions I ask everybody.)

src/rsa.rs
briansmith: When I commit this, I'm going to move the `let blinding...` out of the `unsafe` block. Our convention is to only put lines in the `unsafe` block if they are required to be `unsafe`.

briansmith: Oh...maybe not, because then I'd need another block, I think. Never mind.

ranweiler:
> Great. Are you the copyright holder on this? If not, who is? And do you have permission to license it? (Same questions I ask everybody.)

I am the copyright holder, and I do have permission to license it. I hereby assign copyright in this code to the project, to be licensed under the same terms as the rest of the code.

briansmith:
> I am the copyright holder, and I do have permission to license it. I hereby assign copyright in this code to the project, to be licensed under the same terms as the rest of the code.

Great. I don't require copyright assignment. Just I want to make sure people understand what's going on with the licensing. If any of the copyright stuff is different with your future contributions, please let me know. Otherwise, I'll assume you're the copyright holder for them too.

briansmith:
Landed with very minor changes (expanded some comments, rewrapped some things to 80 columns):

1405732ee8f1e10fb40904247e71be61e3d6b504
4a52d34b81c8f574e157bc44bcea8bdcbd9908c1
36fbff76ebdaf93e5edcad086e8df060f8336db0
f475e5766ea6d54b2bdc9861a986b5b1653f66f5
f1a140a75229e32a1b8bea1637640f48e0611b80

Very good work. Thank you very much for taking this on! I hope we keep working together.

ranweiler:
> Thank you very much for taking this on! I hope we keep working together.

Thanks, looking forward to it!

Issue #202, opened 2016-05-29
Replace crypto/chacha/chacha_test.cc with Rust equivalent


briansmith:
We're trying to get rid of the requirement for a C++ compiler to build *ring*, and this is one of the blockers to that.

The license of crypto/chacha/chacha_test.cc is the same ISC-like license *ring*'s Rust code uses, so feel free to reuse (i.e. transliterate it to Rust) as much of it as is useful to reuse. If any parts are transliterated then make sure the Google copyright/license block is copied exactly.

The test vectors should be put into a chacha_tests.txt file and read using the ring::file_test framework.

NilSet:
This should be closed as of f5be9c34b10cd0d192a5eda394dd655a2db707c4

briansmith:
> This should be closed as of f5be9c3

Thanks!

PR #259, opened 2016-07-25
Replace chacha tests with Rust code.

samscott89:
This addresses #202.


samscott89:
I'm not really sure about how the copyright license part works. I didn't copy any code, but clearly copied the intentions of the original google code. Does that mean we still have to include the copyright notice somewhere?

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,67 @@ mod tests {
...
     }
 
+    #[test]
+    pub fn chacha20_tests(){
+        test::from_file("src/chacha_tests.txt", |section, test_case| {
+            assert_eq!(section, "");
+
+            let key_bytes = test_case.consume_bytes("Key");
+            let mut key = [0u32; super::CHACHA20_KEY_LEN/4];
+            for ki in 0..super::CHACHA20_KEY_LEN/4 {
```
briansmith: Use `use super::*` or whatever above, and don't qualify this with `super::`.

Use spaces around the `/` operator and parentheses around the expression: `(CHACHA20_KEY_LEN / 4)`.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,67 @@ mod tests {
...
+            let output = test_case.consume_bytes("Output");
+
+            for len in 0..input.len()+1 {
+                chacha20_test_case_inner(&key, &ctr, &input[..len], &output[..len], len);
+            }
+            Ok(())
+        });
+    }
+
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64",
```
briansmith: It's important to document why we're doing this: We know the overlapping in/out buffers don't work on x86 and ARM, and we work around them in the code. Include some kind of grepable cross-reference so that it is easy to find the fix-up code above.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,67 @@ mod tests {
                     (CRYPTO_POLY1305_STATE_LEN + 255) / 256);
     }
 
+    #[test]
+    pub fn chacha20_tests(){
```
briansmith: Make sure that the spacing matches the existing conventions throughout your patch, e.g. `s/(){/() {/`.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,67 @@ mod tests {
...
+
+    #[cfg(not(any(target_arch = "x86", target_arch = "x86_64",
+                  target_arch = "arm", target_arch="aarch64")))]
+    const MAX_OFFSET: usize = 259;
+
+    fn chacha20_test_case_inner(key: &[u32; super::CHACHA20_KEY_LEN / 4],
+                                ctr: &[u32; 4], input: &[u8], output: &[u8], len: usize){
+        let mut outbuf = vec![0u8; len];
+
+        // Check regular operation is successful
```
briansmith: Please make comments like this a full sentence and end them with a period. (If you're not a native English speaker, let me know and I'll fix any grammar/spelling mistakes I notice for you.)

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,67 @@ mod tests {
...
+        unsafe {
+          ChaCha20_ctr32(outbuf.as_mut_ptr(), input[..len].as_ptr(), len, key, &ctr);
+        }
+        assert_eq!(outbuf, &output[..len]);
+
+        // Check operation works in place for various alignments/offsets of slices
```
briansmith: Ditto

src/aead/chacha20_poly1305.rs
briansmith: Also, make sure the lines wrap around column 80, or 81 if the last character is punctuation.

briansmith: I feel like this test isn't documented clearly enough. If I didn't already know what it was supposed to do, I would have trouble figuring it out, I think.

The primary point of this test, especially the end part, is to test that the `ChaCha20_ctr32` function properly handles input and output buffers that overlap to various degrees, for various alignment factors. This is an important test because on some platforms the `ChaCha20_ctr32` code doesn't work for partially-overlapping buffers and we have to use a `memmove` to move the input to the output buffer so that they overlap exactly. OpenSSL and BoringSSL don't even try to use `ChaCha20_ctr32` this way so we're on our own when it comes to testing this.

Let's find a way to capture all of the above in a comment.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,67 @@ mod tests {
...
+            let mut ctr = [0u32; 4];
+            for ci in 0..4 {
+                let cb = slice_as_array_ref!(&ctr_bytes[ci * 4..][..4], 4).unwrap();
+                ctr[ci] = polyfill::slice::u32_from_be_u8(cb);
+            }
+
+            let input = test_case.consume_bytes("Input");
+            let output = test_case.consume_bytes("Output");
+
+            for len in 0..input.len()+1 {
```
briansmith: So, I don't know what to do style-wise. It is strange that we don't put spaces around `..`` but we do put spaces around `+`, yet `..` binds less tightly than everything. The way I've always resolved this in the code is to use extra parentheses around either end of the range that has a binary operator in it, and then use spaces around binary operators like normal, e.g. `0..(input.len() + 1)`.  I don't know how others feel about this but this seems like the best way to me. I'm open to other ideas, but I'd like to agree on something so that we're using the same style.

djc: FWIW, that style seems sane to me (and the space-less `/` before also stuck out to me when I looked at this earlier).

briansmith:
> I'm not really sure about how the copyright license part works. I didn't copy any code, but clearly copied the intentions of the original google code. Does that mean we still have to include the copyright notice somewhere?

IDK. I think it is probably OK the way it currently is, unless you don't agree.

samscott89:
Thanks for the review. I've addressed the comments and pushed a new version. I'll try to be more consistent with the coding style in future commits.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,89 @@ mod tests {
...
+    //
+    // On certain architectures (x86 and some branches of ARM), the in-place
+    // operation only succeeds when the input/output buffers are exactly
+    // overlapping due to the assembly code. Hence we do not test offsets for 
+    // x86 and ARM architectures.
+    //
+    // See the body of |open| for more information and the workaround.
+    #[test]
+    pub fn chacha20_tests() {
+        test::from_file("src/chacha_tests.txt", |section, test_case| {
```
briansmith: Please move the test vector file to src/aead/chacha_tests.txt.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,89 @@ mod tests {
...
+            assert_eq!(section, "");
+
+            let key_bytes = test_case.consume_bytes("Key");
+            let mut key = [0u32; CHACHA20_KEY_LEN / 4];
+            for ki in 0..(CHACHA20_KEY_LEN / 4) {
+                let kb =
+                    slice_as_array_ref!(&key_bytes[ki * 4..][..4], 4).unwrap();
+
+                key[ki] = polyfill::slice::u32_from_be_u8(kb);
```
briansmith: ChaCha20 uses little-endian encoding, not big-endian encoding, right?

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,89 @@ mod tests {
...
+    pub fn chacha20_tests() {
+        test::from_file("src/chacha_tests.txt", |section, test_case| {
+            assert_eq!(section, "");
+
+            let key_bytes = test_case.consume_bytes("Key");
+            let mut key = [0u32; CHACHA20_KEY_LEN / 4];
+            for ki in 0..(CHACHA20_KEY_LEN / 4) {
+                let kb =
+                    slice_as_array_ref!(&key_bytes[ki * 4..][..4], 4).unwrap();
+
```
briansmith: Let's get rid of this blank line.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,89 @@ mod tests {
...
+                key[ki] = polyfill::slice::u32_from_be_u8(kb);
+            }
+
+            let ctr_bytes = test_case.consume_bytes("Ctr");
+            let mut ctr = [0u32; 4];
+            for ci in 0..4 {
+                let cb =
+                    slice_as_array_ref!(&ctr_bytes[ci * 4..][..4], 4).unwrap();
+
+                ctr[ci] = polyfill::slice::u32_from_be_u8(cb);
```
briansmith: Ditto blank line and little-endian instead of big-endian.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,89 @@ mod tests {
...
+            }
+
+            let input = test_case.consume_bytes("Input");
+            let output = test_case.consume_bytes("Output");
+
+            // Run the test case over all prefixes of the input
```
briansmith: s/input/input because the behavior of ChaCha20 implementation changes dependent on the length of the input./

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,89 @@ mod tests {
...
+                                         &output[..len], len);
+            }
+            Ok(())
+        });
+    }
+
+    // Do not test offset buffers for x86 and ARM architectures (see above for
+    // rationale).
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64",
+              target_arch = "arm", target_arch = "aarch64"))]
```
briansmith: This condition is wrong. Only the 32-bit x86 code and the 32-bit ARM code needs the workaround. The 64-bit code works fine.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,89 @@ mod tests {
...
+    #[cfg(not(any(target_arch = "x86", target_arch = "x86_64",
+                  target_arch = "arm", target_arch = "aarch64")))]
+    const MAX_OFFSET: usize = 259;
+
+    fn chacha20_test_case_inner(key: &[u32; CHACHA20_KEY_LEN / 4],
+                                ctr: &[u32; 4], input: &[u8], output: &[u8],
+                                len: usize) {
+        let mut outbuf = vec![0u8; len];
+
+        // Straightforward encryption is computed correctly.
```
briansmith: s/encryption/encryption into disjoint buffers/.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,89 @@ mod tests {
...
+        unsafe {
+          ChaCha20_ctr32(outbuf.as_mut_ptr(), input[..len].as_ptr(),
+                         len, key, &ctr);
+        }
+        assert_eq!(outbuf, &output[..len]);
+
+        // Check that in-place encryption works successfully when the pointers
```
briansmith: Define `MAX_OFFSET` locally in this function around here.

src/aead/chacha20_poly1305.rs
```diff
@@ -255,9 +255,10 @@ extern {
 
 #[cfg(test)]
 mod tests {
-    use {aead, c};
+    use {aead, c, test};
+    use super::{ChaCha20_ctr32, CHACHA20_KEY_LEN};
+    use polyfill;
```
briansmith: Put `polyfill` in the `use {aead....}`.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,89 @@ mod tests {
                     (CRYPTO_POLY1305_STATE_LEN + 255) / 256);
     }
 
+    // This verifies the encryption functionality provided by ChaCha20_ctr32
+    // is successful when either computed on distinct input/output buffers,
```
briansmith: s/distinct/disjoint.

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,89 @@ mod tests {
...
+    // or in-place.
```
briansmith: s/./ on fully- or partially- overlapping input/output buffers.

briansmith: I suggest this tweaking:

```
    // This verifies the encryption functionality provided by ChaCha20_ctr32
    // is successful when either computed on disjoint input/output buffers,
    // or on overlapping input/output buffers. On some branches of the 32-bit
    // x86 and ARM code the in-place operation fails in some situations where
    // the input/output buffers are not exactly overlapping. Such failures are
    // dependent not only on the degree of overlapping but also the length of
    // the data. `open()` works around that by moving the input data to the
    // output location so that the buffers exactly overlap, for those targets.
    // This test exists largely as a canary for detecting if/when that type of
    // problem spreads to other platforms.
```

src/aead/chacha20_poly1305.rs
```diff
@@ -278,6 +279,89 @@ mod tests {
...
+            }
+
+            let input = test_case.consume_bytes("Input");
+            let output = test_case.consume_bytes("Output");
+
+            // Run the test case over all prefixes of the input
+            for len in 0..(input.len() + 1) {
+                chacha20_test_case_inner(&key, &ctr, &input[..len],
+                                         &output[..len], len);
+            }
```
briansmith: Blank line here.

src/aead/chacha20_poly1305.rs
samscott89: Sure. Should really move all src/*_tests.txt where corresponding module folders exist.

samscott89:
Made the suggested fixes and rebased on top of master (had some merge conflicts due to adding the u32_from_le_u8 function into polyfill).

One annoyance is how long the test takes due to all the iterations (1024 different input lengths, 16 alignments and 260 offsets). Takes 7s on my system, which probably means even longer on Travis, and I'm aware there are already certain configurations which time out.

briansmith:
> One annoyance is how long the test takes due to all the iterations (1024 different input lengths, 16 alignments and 260 offsets). Takes 7s on my system, which probably means even longer on Travis, and I'm aware there are already certain configurations which time out.

Is it different in release vs. debug mode? We might try to reduce it some way in debug mode if so. I agree 7s is too long.

samscott89:
A bit faster in release mode. Got it down to 4.5s/5.2s in release/debug mode by only allocating a single buffer and re-using it.

samscott89:
That's weird. It's faster when running with other tests:
```
$ time cargo test chacha20_tests
    Finished debug [unoptimized + debuginfo] target(s) in 0.0 secs
     Running target/debug/deps/ring-e906113d8cb3aa6f

running 1 test
test aead::chacha20_poly1305::tests::chacha20_tests ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured

   Doc-tests ring

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured


real	0m5.864s
user	0m5.822s
sys	0m0.034s

```

vs 

```
$ time cargo test chacha
    Finished debug [unoptimized + debuginfo] target(s) in 0.0 secs
     Running target/debug/deps/ring-e906113d8cb3aa6f

running 5 tests
test aead::chacha20_poly1305::tests::test_poly1305_state_len ... ok
test aead::chacha20_poly1305::tests::test_poly1305 ... ok
test aead::chacha20_poly1305::tests::test_chacha20_poly1305_old ... ok
test aead::chacha20_poly1305::tests::test_chacha20_poly1305 ... ok
test aead::chacha20_poly1305::tests::chacha20_tests ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured

   Doc-tests ring

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured


real	0m2.956s
user	0m3.156s
sys	0m0.033s

```

briansmith:
Committed as f5be9c34b10cd0d192a5eda394dd655a2db707c4. I made some tweaks. In particular, it seems that previously we could write `if cfg!(....)` as just `cfg!(....)`, but that doesn't work any more. I probably tweaked some other things too. I also verified that the tests does catch the weirdness of the 32-bit implementations if we don't special-case them.

Thanks a ton for doing this!

PR #260, opened 2016-07-28
Signing algorithm parameters trait

djc:
This is an alternative to #255 using traits instead of an enum, to prevent linking in unused signing/verification algorithms. Note that in my testing (from a different crate), this does not require explicitly importing the specific algorithms. This compiles:

```rust
extern crate ring;
extern crate untrusted;

fn verify() {
    ring::signature::verify(&ring::signature::ED25519,
                            untrusted::Input::from("foo".as_bytes()),
                            untrusted::Input::from("bar".as_bytes()),
                            untrusted::Input::from("crap".as_bytes()));
}
```


briansmith:
Hey, this seems great. Unfrotunately, it broke the `webpki` build, which reminded me about why I didn't do it this way before. But, then, I learned how to fix it, I think! I had to change your trait to be:

```
pub trait VerificationAlgorithm : Sync { ... }
```

Otherwise (webpki)(https://github.com/briansmith/webpki) doesn't build.

WDYT?

src/ec/suite_b/ecdsa.rs
```diff
@@ -216,13 +217,10 @@ macro_rules! ecdsa {
...
         /// 2.2.3](https://tools.ietf.org/html/rfc3279#section-2.2.3).
-        pub static $VERIFY_ALGORITHM: signature::VerificationAlgorithm =
-                signature::VerificationAlgorithm {
-            implementation: &ECDSAVerification {
+        pub static $VERIFY_ALGORITHM: ECDSAVerification = ECDSAVerification {
                 ops: $ecdsa_verify_ops,
                 digest_alg: $digest_alg,
-            }
-        };
+            };
```
briansmith: The indention seems off here.

src/signature.rs
```diff
@@ -194,27 +198,11 @@ impl<'a> Signature {
 }
 
 /// A signature verification algorithm.
```
briansmith: I think we need to add a note that the trait isn't intended to be implemented by things outside the *ring* crate.

briansmith:
Here's a demonstration of the trouble and non-trouble with traits and `use` I had before:

```
mod foo {
   pub trait Trait {
     fn foo(&self);
   }
   
   pub struct Struct { }
   
   impl Trait for Struct {
     fn foo(&self) { println!("foo"); }
   }
   
   pub fn bar(t: &Trait) {
       t.foo();
   }
}

fn main() {
   let s = foo::Struct { };
   foo::bar(&s);
   //s.foo();
}
```

Basically, the compile will let you call `bar()` but it won't let you call `foo()` without a `use`.
'
This works out well for the design of *ring* because we use standalone functions like `verify` often. If we use this pattern in the future then we may need to make more standalone functions that wrap trait methods in the same style as `foo::bar()` above and `ring::signature::verify`.

This means that the only thing we lose, AFAICT, is the simulation of "sealed traits." Like I mentioned in my review comment, we can just document that these traits aren't intended to be implemented outside of *ring* and that *ring* might make assumptions about the implementations assuming that only *ring*'s implementations are in use.

briansmith:
Another downside of this design is that references to `VerificationAlgorithm`s would be two words wide (e.g. 128 bits on 64-bit systems) instead of one word wide as before. Not great but not a significant downside either.

djc:
The `Sync` trait bound seems exactly right.

Addressed your comments and added license disclaimer to commit messages.

briansmith:
Landed as:
b89fed411c642b9c0d3c0b53a9e164bc62db97b2
b12b8a476339efe88c7f2810d7d2f5bb4b418697

Thanks for taking the time to find a good solution for these problems. It's much appreciated.

Updated documentation: https://briansmith.org/rustdoc/ring/signature/.

briansmith:
Note: I tweaked the "do not implement" comment a bit before landing.

Issue #213, opened 2016-06-09
Add RSA-PSS signing


DemiMarie:
If _ring_ is going to support RSA-PKCS#1.5 signing it should support RSA-PSS signing.

briansmith:
This should require just simple modifications to whatever is done in #208.

burdges:
It's maybe worth noting that RSA usage is declining in classical cryptographic applications and increasingly for situations needing homomorphic properties, like Paillier, voting, accumulators, etc., which means lower level access.  [Yey foot guns!](https://twitter.com/isotrumpp/status/707047647568236544)

In [Taler](http://taler.net) (not in Rust) for example, we use RSA blind signatures because Schnorr blind signatures would require an extra round trip and BLS blind signatures employ a Weil pairing.  PSS is useless in that setting.  Instead, one needs a full domain hash and an entirely different proof of security from different underlying assumption.

briansmith:
See https://github.com/briansmith/ring/issues/115#issuecomment-227272106 for the details of how to support the various PSS parameters.

It's probably a good idea to implement PSS verification first.

briansmith:
Regarding how the salts work, the main question is whether we can safely implement a deterministic RSA-PSS scheme that interops with other implementations, particularly implementations that are designed according to TLS 1.3's requirements. For now, I suggest that we always generate a fixed salt that has the same length as the digest's output length, e.g. 32 bytes of zeros for SHA-256 and 48 bytes of zeros for SHA-384. This way, our signatures will be completely deterministic.

Internally, in order to pass the NIST test suite, we'll have to support salts with other values.

References in support of doing this:

* [Optimal Security Proofs for Full Domain Hash, Revisited](https://www.iacr.org/archive/eurocrypt2012/72370533/72370533.pdf): "However, since PSS (with random salt of arbitrary length) is at least as secure as FDH, we obtain as a corollary from Section 3 a tight security proof from lossiness to the security of PSS, with random salt of arbitrary (possibly zero) length."
* [Efficiency Improvements for Signature Schemes
with Tight Security Reductions](https://www.cs.umd.edu/~jkatz/papers/CCCS03_sigs.pdf): "We show a variant of PSS which avoids the random salt altogether yet has an equally-tight security reduction."
* [Another Look at “Provable Security”](http://www.cs.tau.ac.il/~bchor/crypto07/AnotherLook.pdf), section "A Variant of PSS".
[...]

Issue #115, opened 2016-02-20
Implement RSASSA-PSS signature verification


briansmith:
BoringSSL already implemented PSS, but it did so using the C digest API that we removed, so the PSS code was removed. We need to implement the PSS digest calculations in Rust using the _ring_ digest API.

The IETF spec for PSS is at https://tools.ietf.org/html/rfc3447#section-8.

The commit that removed the BoringSSL PSS code is https://github.com/briansmith/ring/commit/4f3535ed43cd577fed1809c3ce32177526bb57e5#diff-03f44f4ac02526e7db1f33340b4642e3L519. Don't transliterate the BoringSSL code, as it is under the OpenSSL license and new Rust code is only be accepted under the ISC license.

It would probably be a good idea to refactor the existing RSA PKCS#1 code in _ring_ (the Rust code and the C code) in order to make this work better.

The [webpki](https://github.com/briansmith/webpki) project contains some test vectors for PSS that should be enabled once this is working. (Though, that requires modifying webpki to parse PSS keys.)

frewsxcv:
I'm currently looking into this.

briansmith:
The BoringSSL code has this:
```
/* Negative sLen has special meanings:
   *	-1	sLen == hLen
   *	-2	salt length is autorecovered from signature
   *	-N	reserved */
[...]
```

Only the `sLen == hLen` case should be supported. In addition, the same digest algorithm that was used to digest the data should be used for the MGF1 function; that is, only the `mgf1Hash == NULL`/``mgf1Hash == Hash` case should be supported.

briansmith:
From the TLS 1.3 draft spec:
```
RSASSA-PSS algorithms
      Indicates a signature algorithm using RSASSA-PSS [RFC3447] with
      MGF1.  The digest used in the mask generation function and the
      digest being signed are both the corresponding hash algorithm as
      defined in [SHS].  When used in signed TLS handshake messages (see
      Section 4.8.1), the length of the salt MUST be equal to the length
      of the digest output.
```

briansmith:
An update on the salt recommendations from above:

[...]

PR #262, opened 2016-08-02
Implement RSA PSS verification and signing.

samscott89:
Steps taken so far:
- Create separate module for padding.
- Refactor existing PKCS padding to use this system.
- Implement PSS padding (encoding and verification).
- Add test cases for PSS signatures.

This is a bit rough at the moment. There are some fairly large changes made by this so I thought it would be best to put it out there for feedback before proceeding.

Relevant issues: #115 and #213


briansmith:
@djc, any chance you could take a look at this? I think you probably have some ideas already.

src/rsa.rs
```diff
@@ -151,7 +98,43 @@ rsa_pkcs1_padding!(RSA_PKCS1_SHA512, &digest::SHA512,
...
+rsa_pss_padding!(RSA_PSS_SHA1, &digest::SHA1,
+                   "Signing using RSA with PSS padding and SHA-1.", 0);
+rsa_pss_padding!(RSA_PSS_SHA256, &digest::SHA256,
+                   "Signing using RSA with PSS padding and SHA-256.", 0);
+rsa_pss_padding!(RSA_PSS_SHA384, &digest::SHA384,
+                   "Signing using RSA with PSS padding and SHA-384.", 0);
+rsa_pss_padding!(RSA_PSS_SHA512, &digest::SHA512,
+                   "Signing using RSA with PSS padding and SHA-512.", 0);
+
+rsa_pss_padding!(RSA_PSS_SHA1_20, &digest::SHA1,
```
samscott89: This is particularly tedious and not very flexible. Could make salt length an additional input to the encode/verify algorithms, but this is not something like to be shared by other padding schemes.

Alternatively, we enforce only a small number of options, e.g., salt length = digest length.

This current list is to cover all of the NIST tests in rsa_pss_{sign,verify}_tests.txt

src/rsa.rs
```diff
@@ -333,14 +360,21 @@ impl RSAKeyPair {
...
                 rng: &rand::SecureRandom, msg: &[u8], signature: &mut [u8])
                 -> Result<(), ()> {
         if signature.len() != self.public_modulus_len() {
             return Err(());
         }
 
-        try!(padding_alg.pad(msg, signature));
+        try!(padding_alg.encode(msg, signature, rng));
+
+        let mod_bits = self.public_modulus_len_bits() - 1;
```
samscott89: A workaround to handle PSS padding which is aligned to the number of bits in the modulus.

An alternative would be to add an input variable to encode which states the length in bits the output should be.

src/pad.rs
```diff
@@ -0,0 +1,363 @@
...
+            //    7.  Let dbMask = MGF(H, emLen - hLen - 1).
+            let mut db = [0u8; MAX_MLEN];
+
+            try!((self.mgf)(h_hash.as_slice_less_safe(), db_len, &mut db, self.digest_alg));
+
+            //    8.  Let DB = maskedDB \xor dbMask.
+
+            try!(masked_db.read_all((), |masked_bytes|{
+                // XXX: First byte may be incorrect due to misaligned number of
+                // bits. This is currently not handled properly
```
samscott89: Related to comment below in the `sign` function. PSS padding is dependent on the number of bits in the modulus.

We would need to add an input variable to this function to specify how long in bits the padding is.

djc:
As is this is pretty hard to review. @samscott89, do you think it's feasible to split this into at least 4 commits? I see the following:

# Move current padding-related functionality into new padding module (which I would call `padding` rather than `pad`, and should probably be RSA-scoped somehow -- see below).
# Implement PSS verification
# Implement PSS signing
# Add test cases

I think it's probably also time to put all RSA stuff in separate directory, so that's `rsa.rs` (which could maybe become `rsa/rsa_signature.rs` or something -- I'm sure @briansmith has a better idea for the name), the `rsa_*_tests.txt` files and the `signature_rsa_*` example keys. I'm also happy to do this in another PR if that makes more sense.

src/pad.rs
```diff
@@ -0,0 +1,363 @@
...
+use rand;
+
+pub trait PaddingAlgorithm {
+    fn encode(&self, msg: &[u8], out: &mut [u8], rng: &rand::SecureRandom) -> Result<(), ()>;
+
+    fn verify(&self, msg: untrusted::Input, encoded: untrusted::Input) 
+              -> Result<(), ()>;
+}
+
+pub struct PKCS1v15Padding {
```
briansmith: I would cut out the "v15".

src/pad.rs
```diff
@@ -0,0 +1,363 @@
...
+// SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+// OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+use untrusted;
+use digest;
+use rand;
+
+pub trait PaddingAlgorithm {
```
briansmith: Perhaps cut "Algorithm" out of the name, making it just "Padding". That would be consistent with PKCS1Padding and PSSPadding.

src/pad.rs
```diff
@@ -0,0 +1,363 @@
...
+
+impl PaddingAlgorithm for PKCS1v15Padding {
+    // Implement padding procedure per EMSA-PKCS1-v1_5,
+    // https://tools.ietf.org/html/rfc3447#section-9.2.
+    fn encode(&self, msg: &[u8], out: &mut [u8], _: &rand::SecureRandom) -> Result<(), ()> {
+        let digest_len =
+            self.digestinfo_prefix.len() + self.digest_alg.output_len;
+
+        // Require at least 8 bytes of padding. Since we disallow keys smaller
+        // than 2048 bits, this should never happen anyway.
```
briansmith: Make sure this comment in `RSA_check_key` gets updated to reference this:

```
  /* The public modulus must be large enough. |RSAPublicKey::sign| depends on
   * this; without it, |RSAPublicKey::sign| would generate padding that is
   * invalid (too few 0xFF bytes) for very small keys.
   *
```

src/pad.rs
```diff
@@ -0,0 +1,363 @@
...
+            }
+            Ok(())
+        })
+    }
+
+
+}
+
+// Maximum supported length of the salt;
+const MAX_SLEN: usize = 512 / 8;
```
briansmith: Should we spell this out as "MAX_SALT_LEN"? I realize "sLen" comes from the spec but I'm not sure it will be clear to readers. Same for `MAX_MLEN` below.

src/pad.rs
```diff
@@ -0,0 +1,363 @@
...
+
+// Maxmimum supported length for intermediate value used in PSS padding.
+const MAX_MLEN: usize = 512;
+
+pub struct PSSPadding {
+    pub digest_alg: &'static digest::Algorithm,
+    pub mgf: fn(seed: &[u8], mask_len: usize, mask: &mut [u8],
```
briansmith: We only need one MGF, right? If so, we don't need to call it through this function pointer.

src/pad.rs
```diff
@@ -0,0 +1,363 @@
...
+                digest_alg: &'static digest::Algorithm) -> Result<(), ()>,
+    pub s_len: usize
```
briansmith: The salt length will always be equal to the digest output length, right? If so, then we can document that and avoid this member.

src/pad.rs
```diff
@@ -0,0 +1,363 @@
...
+        if em_len < min_len {
+            return Err(());
+        }
+        let m_hash = digest::digest(self.digest_alg, msg);
+
+        //    4.  Generate a random octet string salt of length sLen; if sLen = 0,
+        //        then salt is the empty string.
+
+        let mut salt = [0u8; MAX_SLEN];
+        try!(rng.fill(&mut salt[..self.s_len]));
```
briansmith: I would like us to try to find a deterministic method for calculating the salt, similar to [RFC 6979](https://tools.ietf.org/html/rfc6979). See the PSS spec which says:

> RSASSA-PSS is different from other RSA-based signature schemes in
   that it is probabilistic rather than deterministic, incorporating a
   randomly generated salt value.  The salt value enhances the security
   of the scheme by affording a "tighter" security proof than
   deterministic alternatives such as Full Domain Hashing (FDH); see [4]
   for discussion.  However, the randomness is not critical to security.
   In situations where random generation is not possible, a fixed value
   or a sequence number could be employed instead, with the resulting
   provable security similar to that of FDH [12].

I believe for ECDSA we will also be implementing the deterministic scheme.

briansmith: Because this is perhaps a major point of contention, I'd like to see the patches split up so that *signing* using RSA PSS is the *last* patch in the series, if that is at all practical.

src/pad.rs
```diff
@@ -0,0 +1,363 @@
...
+
+        let h_hash = ctx.finish();
+
+        //    7.  Generate an octet string PS consisting of emLen - sLen - hLen - 2
+        //        zero octets.  The length of PS may be 0.
+
+        //    8.  Let DB = PS || 0x01 || salt; DB is an octet string of length
+        //        emLen - hLen - 1.
+
+        let mut db = [0u8; MAX_MLEN];
```
briansmith: Does db really need to be a separate array, or could it just be a slice of `out` in order to avoid the copies below?

src/rsa.rs
briansmith: Yes, the salt length should always be equal to the digest function output length, and the digest for the MGF should always be the same digest used for the hash. Basically, we are constraining ourselves to what TLS 1.3 and other recent IETF specs require: https://github.com/briansmith/ring/issues/115#issuecomment-227272106

briansmith: Simply omit the NIST test vectors that use a different salt length. Add a note to rsa_pss_{sign,verify}_tests.txt that they are omitted and why.

src/rsa.rs
```diff
@@ -151,7 +98,43 @@ rsa_pkcs1_padding!(RSA_PKCS1_SHA512, &digest::SHA512,
...
+                   "Signing using RSA with PSS padding and SHA-1.", 0);
+rsa_pss_padding!(RSA_PSS_SHA256, &digest::SHA256,
+                   "Signing using RSA with PSS padding and SHA-256.", 0);
+rsa_pss_padding!(RSA_PSS_SHA384, &digest::SHA384,
+                   "Signing using RSA with PSS padding and SHA-384.", 0);
+rsa_pss_padding!(RSA_PSS_SHA512, &digest::SHA512,
+                   "Signing using RSA with PSS padding and SHA-512.", 0);
+
+rsa_pss_padding!(RSA_PSS_SHA1_20, &digest::SHA1,
+                   "Signing using RSA with PSS padding, 20 byte salt, and SHA-1.", 20);
```
briansmith: We do not need to support PSS with SHA-1 at all.

src/rsa.rs
```diff
@@ -151,7 +98,43 @@ rsa_pkcs1_padding!(RSA_PKCS1_SHA512, &digest::SHA512,
...
+rsa_pss_padding!(RSA_PSS_SHA256_20, &digest::SHA256,
+                   "Signing using RSA with PSS padding, 20 byte salt, and SHA-256.", 20);
```
briansmith: You can drop the ", XXX byte salt, " part. Instead, in the boilerplate for PSS, we can state that the salt length is equal to the digest length and the MGF digest is the same as the digest function used to hash the message.

src/rsa.rs
briansmith: I am happy to restrict RSA signing to public moduli that are exact multiples of 8 bits.

Regardless of that, do whatever needs to be done to move this logic to `encode()`, whether that means adding a new parameter or whatever.

src/rsa.rs
```diff
@@ -504,6 +539,123 @@ mod tests {
     #[cfg(feature = "rsa_signing")]
     extern { static GFp_BN_BLINDING_COUNTER: u32; }
 
+    #[cfg(feature = "rsa_signing")]
+    #[test]
+    fn test_signature_rsa_pss_sign() {
+        struct DeterministicSalt<'a> {
```
briansmith: Can you just use `ring::rand::test_util::FixedSliceRandom` here? If not, please explain why in a comment.

briansmith:
> As is this is pretty hard to review. @samscott89, do you think it's feasible to split this into at least 4 commits? I see the following:
> 
> Move current padding-related functionality into new padding module (which I would call padding rather than pad, and should probably be RSA-scoped somehow -- see below).

I agree `padding` is a better name and that if such a submodule is going to exist, it should be a submodule of `ring::rsa`. I would also not mind just having this all in `ring::rsa`.

> Implement PSS verification
> Implement PSS signing
> Add test cases

I also prefer this split, but I would prefer the tests for PSS verification to be included in the PSS verification commit, and for the tests for PSS signing to be in the PSS signing commit.

> I think it's probably also time to put all RSA stuff in separate directory, so that's rsa.rs (which could maybe become rsa/rsa_signature.rs or something -- I'm sure @briansmith has a better idea for the name), the rsa_*_tests.txt files and the signature_rsa_* example keys. I'm also happy to do this in another PR if that makes more sense.

If we're going to have `ring::rsa::padding` then I agree that the tests should be moved to `src/rsa/`. I would prefer to use the name `src/rsa/rsa.rs` at least until RSA encryption is added (which we may never do, but if we do then we will rename everything again).

briansmith:
Thanks @djc for providing the great feedback. Also, thanks @samscott89 for an awesome contribution here. I hope my feedback above is enough for you to make forward progress. If not, let me know.

src/pad.rs
```diff
@@ -0,0 +1,363 @@
...
+    // Implement padding procedure per EMSA-PSS,
+    // https://tools.ietf.org/html/rfc3447#section-9.1.
+    fn encode(&self, msg: &[u8], out: &mut [u8], rng: &rand::SecureRandom)
+              -> Result<(), ()> {
+
+        //    Steps:
+
+        //    1.  If the length of M is greater than the input limitation for the
+        //        hash function (2^61 - 1 octets for SHA-1), output "message too
+        //        long" and stop.
```
briansmith: Regarding these comments that quote the spec: At a minimum, let's get rid of the excess indention at the start and rewrap the text to 80 chars. However, I actually would prefer them to be abbreviated to just "Steps 1-3". The reader can read the actual steps in the spec. This way, we don't have to ensure that the text of the actual quotes are correct. See `ring::ec::suite_b::ecdh::ecdh` for an example of the style I mean.

src/pad.rs
briansmith: For reference:

* http://crypto.stackexchange.com/questions/1217/rsa-pss-salt-size.
* [Efficiency Improvements for Signature Schemes
with Tight Security Reductions](https://www.cs.umd.edu/~jkatz/papers/CCCS03_sigs.pdf): "We show a variant of PSS which avoids the random salt altogether yet has an equally-tight security reduction."
* [Another Look at “Provable Security”](http://www.cs.tau.ac.il/~bchor/crypto07/AnotherLook.pdf), section "A Variant of PSS".
* [Another Look at “Provable Security” II](http://cr.yp.to/bib/2006/koblitz-provable.pdf): Section "The implausible magic of one bit"
* http://www5.rz.rub.de:8032/mam/foc/content/publ/rsa-fdh_fullversion.pdf
* https://eprint.iacr.org/2015/517.pdf - Section 4.2

samscott89:
Thank you both for the fantastic feedback, exactly what I was looking for. I'll have another go with the above in mind, and should have a somewhat nicer PR up shortly.

src/rsa.rs
samscott89: Potentially. I wasn't previously aware of `FixedSliceRandom`, which is nice. I wrote it in its current form so that it could be re-used in subsequent calls (I think it's used in RSA blinding). I'll see if I can just use `FixedSliceRandom` or `FixedSliceSequenceRandom`.

samscott89:
Ok, that should be a bit easier to digest/read.

I've assumed the length in bits of the RSA moduli is always a multiple of 8. Is there somewhere this can be enforced?

src/lib.rs
```diff
@@ -124,7 +124,7 @@ mod init;
 pub mod pbkdf2;
 
 pub mod rand;
-#[cfg(feature = "use_heap")] mod rsa;
+#[path = "rsa/rsa.rs"] #[cfg(feature = "use_heap")] mod rsa;
```
djc: This should probably go on multiple lines at this point.

src/rsa/padding.rs
```diff
@@ -0,0 +1,105 @@
...
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+use untrusted;
+use digest;
+
+#[cfg(feature = "rsa_signing")]
+use rand;
+
+#[cfg(feature = "rsa_signing")]
+pub trait Encoding {
```
djc: I'm not sure why you choose to rename `Padding` to `Encoding` here, but I think I don't like it that much. Similarly, I don't think `encode()` is the right name for the method. It seems to me that `Padding` is probably still the best name for the trait, and the method should either be `pad()` or maybe `wrap()` (the latter mostly because for the inverse, `unwrap()` maybe is nicer than `unpad()`).

src/rsa/padding.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn encode(&self, msg: &[u8], out: &mut [u8], rng: &rand::SecureRandom)
+              -> Result<(), ()>;
+}
+
+pub trait Verification {
+    fn verify(&self, msg: untrusted::Input, encoded: untrusted::Input)
```
djc: The "encoded" parameter name does not make much sense to me in the context of a `Verification` trait. The concept that `Verification` is implemented for an `Encoding` is also not that intuitive. It seems better to me that the `Encoding`'s single responsibility is to wrap/unwrap, and that verification will be left to the implementer of the `signature::VerificationAlgorithm` trait.

djc:
As a general comment on 8f8075b, there's still a lot going on that would be easier to digest if split into multiple commits. I.e., I think I see the following related but separate changes:

1. Move `rsa` to `rsa::rsa`
2. Extract padding code into separate `rsa::padding` module
3. Introduction of the `Encoding` trait
4. Introduction of the `Verification` trait
5. Adding the RNG parameter to the `Encoding` trait's method

For 1, I'm not sure `rsa::rsa` is the best name. It seems more idiomatic to have an `rsa/mod.rs` that exposes everything RSA that is used elsewhere. We can either move `rsa.rs` to `rsa/mod.rs` and import stuff from `rsa/padding.rs`, or we keep `rsa/mod.rs` as an import-only module (similar to `lib.rs`) and move the implementation into e.g. `rsa/impl.rs`. In any case, the use of `[#path]` seems to suggest we're doing something non-idiomatic.

For 5, I think this should definitely not be part of any commit where it's not needed yet. It of course needs to be introduced before (or maybe as part of, though, as you noticed, I'm a fan of small commits) the commit that introduces PSS signing, but can be done after verification.

Finally, IMO commit messages in Git work a bit better if you keep the first line shorter. Also, splitting commits as suggested would neatly sidestep the "Also...", "Also..." that you currently have in your commit message.

src/rsa/padding.rs
```diff
@@ -103,3 +103,119 @@ impl Verification for PKCS1 {
...
+
+// Mask-generating function MGF1 as described in:
+// https://tools.ietf.org/html/rfc3447#appendix-B.2.1
+fn mgf1(seed: &[u8], mask_len: usize, mask: &mut [u8],
+        digest_alg: &'static digest::Algorithm) -> Result<(), ()> {
+
+    let digest_len = digest_alg.output_len;
+
+    let ctr_max = 1 + ((mask_len - 1) / digest_len);
+    for i in 0..ctr_max {
```
djc: Maybe consider just putting the expression in the for loop header, rather than using a temporary variable binding? Also, the value of `ctr_max` looks like it will use truncating division at the face of it, maybe add a comment explaining the preconditions for these numbers?

src/rsa/padding.rs
```diff
@@ -103,3 +103,119 @@ impl Verification for PKCS1 {
...
+        let mut ctx = digest::Context::new(digest_alg);
+        ctx.update(seed);
+        ctx.update(&polyfill::slice::be_u8_from_u32(i as u32));
+        let digest = ctx.finish();
+        mask[i*digest_len..][..digest_len].copy_from_slice(digest.as_ref());
```
djc: I think your `[i*digest_len..][..digest_len]` is meant to be equivalent to `[i*digest_len..(i + 1)*digest_len]`, right? I think the latter makes it more obvious what's going on. Also, I think we decided that style would be that we have spaces between binary operator and operands, and then Brian wants parenthesis around the whole thing, I think.

src/rsa/rsa.rs
```diff
@@ -67,11 +67,23 @@ macro_rules! rsa_pkcs1_padding {
         pub static $PADDING_ALGORITHM: padding::PKCS1 =
             padding::PKCS1 {
                 digest_alg: $digest_alg,
-                digestinfo_prefix: $digestinfo_prefix,
+                digestinfo_prefix: $digestinfo_prefix
```
djc: The comma is there so you don't have to change the line if you add another field. Did you have a particular reason for dropping it? Personally I think we should keep it (but then I introduced it, I think).

src/rsa/rsa.rs
```diff
@@ -67,11 +67,23 @@ macro_rules! rsa_pkcs1_padding {
...
     }
 }
 
+macro_rules! rsa_pss_padding {
+    ( $PADDING_ALGORITHM:ident, $digest_alg:expr, $doc_str:expr ) => {
+        #[doc=$doc_str]
+        /// Feature: `rsa_signing`.
+        pub static $PADDING_ALGORITHM: padding::PSS =
+            padding::PSS {
+                digest_alg: $digest_alg
```
djc: Not sure what style we have for this, but in this case I would just put the whole declaration of `padding::PSS { }` on a single line.

src/rsa/rsa.rs
```diff
@@ -85,7 +97,14 @@ rsa_pkcs1_padding!(RSA_PKCS1_SHA512, &digest::SHA512,
...
 
-macro_rules! rsa_pkcs1 {
+rsa_pss_padding!(RSA_PSS_SHA256, &digest::SHA256,
+                   "Signing using RSA with PSS padding and SHA-256.");
+rsa_pss_padding!(RSA_PSS_SHA384, &digest::SHA384,
+                   "Signing using RSA with PSS padding and SHA-384.");
+rsa_pss_padding!(RSA_PSS_SHA512, &digest::SHA512,
+                   "Signing using RSA with PSS padding and SHA-512.");
+
+macro_rules! rsa_impl {
```
djc: More from the Small Commit Propaganda Committee: I would split the rename of the `rsa_pkcs1` macro out into a separate commit.

util/generate-rsa-signing-tests.py
```diff
@@ -26,10 +26,21 @@
 from cryptography.hazmat.primitives.asymmetric import rsa, padding
 import sys, copy
 
-def parse(fn):
+digest_output_lens = {
```
djc: I would rename this to `DIGEST_OUTPUT_LENGTHS`. The use of upper case is customary for constants in Python, and `lens` is hard to read as a multiple of `len`. Also, I would add the trailing comma for the last key-value pair.

util/generate-rsa-signing-tests.py
```diff
@@ -40,57 +51,161 @@ def parse(fn):
...
+    "# http://csrc.nist.gov/groups/STM/cavp/digital-signatures.html#test-vectors.\n" + \
+    "# Generated with the help of the util/generate-rsa-signing-tests.py script.\n" + \
+    "#\n" + \
+    "# Digest = SHAAlg.\n" + \
+    "# Key is a DER-formatted PKCS1-encoded RSA key.\n" + \
+    "# Sig = S.\n\n"
+
+    num_cases = 0
+
+    # Each test type has a different field as the last entry per case
```
djc: This if-block seems a bit iffy, since it uses if/elif/else for conditions that don't seem mutually exclusive. Maybe add a comment on each condition line to clarify? Also, I would add an empty line after the if-block, and initialize `num_cases` after it (right before the for-loop).

util/generate-rsa-signing-tests.py
```diff
@@ -40,57 +51,161 @@ def parse(fn):
...
+
+        if test_type == 'sign':
+            # Serialize the private key in DER format.
+            der = key.private_bytes(serialization.Encoding.DER,
+                                    serialization.PrivateFormat.TraditionalOpenSSL,
+                                    serialization.NoEncryption())
+        elif test_type == 'verify':
+            # Serialize the private key in DER format.
+            der = key.public_bytes(serialization.Encoding.DER,
+                                    serialization.PublicFormat.PKCS1)
```
djc: Nit: off-by-one alignment.

util/generate-rsa-signing-tests.py
```diff
@@ -40,57 +51,161 @@ def parse(fn):
...
+            dmq1 = rsa.rsa_crt_dmq1(d, q)
+            iqmp = rsa.rsa_crt_iqmp(p, q)
 
         # Create a private key instance.
         pub = rsa.RSAPublicNumbers(e, n)
-        priv = rsa.RSAPrivateNumbers(p, q, d, dmp1, dmq1, iqmp, pub)
-        key = priv.private_key(default_backend())
 
-        # Recalculate and compare the signature to validate our processing.
+        if test_type == 'sign':
```
djc: Since this is first of three subsequent `if test_type == 'sign'`s, I would suggest it's probably better to factor out the different types into different functions, e.g. have `print_sign_test()` and `print_verify_test()` that you can then dispatch to. You'll probably then want a helper function that has some of the `print` statements that are common to both.

util/generate-rsa-signing-tests.py
```diff
@@ -26,10 +26,21 @@
 from cryptography.hazmat.primitives.asymmetric import rsa, padding
```
djc: I think it would make sense to put the changes in the Python script in a separate commit. (Do you know that you can easily split a commit along file or even hunk lines with interactive rebase? Just start by `edit`ing the commit in question, then `git reset HEAD^`, then `git commit` per file or `git commit -p` until you've committed everything again, then `git rebase --continue`.)

src/rsa/padding.rs
samscott89: Encode/Verify is the terminology used in the specs: https://tools.ietf.org/html/rfc3447#section-9.1.1.

Pad is misleading because it implies that you are extending the message by adding some redundancy. Usually to fit a specified block length. This is not the case here. Messages are hashed down and then expanded to fill a single block.

Similarly, the inverse operation does not recover the message, but checks that the recovered block is a valid encoding.

samscott89:
> For 1, I'm not sure rsa::rsa is the best name. It seems more idiomatic to have an rsa/mod.rs that exposes everything RSA that is used elsewhere. We can either move rsa.rs to rsa/mod.rs and import stuff from rsa/padding.rs, or we keep rsa/mod.rs as an import-only module (similar to lib.rs) and move the implementation into e.g. rsa/impl.rs. In any case, the use of [#path] seems to suggest we're doing something non-idiomatic.

This is a project-wide decision (which I happen to agree with). See: aead.rs, digest.rs, ec.rs. 
https://github.com/briansmith/ring/blob/master/STYLE.md#submodules-and-file-naming

> For 5, I think this should definitely not be part of any commit where it's not needed yet. It of course needs to be introduced before (or maybe as part of, though, as you noticed, I'm a fan of small commits) the commit that introduces PSS signing, but can be done after verification.

Personally I would argue that by writing an encoding trait you should include the functionality you intend to support. Encoding is commonly a randomised algorithm and hence the api should reflect that. I don't feel strongly about this though and happy to change this.

> Finally, IMO commit messages in Git work a bit better if you keep the first line shorter. Also, splitting commits as suggested would neatly sidestep the "Also...", "Also..." that you currently have in your commit message.

Apologies, that was sloppy. That was even a self-inflicted line break in the title which I absent-mindedly added without realising :/

I will split it up into more commits.

djc:
> This is a project-wide decision (which I happen to agree with). See: aead.rs, digest.rs, ec.rs.
https://github.com/briansmith/ring/blob/master/STYLE.md#submodules-and-file-naming

Okay, I had forgotten about that part.

> Personally I would argue that by writing an encoding trait you should include the functionality you intend to support. Encoding is commonly a randomised algorithm and hence the api should reflect that. I don't feel strongly about this though and happy to change this.

Fair enough; I also don't feel that strongly about it.

src/rsa/padding.rs
djc: Okay, maybe that makes sense for PSS. Still, this doesn't seem as good a description of the PKCS1 padding operations, which also need to be represented, right?

At a high level, we're fitting a hash with variable size into a larger-sized block (by key size). To me, "encoding" invokes the idea of something that keeps the size the same, or has a linear relation between input and output size.

In the end, I guess this isn't public API anyway, so the high-level/end-user view is maybe less important than following the spec/using terminology crypto-engineers use.

samscott89:
I've found the presence of both `RSAParameters` and `RSAPadding` slightly confusing in the rsa code. Especially since the comment of the `RSAParameters` struct claims it is used for "Parameters for RSA signing and verification." when signing only uses `RSAPadding`.

Are there any objections to making the `RSAParameters` also the input to the `sign` function? This has the added benefit that the padding algorithms don't need to be exported in `signature.rs`?

The minor downside is that `sign` is effectively taking in an extra unnecessary parameter (the `min_bits` field), which is reflected in, for example, choosing between `RSA_PKCS1_2048_8192_SHA384` and `RSA_PKCS1_3072_8192_SHA384` for signing even though there would be (currently) no difference.

Though it might also be reasonable to enforce a certain modulus length when signing?

djc:
I would be a fan of that change; however, I seem to remember that it was impractical before. That might have changed, though.

If we do so, we should probably also enforce the modulus length when signing.

samscott89:
Hmm, actually it might not make too much sense since later there will be separate Encoding/Verification traits anyway. At which point the syntax is a bit clearer anyway: `RSAParameters` includes a `padding::Verification` algorithm and `sign` takes as input a `padding::Encoding` algorithm.

src/rsa/padding.rs
```diff
@@ -150,3 +150,138 @@ impl Verification for PKCS1 {
...
+
+            Ok(())
+        })
+    }
+}
+
+// Mask-generating function MGF1 as described in:
+// https://tools.ietf.org/html/rfc3447#appendix-B.2.1
+fn mgf1(seed: &[u8], mask_len: usize, mask: &mut [u8],
+        digest_alg: &'static digest::Algorithm) -> Result<(), ()> {
```
briansmith: The `digest_alg` function should come first. Instead of having the caller pass `mask_len`, use `mask.len()` and have the caller provide a slice of the appropriate length.

src/rsa/padding.rs
```diff
@@ -150,3 +150,138 @@ impl Verification for PKCS1 {
...
+
+    let digest_len = digest_alg.output_len;
+
+    // Maximum counter value is the value of (mask_len / digest_len) rounded up.
+    let ctr_max = 1 + ((mask_len - 1) / digest_len);
```
briansmith: `assert!(ctr_max <= std::u32.min_value());` here.

src/rsa/padding.rs
```diff
@@ -150,3 +150,138 @@ impl Verification for PKCS1 {
...
+/// PSS Padding as described in https://tools.ietf.org/html/rfc3447#section-9.1.
+/// Salt length is fixed to be the length of the output of the digest algorithm.
+/// The only mask generating function supported is MGF1.
+/// The size (in bits) of the RSA moduli is assumed to be a multiple of 8.
+pub struct PSS {
+    pub digest_alg: &'static digest::Algorithm,
+}
+
+// Maxmimum supported output length for PSS padding.
+const MAX_OUTPUT_LEN: usize = 4096;
```
briansmith: Where does this number come from? The max modulus size we support is 8192 bits (currently), so I would expect this to be `8192 / 8`. Please add a comment about what this is for.

Maybe we need to move `const MAX_BITS: usize = 8192` out of `RSAParameters::verify` so everything can use it?

src/rsa/padding.rs
```diff
@@ -150,3 +150,138 @@ impl Verification for PKCS1 {
...
+            try!(mgf1(h_hash.as_slice_less_safe(), db_len, &mut db,
+                      self.digest_alg));
+
+            try!(masked_db.read_all((), |masked_bytes| {
+                // Step 6. Moduli lengths are always a multiple of 8 so the top
+                // bit must be 0.
+                let b = try!(masked_bytes.read_byte());
+                if b & 0x80 != 0 {
+                    return Err(());
+                } else {
```
briansmith: No `else` after `return`.

src/rsa/padding.rs
```diff
@@ -150,3 +150,138 @@ impl Verification for PKCS1 {
...
+            //     masked_db || h_hash || 0xbc
+            let db_len = em_len - digest_len - 1;
+            let masked_db = try!(em.skip_and_get_input(db_len));
+            let h_hash = try!(em.skip_and_get_input(digest_len));
+            if try!(em.read_byte()) != 0xbc {
+                return Err(());
+            }
+
+            // Step 7.
+            let mut db = [0u8; MAX_OUTPUT_LEN];
```
briansmith: `let db = db[..db_len];`. Then you can use `db.len()` everywhere instead of `db_len`.

src/rsa/padding.rs
```diff
@@ -150,3 +150,138 @@ impl Verification for PKCS1 {
...
+                return Err(());
+            }
+
+            // Step 11.
+            let salt = &db[db_len - digest_len..][..digest_len];
+
+            // Step 12 and 13: compute hash value of:
+            //     (0x)00 00 00 00 00 00 00 00 || m_hash || salt
+            let mut ctx = digest::Context::new(self.digest_alg);
+            let prefix = [0u8; 8];
```
briansmith: this should be `const PSS_PREFIX_ZEROS: [u8; 8] = [0u8; 8]` and should be shared between the encoding and decoding functions. And/or, maybe this whole block from `let mut ctx...ctx.finish()` should be factored out into a shared function?

src/rsa/padding.rs
briansmith: Please add a `// TODO: Can we avoid requiring this `MAX_OUTPUT_LEN` buffer?` comment. I think shrinking `MAX_OUTPUT_LEN` from 4096 bytes to 1024 (8192/8) bytes is good enough for now. In the future we may want to reduce stack usage further.

src/rsa/padding.rs
```diff
@@ -150,3 +150,138 @@ impl Verification for PKCS1 {
         })
     }
 }
+
+/// PSS Padding as described in https://tools.ietf.org/html/rfc3447#section-9.1.
+/// Salt length is fixed to be the length of the output of the digest algorithm.
+/// The only mask generating function supported is MGF1.
+/// The size (in bits) of the RSA moduli is assumed to be a multiple of 8.
```
briansmith: Let's rewrap all of this into one paragraph at 80 chars.

src/rsa/padding.rs
```diff
@@ -150,3 +150,138 @@ impl Verification for PKCS1 {
...
+            }
+
+            // Step 7.
+            let mut db = [0u8; MAX_OUTPUT_LEN];
+            try!(mgf1(h_hash.as_slice_less_safe(), db_len, &mut db,
+                      self.digest_alg));
+
+            try!(masked_db.read_all((), |masked_bytes| {
+                // Step 6. Moduli lengths are always a multiple of 8 so the top
+                // bit must be 0.
```
briansmith: Sorry, I misled you here. It is OK for *signing* to require that the modulus be a multiple of 8 bits (or even 64 bits or whatever we need for an efficient implementation of all the primitives), but we should still accept all the oddball moduli for *verification*.

It is definitely OK to add the modulus length in bits, or perhaps the number of leading zero bits in the top byte of the modulus, or perhaps the index of the highest set bit in the top byte of the modulus, or whatever is convenient, as an argument to the padding verification trait function.

src/rsa/padding.rs
```diff
@@ -150,3 +150,138 @@ impl Verification for PKCS1 {
...
+                if b & 0x80 != 0 {
+                    return Err(());
+                } else {
+                    db[0] ^= b;
+                }
+
+                // Step 8.
+                for i in 1..db_len {
+                    let b = try!(masked_bytes.read_byte());
+                    db[i] ^= b;
```
briansmith: You can combine these like so: `db[i] ^= try!(...);`.

src/rsa/padding.rs
briansmith: Another option is to have `seed: &digest::Digest`. Then you wouldn't need the `digest_alg` parameter at all because you could get it as `seed.algorithm()`.

src/rsa/padding.rs
```diff
@@ -164,9 +164,79 @@ pub struct PSS {
     pub digest_alg: &'static digest::Algorithm,
 }
 
-// Maxmimum supported output length for PSS padding.
+#[cfg(feature = "rsa_signing")]
+// Maximum supported length of the salt in bytes.
+// In practice, this is constrained by the maximum digest length.
+const MAX_SALT_LEN: usize = 512 / 8;
```
briansmith: I don't think we need this. We can just use `digest::MAX_OUTPUT_LEN` where necessary.

src/rsa/padding.rs
```diff
@@ -164,9 +164,79 @@ pub struct PSS {
...
+// Maxmimum supported output length for PSS padding in bytes.
 const MAX_OUTPUT_LEN: usize = 4096;
 
+#[cfg(feature = "rsa_signing")]
+impl Encoding for PSS {
+    // Implement padding procedure per EMSA-PSS,
+    // https://tools.ietf.org/html/rfc3447#section-9.1.
+    fn encode(&self, msg: &[u8], out: &mut [u8], rng: &rand::SecureRandom)
+              -> Result<(), ()> {
+
```
briansmith: In all your commits, please remove this blank line at the top of each function where one exists.

src/rsa/padding.rs
```diff
@@ -164,9 +164,79 @@ pub struct PSS {
...
+        let digest_len = self.digest_alg.output_len;
+        let em_len = out.len();
+
+        debug_assert!(out.len() <= MAX_OUTPUT_LEN);
```
briansmith: This type of assertion (precondition) should appear first.

src/rsa/padding.rs
```diff
@@ -164,9 +164,79 @@ pub struct PSS {
...
+        // Step 2.
+        let m_hash = digest::digest(self.digest_alg, msg);
+
+        // Step 3: where we assume the digest and salt are of equal length.
+        if em_len < 2 + (2 * digest_len) {
+            return Err(());
+        }
+
+        // Step 4.
+        let mut salt = [0u8; MAX_SALT_LEN];
```
briansmith:
```
let salt = &mut salt[..digest_len];
try!(rng.fill(&mut salt));
```

src/rsa/padding.rs
```diff
@@ -164,9 +164,79 @@ pub struct PSS {
...
+        try!(rng.fill(&mut salt[..digest_len]));
+
+        // Step 5 and 6: compute hash value of:
+        //     (0x)00 00 00 00 00 00 00 00 || m_hash || salt
+        let mut ctx = digest::Context::new(self.digest_alg);
+        let prefix = [0u8; 8];
+        ctx.update(&prefix);
+        ctx.update(m_hash.as_ref());
+        ctx.update(&salt[..digest_len]);
```
briansmith: s/`[..digest_len]`//

src/rsa/padding.rs
```diff
@@ -164,9 +164,79 @@ pub struct PSS {
...
+        let pad_len = db_len - digest_len - 1;
+        for i in 0..pad_len {
+            out[i] = 0;
+        }
+        out[pad_len] = 0x01;
+        out[pad_len + 1..][..digest_len].copy_from_slice(&salt[..digest_len]);
+
+        // Step 9.
+        let mut db_mask = [0u8; MAX_OUTPUT_LEN];
+        try!(mgf1(h_hash.as_ref(), db_len, &mut db_mask, self.digest_alg));
```
briansmith: Instead of setting `out` to "PS || 0x01 || salt" and then masking it with `db_mask`, why not fill `out` with the MGF1 output and then mask "PS || 0x01 || salt" onto it? Wouldn't this avoid the need for the extra buffer?

src/rsa/padding.rs
```diff
@@ -164,9 +164,79 @@ pub struct PSS {
...
+        let h_hash = ctx.finish();
+
+        // Steps 7 and 8: encode into output the value of db:
+        //     PS || 0x01 || salt
+        // Where PS is all zeros.
+        let db_len = em_len - digest_len - 1;
+        let pad_len = db_len - digest_len - 1;
+        for i in 0..pad_len {
+            out[i] = 0;
+        }
```
briansmith:
```
let pad = out[..(db_len - digest_len - 1)];
for p in pad {
    *p = 0;
}
```

In general, instead of having `xxx_len` variables, have slices named `xxx` and use `xxx.len()`.

briansmith:
I rebased and pushed to master (with minor modifications) the first commit: "Move rsa.rs and other rsa files to separate folder.

@djc, in general we avoid using the name "mod.rs", which is why we have all the `#[path]` gunk in lib.rs. Unfortunate all around.

briansmith:
I am rebasing the other commits on top of the current master and adding some suggestions. I will give you a link to the rebased version when I'm done.

briansmith:
OK, see https://github.com/briansmith/ring/tree/padding-rebased-and-tweaked.

src/rsa/padding.rs
```diff
@@ -0,0 +1,101 @@
...
+// SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+// OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+use {der, digest};
+
+pub struct RSAPadding {
+    pub digest_alg: &'static digest::Algorithm,
+    pub digestinfo_prefix: &'static [u8],
```
briansmith: Let's not make these public. I realize that you had to make them public because of how the commits are split up. But, instead let's just squash the commits together (this commit through "Refactor RSA padding to use Verification trait.").

src/rsa/rsa.rs
```diff
@@ -26,48 +26,18 @@ use std;
...
-        let (digest_prefix, digest_dst) = out[3 + pad_len..].split_at_mut(
-            self.digestinfo_prefix.len());
-        digest_prefix.copy_from_slice(self.digestinfo_prefix);
-        digest_dst.copy_from_slice(
-            digest::digest(self.digest_alg, msg).as_ref());
-        Ok(())
-    }
-}
+// `RSA_PKCS1_SHA1` is intentionally not exposed.
+use self::padding::{RSAPadding, RSA_PKCS1_SHA1};
```
briansmith: Style nit: Avoid this `use` and refer to then qualified `padding::`.

src/rsa/rsa.rs
```diff
@@ -26,48 +26,18 @@ use std;
...
+pub use self::padding::{RSA_PKCS1_SHA256, RSA_PKCS1_SHA384, RSA_PKCS1_SHA512};
```
briansmith: I understand we need this `pub use` to re-export them. But, if we qualify the references to the non-exported other ones using `padding::*` then let's qualify these below too.

src/rsa/padding.rs
```diff
@@ -14,16 +14,19 @@
 
 use {der, digest};
 
+pub trait Encoding {
```
briansmith: This should have `cfg[(feature = "rsa_signing")]` I think.

briansmith: We should add a comment about why we use the terms "Encoding" and "Verification" instead of terms that seem more like duals like Encoding/Decoding or Signing/Verification. See my suggested text in my rebase-and-tweak branch.

src/rsa/padding.rs
```diff
@@ -38,8 +41,8 @@ impl RSAPadding {
         }
         out[2 + pad_len] = 0;
 
-        let (digest_prefix, digest_dst) = out[3 + pad_len..].split_at_mut(
-            self.digestinfo_prefix.len());
+        let (digest_prefix, digest_dst) =
+                out[3 + pad_len..].split_at_mut(self.digestinfo_prefix.len());
```
briansmith: What's the point of this change? If you think this new formatting is better, please split it into a separate commit that consists just of reformatting changes.

src/rsa/rsa.rs
```diff
@@ -29,12 +29,12 @@ use untrusted;
...
 
 // `RSA_PKCS1_SHA1` is intentionally not exposed.
-use self::padding::{RSAPadding, RSA_PKCS1_SHA1};
+use self::padding::RSA_PKCS1_SHA1;
 pub use self::padding::{RSA_PKCS1_SHA256, RSA_PKCS1_SHA384, RSA_PKCS1_SHA512};
 
 /// Parameters for RSA verification.
 pub struct RSAParameters {
-    padding_alg: &'static RSAPadding,
+    padding_alg: &'static (padding::Verification + Sync),
```
briansmith: Let's just make `padding::Verification` `Sync`; see my suggested change in the rebase-and-tweak branch. We don't want everybody that references this stuff to have to specify `+ Sync`.

src/rsa/rsa.rs
```diff
@@ -76,19 +76,19 @@ macro_rules! rsa_pkcs1 {
     }
 }
 
-rsa_pkcs1!(RSA_PKCS1_2048_8192_SHA1, 2048, RSA_PKCS1_SHA1,
+rsa_impl!(RSA_PKCS1_2048_8192_SHA1, 2048, RSA_PKCS1_SHA1,
```
briansmith: I don't understand the point of this change unless it is supposed to be useful after we add PSS. If so, then let's move it to where PSS is added. Also, the indention of the text is wrong.

briansmith:
Squash these together so that the tests are in the same commit as the new code they test:
  * Implement RSA-PSS verification.  …
  * Add tests for RSA-PSS verification from NIST test cases.  …

Squash these together so that the tests are in the same commit as the new code they test. Also, no point in making Encoding randomized in a separate commit, so it can be squashed in too.
  * Refactor Encoding trait to be randomised.  …
  * Implement RSA-PSS signing.  …
  * Add RSA-PSS signing test cases.  …

briansmith:
It will take some time for me to review the actual code that adds the PSS verification and signing, but it generally seems OK so far. I'm very interested in getting the rest of the prerequisite refactoring changes landed, because whatever refactoring we are going to do is going to affect several in-flight PRs.

src/rsa/rsa.rs
djc: I was thinking `rsa_params` might be a better name at this point.

samscott89:
I've pushed the latest fixes/suggestions.

I think it *might* be possible to completely remove the need for allocating buffers in the PSS encode/verify by modifying MGF1 to directly XOR output into the provided array. However, for `verify`, that would mean performing the XOR on `masked_db` which is currently an immutable `untrusted::Input`.

I'm going to hold off on making any more changes for the time being since there seems to be a few RSA related PRs up in the air. I can either help try and push this one through first, or will wait until the dust settles to rebase this.

briansmith:
> I think it might be possible to completely remove the need for allocating buffers in the PSS encode/verify by modifying MGF1 to directly XOR output into the provided array. However, for verify, that would mean performing the XOR on masked_db which is currently an immutable untrusted::Input.

I field #298 for this optimization.

briansmith:
> 'm going to hold off on making any more changes for the time being since there seems to be a few RSA related PRs up in the air. I can either help try and push this one through first, or will wait until the dust settles to rebase this.

I think that the changes that create non-trivial merge conflicts have all landed now.

src/rsa/rsa.rs
```diff
@@ -43,6 +46,7 @@ fn parse_public_key(input: untrusted::Input)
 
 struct PositiveInteger {
     value: Option<*mut BIGNUM>,
+    bit_len: usize,
```
samscott89: This, plus the logic in `from_be_bytes` to initialise the field, is the only non-trivial change I've made in the rebase.

samscott89:
Okay, rebased this branch on top of master. As I noted [here](https://github.com/briansmith/ring/pull/262/files#r78358489), the only non-trival change I've had to make to the code in the rebase was to add a `bit_len` field to `PositiveInteger`. This is necessary for the PSS padding logic, and seemed like the most natural place to put the logic since the top byte is already parsed separately to handle leading zeros.

briansmith:
The following parts landed
8f14460 Bump version number for backward-incompatible API change.
cb3ce35 Refactor RSA padding to new Encoding/Verification traits.
f1000c7 Normalize formatting of RSA code with rustfmt.
ec56f3f Remove duplicate definition of `pkcs1_digestinfo_prefix!`.

cb3ce35 contains the changes to add the `Verification` and `Encoding` traits and to rename `RSAPadding` to `PKCS1`. I tweaked these changes to put the `impl Verification for PKCS1` directly underneath the `impl Encoding for PKCS1`, to match the organization of the PSS code that will land in the following commits. I also tweaked the formatting using rustfmt.

I also updated my copy of the branch: https://github.com/briansmith/ring/commits/padding, rebased on the above.

src/rsa/padding.rs
```diff
@@ -24,8 +24,8 @@ pub trait Encoding : Sync {
 
 /// The term "Verification" comes from RFC 3447.
 pub trait Verification : Sync {
-    fn verify(&self, msg: untrusted::Input, encoded: untrusted::Input)
-              -> Result<(), error::Unspecified>;
+    fn verify(&self, msg: untrusted::Input, encoded: untrusted::Input,
+              bit_len: usize) -> Result<(), error::Unspecified>;
```
briansmith: I think `public_modulus_len_in_bits` or similar would be a better name.

src/rsa/padding.rs
```diff
@@ -116,8 +116,9 @@ pkcs1_digestinfo_prefix!(
     [ 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x03 ]);
 
 impl Verification for PKCS1 {
-    fn verify(&self, msg: untrusted::Input, encoded: untrusted::Input)
-              -> Result<(), error::Unspecified> {
+    fn verify(&self, msg: untrusted::Input, encoded: untrusted::Input,
+              bit_len: usize) -> Result<(), error::Unspecified> {
+        assert_eq!(encoded.len() * 8, bit_len);
```
briansmith: Is this assertion useful? It seems like we can just add a `_` prefix to the parameter name to show that it is unused so that we don't get an unused parameter warning.

Also, I think in some cases this assertion isn't correct? If it were correct then we wouldn't need the `bit_len` parameter at all, right?

src/rsa/padding.rs
```diff
@@ -154,3 +155,150 @@ impl Verification for PKCS1 {
         })
     }
 }
+
+/// PSS Padding as described in https://tools.ietf.org/html/rfc3447#section-9.1.
+/// It generates a random salt equal in length to the output of the specified
+/// digest algorithm and uses MGF1 as the mask generating function.
```
briansmith: s/MGF1/MGF1 with that digest algorithm/.

src/rsa/padding.rs
```diff
@@ -154,3 +155,150 @@ impl Verification for PKCS1 {
...
+pub struct PSS {
+    digest_alg: &'static digest::Algorithm,
+}
+
+// The maxmimum supported output length for PSS padding is equal to the maximum
```
briansmith: spelling: maximum.

src/rsa/padding.rs
```diff
@@ -154,3 +155,150 @@ impl Verification for PKCS1 {
...
+// supported RSA modulus length.
+// TODO: Can we avoid requiring this MAX_OUTPUT_LEN buffer?
+const MAX_OUTPUT_LEN: usize = 8192 / 8;
```
briansmith: This is actually the maximum public modulus length, right? We should put this in `ring::rsa` since it isn't a padding-specific value. Also, we should replace `MAX_BITS` in `verify_rsa` with a reference to this value. `PUBLIC_MODULUS_MAX_LEN` would be a good name for this.

briansmith: We can drop the TODO because we already have an issue filed for removing the buffer.

src/rsa/padding.rs
```diff
@@ -154,3 +155,150 @@ impl Verification for PKCS1 {
...
+impl Verification for PSS {
+    // PSS verification as specified in
+    // https://tools.ietf.org/html/rfc3447#section-9.1.2
+    fn verify(&self, msg: untrusted::Input, encoded: untrusted::Input,
+              bit_len: usize) -> Result<(), error::Unspecified> {
+        // Number of bytes required for bit_len bits.
+        let em_len = 1 + (bit_len - 1) / 8;
+        assert_eq!(encoded.len(), em_len);
+        // Create a bit mask to match the size of the modulus used.
+        let top_byte_mask = (0xffu16 >> 1 + bit_len % 8) as u8;
```
briansmith: We need to have unit tests for this particular calculation, with vectors for each possible value of `bit_len % 8`. These tests don't need to do the RSA part; they just need to call this function and/or call a function that does just this calculation.

Clippy says:
> operator precedence can trip the unwary. Consider parenthesizing your expression: `0xffu16 >> (1 + bit_len % 8)`,

That sounds like a good recommendation.

src/rsa/padding.rs
```diff
@@ -154,3 +155,150 @@ impl Verification for PKCS1 {
...
+        encoded.read_all(error::Unspecified, |em| {
+            let digest_len = self.digest_alg.output_len;
+
+            // Step 2.
+            let m_hash = digest::digest(self.digest_alg,
+                                        msg.as_slice_less_safe());
+
+            // Step 3: where we assume the digest and salt are of equal length.
+            if em_len < 2 + (2 * digest_len) {
+                return Err(error::Unspecified)
```
briansmith: Missing semicolon here.

src/rsa/padding.rs
```diff
@@ -154,3 +155,150 @@ impl Verification for PKCS1 {
...
+}
+
+// Mask-generating function MGF1 as described in
+// https://tools.ietf.org/html/rfc3447#appendix-B.2.1.
+fn mgf1(digest_alg: &'static digest::Algorithm, seed: &[u8], mask: &mut [u8])
+        -> Result<(), error::Unspecified> {
+    let digest_len = digest_alg.output_len;
+
+    // Maximum counter value is the value of (mask_len / digest_len) rounded up.
+    let ctr_max = (mask.len() - 1) / digest_len;
```
briansmith: We should get rid of this code duplication between the loop and the block that immediately follows the loop. Maybe we can do this instead?:

````
for (i, mask_chunk) in mask.chunks_mut(digest_len).enumerate() {
    let mut ctx = digest::Context::new(digest_alg);
    ctx.update(seed);
    ctx.update(&polyfill::slice::be_u8_from_u32(i as u32));
    let digest = ctx.finish();
    mask_chunk.copy_from_slice(digest.as_ref()[..mask_chunk.len()];
}

File: src/rsa/rsa.rs [hunk 61]
```diff
@@ -20,7 +20,10 @@ use untrusted;
 mod padding;
 
 // `RSA_PKCS1_SHA1` is intentionally not exposed.
-pub use self::padding::{RSA_PKCS1_SHA256, RSA_PKCS1_SHA384, RSA_PKCS1_SHA512};
+pub use self::padding::{
```
briansmith: Reformat this part with rustfmt.

src/rsa/rsa.rs
briansmith: Why add bit_len as a field, instead of implementing a method on `PositiveInteger` that calls `BN_num_bits()` or similar? It doesn't seem great to store this extra field here when it's value is only needed in during RSA signature verification. It's not a huge field but `RSAKeyPair` will eventually contain a bunch of these so it seems wasteful.

src/rsa/rsa.rs
```diff
@@ -63,17 +67,19 @@ impl PositiveInteger {
         }
         // Reject leading zeros. Also reject the value zero ([0]) because zero
         // isn't positive.
-        if untrusted::Reader::new(input).peek(0) {
+        let top_byte =  try!(untrusted::Reader::new(input).read_byte());
+        if top_byte == 0 {
             return Err(error::Unspecified);
         }
+        let bit_len = input.len() * 8 - top_byte.leading_zeros() as usize;
```
briansmith: `input.len() * 8` could overflow. We need to prevent that.

Maybe we should just calculate/store the number of leading zeros in the top byte directly, instead of calculating the length of the modulus in bits?

src/rsa/padding.rs
```diff
@@ -37,8 +39,9 @@ pub struct PKCS1 {
 impl Encoding for PKCS1 {
     // Implement padding procedure per EMSA-PKCS1-v1_5,
     // https://tools.ietf.org/html/rfc3447#section-9.2.
-    fn encode(&self, msg: &[u8], out: &mut [u8])
-              -> Result<(), error::Unspecified> {
+    fn encode(&self, msg: &[u8], out: &mut [u8], bit_len: usize,
+              _: &rand::SecureRandom) -> Result<(), error::Unspecified> {
+        assert_eq!(out.len() * 8, bit_len);
```
briansmith: Again, I don't think this assertion is correct.

src/rsa/padding.rs
```diff
@@ -163,6 +166,11 @@ pub struct PSS {
     digest_alg: &'static digest::Algorithm,
 }
 
+#[cfg(feature = "rsa_signing")]
+// Maximum supported length of the salt in bytes.
+// In practice, this is constrained by the maximum digest length.
+const MAX_SALT_LEN: usize = 512 / 8;
```
briansmith: `const MAX_SALT_LEN: usize = digest::MAX_OUTPUT_LEN`.

briansmith:
Are there any test cases in the test suite for PSS *verification* that have public moduli with the high bit not set?

samscott89:
Ok, I think I've addressed the suggestions/problems from before and added tests for PSS verification for moduli of variable bit lengths mod 8.

ctz:
_gentle bump_. I could really use this for rustls TLS1.3 support.

util/generate-rsa-signing-tests.py
```diff
@@ -26,10 +26,26 @@
...
 import sys, copy
 
-def parse(fn):
+DIGEST_OUTPUT_LENGTHS = {
+    'SHA1': 80,
+    'SHA224': 112,
```
briansmith: 'SHA224` should be removed since it's not a supported algorithm and it is never used, IIUC.

util/generate-rsa-signing-tests.py
```diff
@@ -40,57 +56,163 @@ def parse(fn):
...
                 cur = copy.copy(cur)
     return cases
 
-def main(fn):
-    for case in parse(fn):
+def print_sign_test(case, n, e, d, padding_alg):
+    # Can only support moduli which are a whole number of bytes
+    if n.bit_length() % 8 != 0:
+        debug("Skipping due to modulus (not a whole number of bytes).", DEBUG)
+        return
```
briansmith: Sorry, I gave you bad advice. In the review of #283 we did the math and found out that even if we require the private key primes `p` and `q` to be multiples of 8 bits or even multiples of 1024 bits, it's still possible and even pretty likely that the public modulus will be 2047 bits; i.e. `n.bit_length() % 8 == 7`. So, we in fact do need to support public moduli that aren't an even number of bytes, even for signing. Sorry!

util/generate-rsa-signing-tests.py
```diff
@@ -40,57 +56,163 @@ def parse(fn):
...
+
+    msg = case['Msg'].decode('hex')
+
+    # Recalculate and compare the signature to validate our processing.
+    if padding_alg == 'PKCS':
+        sig = key.sign(msg, padding.PKCS1v15(),
+                       getattr(hashes, case['SHAAlg'])())
+        hex_sig = ''.join('{:02x}'.format(ord(c)) for c in sig)
+        assert hex_sig == case['S']
+    elif padding_alg != "PSS":
```
briansmith: Please write this like this:

```python
elif padding_alg == "PSS":
    # PSS is randomised, can't recompute this way
    pass
else:
    print "Invalid padding algorithm"
    quit()
```

util/generate-rsa-signing-tests.py
```diff
@@ -40,57 +56,163 @@ def parse(fn):
...
-
-        # Recalculate and compare the signature to validate our processing.
-        msg = case['Msg'].decode('hex')
-        sig = key.sign(msg, padding.PKCS1v15(),
-                       getattr(hashes, case['SHAAlg'])())
-        hex_sig = ''.join('{:02x}'.format(ord(c)) for c in sig)
-        assert hex_sig == case['S']
+        if n.bit_length() < 2048:
+            debug("Skipping due to modulus length (too small).", DEBUG)
+            continue
```
briansmith: This check should be `n.bit_length() // 8 < 2048 // 8`. In particular, *ring* considers any modulus where the length in bytes * 8 == 2048 to be large enough.

util/generate-rsa-signing-tests.py
```diff
@@ -40,57 +56,163 @@ def parse(fn):
...
+
+        if n.bit_length() > 4096:
```
briansmith: This should only be done for signing tests, not verification tests, right?

briansmith:
FYI, I'm updating the test vector conversion script now. I found some additional things that should be changed when I tested it. I'll commit the updated version and then finish the review of the remaining stuff.

src/rsa/pss_verify_tests.txt
```diff
@@ -0,0 +1,60 @@
+# Test vectors for RSA-PSS padding for variable public modulus lengths
```
briansmith: I'll move these test vectors into the same file as the NIST test vectors. In the PKCS#1 test vectors and other places we already mix imported test vectors with others that we've created, so there's always one file per algorithm (basically). It makes the regeneration of the converted-from-NIST files less convenient but it's not worth having separate files.

Plus, this file doesn't have the right naming convention.

src/rsa/verification.rs
```diff
@@ -89,10 +101,8 @@ pub fn verify_rsa(params: &RSAParameters,
                   (n, e): (untrusted::Input, untrusted::Input),
                   msg: untrusted::Input, signature: untrusted::Input)
                   -> Result<(), error::Unspecified> {
-    const MAX_BITS: usize = 8192;
-
     let signature = signature.as_slice_less_safe();
-    let mut decoded = [0u8; (MAX_BITS + 7) / 8];
+    let mut decoded = [0u8; (PUBLIC_MODULUS_MAX_LEN + 7) / 8];
```
briansmith: I'll rename this variable to `PUBLIC_MODULUS_MAX_BITS` for clarity. I expect `*_LEN` to be in terms of array elements, which can't be bits.

src/rsa/rsa_pss_verify_tests.txt
```diff
@@ -0,0 +1,338 @@
+# NIST RSA PSS Test Vectors for FIPS 186-4 from SigVerPSS_186-3.rsp in
```
briansmith: I'll regenerate this file using the script that was updated with my changes.

src/rsa/pss_verify_tests.txt
```diff
@@ -0,0 +1,60 @@
...
+#
```
briansmith: All the test vectors here are passing test vectors. We'll need to add some failing test vectors for these cases too.

src/rsa/pss_verify_tests.txt
briansmith: OK, I misunderstood the tests. They don't contain a public key so I can't merge them into rsa_pss_verify_tests.txt.

Instead, let's rename the file rsa_pss_verify_padding_tests.txt to match the other files that all start with rsa_.

briansmith: In addition to these test vectors, we will still need to have end-to-end tests of RSA-PSS with public moduli that aren't a multiple of 8, both positive and negative.

briansmith:
I moved the script that converts the NIST test vectors to src/rsa/convert_nist_rsa_test_vectors.py because (1) the old name implied that it was for the signing tests only, but actually it is for both signing and verification, and (2) to keep all the RSA-related stuff together.

I landed your updated NIST test vector conversion script in bbbb2fe17283cb83cb57d731bfdb50a3b53aac65. I removed the unused copies of the original NIST files. My tweaks, which address the issue I noted above and more, are in 563c2c3be5416b7f889f05cb91b8d1a3e9365621. I also added the SHA-384 digest of the NIST test vector files to the output.

I redid the existing PKCS#1 1.5 test vectors based on the new script: 1045c8f89a25ecc865f424e604db8e8570cee02f.

Let me know if anything looks off. Otherwise, I think we're done with the Python coding in this PR.

briansmith:
I forgot to fold the SHA-384 digest part of the test vector conversion script into my aforementioned commit. It's now in 03ce40dcb9bcfdd0d501084f3730316e91020253.

src/rsa/padding.rs
```diff
@@ -159,3 +160,179 @@ pkcs1_digestinfo_prefix!(
...
+impl Verification for PSS {
+    // PSS verification as specified in
+    // https://tools.ietf.org/html/rfc3447#section-9.1.2
+    fn verify(&self, msg: untrusted::Input, encoded: untrusted::Input,
+              public_modulus_len_in_bits: usize)
+              -> Result<(), error::Unspecified> {
+        // Number of bytes required to encode message. The maximum length is
+        // given by the length of the public modulus in bits minus 1.
+        let em_len = 1 + (public_modulus_len_in_bits - 2) / 8;
+        assert_eq!(encoded.len(), em_len);
```
briansmith: I don't think this is correct. Let's say we have a 2049 bit modulus. Then `encoded.len() == (2049 + 7) / 8`, which is 257. However, `emBits == modLen - 1` == 2048, and `emLen == ceil(emBits / 8)`, which is 256. Thus we have `encoded.len() != em_len` here.

I have a patch that fixes this.

briansmith:
I spent quite a bit of time reviewing this PR. In the course of doing so, I've written some patches that (1) factor out the duplicate logic between verification and encoding, and (2) align the code more closely to the specification. I still need to create more test vectors, including in particular negative test vectors.

src/rsa/pss_verify_tests.txt
```diff
@@ -0,0 +1,60 @@
...
+Len = 2048
+Result = P
+
+Digest = SHA256
+Msg = e002377affb04f0fe4598de9d92d31d6c786040d5776976556a2cfc55e54a1dcb3cb1b126bd6a4bed2a184990ccea773fcc79d246553e6c64f686d21ad4152673cafec22aeb40f6a084e8a5b4991f4c64cf8a927effd0fd775e71e8329e41fdd4457b3911173187b4f09a817d79ea2397fc12dfe3d9c9a0290c8ead31b6690a6
+Encoded = c9cbb20ff49558e10505fea39e85e30f64f7b71a6307e51d58d1b09139e623ca00800224f6024fb84c7f5407f3d4045e4474def670e32d62590861e10612567eb8f51ba7d8b095553c214d75bb9960f33aa1aa173d020558e35d2b6d7385bd04223b707083d3311b37fea10f73fcb88ec5a9f1920fdef4dba6965756cf78a89ba4cea0fc29b834849c05fa6f8ba823e9c47df8b563f56aaa9a20f69dbee63bb56306c467963703e6671e1eb2f33021404f8ee9332b364f33b6468272afe85dd0ff8f94e7eb2a28876a975d51ec87af3e5fba4a50374daa2a0cde9eb8b675d9fab3a8ec5deee1cace4acb6e0df2ae12f6c63cf28c96492f1a742dcf91186cc0bc
+Len = 2049
```
samscott89: It has been a while now since I wrote that code, but I think it's correct in this form (per the spec).

From https://tools.ietf.org/html/rfc3447#section-8.1.1:
> Apply the EMSA-PSS encoding operation (Section
      9.1.1) to the message M to produce an encoded message EM of length
      `\ceil ((modBits - 1)/8)` octets such that the bit length of the
      integer OS2IP (EM) (see Section 4.2) is at most `modBits - 1`, where
      modBits is the length in bits of the RSA modulus n:
     `    EM = EMSA-PSS-ENCODE (M, modBits - 1)`.
      Note that the octet length of EM will be one less than k if
      modBits - 1 is divisible by 8 and equal to k otherwise.

(Here k is the length in bytes of the public modulus).

Using your example values, modBits is 2049, k is 257, so the encoded messages should be \ceil((2049 - 1)/8) = 256 bytes. This is is the exceptional case noted in the last sentence.

samscott89:
Thanks for the review. The python changes all look good. It's a shame there aren't more test vectors out there we can use to clarify the length issue. Let me know if there's anything more I can do otherwise.

Small nit in python file: filename in the usage help string is incorrect ([here](https://github.com/briansmith/ring/blob/master/src/rsa/convert_nist_rsa_test_vectors.py#L220)).

src/rsa/padding.rs
samscott89: See my other comment for why I think this is correct. It should be the number of bytes needed to represent a `public_modulus_len_in_bits - 1`-bit value. Could probably be simplified to `public_modulus_len_in_bits + 6) / 8` however.

src/rsa/pss_verify_tests.txt
briansmith: I'm open to the idea that I'm wrong, but here's my thinking:

PKCS#1 and PSS both get the same inputs, namely for a 2049-bit modulus they are both going to get a value of `encoded()` that is 257 bytes long. Note in particular that the caller of `Verification::verify()` always converts the `BIGNUM` value `M` into a big-endian value of length `k`. Now, for PSS the value of `M` only requires `emLen` bytes, which, as you say, might be one less than `k`. But, the caller of `Verification::verify()` is going to left-pad `encoded` with an extra zero byte, and so in order to get the value `EM` the implementation of `Verification::verify()` must strip off one zero byte if and only if `(modBits - 1) % 8 == 0`.

WDYT?

samscott89: > PKCS#1 and PSS both get the same inputs, namely for a 2049-bit modulus they are both going to get a value of encoded that is 257 bytes long.

Ah because the upper layer producing the encoded message is always going to return 257 bytes. That makes sense. This is covered step 2c in the spec [section 8.1.2](https://tools.ietf.org/html/rfc3447#section-8.1.2), which technically should be handled before calling verify, but definitely looks more sensible to include it within the body of verify.

I think we should probably put a small comment to that effect somewhere.

samscott89: The test case was simply one of those used in the NIST tests. Empty string seems like a sensible alternative.

src/rsa/padding.rs
samscott89: Also note that we will need to do something similar for the `encode` function. If `(modbits - 1) % 8 ==0` then the padding will need to set the first byte to be 0 and output the padding starting from byte 1.

briansmith:
For people following along, I posted some review comments in the form of commits in #344, with the expectation that those commits will get incorporated back into this PR.

briansmith:
Thanks very much for doing this. I landed all these changes plus the additional changes in #344 onto master. There is some follow-up work to do before the 0.6.0 release regarding testing and clarifying the code further, but I'd rather do that in separate PRs that are easier to manage. Let's continue discussion in  #115 (PSS verification) and #213 (PSS signing).

4e5145157ccf813b76b86a0f7064b0ae3ebb48f7 <- From this PR.
7520d283ebc6516f9959b91ba303c07ae288c419 <- Bug fix.
0ce0643f9515065671c58b8440d94a7002edf9d5
ba8199b17559d8a60186446bf5daad9eacac0826
8cb367341707cfb78fa0be9085c59db71bdaefcb
2e9ef69c23645642aa89f9c00b2f4e27ee2fd191 <- From this PR.
5be8b7f9606b98e8c3b1efcc7dd730e582b0e702 <- Bug fix.
18ff68f90cb2255b634c38c4ac5bbda57cfaabaa
eeb99a780ebce0736d9b0b8d17722be9c2d335ce
4c9a9c875192ea91b670b20838d18a5f9182d4b0
be08baa959330d1e9e2c52b77cb5ac7600d2959d <- Bug fix.
5fb1fc8a5e65b65cb7247eb5b50905a0950986b6

PR #265, opened 2016-08-16
Only build `test` module if we are in test mode.

frewsxcv:

### DIFF
--- a/src/digest/digest.rs
+++ b/src/digest/digest.rs
@@ -221,7 +221,7 @@ impl Clone for Context {
 /// # Examples:
 ///
 /// ```
-/// # #[cfg(any(feature = "use_heap", test))]
+/// # #[cfg(all(feature = "use_heap", test))]
 /// # fn main() {
 /// use ring::{digest, test};
 ///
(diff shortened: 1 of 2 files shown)


src/digest/digest.rs
```diff
@@ -221,7 +221,7 @@ impl Clone for Context {
 /// # Examples:
 ///
 /// ```
-/// # #[cfg(any(feature = "use_heap", test))]
+/// # #[cfg(all(feature = "use_heap", test))]
```
briansmith: Good catch.

I believe this should just be `#[cfg(feature = "use_heap")]`, and same below. We don't need to condition this on `test` at all because that is handled by the test driver, right?

src/lib.rs
```diff
@@ -136,7 +136,7 @@ mod rsa;
 
 pub mod signature;
 
-#[cfg(any(feature = "use_heap", test))]
+#[cfg(all(feature = "use_heap", test))]
```
briansmith: `ring::test` is a public module that exposes an API for others to write their own tests in the same style as *ring*'s, using data files with test vectors instead of hard-coding test data into the Rust code. Thus, I believe `any` is correct, but maybe I am mistaken.

In particular, consider a crate Foo that depends on *ring*. When we do `cargo test` for Foo, with `*ring*` be built using the equivalent of `cargo test` or `cargo build`? In other words, would `test` be active when building *ring* when doing `cargo test` on another crate?

frewsxcv:
From #cargo on IRC:

```
frewsxcv> If I run `cargo test` for a crate, will dependencies be rebuilt with #[cfg(test)] or does that only happen for the immediate crate?
sfackler> only the immediate crate
```

Considering you want `ring::test` to be available to crates outside *ring* (which I didn't know), using `any` is probably the right thing.

frewsxcv:
Actually, you probably do want just `[cfg(feature = "use_heap")]`. I can open a PR for that later if you don't already get to it.

PR #266, opened 2016-08-16
Test on '1.9.0' instead of 'stable' channel.

frewsxcv:
This will make it easier to track what minimum version of Rust
*ring* requires.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,27 +18,27 @@ matrix:
     # BEGIN GENERATED
 
     - env: TARGET_X=x86_64-apple-darwin CC_X=clang CXX_X=clang++ FEATURES_X=--features=rsa_signing MODE_X=DEBUG KCOV=0
-      rust: stable
+      rust: 1.9.0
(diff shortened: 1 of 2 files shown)


briansmith:
Thanks for the PR. The only version “supported” by *me* is the latest stable version. As soon as a new features is added to rustc or cargo in a release version, we'll consider using it in *ring*. I plan to bump the version number of *ring* when I do things like that. One reason I added *ring* to crates.io is so that people can use SemVer to manage changes like this that might have a negative impact on them.

I also keep track of the builds for the latest beta and latest nightly version, mostly to get a head start on any upcoming breaking changes that would affect the latest stable version.

I recommend that people who depend on *ring* in their own projects also run the *ring* test suite during their own CI builds with whatever versions of the tools they are using. I believe this can be done with `cargo test -p ring`. If this doesn't work, then I would be happy to take PRs to make it work (on that latest stable version of Rust).

Again, thanks for the PR.

PR #269, opened 2016-08-17
Fix incorrect compilation flag in doc example.

frewsxcv:

### DIFF
--- a/src/digest/digest.rs
+++ b/src/digest/digest.rs
@@ -221,7 +221,7 @@ impl Clone for Context {
 /// # Examples:
 ///
 /// ```
-/// # #[cfg(any(feature = "use_heap", test))]
+/// # #[cfg(feature = "use_heap")]
 /// # fn main() {
 /// use ring::{digest, test};
 ///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -136,7 +136,7 @@ mod rsa;
 
 pub mod signature;
 
-#[cfg(any(feature = "use_heap", test))]
+#[cfg(feature = "use_heap")]
```
briansmith: Consider the case where the library will be linked into an executable that doesn't have `malloc`. That is what disabling `use_heap` is supposed to allow.

But, we still want people to be able to run the *ring* test suite for that configuration, in another executable (the one generated by `cargo test`. Right now, that means they have to have a working `malloc`. In the future, we'll hopefully remove this requirement.

tl;dr: I think the original condition is what we want.

src/digest/digest.rs
```diff
@@ -221,7 +221,7 @@ impl Clone for Context {
 /// # Examples:
 ///
 /// ```
-/// # #[cfg(any(feature = "use_heap", test))]
+/// # #[cfg(feature = "use_heap")]
```
briansmith: I agree this and the following identical change are good.

briansmith:
Thanks!

PR #273, opened 2016-08-18
Refactor RSA signing API to allow concurrent signing with the same RSA key

briansmith:
Also, make `RSAKeyPair : Sync + Send` and `RSASigningState: Send`.

### DIFF
--- a/src/rsa/rsa.rs
+++ b/src/rsa/rsa.rs
@@ -12,122 +12,21 @@
 // OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
-#![allow(unsafe_code)]
-
 /// RSA PKCS#1 1.5 signatures.
 
-use {bssl, c, der, digest, error, private, signature};
-
-#[cfg(feature = "rsa_signing")]
-use rand;
-
-#[cfg(feature = "rsa_signing")]
-use std;
-
+use {der, digest, error};
 use untrusted;
 
+#[cfg(feature = "rsa_signing")]
+use c;
+
+#[cfg(feature = "rsa_signing")]
+use core;
 
 pub struct RSAPadding {
     digest_alg: &'static digest::Algorithm,
     digestinfo_prefix: &'static [u8],
 }
-
-#[cfg(feature = "rsa_signing")]
-impl RSAPadding {
-    // Implement padding procedure per EMSA-PKCS1-v1_5,
-    // https://tools.ietf.org/html/rfc3447#section-9.2.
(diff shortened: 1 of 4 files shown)


djc:
I like the split between `signing` and `verification`, but it would seem nice to me if we hide such module internals from the rest of *ring*, such that `signature` can be oblivious to changes in structure inside the `rsa` module (and similar for e.g. `ec`, I guess).

As for the introduction of the `RSASigningState`, would it be possible to have `RSAKeyPair` construction return a bare`RSAKeyPair` and then have the `RSASigningState` wrap it in the `Arc`, instead? I find it too bad that the API is significantly more complicated, but it does seem to fit *ring*'s approach of making stuff explicit and not impeding performance.

briansmith:
> I like the split between signing and verification, but it would seem nice to me if we hide such module internals from the rest of ring, such that signature can be oblivious to changes in structure inside the rsa module (and similar for e.g. ec, I guess).

I'm not sure this is worthwhile. It would make more work when we need to change/add/remove something internally, since we'd have to modify the re-exports.

> As for the introduction of the RSASigningState, would it be possible to have RSAKeyPair construction return a bare RSAKeyPair and then have the RSASigningState wrap it in the Arc, instead?

Sure. I will try this.

> I find it too bad that the API is significantly more complicated, but it does seem to fit ring's approach of making stuff explicit and not impeding performance.

We could actually keep `RSAKeyPair::sign`, implemented as the construction of a new temporary `RSASigningState` and then calling `sign` on it. This would keep the convenience for simple callers.

djc:
> I'm not sure this is worthwhile. It would make more work when we need to change/add/remove something internally, since we'd have to modify the re-exports.

Not sure I understand. It seems that, if you adjust something internal to `ring::rsa`, you either have to (a) adjust all the users inside the entirety of `ring`, or (b) only adjust the `rsa`-internal re-exports. I would assume the latter would make for a smaller change than the former.

> We could actually keep `RSAKeyPair::sign`, implemented as the construction of a new temporary `RSASigningState` and then calling sign on it. This would keep the convenience for simple callers.

That sounds great! We can point people to the more-performant API from documentation near the simple one.

burdges:
Wait, there is no optimization for parallel signing with the same key here, ala Ed25519 tricks, right?

briansmith:
Landed:

5b75fb1 Refactor RSA signing API to allow concurrent use of `RSAKeyPair`.
d478370 Remove unnecssary boxing of `RSA` within `RSAKeyPair`.
bf6ccac Split RSA verification- & signing- specific code into submodules.

I didn't add an `RSASigningKey::sign` because I didn't have time and because it was a little bit trickier than I expected. We can do that in a follow-up. I did change `RSASigningKey::new()` so that it returns an `RSASigningKey` instead of a `Arc<RSASigningKey>`.

PR #274, opened 2016-08-18
hkdf test was not comparing actual to expected output

PeterReid:


briansmith:
Thanks a lot! Fixed in e7cd8ee76646e3e46ea6d3238761be456da44772!

PR #277, opened 2016-08-22
fixed some clippy warnings

llogiq:
Mostly readability nits, some simplifications of the code. I have abstained from changing indexed loops to iterators for now because I don't know if you need a certain timing profile.

### DIFF
--- a/src/digest/digest.rs
+++ b/src/digest/digest.rs
@@ -435,7 +435,7 @@ fn sha512_format_output(input: &[u64; MAX_CHAINING_LEN / 8])
     ]
 }
 
-/// SHA512_4 calculates the SHA-512 digest of the concatenation of |part1|
+/// `SHA512_4` calculates the SHA-512 digest of the concatenation of |part1|
 /// through |part4|. Any part<N> may be null if and only if the corresponding
 /// part<N>_len is zero. This ugliness exists in order to allow some of the
 /// C ECC code to calculate SHA-512 digests.
--- a/src/ec/eddsa.rs
+++ b/src/ec/eddsa.rs
@@ -96,9 +96,7 @@ impl<'a> Ed25519KeyPair {
 
     fn from_bytes_unchecked(private_key: &[u8], public_key: &[u8])
                             -> Result<Ed25519KeyPair, error::Unspecified> {
-        if private_key.len() != 32 {
-            return Err(error::Unspecified);
-        } else if public_key.len() != 32 {
+        if private_key.len() != 32 || public_key.len() != 32 {
             return Err(error::Unspecified);
         }
         let mut pair = Ed25519KeyPair { private_public: [0; 64] };
--- a/src/ec/suite_b/ops/p256.rs
+++ b/src/ec/suite_b/ops/p256.rs
@@ -88,14 +88,14 @@ fn p256_elem_inv(a: &ElemUnreduced) -> ElemUnreduced {
     }
 
     let b_1 = &a;
-    let b_11     = sqr_mul(&b_1,  0 +  1, &b_1);
+    let b_11     = sqr_mul(b_1,   0 +  1, b_1);
     let f        = sqr_mul(&b_11, 0 +  2, &b_11);
     let ff       = sqr_mul(&f,    0 +  4, &f);
     let ffff     = sqr_mul(&ff,   0 +  8, &ff);
     let ffffffff = sqr_mul(&ffff, 0 + 16, &ffff);
 
     // ffffffff00000001
-    let mut acc = sqr_mul(&ffffffff, 31 + 1, &b_1);
+    let mut acc = sqr_mul(&ffffffff, 31 + 1, b_1);
 
     // ffffffff00000001000000000000000000000000ffffffff
     sqr_mul_acc(&mut acc, 96 + 32, &ffffffff);
@@ -114,7 +114,7 @@ fn p256_elem_inv(a: &ElemUnreduced) -> ElemUnreduced {
 
     // ffffffff00000001000000000000000000000000fffffffffffffffffffffffd
     sqr_mul_acc(&mut acc, 0 + 2, &b_11);
-    sqr_mul(&acc, 1 + 1, &b_1)
+    sqr_mul(&acc, 1 + 1, b_1)
 }
 
 fn p256_point_mul_base_impl(g_scalar: &Scalar) -> Point {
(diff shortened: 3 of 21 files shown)


src/digest/digest.rs
```diff
@@ -435,7 +435,7 @@ fn sha512_format_output(input: &[u64; MAX_CHAINING_LEN / 8])
     ]
 }
 
-/// SHA512_4 calculates the SHA-512 digest of the concatenation of |part1|
+/// `SHA512_4` calculates the SHA-512 digest of the concatenation of |part1|
```
briansmith: I'll just s/`SHA512_4` c/C/.

src/ec/eddsa.rs
```diff
@@ -96,9 +96,7 @@ impl<'a> Ed25519KeyPair {
 
     fn from_bytes_unchecked(private_key: &[u8], public_key: &[u8])
                             -> Result<Ed25519KeyPair, error::Unspecified> {
-        if private_key.len() != 32 {
-            return Err(error::Unspecified);
-        } else if public_key.len() != 32 {
+        if private_key.len() != 32 || public_key.len() != 32 {
```
briansmith: I still prefer to keep this as two if statements. Let's replace "} else " with whitespace to move the "if public_key_len() != 32" to the next line.

Separate statements help with code coverage.

src/ec/suite_b/ops/p256.rs
```diff
@@ -88,14 +88,14 @@ fn p256_elem_inv(a: &ElemUnreduced) -> ElemUnreduced {
     }
 
     let b_1 = &a;
-    let b_11     = sqr_mul(&b_1,  0 +  1, &b_1);
+    let b_11     = sqr_mul(b_1,   0 +  1, b_1);
```
briansmith: I did this one intentionally, but OK, if Clippby complains, let's get rid of it.

src/lib.rs
```diff
@@ -86,7 +86,7 @@
     unused_unsafe,
     unused_variables,
     variant_size_differences,
-    warnings,
+//    warnings,
```
briansmith: Is this intentional, or was it supposed to be removed before you submitted the PR?

src/pbkdf2.rs
```diff
@@ -158,7 +158,9 @@ pub fn derive(prf: &'static PRF, iterations: usize, salt: &[u8], secret: &[u8],
     let mut idx: u32 = 0;
 
     for chunk in out.chunks_mut(output_len) {
-        idx = idx.checked_add(1).expect("derived key too long");
+        idx = idx.checked_add(1).expect("derived kmanual implementation of an assign operation, #[warn(assign_op_pattern)] on by default
+   --> src/pbkdf2.rs:248:9
+ey too long");
```
briansmith: Looks like this need to be fixed?

src/test.rs
```diff
@@ -218,13 +218,8 @@ pub fn from_file<F>(test_data_relative_file_path: &str, mut f: F)
     let mut current_section = String::from("");
     let mut failed = false;
 
-    loop {
-        let mut test_case =
-                match parse_test_case(&mut current_section, &mut lines) {
-            Some(test_case) => test_case,
-            None => { break; },
-        };
-
```
briansmith: Nice. I didn't know Clippy could do this.

briansmith:
Thanks for doing this!

Some general comments:

* Should I assume you the copyright holder for this contribution and future contributions from you?
* Please add the contribution statement to the end of your commit message. See https://github.com/briansmith/ring#contributing.
* I would have preferred to have the comment-fixing stuff in a separate commit from the stuff that changes actual code, but this is OK for now.

As for the iterator stuff, I am open to doing it in a separate PR. I am a bit skeptical that it will improve readability, but OTOH it could. I'll run the benchmarks to make sure it doesn't regress performance. (The benchmarks are in [crypto-bench](https://github.com/briansmith/crypto-bench).

src/ec/suite_b/public_key.rs
```diff
@@ -28,10 +28,10 @@ use untrusted;
...
 /// [NIST SP 800-56A, revision 2]:
 ///     http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar2.pdf
-pub fn parse_uncompressed_point<'a>(ops: &PublicKeyOps,
-                                    input: untrusted::Input<'a>)
-                                    -> Result<(Elem, Elem),
-                                              error::Unspecified> {
+pub fn parse_uncompressed_point(ops: &PublicKeyOps,
+                                input: untrusted::Input)
+                                -> Result<(Elem, Elem),
+                                          error::Unspecified> {
```
Ms2ger: This can fit inside 80 columns now. (That is the policy, right? I didn't find it in STYLE.md.)

briansmith: I just kind of inherited the convention from BoringSSL, which seems to be "approximately 80" in a way I can't fully describe. Apparently punctuation can overhang by one character? Not sure. I'm looking forward to figuring out how to get rustfmt to format my Rust code in the "wrap to fewest number of lines" style that is used in *ring* and BoringSSL (As opposed to the "one thing per line unless everything fits on one line" style that seems to be the default.)

briansmith:
Thanks! I split up the patches slightly and made some tweaks mentioned above and a couple other tweaks. Landed as:

89ab6d2 Simplify test code by using `while let`.
3db3bf4 Use compound assignment operator when possible.
789cb41 Use characters instead of strings for single-character comparisons.
c260b35 Avoid explicit specificiation of lifetimes unless necessary.
80b7263 use `x.is_empty()` instead of `x.len() == 0`.
26e7309 Use implicit return instead of explicit return.
9b04233 Add backticks in doc comments to appease Clippy.
48bb3ed Remove `else` after return.
19d0d79 Drop mention of function name in SHA512_4 documentation.
57505cd Remove unnecessary uses of `&` on references.

llogiq:
Just when I was going to return to fix the problems with the PR. :smile: Anyway, glad I could be of help.

briansmith:
@llogiq Could you provide an example of the iterator stuff? Do you think it it would make the code better?

llogiq:
For example, in https://github.com/briansmith/ring/blob/master/src/pbkdf2.rs#L180-L182 , you could write
```Rust
for (i, mut item) in out.iter_mut().enumerate() {
     *item ^= u.as_ref()[i];
}
```
to elide the bounds check on accessing `out`.

briansmith:
> for (i, mut item) in out.iter_mut().enumerate() {
>      *item ^= u.as_ref()[i];
> }
> to elide the bounds check on accessing out.

I see. I guess the idiomatic iterator- based way is actually something like this, right?:

```
for (mut out, u) in out.iter_mut().zip(u.as_ref().iter()) {
    *out |= *u;
}
```

I guess it's not clearly better to me. The rust compiler should (but probably doesn't) optimize them into the same thing. I guess I might be OK with leaving things as they are for now, since I think the current code is easier to understand at least for C(rypto) programmers new to Rust.

briansmith:
Thanks for the info!

llogiq:
You're welcome. Yes, that's what I thought.

PR #282, opened 2016-08-24
Fix strawberryperl link in BUILDING.md

pravic:
Currently it points to https://github.com/briansmith/ring/blob/master/strawberryperl.com.


briansmith:
Thanks!

briansmith:
Note that if you use *ring* from crates.io, you don't need any Perl or Yasm. It's only for building from Git now. I'll update the documentation when I have time.

PR #283, opened 2016-08-24
Oxidize/restore special-case handling for `0b100...` range bounds

djudd:
... in RSA blinding keygen, lost when OpenSSL code was oxidized

There are some TODOs in the code regarding `as`, but also a larger one, where I want to agree on a strategy before moving forward.

I pulled the `0b100` check and mask calculation for Range sampling into the Range constructor, where it's easier to test and, I think, clearer. However, this means that the EC code which uses Range just for the `are_limbs_within` check now is unnecessarily initializing those values too. Even though it skips the complex path, I'm not all that happy with it. Two options:
- [x] Remove any dependency on EC code on Range and give up on being DRY;
- [ ] Use Range code for EC private key generation too (as in an earlier iteration of my prior PR).

In the former case, we'd be able to move Range and everything associated out into `rsa` land.

In the latter case, we'll still in some sense be making the EC keygen code more complex than it is now, but using a different constructor with explicit flags should make it fairly clear that the complex paths (3 \* 0b100 reduction and mask calculation) are not used.

Thoughts?

I am the sole copyright holder of all added code, and agree to license it per the terms in each file changed.

### DIFF
--- a/src/ec/suite_b/ops/ops.rs
+++ b/src/ec/suite_b/ops/ops.rs
@@ -12,7 +12,7 @@
 // OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
-use {c, der, error};
+use {der, error};
 use core;
 use untrusted;
 
@@ -243,10 +243,7 @@ impl CommonOps {
     fn reduced_limbs(&self, a: &[Limb; MAX_LIMBS], p: &[Limb; MAX_LIMBS])
                      -> [Limb; MAX_LIMBS] {
         let mut r = *a;
-        unsafe {
-            GFp_constant_time_limbs_reduce_once(r.as_mut_ptr(), p.as_ptr(),
-                                                self.num_limbs);
-        }
+        limbs_reduce_once_constant_time(&mut r, p);
         r
     }
 }
@@ -531,12 +528,6 @@ pub fn limbs_less_than_limbs(a: &[Limb], b: &[Limb]) -> bool {
         }
     }
     false
-}
-
-
-extern {
-    fn GFp_constant_time_limbs_reduce_once(r: *mut Limb, m: *const Limb,
-                                           num_limbs: c::size_t);
 }
 
 
--- a/src/ec/suite_b/private_key.rs
+++ b/src/ec/suite_b/private_key.rs
@@ -37,7 +37,7 @@ pub fn generate_private_key(ops: &PrivateKeyOps, rng: &rand::SecureRandom)
     // and switch to the other mechanism.
 
     let num_limbs = ops.common.num_limbs;
-    let max_exclusive =  &ops.common.n.limbs[..num_limbs];
+    let max_exclusive = &ops.common.n.limbs[..num_limbs];
 
     // XXX: The value 100 was chosen to match OpenSSL due to uncertainty of
     // what specific value would be better, but it seems bad to try 100 times.
--- a/src/ec/suite_b/private_key.rs
+++ b/src/ec/suite_b/private_key.rs
@@ -37,7 +37,7 @@ pub fn generate_private_key(ops: &PrivateKeyOps, rng: &rand::SecureRandom)
     // and switch to the other mechanism.
 
     let num_limbs = ops.common.num_limbs;
-    let range = Range::from_max_exclusive(&ops.common.n.limbs[..num_limbs]);
+    let max_exclusive =  &ops.common.n.limbs[..num_limbs];
 
     // XXX: The value 100 was chosen to match OpenSSL due to uncertainty of
     // what specific value would be better, but it seems bad to try 100 times.
@@ -70,7 +70,7 @@ pub fn generate_private_key(ops: &PrivateKeyOps, rng: &rand::SecureRandom)
         // other than being less error prone w.r.t. accidentally generating
         // zero-valued keys.
         let scalar = private_key_as_scalar_(ops, &candidate_private_key);
-        if !range.are_limbs_within(&scalar.limbs[..num_limbs]) {
+        if !is_scalar_within_range(&scalar, max_exclusive) {
             continue;
         }
 
@@ -92,10 +92,10 @@ pub fn generate_private_key(ops: &PrivateKeyOps, rng: &rand::SecureRandom)
 pub fn private_key_as_scalar(ops: &PrivateKeyOps,
                              private_key: &ec::PrivateKey) -> Scalar {
     let num_limbs = ops.common.num_limbs;
-    let range = Range::from_max_exclusive(&ops.common.n.limbs[..num_limbs]);
+    let max_exclusive = &ops.common.n.limbs[..num_limbs];
 
     let r = private_key_as_scalar_(ops, private_key);
-    assert!(range.are_limbs_within(&r.limbs[..num_limbs]));
+    assert!(is_scalar_within_range(&r, max_exclusive));
     r
 }
 
@@ -115,6 +115,14 @@ fn private_key_as_scalar_(ops: &PrivateKeyOps, private_key: &ec::PrivateKey)
         limbs[i] = limb;
     }
     Scalar::from_limbs_unchecked(&limbs)
+}
+
+// Is scalar within (0, max_exclusive)?
+fn is_scalar_within_range(scalar: &Scalar, max_exclusive: &[Limb]) -> bool {
+    let limbs = &scalar.limbs[..max_exclusive.len()];
+    let eq_zero = limbs_are_zero_constant_time(limbs);
+    let lt_bound = limbs_less_than_limbs_constant_time(limbs, max_exclusive);
+    eq_zero == LimbMask::False && lt_bound == LimbMask::True
 }
 
 pub fn public_from_private(ops: &PrivateKeyOps, public_out: &mut [u8],
--- a/src/limb.rs
+++ b/src/limb.rs
@@ -18,15 +18,13 @@
 //! Limbs ordered least-significant-limb to most-significant-limb. The bits
 //! limbs use the native endianness.
 
-use {rand, polyfill, c, core, error};
+use {polyfill, c};
 
 // XXX: Not correct for x32 ABIs.
 #[cfg(target_pointer_width = "64")] pub type Limb = u64;
 #[cfg(target_pointer_width = "32")] pub type Limb = u32;
 #[cfg(target_pointer_width = "64")] pub const LIMB_BITS: usize = 64;
 #[cfg(target_pointer_width = "32")] pub const LIMB_BITS: usize = 32;
-#[cfg(target_pointer_width = "64")] const LIMB_BITS_U32: u32 = 64;
-#[cfg(target_pointer_width = "32")] const LIMB_BITS_U32: u32 = 32;
 
 #[cfg(target_pointer_width = "64")]
 #[allow(trivial_numeric_casts)] // XXX: workaround compiler bug.
@@ -48,131 +46,27 @@ pub enum LimbMask {
 
 pub const LIMB_BYTES: usize = (LIMB_BITS + 7) / 8;
 
-/// References a positive integer range `[1..max_exclusive)`.
-/// `max_exclusive` is assumed to be public, not secret.
-pub struct Range<'a> {
-    pub max_exclusive: &'a [Limb],
-}
-
-impl<'a> Range<'a> {
-    pub fn from_max_exclusive(max_exclusive: &[Limb]) -> Range {
-        debug_assert_eq!(limbs_are_zero_constant_time(max_exclusive),
-                         LimbMask::False);
-        debug_assert!(max_exclusive[max_exclusive.len() - 1] > 0);
-
-        Range { max_exclusive: max_exclusive }
-    }
-
-    /// Are these little-endian limbs within the range?
-    ///
-    /// Checks in constant time.
-    pub fn are_limbs_within(&self, limbs: &[Limb]) -> bool {
-        assert_eq!(self.max_exclusive.len(), limbs.len());
-
-        let is_zero = limbs_are_zero_constant_time(limbs);
-        let is_lt_max = limbs_less_than_limbs_constant_time(limbs,
-                                                            self.max_exclusive);
-
-        is_zero == LimbMask::False && is_lt_max == LimbMask::True
-    }
-
-    /// Chooses a positive integer within the range and stores it into `dest`.
-    ///
-    /// This function is intended to be suitable for generating private keys.
-    //
-    // TODO: DRY-up with `ec::suite_b::private_key::generate_private_key`.
-    fn sample_into_limbs(&self, dest: &mut [Limb], rng: &rand::SecureRandom)
-                         -> Result<(), error::Unspecified> {
-        // Loosely based on [NSA Suite B Implementer's Guide to ECDSA]
-        // Appendix A.1.2, and
-        // [NSA Suite B Implementer's Guide to NIST SP 800-56A] Appendix B.2,
-        // "Key Pair Generation by Testing Candidates".
-        //
-        // [NSA Suite B Implementer's Guide to ECDSA]: doc/ecdsa.pdf.
-        // [NSA Suite B Implementer's Guide to NIST SP 800-56A]: doc/ecdh.pdf.
-
-        assert_eq!(self.max_exclusive.len(), dest.len());
-
-        let most_significant_limb =
-            self.max_exclusive[self.max_exclusive.len() - 1];
-        debug_assert!(most_significant_limb > 0);
-        let most_significant_limb_mask =
-            most_significant_limb_mask_variable_time(most_significant_limb);
-        debug_assert!(most_significant_limb_mask != 0);
-
-        // XXX: The value 100 was chosen to match OpenSSL due to uncertainty of
-        // what specific value would be better, but it seems bad to try 100
-        // times.
-        for _ in 0..100 {
-            {
-                let mut dest_as_bytes = limbs_as_bytes_mut(dest);
-                try!(rng.fill(&mut dest_as_bytes));
-            }
-            dest[self.max_exclusive.len() - 1] &= most_significant_limb_mask;
-
-            if self.are_limbs_within(&dest) {
-                return Ok(());
-            }
-        }
-
-        Err(error::Unspecified)
-    }
-}
-
-/// Returns a mask that has the same number of leading zeros as
-/// `most_significant_limb`, with all the following bits set.
-fn most_significant_limb_mask_variable_time(most_significant_limb: Limb)
-                                            -> Limb {
-    const ONE: Limb = 1;
-
-    let bits_to_mask_off = most_significant_limb.leading_zeros();
-    if bits_to_mask_off == 0 {
-        Limb::max_value()
-    } else {
-        (ONE << (LIMB_BITS_U32 - bits_to_mask_off)) - 1
-    }
-}
-
-#[allow(non_snake_case)]
-#[doc(hidden)]
-#[no_mangle]
-pub unsafe extern fn GFp_rand_mod(dest: *mut Limb, max_exclusive: *const Limb,
-                                  num_limbs: c::size_t, rng: *mut rand::RAND)
-                                  -> c::int {
-    const ERR: c::int = 0;
-    const SUCCESS: c::int = 1;
-
-    let range =
-        Range::from_max_exclusive(core::slice::from_raw_parts(max_exclusive,
-                                                              num_limbs));
-    let mut dest = core::slice::from_raw_parts_mut(dest, num_limbs);
-
-    let result = range.sample_into_limbs(&mut dest, (*rng).rng);
-    if result.is_err() {
-        return ERR;
-    }
-
-    SUCCESS
-}
-
-
 #[cfg(all(test, target_pointer_width = "64"))]
-fn limbs_as_bytes<'a>(src: &'a [Limb]) -> &'a [u8] {
+#[inline]
+pub fn limbs_as_bytes<'a>(src: &'a [Limb]) -> &'a [u8] {
     polyfill::slice::u64_as_u8(src)
 }
 
 #[cfg(target_pointer_width = "64")]
-fn limbs_as_bytes_mut<'a>(src: &'a mut [Limb]) -> &'a mut [u8] {
+#[inline]
+pub fn limbs_as_bytes_mut<'a>(src: &'a mut [Limb]) -> &'a mut [u8] {
     polyfill::slice::u64_as_u8_mut(src)
 }
 
 #[cfg(all(test, target_pointer_width = "32"))]
-fn limbs_as_bytes<'a>(src: &'a [Limb]) -> &'a [u8] {
+#[inline]
+pub fn limbs_as_bytes<'a>(src: &'a [Limb]) -> &'a [u8] {
     polyfill::slice::u32_as_u8(src)
 }
 
 #[cfg(target_pointer_width = "32")]
-fn limbs_as_bytes_mut<'a>(src: &'a mut [Limb]) -> &'a mut [u8] {
+#[inline]
+pub fn limbs_as_bytes_mut<'a>(src: &'a mut [Limb]) -> &'a mut [u8] {
     polyfill::slice::u32_as_u8_mut(src)
 }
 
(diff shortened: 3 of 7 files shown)


src/ec/suite_b/ops/ops.rs
```diff
@@ -252,13 +252,12 @@ impl CommonOps {
     // Gueron for the details. This is the case for both the field order and
     // group order for both P-256 and P-384, but it is not the case for all
     // curves. For example, it is not true for P-521.
+    //
+    // TODO(DJ): This interface is misleading: `a` is actually mutated.
```
briansmith: Are you sure? Doesn't `let mut r = *a` copy (via `Copy`) a?

src/limb.rs
```diff
@@ -51,17 +49,62 @@ pub const LIMB_BYTES: usize = (LIMB_BITS + 7) / 8;
...
+        let most_significant_limb = max_exclusive[max_exclusive.len() - 1];
+
+        debug_assert!(most_significant_limb > 0);
+
+        // TODO(DJ) Eliminate `as`
+        let leading_zeros = most_significant_limb.leading_zeros() as usize;
+        if leading_zeros == 0 {
+            // If there are no leading zeros, we'll never use the special
+            // `0b100...` behavior, because it would require us to allocate
+            // an extra limb into which to sample.
```
briansmith: This seems problematic to me. Isn't it going to be the case usually that there will be zero leading zeros, because the modulus will be 2048 bits, which is the most common length for a public RSA modulus?

Maybe we should instead generate the random value into a stack-allocated buffer that has the extra space, and then copy it over?

src/limb.rs
```diff
@@ -110,31 +146,70 @@ impl <'a> Range<'a> {
...
                 try!(rng.fill(&mut dest_as_bytes));
             }
-            dest[self.max_exclusive.len() - 1] &= most_significant_limb_mask;
+            dest[self.max_exclusive.len() - 1] &= self.sampling_most_sig_limb_mask;
 
             if self.are_limbs_within(&dest) {
                 return Ok(());
             }
+
+            if self.sampling_reduce_when_over_bound {
```
briansmith: It may be worth putting this in its own function so that (a) we can test it, and (b) to reduce the indention for the giant comment.

briansmith:
> I pulled the 0b100 check and mask calculation for Range sampling into the Range constructor, where it's easier to test and, I think, clearer. However, this means that the EC code which uses Range just for the are_limbs_within check now is unnecessarily initializing those values too. Even though it skips the complex path, I'm not all that happy with it. Two options:
> 
> 1. Remove any dependency on EC code on Range and give up on being DRY;
> 2.  Use Range code for EC private key generation too (as in an earlier iteration of my prior PR).
> In the former case, we'd be able to move Range and everything associated out into rsa land.

I prefer #1. I like the DRY principle but IMO it is more important that the ECC code be as clear and as small as possible. I eventually want to literally prove the ECC code is correct.

> I am the sole copyright holder of all added code, and agree to license it per the terms in each file changed.

No need to put this in the PR comments, just in the commit messages.

src/ec/suite_b/ops/ops.rs
djudd: You're right. I was confused and thinking of `*a` as if it was taking a raw pointer, but of course it's not.

src/limb.rs
djudd: I worry breaking the series of `if self.are_limbs_within ... return Ok` checks into separate functions would make the pattern harder to follow, and I'm not sure how much sense tests of this part in isolation would make, but i'll poke at it.

Regardless, the indentation on the comment is definitely annoying. I wonder if it makes sense to move most of it into a sort of footnote?

src/limb.rs
djudd: Good point. I'll microbenchmark again with copying before committing to it. It especially sucks because either we're doing the extra copying even when the number isn't of the form `0b100`, or we're complicating the code even further. But maybe the cost will turn out to be negligible.

djudd: Hrm, the bigger problem is actually that we don't know the length of `max_exclusive` at compile time, and AFIACT Rust doesn't support dynamically-sized stack-allocated buffers, so we'd be talking about heap allocation, which we've tried pretty hard to avoid. Now that this will be used in RSA code exclusively, maybe heap usage is ok?

We only need one extra bit so it feels like it should be possible to just generate that into a single Limb or something, but that's going to make our bignum arithmetic more complicated and possibly no longer constant-time.

djudd:
@briansmith I believe this is ready for re-review.

src/ec/suite_b/ops/ops.rs
```diff
@@ -14,7 +14,7 @@
 
 #![allow(unsafe_code)]
 
-use {c, der, error};
+use {der, error};
```
briansmith: Awesome!

build.rs
```diff
@@ -19,7 +19,6 @@
     const_err,
     dead_code,
     deprecated,
-    drop_with_repr_extern,
```
briansmith: This commit can be removed, because I fixed this already. Thanks though!

src/ec/suite_b/private_key.rs
```diff
@@ -119,6 +119,13 @@ fn private_key_as_scalar_(ops: &PrivateKeyOps, private_key: &ec::PrivateKey)
     Scalar::from_limbs_unchecked(&limbs)
 }
 
+// Are limbs within (0, max_exclusive)?
+fn are_limbs_within_range(limbs: &[Limb], max_exclusive: &[Limb]) -> bool {
```
briansmith: Either this should take a `Scalar` or `PrivateKey` argument or we should find a way to share it between the RSA and ECC code, since I think it is exactly the same, even if we don't share the `Range` stuff.

src/limb.rs
```diff
@@ -48,133 +46,23 @@ pub enum LimbMask {
...
-        return ERR;
-    }
-
-    SUCCESS
-}
-
-
 #[cfg(all(test, target_pointer_width = "64"))]
-fn limbs_as_bytes<'a>(src: &'a [Limb]) -> &'a [u8] {
+pub fn limbs_as_bytes<'a>(src: &'a [Limb]) -> &'a [u8] {
```
briansmith: I suspect these small functions should be annotated `#[inline]` to make sure they get inlined cross-module.

src/limb.rs
```diff
@@ -196,181 +84,22 @@ pub fn limbs_are_zero_constant_time(limbs: &[Limb]) -> LimbMask {
     }
 }
 
+/// Equivalent to `if (r >= m) { r -= m; }`
+#[allow(unsafe_code)]
```
briansmith: Ditto. This should be inline.

src/rsa/rsa.rs
```diff
@@ -26,6 +26,7 @@ use std;
 
 use untrusted;
 
+pub const RSA_KEY_MAX_BITS: usize = 4096;
```
briansmith: I think this should be named so that it is clear it refers to the public key modulus, e.g. `RSA_PUBLIC_KEY_MODULUS_BITS_MAX`.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,644 @@
...
+    let starts_with_0b100 = starts_with_0b100_variable_time(max_exclusive);
+    let leading_zeros = max_exclusive[max_exclusive.len() - 1].leading_zeros();
+
+    if starts_with_0b100 && leading_zeros == 0 {
+        SamplingParams {
+            most_sig_limb_mask: 1, // effectively a carry into a new, more-significant limb
+            reduce_when_over_bound: true,
+            extend_limbs_by_one: true,
+        }
+    } else if starts_with_0b100 {
```
briansmith: I feel like we don't need to treat this `else if` case differently than the `else` case below. It is very unlikely that the modulus isn't going to be a multiple of 64-bits in any case where the application cares about performance. This means we can combine `reduce_when_over_bound` and `extend_limbs_by_one` into a single variable.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,644 @@
...
+    let buf_needed = dest.len() + 1;
+    debug_assert_eq!(buf_needed, max_exclusive.len() + 1); // equivalent definition
+
+    assert!(buf_needed <= BUF_SIZE);
+
+    let mut tmp_max: [Limb; BUF_SIZE] = [0; BUF_SIZE];
+    let mut tmp_dest: [Limb; BUF_SIZE] = [0; BUF_SIZE];
+
+    tmp_dest[..dest.len()].clone_from_slice(&dest);
+    tmp_max[..max_exclusive.len()].clone_from_slice(&max_exclusive);
```
briansmith: Let's use `copy_from_slice`.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,644 @@
...
+/// and stores it into `dest`.
+fn set_to_rand_mod(dest: &mut [Limb], max_exclusive: &[Limb], rng: &rand::SecureRandom)
+            -> Result<(), error::Unspecified> {
+
+    debug_assert_eq!(dest.len(), max_exclusive.len());
+
+    let sampling_params = select_sampling_params(max_exclusive);
+
+    if sampling_params.extend_limbs_by_one {
+        with_extended_buffers(dest, max_exclusive, |ext_dest, ext_max| {
```
briansmith: I suggest that you always just copy the stuff into the extended buffers and then just inline `with_extended_buffers` into this function. The most common case is a 2048-bit modulus that starts with `0b100.....`, right? If that case is too slow then we'd need to fix things. Conversely, if that case isn't to slow then there's no point in optimizing anything about the other case. In general, I feel like this is pretty over-engineered. I suggest trying to write it the simplest possible way that works for the 2048-bit case where the public modulus starts with `0b100...`, and then do the simplest possible (if suboptimal) thing for all other cases.

src/ec/suite_b/private_key.rs
djudd: I'll make it take a Scalar for now. I think there's a clean way to share the random generation loop without adding any complication here once specialization gets stabilized, but till then, I don't think it's worth trying too hard for a 3-line method.

src/rsa/blinding.rs
djudd: I'm not particularly familiar with real-world RSA use, so I have to take your word for it that the common case has a number of bits which is a multiple of LIMB_BITS, and more often than not starts with `0b100`. Given that, I'm happy to optimize for that case and simplify this logic. I'm not a fan of that giant `starts_with_0b100_variable_time` function either!

But, I actually think pulling `with_extended_buffers` out into its own function here makes the code easier to read even if that function gets called every time, since the buffer copying is pretty verbose and the noise could hide what we're actually trying to do.

briansmith: > I'm not particularly familiar with real-world RSA use, so I have to take your word for it that the common case has a number of bits which is a multiple of LIMB_BITS, and more often than not starts with 0b100.

I would say they probably don't start with 0b100 more often than any other 0b1xx combination, but the savings of any optimization we do for the non-0b100 case would be so minimal that it isn't worth any extra code.

By far the most common RSA keys are 2048-bit modulus (the minimum allowed by CAB/Forum requirements). The public modulus is usually (always, in *ring*) the product of two 1024-bit integers:

```python
>>> min = (1 << 1023) + 1
>>> max = (1 << 1024) - 1
>>> avg = (min + max) / 2
>>> min.bit_length()
1024
>>> max.bit_length()
1024
>>> avg.bit_length()
1024
>>> (min**2).bit_length()
2047
>>> (max**2).bit_length()
2048
>>> (avg**2).bit_length()
2048
>>> (((9*min + 7*max)/16)**2).bit_length()
2048
>>> (((19*min + 13*max)/32)**2).bit_length()
2047
```

So, it might be worthwhile to also do the optimization for the case where the high limb starts with `0b0100...`, in addition to the case where it starts with `0b100...`. The important thing is that you don't have to worry about the case where the `0b100` part is split across two limbs.

briansmith: > But, I actually think pulling with_extended_buffers out into its own function here makes the code easier to read even if that function gets called every time, since the buffer copying is pretty verbose and the noise could hide what we're actually trying to do.

OK.

djudd: If we always copy into the extended buffers, that actually complicates our Range code, because it requires us to decide inside Range whether we are going to apply the mask to the most-significant limb or the next-most-significant limb. (Or pass in two masks, or something.)

Given that, I slightly prefer leaving the branching logic here, if only because that's how I already wrote it.

djudd:
Ready for re-review, unless you'd like me to squash the new (/ all) commits first. 

I'll do that before I merge in any case, and add the licensing agreement.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+                return Ok(());
+            }
+
+            if self.sampling_params.reduce_when_over_bound {
+                // `dest` is not in (0, max) but maybe we can fix that.
+                // (See above for explanation of why this is safe.)
+
+                limbs_reduce_once_constant_time(dest, self.max_exclusive);
+                if self.are_limbs_within(&dest) {
+                    // `dest` was in [max, 2*max)
```
briansmith: This should be `(max, 2*max)`, right? If the value were `max` then `are_limbs_within(&dest)` would return false because `max` - `max` == 0, which isn't allowed.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+                    return Ok(());
+                }
+
+                limbs_reduce_once_constant_time(dest, self.max_exclusive);
+                if self.are_limbs_within(&dest) {
+                    // `dest` was in [2*max, 3*max)
```
briansmith: `(2*max, 3*max)`, right?

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+                    return Ok(());
+                }
+
+                // `dest` was in [3*max, 2**(n+1)) or zero
```
briansmith: It is either in that subrange, or it was zero, or it was `max` or `2*max`, right?

src/rsa/blinding.rs
briansmith: Should we change `are_limbs_with()` to work like this?:

```
if limbs_less_than_limbs_constant_time(limbs, self.max_exclusive) != LimbMask::True {
    return false;
}
return limbs_are_zero_constant_time(limbs) == LimbMask::False;
```

In particular this would reduce the number of calls to `limbs_are_zero_constant_time` for the cases where we have to do one or more reductions.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+// TODO(djudd) Part of this code can potentially be pulled back into
+// `super::limb` and shared with EC key generation, without unnecessarily
+// complicating that, once specialization is stabilized.
+struct Range<'a> {
+    max_exclusive: &'a [Limb],
+    sampling_params: &'a SamplingParams,
+}
+
+impl <'a> Range<'a> {
+
```
briansmith: Remove extra blank line.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+    SUCCESS
+}
+
+/// Chooses a positive integer less than `max_exclusive` uniformly at random
+/// and stores it into `dest`.
+fn set_to_rand_mod(dest: &mut [Limb], max_exclusive: &[Limb], rng: &rand::SecureRandom)
+            -> Result<(), error::Unspecified> {
+
```
briansmith: Remove extra blank line.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+// OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+//! RSA blinding parameter generation.
+
+use {c, core, rand, error, rsa};
+use limb::*;
+
+const RSA_KEY_MAX_LIMBS: usize = rsa::RSA_PUBLIC_KEY_MODULUS_BITS_MAX / LIMB_BITS;
```
briansmith: When re-reviewing this, I realized that this should be defined next to `RSA_PUBLIC_KEY_MODULUS_BITS_MAX` and we should add a comment explaining that we assume that `RSA_PUBLIC_KEY_MODULUS_BITS_MAX` is always a multiple of `LIMB_BITS`.

Preferably, use the rounding calculation:

```
(rsa::RSA_PUBLIC_KEY_MODULUS_BITS_MAX + LIMB_BITS - 1) / LIMB_BITS.

File: src/rsa/blinding.rs (hunk 15 above)
Comment (briansmith) [inline src/rsa/blinding.rs]: I suggest that you append ". Continue looping to generate a new random value." Especially because the period is missing, it looks like there might be some code unintentionally missing here.

File: src/rsa/blinding.rs [hunk 19]
```diff
@@ -0,0 +1,585 @@
...
+
+                // `dest` was in [3*max, 2**(n+1)) or zero
+            }
+        }
+
+        Err(error::Unspecified)
+    }
+}
+
+#[allow(unsafe_code)]
```
briansmith: You can drop this `#[allow(unsafe_code)]` now.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+
+        let limbs = &[2];
+        let range = Range { max_exclusive: limbs, sampling_params: &params };
+        assert!(!range.are_limbs_within(&[0]));
+        assert!(range.are_limbs_within(&[1]));
+        assert!(!range.are_limbs_within(&[2]));
+    }
+
+    #[test]
+    #[allow(unused_results)]
```
briansmith: Let's not use `allow(unused_results)`, even in tests. Instead we can use the `let _ = ` pattern.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use {core,rand};
+    use limb::*;
+
+    #[test]
+    fn test_select_sampling_params() {
```
briansmith: All of these test cases are one limb long, so they are not verifying that we're looking at the high limb instead of, say, the low limb.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+                }
+            }
+        }
+
+        for i in 0..(LIMB_BITS - 2) {
+            let l = 1 << i;
+
+            assert_normal!(i, l);
+            assert_normal!(i, l | 1);
+            assert_normal!(i, l | l >> 3);
```
briansmith: I don't understand the significance of these test cases, especially the last one. Add some comments to show why these values are important to test.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+                    let p = select_sampling_params(&x[..]);
+                    let mask = Limb::max_value() >> (LIMB_BITS - 1 - $i);
+                    assert!(!p.extend_limbs_by_one);
+                    assert!(!p.reduce_when_over_bound);
+                    assert_eq!(mask, p.most_sig_limb_mask);
+                }
+            }
+        }
+
+        for i in 0..(LIMB_BITS - 2) {
```
briansmith: Add a comment explaining why you stoped at `LIMB_BITS - 2`. The reason is that you want to ensure that the value never starts with `0b100` or `0b0100`.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+            assert_normal!(i, l | 1);
+            assert_normal!(i, l | l >> 3);
+        }
+
+        for i in 0..LIMB_BITS {
+            let l = 1 << i;
+
+            assert_normal!(i, l | l >> 1);
+            assert_normal!(i, l | l >> 2);
+            assert_normal!(i, Limb::max_value() >> (LIMB_BITS - 1 - i));
```
briansmith: Again, this is too clever for me, so we should add some comments about what exactly we're testing here.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+}
+
+#[cfg(feature = "internal_benches")]
+mod bench {
+    use {bench, rand};
+    use limb::*;
+    use super::{RSA_KEY_MAX_LIMBS, Range, SamplingParams, with_extended_buffers};
+
+    #[bench]
+    fn bench_sample_into_limbs_simple(b: &mut bench::Bencher) {
```
briansmith: Some comments about the purpose of this benchmark would be useful. In particular, why does the benchmark fill in its own `SamplingParams` and `Range` instead of having the real library code calculate them? What is meant by "simple" in the function name?

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+            sampling_params: &params,
+        };
+
+        b.iter(|| {
+            range.sample_into_limbs(&mut dest, &rng)
+        });
+    }
+
+    #[bench]
+    fn bench_sample_into_limbs_reduce_no_copy(b: &mut bench::Bencher) {
```
briansmith: Again, please add a comment about the purpose of this benchmark.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+#[cfg(feature = "internal_benches")]
+mod bench {
+    use {bench, rand};
+    use limb::*;
+    use super::{RSA_KEY_MAX_LIMBS, Range, SamplingParams, with_extended_buffers};
+
+    #[bench]
+    fn bench_sample_into_limbs_simple(b: &mut bench::Bencher) {
+        let mut max: [Limb; RSA_KEY_MAX_LIMBS] = [0; RSA_KEY_MAX_LIMBS];
+        max[RSA_KEY_MAX_LIMBS - 1] = 1 << (LIMB_BITS - 1);
```
briansmith: Add a comment such as "construct a value such that the high limb starts with `0b100...`

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+
+        b.iter(|| {
+            range.sample_into_limbs(&mut dest, &rng)
+        });
+    }
+
+    #[bench]
+    fn bench_sample_into_limbs_reduce_no_copy(b: &mut bench::Bencher) {
+        let mut max: [Limb; RSA_KEY_MAX_LIMBS] = [0; RSA_KEY_MAX_LIMBS];
+        max[RSA_KEY_MAX_LIMBS - 1] = 1 << (LIMB_BITS - 2);
```
briansmith: Please add a comment and an assertion such as `assert_eq!(max[RSA_KEY_MAX_LIMBS - 1] >> LIMB_BITS - 3, 0b100)` to indicate the expected form of `max`.

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+
+        b.iter(|| {
+            range.sample_into_limbs(&mut dest, &rng)
+        });
+    }
+
+    #[bench]
+    fn bench_sample_into_limbs_reduce_with_copy(b: &mut bench::Bencher) {
+        let mut max: [Limb; RSA_KEY_MAX_LIMBS] = [0; RSA_KEY_MAX_LIMBS];
+        max[RSA_KEY_MAX_LIMBS - 1] = 1 << (LIMB_BITS - 1);
```
briansmith: Please add a comment and an assertion such as `assert_eq!(max[RSA_KEY_MAX_LIMBS - 1] >> LIMB_BITS - 3, 0b100)` to indicate the expected form of `max`.

briansmith:
> Rust code should have input parameters first, then rng, then output parameters.

Just to follow-up on this: I went and looked and we're already not as consistent as I was hoping. In particular, it's more the case that we put the `modulus`/`max` parameter and `rng` parameter last, no matter what. So, it's fine to continue to do that in this PR.

src/rsa/blinding.rs
djudd: Yes, good catch

src/rsa/blinding.rs
djudd: Makes sense. I'd actually already gone back and forth on whether expanding this comment was redundant or not

src/rsa/blinding.rs
```diff
@@ -0,0 +1,585 @@
...
+    if result.is_err() {
+        return ERR;
+    }
+
+    SUCCESS
+}
+
+/// Chooses a positive integer less than `max_exclusive` uniformly at random
+/// and stores it into `dest`.
+fn set_to_rand_mod(dest: &mut [Limb], max_exclusive: &[Limb], rng: &rand::SecureRandom)
```
djudd: Fixed by cherry-picking in your rewrite of the function.

djudd:
I'll squash the changes before requesting re-review, and I cherry-picked in your commit (I hadn't thought of your `extra_limb` solution to the branch, or hadn't understood it if you had suggested it.)

Re line length comments & some others, should I be running `rustfmt` on the files I'm touching? I recall a comment on my first PR that said this project follows BoringSSL style at least in some respects, not `rustfmt` style, but I've seen some more references to `rustfmt` going by recently.

src/rsa/blinding.rs
djudd: The last is testing that we don't treat `0b001001`, `0b0001001`, etc like we do `0b1001` and `0b01001`. But every case I pick from the wide set of non-`0b100` numbers feels a bit arbitrary to me; if you find these redundant, I'm happy to remove them.

src/rsa/blinding.rs
djudd: We're testing `0b11...`, `0b101`..., and `0b111`..., any of which, per above, I'm happy to remove if you find them redundant, since I don't have any particular justification. Would it make more sense to just test `0b11`, `0b101`, and `0b001...` and drop the last case here and last two cases above?

djudd:
Ready for re-review, modulo possible `rustfmt`. (I went ahead and removed some of the probably-redundant test cases that I'd asked about.)

briansmith:
> We're testing 0b11..., 0b101..., and 0b111..., any of which, per above, I'm happy to remove if you find them redundant, since I don't have any particular justification. Would it make more sense to just test 0b11, 0b101, and 0b001... and drop the last case here and last two cases above?

No, it's not that I want you to remove the tests. More, it would be nice to find a way to make the tests more obvious, by naming some of the calculated values with clear names and/or adding comments.

briansmith:
> Re line length comments & some others, should I be running rustfmt on the files I'm touching? I recall a comment on my first PR that said this project follows BoringSSL style at least in some respects, not rustfmt style, but I've seen some more references to rustfmt going by recently.

I have a rustfmt configuration file that makes things more like the BoringSSL style. Unfortunately, rustfmt likes to put function arguments each on their own line, so I end up having to reformat some things after running rustfmt. So, it isn't quite 100% formatted according to rustfmt but it might be helpful to use it.

djudd:
Bump - AFAIK this is ready (apologies if I missed some comment).

briansmith:
Yes, I've already reviewed it earlier this week and I'm planning to push it with minor changes sometime this weekend.

djudd:
Great. No hurry, just wasn't sure of the status. Thanks!

briansmith:
Please see the tweaks at https://github.com/briansmith/ring/commits/rsa-blinding-0b100. The first commit I stacked on top of yours is just reordering your code so that the code is in top-down(-ish) order. The second commit is just a bunch of tweaks. LMK if I created any problem.

djudd:
Aside from a typo in a comment, looks good. Thanks!

briansmith:
Thanks for doing this. This landed with some tweaks as f014d94af646d7d3cf3427278ba111c7be191513. (I corrected the typo you pointed out.)

I also filed #353 to improve the performance further.

PR #288, opened 2016-08-27
Added rust tests for ChaCha20_ctr32. Fixed #202

### LINKED ISSUE #202: Replace crypto/chacha/chacha_test.cc with Rust equivalent

newpavlov:
Fixed #202


newpavlov:
Haven't noticed samscott89's pull request.

briansmith:
Thanks for the effort!

PR #292, opened 2016-09-03
Fixed #141 -- replace the fill_from_slice polyfill with copy_from_slice

### LINKED ISSUE #141: Replace use of `ring::polyfill::slice::fill_from_slice` with `copy_from_slice`
Depends on `copy_from_slice` becoming stable. See https://github.com/rust-lang/rust/issues/31755.

alex:


briansmith:
Thanks for the PR.

Please follow the directions at https://github.com/briansmith/ring#contributing regarding the commit message.

Questions about the performance impact:

1. Did you do an A/B run of the benchmarks for *ring* using https://github.com/briansmith/crypto-bench? 
2. Did you measure any difference in debug builds when doing `cargo test --features=rsa_signing`?
3. Did you measure any difference in debug builds when doing `cargo test --release --features=rsa_signing`?

I'll watch the Travis CI to see if it causes any tests to time out. In the past, tests timing out (in debug builds, IIRC) was the reason we couldn't use `clone_from_slice()`, so we need to be sure that `copy_from_slice()` doesn't have the same problems.

alex:
Amended the commit to include the required text. I have not performed any benchmarking/measurement yet.

briansmith:
> Amended the commit to include the required text. I have not performed any benchmarking/measurement yet.

Thanks! I didn't observe any (bad) performance implications of this and I double-checked that `copy_from_slice` is implemented exactly like our `polyfill::fill_from_slice` except for `#[inline]` vs `#[inline(always)]`.

Committed in cbfd9fb72c8b730fe72b37a5435eacd15f5db24c.

briansmith:
I forgot to mention that I reformatted some of the changed lines using rustfmt before committing.

PR #303, opened 2016-09-15
Prehashed signature verification (closes #253).

### LINKED ISSUE #253: Add API for computing RSA/ECDSA signatures from a pre-computed digest
This is necessary to support client authentication without having to store all handshake messages in TLS 1.2 - see discussion in https://github.com/ctz/rustls/issues/3.

briansmith:
Closes #253.

### DIFF
--- a/src/digest/digest.rs
+++ b/src/digest/digest.rs
@@ -299,6 +299,13 @@ pub struct Algorithm {
     initial_state: [u64; MAX_CHAINING_LEN / 8],
 }
 
+impl PartialEq for Algorithm {
+    #[inline]
+    fn eq(&self, other: &Algorithm) -> bool {
+        self.initial_state == other.initial_state
+    }
+}
+
 impl core::fmt::Debug for Algorithm {
     fn fmt(&self, fmt: &mut core::fmt::Formatter) -> core::fmt::Result {
         // This would have to change if/when we add other algorithms with the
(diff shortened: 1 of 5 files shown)


briansmith:
The current version is a WIP.

djc:
It looks nice!

Why can't we do this for EdDSA, again? I was working on ed25519 signing/verification for rustup recently, which could benefit from this.

briansmith:
@djc, Sorry, when I tried to reply to your comment, I ended up editing your comment instead of adding my own reply! I tried to put your comment back the way it was and here's the reply, seperately:

> Why can't we do this for EdDSA, again?

Read how an EdDSA signature is constructed. It needs to digest  the public key as a prefix to the data. If the data is given as a digest then it can't insert the public key ahead of the data.

> I was working on ed25519 signing/verification for rustup recently, which could benefit from this.

Cool. Though, even for RSA, rustup doesn't need prehashing if it uses mmap, and that's what I'd recommend it use. Secondly, if we don't want to require applications to use mmap to stream a large item that won't fit in a single slice, we should define a `std::io::Read`-based interface that can be used for those cases.

djc:
A `std::io::Read`-based interface sounds nice to me as a future enhancement.

briansmith:
> A std::io::Read-based interface sounds nice to me as a future enhancement.

We seem like we'd need something like that, but I'd like to find a solution that doesn't depend on std::io. Maybe using an `Iterator<&[u8]>` or maybe defining a trait that is like `std::io::Read` but which doesn't depend on libstd; presumably we would then `impl <ThatTrait> for std::io::Read` when libstd is available.

Regardless, I originally started this PR because I thought it was needed to implement SSH. However, I was wrong, so I'm going to close this, as there doesn't seem to be any immediate need for it.

Issue #143, opened 2016-03-09
Remove last usage of <stdarg> and vararg stuff


briansmith:
Right now, AFAICT, the only use of #include <stdarg.h> is in file_test.cc (testing code), and it looks like it would be easy to switch to a non-vararg function.

When we do that, we should add documentation to STYLE.md that the definition or use of vararg functions is disallowed.

Besides making static analysis easier, this would also make a good dent in the the minimum required C/C++ standard library subset.

briansmith:
See http://blog.regehr.org/archives/1303.

frewsxcv:
What do you propose we replace the [last vararg usage](https://github.com/briansmith/ring/blob/c5fc35ff04ca95fef064cbd615c4db95cb2b03b0/crypto/test/file_test.cc#L132) with?

briansmith:
I don't really care. It's in test code. IIRC, there's never more than one extra arg passed to that function, so it can probably be changed to just take another arg as std::string& or const char* or something. And/or we can overload the function with a version that takes a number and a version that takes a string. Or, whatever.

briansmith:
Fixed with #306.

PR #305, opened 2016-09-24
Remove last usage of <stdarg.h> and vararg by reimplement test printer

yurizhykin:
Fix for #143 . Assumed that it's enough for test-printer function to accept only up to 3 string parts.


yurizhykin:
Accidentally did reset for 2 commits and commited one of @briansmith 's, so will try one more time:)

Issue #246, opened 2016-07-09
Replace RSA verification function `GFp_rsa_public_decrypt` with Rust code


briansmith:
After the ECC implementation is improved, RSA signing and RSA verification will be the only two remaining operations that use `BIGNUM`. In particular, these will be the only two operations that use the heap. We would like to enable RSA signature verification in environemnts where there is no heap available too.

We should just write a new modular exponentiation function that always uses square-and-multiply and which avoids using `BIGNUM`. Note that verification doesn't need to be constant time. We can use `bn_mod_mul` for the modular multiplication.

Typically, the public exponent is `0x10001`. Sometimes it is 3 (0b11). It is rarely any other value.  Note that the typical exponents `0x10001` and `0b11` have hamming weight of 2. When the hamming weight is 2, the optimal algorithm is the simplest one: square and multiply.

Even if the exponent has a large hamming weight, the largest hamming weight possible is 33, because the RSA verification logic already restricts the public exponent to a maximum of 33 bits. This maximum was also contributed to BoringSSL so that Google Chrome could field-test that a 33-bit maximum is web-compatible. 

`BN_window_bits_for_exponent_size` already uses the basic square-and-multiply algorithm for exponents that are less than 24 bits; i.e. almost all the time. It only uses window-based exponentiation for exponents of 24 bits or more, and even then it only uses 3-bit windows.

Thus, the difference in performance in the best case should be minimal, and the difference in the worst case, which is extremely rare, will be only a little bit worse.

A good way to approach this project is to:
- [ ] Replace `BN_mod_exp_mont` with Rust code takes its input and output parameters as `&[Limb]` instead of `BIGNUM` and which uses `bn_mod_mul` for both squaring and multiplying, using the simple square-and-multiple algorithm. `Limb` should be moved from the `ring::ec` submodule to a new `ring::bn` module.
- [ ] Replace `GFp_rsa_public_decrypt` with Rust code that uses `BN_MONT_CTX_new()`/`BN_MONT_CTX_set()`/`BN_MOD_CTX_free()` and the new modular exponentiation function.
- [ ] Create a new variable-time modular inversion function that operates on `[Limb]`s (or, if in C, `BN_ULONG[]`) instead of `BIGNUM`, which will replace the variable-time `BN_mod_inverse_ex`.
[...]

PR #307, opened 2016-09-26
Port `rsa_public_decrypt`

ranweiler:
Toward #246.

### DIFF
--- a/include/openssl/bn.h
+++ b/include/openssl/bn.h
@@ -452,6 +452,7 @@ OPENSSL_EXPORT int GFp_BN_mod_exp_mont_consttime(BIGNUM *rr, const BIGNUM *a,
 
 /* Private functions */
 
+/* Needs to be kept in sync with `struct BIGNUM` (in `src/bn.rs`) */
 struct bignum_st {
   BN_ULONG *d; /* Pointer to an array of 'BN_BITS2' bit chunks in little-endian
                   order. */
--- a/src/rsa/verification.rs
+++ b/src/rsa/verification.rs
@@ -71,9 +71,9 @@ fn rsa_public_decrypt(out: &mut [u8], public_key_n: &bn::PositiveInteger,
                       public_key_e: &bn::PositiveInteger, ciphertext: &[u8],
                       min_bits: usize, max_bits: usize)
                       -> Result<(), error::Unspecified> {
-    // We define this closure as a translation of `goto`. We still want
-    // to use an early-return code style, but we also need to manually
-    // free any |BIGNUM| heap-allocated using a foreign call.
+    // We define this closure as a translation of `goto`. We want to use
+    // an early-return code style, but we also need to manually free any
+    // heap-allocated |BIGNUM| substructures using a foreign call.
     let mut decrypt = |f: *mut bn::BIGNUM, result: *mut bn::BIGNUM| {
         let rsa_size: usize = unsafe {
             bn::GFp_BN_num_bytes(public_key_n.as_ref())
@@ -120,21 +120,29 @@ fn rsa_public_decrypt(out: &mut [u8], public_key_n: &bn::PositiveInteger,
         Ok(())
     };
 
-    let f: *mut bn::BIGNUM = unsafe { bn::GFp_BN_new() };
-    if f.is_null() {
-        return Err(error::Unspecified);
-    }
+    let mut f = bn::BIGNUM {
+        d: core::ptr::null_mut(),
+        top: 0,
+        dmax: 0,
+        neg: 0,
+        flags: 0,
+    };
+    unsafe { bn::GFp_BN_init(&mut f) };
 
-    let result: *mut bn::BIGNUM = unsafe { bn::GFp_BN_new() };
-    if result.is_null() {
-        return Err(error::Unspecified);
-    }
+    let mut result = bn::BIGNUM {
+        d: core::ptr::null_mut(),
+        top: 0,
+        dmax: 0,
+        neg: 0,
+        flags: 0,
+    };
+    unsafe { bn::GFp_BN_init(&mut result) };
 
-    let res = decrypt(f, result);
+    let res = decrypt(&mut f, &mut result);
 
     unsafe {
-        bn::GFp_BN_free(result);
-        bn::GFp_BN_free(f);
+        bn::GFp_BN_free(&mut result);
+        bn::GFp_BN_free(&mut f);
     }
 
     res
--- a/src/rsa/verification.rs
+++ b/src/rsa/verification.rs
@@ -14,8 +14,7 @@
 
 /// RSA PKCS#1 1.5 signatures.
 
-use {bssl, c, error, private, signature};
-use bn::{BIGNUM, PositiveInteger};
+use {bn, bssl, c, core, error, private, signature};
 use super::{RSAParameters, parse_public_key};
 use untrusted;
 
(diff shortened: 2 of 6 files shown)


Ms2ger:
Don't forget to copy the license header into the new file.

ranweiler:
@briansmith, finally found some time for this, ready for a first look.

I'd like to point out that, in https://github.com/briansmith/ring/pull/307/commits/5cfc9de9cbf8125e97e6dd2e2ee64b860b68bddb, I replace two stack allocations (of `f` and `result`) with heap allocations. I revert this in https://github.com/briansmith/ring/pull/307/commits/69137656f7bd11967711acb6cad53641da458f39, at the cost of elaborating the fields of the Rust `BIGNUM` type. Later, in https://github.com/briansmith/ring/pull/307/commits/fc20d99f342b56a39fdd1142431062ceb0ae849b, I add a ctor to clean this up, and finally even tried moving the `GFp_BN_init` call into the ctor in https://github.com/briansmith/ring/pull/307/commits/400c2cbe25eb4c8860805dbdb34931000a692e7e.

In a sense, as we step through those commits, I'm sketching a solution where we create another `BIGNUM` wrapper named `Integer`, which is like `PositiveInteger`, but without constraints on its value. In this way, we could hide both the `GFp_BN_init` and `GFp_BN_free` calls behind the ctor and a `Drop` impl, which we can't give the `repr(C)`-annotated `BIGNUM`. We may also prefer to always heap-allocate the `BIGNUM` using `GFp_BN_new` in the Rust ctor, since heap-allocation must always occur for this type anyway (via `_init`). This would let us remove the new struct-syncing burden.

I think we _shouldn't_ have the foreign call in `BIGNUM::new` unless we can also implement `Drop`, but I wanted to talk this out with you before going in another direction. What do you think? Long-term, we still want to allow RSA without a heap, right?

src/rsa/verification.rs
```diff
@@ -98,26 +177,16 @@ pub fn verify_rsa(params: &RSAParameters,
         return Err(error::Unspecified);
     }
 
-    let n = try!(PositiveInteger::from_be_bytes(n));
-    let e = try!(PositiveInteger::from_be_bytes(e));
+    let n = try!(bn::PositiveInteger::from_be_bytes(n));
```
djc: These actually belong in the other commit, right?

ranweiler: Those lines definitely _could've_ been in the other commit.

Before my first commit, `PositiveInteger` was used unqualified everywhere. I didn't want to change that in _that_ commit because I was trying to make it clear I was just moving existing code over into a new module.

It may just be better to leave the `bn` _types_ as unqualified imports. What do you think?

ranweiler:
> I think this could better be folded into the previous commit.
> This probably also makes more sense to squash into previous.

Sorry @djc, I can't find the line # context for the above comments. Whatever the context was, I likely agree with you— I tried to organize my commits for easy review and _individual_ acceptance/rejection, with the expectation that a bunch of squashing will occur at the end.

djc:
Yeah, I guess I still have to get used to GitHub's new review stuff. Anyway, I think your last three commits (where you have a Rust BIGNUM type) would be easier to review if squashed together.

src/rsa/verification.rs
```diff
@@ -61,6 +60,86 @@ rsa_pkcs1!(RSA_PKCS1_3072_8192_SHA384, 3072, &super::RSA_PKCS1_SHA384,
...
+                                          e: *const bn::BIGNUM,
+                                          min_bits: c::size_t,
+                                          max_bits: c::size_t) -> c::int;
+}
+
+fn rsa_public_decrypt(out: &mut [u8], public_key_n: &bn::PositiveInteger,
+                      public_key_e: &bn::PositiveInteger, ciphertext: &[u8],
+                      min_bits: usize, max_bits: usize)
+                      -> Result<(), error::Unspecified> {
+    // We define this closure as a translation of `goto`. We still want
```
briansmith: My main worry with starting with a literal transliteration of the C code into Rust is that we'd be stuck with needing to use the license that the C code uses. So, even though your approach here is very logical in showing that the Rust code is line-for-line the same the C code, that's actually undesirable. 

Of course, at the end of the day we need to do basically the same thing so there's going to be some similarity, but I'd rather get rid of this intermediate state completely, so that the licensing issue doesn't loom over us.

We know that we need to do the following things:
* Verify that the public modulus and exponent are valid.
* Check that the input and output lengths are exactly equal to the length in bytes of the public modulus.
* Verify that the input, when interpreted as a big-endian-encoded number, is less than the public modulus.
* Do the modular exponentiation.
* Encode the result of the exponentiation in big-endian.

For some of these steps, we already have Rust code available. For example we already have `PositiveInteger::from_be_bytes` so we should be able to already avoid using `GFp_BN_new()` and `GFp_BN_bin2bn()` directly. So, I suggest that you first write the Rust code for the function from scratch such that its body doesn't use any `unsafe` directly. If there is some function that we don't have Rust code for, like the modular exponentiation, then we can create a function, e.g. `fn mod_exp_vartime(in_out: &mut PositiveInteger, exponent: u64, modulus: &PositiveInteger) -> Result<(), error::Unspecified`, that uses `unsafe` under the hood to call one of the `GFp_` functions.

WDYT?

src/rsa/verification.rs
```diff
@@ -61,6 +60,86 @@ rsa_pkcs1!(RSA_PKCS1_3072_8192_SHA384, 3072, &super::RSA_PKCS1_SHA384,
...
+extern {
+    fn GFp_rsa_check_modulus_and_exponent(n: *const bn::BIGNUM,
+                                          e: *const bn::BIGNUM,
+                                          min_bits: c::size_t,
+                                          max_bits: c::size_t) -> c::int;
+}
+
+fn rsa_public_decrypt(out: &mut [u8], public_key_n: &bn::PositiveInteger,
+                      public_key_e: &bn::PositiveInteger, ciphertext: &[u8],
+                      min_bits: usize, max_bits: usize)
```
briansmith: I think we can lose the `rsa_` prefix and in fact I think, especially if we take the approach outlined below, that we should probably just inline this whole function into `ring::rsa::verification::verify_rsa()`.

include/openssl/bn.h
```diff
@@ -452,6 +452,7 @@ OPENSSL_EXPORT int GFp_BN_mod_exp_mont_consttime(BIGNUM *rr, const BIGNUM *a,
 
 /* Private functions */
 
+/* Needs to be kept in sync with `struct BIGNUM` (in `src/bn.rs`) */
```
briansmith: Your plan for creating an `Integer` that isn't constrained to be positive and that doesn't use the heap (IIUC) seems good to me. But, in the interim, an extra heap allocation where we're already doing a heap allocation doesn't matter when we know we're going to remove them all soon, so I suggest that we just KISS and jump directly to creating your `Integer` type, perhaps starting with it being a wrapper around a `BIGNUM*` and then later getting rid of its use of `BIGNUM` completely at the same time it stops using the heap.

ranweiler: Your "interim" suggestion is exactly what I had in mind as a next step. I'll proceed with that.

src/rsa/verification.rs
ranweiler: Ah right, the licensing burden. Okay, as you suggest, I'll write a new, equivalent Rust function from scratch. It'll be a good way to further exercise the (`Positive`)`Integer` types.

briansmith:
> Okay, as you suggest, I'll write a new, equivalent Rust function from scratch. It'll be a good way to further exercise the (Positive)Integer types.

> Your "interim" suggestion is exactly what I had in mind as a next step. I'll proceed with that.

Thanks! I'm going to close this PR so that the new code can have a fresh start.

PR #309, opened 2016-10-05
Remove use of compatibility lint `match_of_unit_variant_via_paren_dotdot`

petrochenkov:
https://github.com/rust-lang/rust/pull/36871 removes this compatibility lint and therefore breaks ring (but not its dependencies) because ring uses `#![deny(unknown_lints)]`.
FYI, `#![deny(future_incompatible)]` can be used instead of concrete compatibility lints, it's not going to be removed.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -22,7 +22,6 @@
     exceeding_bitshifts,
     fat_ptr_transmutes,
     improper_ctypes,
-    match_of_unit_variant_via_paren_dotdot,
     missing_copy_implementations,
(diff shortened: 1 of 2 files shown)


briansmith:
Thanks! Landed in 5a2c0a2006165db9da3721c4b49300ded4dacd5b.

petrochenkov:
Thanks!
Could you also publish a new version to crates.io so it doesn't show up as a regression again?

PR #321, opened 2016-10-25
[WIP] Port build system to use only build.rs on unix

### LINKED ISSUE #320: Build fails when there are spaces in the path to the project
When using ring as dependency of a rust project the build is failing if the path to the project contains a white space, for example `/tmp/ring\ build\ fail`.

Output:

```
root@215b2bafd8c8:/tmp/ring build fail# cargo build
    Updating registry `https://github.com/rust-lang/crates.io-index`
   Compiling ring v0.4.3
   Compiling untrusted v0.3.1
   Compiling lazy_static v0.2.1
error: failed to run custom build command for `ring v0.4.3`
process didn't exit successfully: `/tmp/ring build fail/target/debug/build/ring-540e5df33a7f7e62/build-script-build` (exit code: 101)
--- stdout
HOSTNAME: 215b2bafd8c8
TERM: xterm
OLDPWD: /tmp
USER: root
PATH: /root/.cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/root/.cargo/bin:/root/.multirust/toolchains/nightly-2016-10-18-x86_64-unknown-linux-gnu/bin
PWD: /tmp/ring build fail
SHLVL: 1
HOME: /root
_: /root/.cargo/bin/cargo
LD_LIBRARY_PATH: /tmp/ring build fail/target/debug/deps:/root/.multirust/toolchains/nightly-2016-10-18-x86_64-unknown-linux-gnu/lib
CARGO_HOME: /root/.cargo
RUST_RECURSION_COUNT: 1
RUSTUP_TOOLCHAIN: nightly-2016-10-18-x86_64-unknown-linux-gnu
RUSTUP_HOME: /root/.multirust
SSL_CERT_FILE: /usr/lib/ssl/certs/ca-certificates.crt
SSL_CERT_DIR: /usr/lib/ssl/certs
CARGO_FEATURE_DEFAULT: 1
RUSTC: rustc
CARGO_PKG_VERSION_MAJOR: 0
DEBUG: true
CARGO_FEATURE_DEV_URANDOM_FALLBACK: 1
RUSTDOC: rustdoc
CARGO_PKG_HOMEPAGE: 
CARGO_MANIFEST_DIR: /root/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.4.3
CARGO_PKG_AUTHORS: Brian Smith <brian@briansmith.org>
TARGET: x86_64-unknown-linux-gnu
PROFILE: debug
CARGO_PKG_NAME: ring
NUM_JOBS: 4
CARGO_FEATURE_USE_HEAP: 1
CARGO_PKG_VERSION: 0.4.3
HOST: x86_64-unknown-linux-gnu
CARGO_PKG_VERSION_MINOR: 4
OPT_LEVEL: 0
CARGO_PKG_DESCRIPTION: Safe, fast, small crypto using Rust.
CARGO_PKG_VERSION_PATCH: 3
OUT_DIR: /tmp/ring build fail/target/debug/build/ring-540e5df33a7f7e62/out
CARGO_PKG_VERSION_PRE: 
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/lib/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/aes/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/aes/asm/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/bn/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/bn/asm/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/chacha/asm/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/cipher/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/curve25519/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/curve25519/asm/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/ec/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/ec/asm/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/modes/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/modes/asm/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/poly1305/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/poly1305/asm/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/rand/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/rsa/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/sha/asm/
mkdir -p fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/crypto/test/

--- stderr
mk/ring.mk:173: warning: overriding recipe for target '/tmp/ring'
mk/ring.mk:169: warning: ignoring old recipe for target '/tmp/ring'
mk/ring.mk:173: warning: overriding recipe for target 'build'
mk/ring.mk:169: warning: ignoring old recipe for target 'build'
mk/ring.mk:224: warning: overriding recipe for target '/tmp/ring'
mk/ring.mk:173: warning: ignoring old recipe for target '/tmp/ring'
mk/ring.mk:224: warning: overriding recipe for target 'build'
mk/ring.mk:173: warning: ignoring old recipe for target 'build'
mk/ring.mk:223: *** mixed implicit and normal rules: deprecated syntax
mk/bottom_of_makefile.mk:35: warning: overriding recipe for target '/tmp/ring'
mk/ring.mk:224: warning: ignoring old recipe for target '/tmp/ring'
mk/bottom_of_makefile.mk:35: warning: overriding recipe for target 'build'
mk/ring.mk:224: warning: ignoring old recipe for target 'build'
mk/bottom_of_makefile.mk:34: *** mixed implicit and normal rules: deprecated syntax
mk/bottom_of_makefile.mk:37: warning: overriding recipe for target '/tmp/ring'
mk/bottom_of_makefile.mk:35: warning: ignoring old recipe for target '/tmp/ring'
mk/bottom_of_makefile.mk:37: warning: overriding recipe for target 'build'
mk/bottom_of_makefile.mk:35: warning: ignoring old recipe for target 'build'
mk/bottom_of_makefile.mk:36: *** mixed implicit and normal rules: deprecated syntax
mk/bottom_of_makefile.mk:37: warning: overriding recipe for target 'fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/%.o'
mk/bottom_of_makefile.mk:35: warning: ignoring old recipe for target 'fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/%.o'
mk/bottom_of_makefile.mk:39: warning: overriding recipe for target '/tmp/ring'
mk/bottom_of_makefile.mk:37: warning: ignoring old recipe for target '/tmp/ring'
mk/bottom_of_makefile.mk:39: warning: overriding recipe for target 'build'
mk/bottom_of_makefile.mk:37: warning: ignoring old recipe for target 'build'
mk/bottom_of_makefile.mk:38: *** mixed implicit and normal rules: deprecated syntax
mk/bottom_of_makefile.mk:39: warning: overriding recipe for target 'fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/%.o'
mk/bottom_of_makefile.mk:37: warning: ignoring old recipe for target 'fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/%.o'
mk/bottom_of_makefile.mk:41: warning: overriding recipe for target '/tmp/ring'
mk/bottom_of_makefile.mk:39: warning: ignoring old recipe for target '/tmp/ring'
mk/bottom_of_makefile.mk:41: warning: overriding recipe for target 'build'
mk/bottom_of_makefile.mk:39: warning: ignoring old recipe for target 'build'
mk/bottom_of_makefile.mk:40: *** mixed implicit and normal rules: deprecated syntax
mk/bottom_of_makefile.mk:41: warning: overriding recipe for target 'fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/%.o'
mk/bottom_of_makefile.mk:39: warning: ignoring old recipe for target 'fail/target/debug/build/ring-540e5df33a7f7e62/out/obj/%.o'
make: *** No rule to make target '%.cc', needed by '/tmp/ring'.  Stop.
thread 'main' panicked at 'make execution failed', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.4.3/build.rs:211
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

weiznich:
- Fix #320


weiznich:
Any ideas why the windows builder is failing?

briansmith:
Wow! Thanks for this!

Three questions to start with:

* Is this code (or similar) being used in another open-source project? If so, which ones? It would help speed up the review if I can look at the history of the code, if there is any history.

* How necessary is the target_build_utils dependency? Would it be a lot of work to do this without that dependency?

* Does the gcc crate honor the `CC`, `CXX`, `AS`, etc. variables on Unix-ish hosts? In the current build system, `CC` and `CXX` are used to set the compiler, e.g. clang or GCC, or specific versions.

weiznich:
> Is this code (or similar) being used in another open-source project? If so, which ones? It would help speed up the review if I can look at the history of the code, if there is any history.

I've developed the code to fix the issue with the white spaces. So unfortunately there is no history for this.

>    How necessary is the target_build_utils dependency? Would it be a lot of work to do this without that dependency?

target_build_utils simplifies the handling of the target_triples. Especially it removes the parsing and unifying of those triples (For example it maps i586, i386 and i686 to x86). It should be possible to implement this without this crate, but it would result in quite a bit additional code that needs to be written. Furthermore it's only a build_dependency, which means its not added in the final library. Also target_build_utils and all of it's dependencies only contains pure rust code (as far as I can see). So it should build fine on every platform.

>   Does the gcc crate honor the CC, CXX, AS, etc. variables on Unix-ish hosts? In the current build system, CC and CXX are used to set the compiler, e.g. clang or GCC, or specific versions.

The gcc crate honors CC, CXX, AS, etc (and many more). It seems like it even detects if there it's cross compiling to ios and sets the according sdk path. Take a look at the [code](https://github.com/alexcrichton/gcc-rs/blob/master/src/lib.rs#L676) for the compiler selection. We could also set the compiler manually using this [function](https://docs.rs/gcc/0.3.38/gcc/struct.Config.html#method.compiler).

briansmith:
@frewsxcv Do you have an opinion about this?

My main concerns about this kind of switch are:

* Legibiility: Is the new thing as easy to read as the Makefile-based approach?

* Performance for people hacking on *ring*: Is this going to slow down the build times for people who are actively hacking on *ring*, by forcing every C/asm file to be recompiled/regenerated every time anything changes? The Makefile-based approach allows incremental building of all the C/asm stuff?

Despite those concerns, I think it makes sense to find a way to make the build work without requiring GNU make.

crypto/test/bn_test_lib.c
```diff
@@ -133,7 +133,7 @@ int BN_rand(BIGNUM *rnd, int bits, RAND *rng) {
   bit = (bits - 1) % 8;
   mask = 0xff << (bit + 1);
 
-  buf = OPENSSL_malloc(bytes);
+  buf = (uint8_t*) OPENSSL_malloc(bytes);
```
briansmith: Why did you add this cast? The policy so far is to not cast `(void *)` since C doesn't require the cast, because in general we have a goal of eliminating every single cast.

crypto/constant_time_test.c
```diff
@@ -42,6 +42,9 @@
  * copied and put under another distribution licence
  * [including the GNU Public Licence.]
  */
+#ifdef __cplusplus
+extern "C" {
+#endif
```
briansmith: If this change is important then please make it in a separate PR that explains why it was done. It shouldn't be needed, AFAICT.

build.rs
```diff
@@ -64,159 +64,488 @@
...
+            panic!("failed to execute {}: {}",
+                   command_name.as_ref().to_str().unwrap(),
+                   e);
+        })
+        .success() {
         panic!("{} execution failed", command_name.as_ref().to_str().unwrap());
     }
 }
+
+fn compile(file: &str, target_info: &TargetInfo, mut out_dir: PathBuf)
```
briansmith: My understanding is that the gcc crate automatically parallelizes the build if you pass in multiple files, and so did the `make`-based approach.  So, we need to decide whether we want to use the gcc crate's parallelism support, or whether we're going to (perhaps in a separate PR, and perhaps not necessarily you writing it) do the parallelism ourselves.

The benefit of doing the parallelism ourselves is that we can do it for  the `*.pl` -> `*.S` and `.S` -> `.o` steps too. The downside to doing the parallelism ourselves is that it is more work than relying on the gcc crate's built-in stuff.

WDYT?

briansmith:
Thinking about this more:

The first step, I think, is agreeing on a plan. My main concern is whether we can (eventually) support incremental builds and parallelized builds. I'd like to hear your thoughts and make sure we agree on how that could eventually work, even if we don't do that in this PR.

After that, I'd like to see the PR broken up into at least these parts:

0. The changes to the C and C++ files, which should be their own PRs.
1. rustfmt on build.rs.
2. Factor out the `build_msvc` function, underneath `main`.
3. The `RUST_SRCS` and similar constants + generation of the `cargo:rerun-if-changed` output (on non-MSVC, at least; include Makefile and mk/*.mk).
4. The rest of the new stuff, underneath `main`, in the top-down style.

I would also be happy if #4 could be broken into multiple parts (e.g. dealing with the perlasm stuff in a separate commit), if that isn't an unreasonable amount of work.

Does this make sense to you?

crypto/test/bn_test_lib.c
weiznich: I've tried to compile this as c++ code. In this case g++ complains that `void*` is no `uint8_t*`.
```error: invalid conversion from ‘void*’ to ‘uint8_t*‘[-fpermissive]```
It's not needed now, so I've removed this from the current version.

File: build.rs [hunk 4]
```diff
@@ -64,159 +64,488 @@
...
+      "crypto/sha/asm/sha512-armv8.pl"];
+
+const RING_TEST_SRCS: &'static [&'static str] =
+    &["crypto/bn/bn_test.cc",
+      "crypto/constant_time_test.c",
+      "crypto/test/bn_test_convert.c",
+      "crypto/test/bn_test_lib.c",
+      "crypto/test/file_test.cc"];
+
+const C_FLAGS: &'static [&'static str] = &["-std=c1x",
```
weiznich: It's produced by `rustfmt` in this way. :disappointed:

build.rs
weiznich: The gcc crate is not capable to distinguish between different kind of files. If I pass all RING_TEST_SRCS to the gcc crate I need to set the [cpp flag](https://docs.rs/gcc/0.3.38/gcc/struct.Config.html#method.cpp). This means everything is compiled as c++ file. (The gcc crate simply calls CC or CXX on all passed files.) At this point some of the supported compilers are beginning to complain about incorrect compiler and linker flags. For example see this [build](https://travis-ci.org/briansmith/ring/jobs/170415102).
Because of this I decided to compile every file on it's own. It should be simple to call this in parallel by using par_iter form [rayon](https://crates.io/crates/rayon) like the gcc crate.

weiznich:
> My main concern is whether we can (eventually) support incremental builds and parallelized builds.

The current version already checks if a file needs to be recompiled. (See the need_run method.). If someone changes a header file everything is recompiled.
For parallel builds we could simply use `par_iter` from [rayon](https://crates.io/crates/rayon) as mentioned above.

I've split the commit into smaller parts. Hopefully this is now easier to review.

briansmith:
@weiznich I did the reformatting myself because it involves some rearranging the code to make it less horrific: 61e6afc87b9466f272695ab8391d666404d9ac20.

build.rs
```diff
@@ -125,85 +125,96 @@ fn build_c_code(out_dir: &str) -> Result<(), std::env::VarError> {
...
+    let arch = target_triple[0];
+    let (platform, optional_amd64) = match arch {
+        "i686" => ("Win32", None),
+        "x86_64" => ("x64", Some("amd64")),
+        _ => panic!("unexpected ARCH: {}", arch),
+    };
+
+    fn find_msbuild_exe(program_files_env_var: &str,
+                        optional_amd64: Option<&str>)
+                        -> Result<std::ffi::OsString, ()> {
```
briansmith: We never return `Err(...)` here, so the return type should be removed.

build.rs
```diff
@@ -125,85 +125,96 @@ fn build_c_code(out_dir: &str) -> Result<(), std::env::VarError> {
...
-        let (platform, optional_amd64) = match arch {
-            "i686" => ("Win32", None),
-            "x86_64" => ("x64", Some("amd64")),
-            _ => panic!("unexpected ARCH: {}", arch),
-        };
+        try!(build_msvc(&target_triple,
+                        disable_opt,
+                        &num_jobs,
+                        lib_path.clone(),
+                        out_dir));
```
briansmith: We never return `Err()` so we should remove the `try!(...)`.

build.rs
```diff
@@ -65,7 +65,7 @@
 )]
 
 use std::env;
-use std::path::Path;
+use std::path::{Path, PathBuf};
```
briansmith: There are some uses of `std::path::PathBuf` below that should have the `std::path::` qualification removed.

build.rs
```diff
@@ -125,85 +125,96 @@ fn build_c_code(out_dir: &str) -> Result<(), std::env::VarError> {
...
 
-        let configuration = if disable_opt { "Debug" } else { "Release" };
-        let args = vec![
+    let configuration = if disable_opt { "Debug" } else { "Release" };
+    let args = vec![
             format!("/m:{}", num_jobs),
             format!("/p:Platform={}", platform),
             format!("/p:Configuration={}", configuration),
             format!("/p:OutRootDir={}/", out_dir),
         ];
```
briansmith: These lines are too indented. (When I reduce the indention by 4 spaces in my copy and then run `rustfmt` it keeps it as-is.)

build.rs
briansmith: I don't know how to configure `rustfmt` to format this in the compact style, but it should be like this:

```rust
        build_msvc(&target_triple, disable_opt, &num_jobs, lib_path.clone(),
                   out_dir);
```

build.rs
```diff
@@ -67,9 +67,145 @@
 use std::env;
 use std::path::{Path, PathBuf};
 
-
 const LIB_NAME: &'static str = "ring";
 
+const RING_SRC: &'static [&'static str] = &["crypto/aes/aes.c",
```
briansmith: I can't figure out how to make rustfmt do something reasonable with this. Let's format all of these variables containing lists of files like `RING_X86_64_SRC` and add `#[cfg_attr(rustfmt, rustfmt_skip)]` to each of them.

build.rs
```diff
@@ -67,9 +67,145 @@
...
+                                            "crypto/ec/wnaf.c",
+                                            "crypto/mem.c",
+                                            "crypto/modes/gcm.c",
+                                            "crypto/poly1305/poly1305.c",
+                                            "crypto/rand/sysrand.c",
+                                            "crypto/rsa/blinding.c",
+                                            "crypto/rsa/rsa.c",
+                                            "crypto/rsa/rsa_impl.c"];
+
+const RING_INTEL_SHARED_SRCS: &'static [&'static str] = &["crypto/cpu-intel.c"];
```
briansmith: This one too (value on its own line).

build.rs
```diff
@@ -67,9 +67,145 @@
...
+                                                "include/openssl/aes.h",
+                                                "include/openssl/base.h",
+                                                "include/openssl/err.h"];
+
+const RING_TEST_HEADERS: &'static [&'static str] =
+    &["crypto/test/bn_test_lib.h",
+      "crypto/test/file_test.h",
+      "crypto/test/bn_test_util.h"];
+
+const RING_PERL_INCLUDES: &'static [&'static str] = &["crypto/sha/asm/sha-x86_64.pl"];
```
briansmith: Also this one.

build.rs
```diff
@@ -67,9 +67,145 @@
...
+      "crypto/sha/asm/sha512-armv8.pl"];
+
+const RING_TEST_SRCS: &'static [&'static str] =
+    &["crypto/bn/bn_test.cc",
+      "crypto/constant_time_test.c",
+      "crypto/test/bn_test_convert.c",
+      "crypto/test/bn_test_lib.c",
+      "crypto/test/file_test.cc"];
+
+const RING_HEADERS: &'static [&'static str] = &["crypto/poly1305/internal.h",
```
briansmith: This would be impossible to keep in sync with the source if we have to do it 100% manually. Already, the `*.inl` files are missing from this list, and so is `find crypto/perlasm -name "*.pl"` is missing, at least.

On the other hand, we can't have build.rs build the list dynamically from doing something like `find crypto include mk -name "*.h" -o -name "*.inl" -o -name "*.mk"` itself, because that won't account for removed files.

Maybe we should just always run the build script, and have the build script do conditional rebuild logic on a per-source-file so that it isn't ever too slow? After all, we want an incremental build where only a `*.c` file changed to be about as fast as a build where only a `*.rs` file changed. If the difference between running build.rs or not is large enough to be worth skipping then build.rs isn't fast enough.

WDYT?

build.rs
briansmith: `crypto/sha/asm/sha-armv8.pl` is another thing that should be added to this list.

build.rs
```diff
@@ -199,96 +203,86 @@ const RING_TEST_HEADERS: &'static [&'static str] =
...
                                                   "mk/ring.mk"];
 
+const C_FLAGS: &'static [&'static str] = &["-std=c1x",
+                                           "-Wbad-function-cast",
+                                           "-Wmissing-field-initializers",
+                                           "-Wmissing-prototypes",
+                                           "-Wnested-externs",
+                                           "-Wstrict-prototypes"];
+
+const CPP_FLAGS: &'static [&'static str] = &["-std=c++0x",
```
briansmith: I noticed you copied `CFLAGS` and `CPPFLAGS` from mk/top_of_makefile.mk, but you didn't copy `CXXFLAGS`. Instead of looks like you confused `CPPFLAGS` and `CXXFLAGS`. `CPPFLAGS` should be passed to every compilation (*.S, *.c, *.cc), `CFLAGS` should additionally be passed for *.c only, and `CXXFLAGS` should be passed for *.cc.

In particular, `-std=c++0x` should be in `CXXFLAGS` but most of the other stuff here should remain in `CPPFLAGS`.

build.rs
```diff
@@ -351,7 +347,105 @@ fn build_msvc(target_triple: &[&str], disable_opt: bool, num_jobs: &str,
...
+    test_target.push("libring-test.a");
+    let test_target = test_target.as_path();
+
+    let lib_header_change = RING_HEADERS.iter()
+        .map(Path::new)
+        .any(|p| need_run(&p, lib_target));
+    let test_header_change = RING_TEST_HEADERS.iter()
+        .map(Path::new)
+        .any(|p| need_run(&p, test_target)) ||
+                             lib_header_change;
```
briansmith: I suggest that we just check if any of `find crypto include mk -name "*.h" -o -name "*.inl" -o -name "*.mk"` changed, and if so, rebuild all the libraries. Especially since the test lib will go away sometime soon.

build.rs
```diff
@@ -351,7 +347,105 @@ fn build_msvc(target_triple: &[&str], disable_opt: bool, num_jobs: &str,
...
+
+    if objs.iter().map(|f| Path::new(f)).any(|p| need_run(&p, lib_target)) {
+
+        let mut c = gcc::Config::new();
+        for o in objs {
+            let _ = c.object(o);
+        }
+        c.compile("libring-core.a");
+    }
+
```
briansmith: Please factor out the duplicate logic between the code in this function above this line and the code below this line.

build.rs
```diff
@@ -370,6 +464,45 @@ fn run_command_with_args<S>(command_name: S, args: &Vec<String>)
     }
 }
 
+fn make_asm(source: &str, mut dst: PathBuf, target_info: &TargetInfo)
+            -> String {
+    let p = Path::new(source);
```
briansmith: Here we need to consider `crypto/perlasm/*.pl` and `crypto/sha/{sha-armv8.pl, sha-x86_64.pl}` the same way we consider *.h and *.inl in the `compile()` function.

briansmith:
Are you the copyright holder for these changes, or is there an employer or somebody else that owns the copyright? If somebody else, who?

Please add the contributor statement to the bottom of each commit; See https://github.com/briansmith/ring#contributing.

If you add the contributor statement to the commit that factors out the `build_msvc()` function then I'll commit it to master while we work on the rest.

weiznich:
I've updated the pull request according to your comments.

briansmith:
Thanks!

I landed the "Move msvc build logic to own function" commit as a0ee4b576359fa1316aae04def13769faa47ea10, tweaked as I suggested above.

briansmith:
@weiznich Just to make sure you're aware, I also wrote many comments above that are now hidden behind the "show outdated" links.

build.rs
weiznich: I've added some logic to check for files in crypto, include and mk. Build.rs will now refuse to build the native code if there are untracked files (*.c, *.cc, *.pl, *.S,  and *.mk) in there.

weiznich:
> @weiznich Just to make sure you're aware, I also wrote many comments above that are now hidden behind the "show outdated" links.

Thanks, I've missed them.


The current version should address those comments. Also I added a commit which introducing parallel compilation of the native stuff. But I'm not sure if this improves the compiletime.

build.rs
```diff
@@ -66,95 +66,265 @@
...
 
-    Ok(())
+    check_all_files_tracked();
+
+    let out_dir = env::var("OUT_DIR").unwrap();
+
+    build_c_code(&out_dir);
+}
+
+fn build_c_code(out_dir: &str) {
```
frewsxcv: Minor, but from an API standpoint, it seems like this should take a `Path` or `PathBuf`

frewsxcv:
@weiznich Out of curiosity, do you see any significant build time changes with these changes (from a clean build via `cargo clean`) versus the Makefile approach on `master`?

> @frewsxcv Do you have an opinion about this?

My opinion is pretty subjective: I (primarily) come from Python-land and JavaScript-land, so I haven't used Makefiles that much, so I feel like I never forced myself to learn the syntax and semantics thoroughly. Because of this, it might be the case that when I run into issues with Makefiles, that I find debugging them not particularly enjoyable; contrasting to native-language-based-build-files (like this PR) where (for me) the flow of the build is much easier to follow.  On the other hand, I do agree there is some value in the terseness and simplicity of a Makefile, as the diff count illustrates.

weiznich:
>@weiznich Out of curiosity, do you see any significant build time changes with these changes (from a clean build via cargo clean) versus the Makefile approach on master?


@frewsxcv I've tried a few build scenarios:

x | master | sequential build | parallel build
------------ | ------------- | ----|----
clean build(`cargo clean && cargo build`) |10.2 s | 39.9 s| 42.5 s
rebuild nativ(`rm target/debug/build/ring-* && cargo build`)| 9.9 s | 15.4 s | 15.8 s
change source(`touch crypto/aes/aes.c && cargo build`)| 4.0 s | 4.4 s | 3.9 s
change header(`touch include/openssl/aes.h && cargo build`)| 5.6 s | 10.5 s | 5.6 s

> On the other hand, I do agree there is some value in the terseness and simplicity of a Makefile, as the diff count illustrates.

I did not remove the makefiles, so the diff count is a bit misleading. There are 799 lines build code on master (Makfile + mk/top_of_makefile.mk + mk/ring.mk + mk/bottom_of_makefile.mk + build.rs), whereas the new system only uses 754 lines in build.rs

weiznich:
Any news on this?

build.rs
```diff
@@ -230,3 +500,203 @@ fn run_command_with_args<S>(command_name: S, args: &Vec<String>)
...
+            if path.is_dir() {
+                walk_dir(&path, cb);
+            } else {
+                cb(&entry);
+            }
+        }
+    }
+}
+
+trait BuildHelper {
```
briansmith: What is the purpose of this trait? In particular, srcs() is only called in one place, asmflags() is only called in one place, etc. And there's only one implementation of the trait. So, why not just get rid of the trait and calculate these values within the functions that currently call these methods? That would be much clearer to me because each function would be more self-contained.

build.rs
```diff
@@ -230,3 +500,203 @@ fn run_command_with_args<S>(command_name: S, args: &Vec<String>)
...
+}
+
+fn need_run(source: &Path, target: &Path) -> bool {
+    let s = std::fs::metadata(source);
+    let t = std::fs::metadata(target);
+    if s.is_err() || t.is_err() {
+        true
+    } else {
+        match (s.unwrap().modified(), t.unwrap().modified()) {
+            (Ok(s), Ok(t)) => s > t,
```
briansmith: Shouldn't this be `s >= t`? In particular, if you only have one-second-precision timestamps then it is very likely that you can't reliably know if the source was modified after the target, so you have to play it safe.

build.rs
```diff
@@ -410,34 +414,60 @@ fn build_unix(target_info: &TargetInfo, out_dir: PathBuf) {
...
-        for src in additional_src.iter() {
-            additional.push(make_asm(src, out_dir.clone(), &target_info));
+    // copied from gcc
+    let mut cfg = rayon::Configuration::new();
+    if let Ok(amt) = env::var("NUM_JOBS") {
+        if let Ok(amt) = amt.parse() {
+            cfg = cfg.set_num_threads(amt);
         }
     }
+    drop(rayon::initialize(cfg));
```
briansmith: What is the purpose of the `drop()` here?

build.rs
```diff
@@ -410,34 +414,60 @@ fn build_unix(target_info: &TargetInfo, out_dir: PathBuf) {
         .any(|p| need_run(&p, test_target)) ||
                              lib_header_change;
```
briansmith: The header change test can/should also be done in parallel, right?

build.rs
```diff
@@ -66,95 +66,261 @@
...
+const AS_FLAGS: &'static [&'static str] = &["-Wa,--noexecstack,-gdwarf"];
+
+
+fn main() {
+    for (key, value) in env::vars() {
+        println!("{}: {}", key, value);
     }
 
-    Ok(())
+    check_all_files_tracked();
```
briansmith: The parallelism commit should also make `check_all_files_tracked()` function work in parallel.

Also, I suggest that we put the `check_all_files_tracked()` call after `build_c_code();`. Imagine that there is an error in the C code; we'd rather get that error sooner, rather than wait for the `check_all_files_tracked()` function.

And/or, is there a way to have Rayon do the `check_all_files_tracked()` work in parallel with `build_c_code()`, but giving tasks from `build_c_code()` higher priority?

briansmith:
Hi, I tried to review each commit individually, but unfortunately the later reviews made the earlier reviews "outdated" in GitHub's eyes. So, please use the "show outdated" above.

Everything needs to be considered dirty if build.rs changes. I'm not sure that's accounted for here.

I will re-review the changes again after I see an updated version. In particular, I haven't done a line-by-line comparison of the makefile logic and the build.rs logic to make sure they match. Right now I'm assuming they do match but I'll double-check that before merging.

build.rs
weiznich: This code is coming from the [gcc crate](https://github.com/alexcrichton/gcc-rs/blob/master/src/lib.rs#L372-L378). I'm not sure why drop is called here. (I think it should not be needed)

weiznich:
> Hi, I tried to review each commit individually, but unfortunately the later reviews made the earlier reviews "outdated" in GitHub's eyes. So, please use the "show outdated" above.

Thanks, hopefully everything is now fixed.

> Everything needs to be considered dirty if build.rs changes. I'm not sure that's accounted for here.

[It's already there](https://github.com/weiznich/ring/blob/504c589567138ea763d35520e1fabb424b09ea7c/build.rs#L405)

> I will re-review the changes again after I see an updated version. In particular, I haven't done a line-by-line comparison of the makefile logic and the build.rs logic to make sure they match. Right now I'm assuming they do match but I'll double-check that before merging.

Before merging also [this TODO](https://github.com/weiznich/ring/blob/504c589567138ea763d35520e1fabb424b09ea7c/build.rs#L363) needs to be fixed. I don't know where this needs to be changed.

weiznich:
Any news on this?

briansmith:
I'm expecting this to land in *ring* 0.7. I want to release *ring* 0.6 without this first.

weiznich:
I've rebased the pull request.
I'm not sure why some of the ci-builders are failing:
* [MacOS](https://travis-ci.org/briansmith/ring/jobs/192323129): Linker complains about undefined Symbols from c++-stdlib
* [TARGET_X=x86_64-unknown-linux-gnu CC_X=gcc-4.6](https://travis-ci.org/briansmith/ring/jobs/192323151), [TARGET_X=i686-unknown-linux-gnu CC_X=gcc-4.6](https://travis-ci.org/briansmith/ring/jobs/192323171), [TARGET_X=arm-unknown-linux-gnueabihf CC_X=arm-linux-gnueabihf-gcc](https://travis-ci.org/briansmith/ring/jobs/192323188) and [TARGET_X=aarch64-unknown-linux-gnu CC_X=aarch64-linux-gnu-gcc](https://travis-ci.org/briansmith/ring/jobs/192323168) fails with: `crypto/ec/ecp_nistz.h:106:20: error: inlining failed in call to ‘booth_recode.part.0’: call is unlikely and code size would grow [-Werror=inline]`
* [TARGET_X=aarch64-unknown-linux-gnu CC_X=aarch64-linux-gnu-gcc](https://travis-ci.org/briansmith/ring/jobs/192323166) fails while setting up the image: `WARNING: The following packages cannot be authenticated!
  qemu-user qemu-user-binfmt
E: There are problems and -y was used without --force-yes`
* [TARGET_X=aarch64-unknown-linux-gnu CC_X=aarch64-linux-gnu-gcc](https://travis-ci.org/briansmith/ring/jobs/192323167) fails with error about invalid ppa name: `Cannot add PPA: 'ppa:ubuntu-toolchain-r/test'.
Please check that the PPA name or format is correct.`
* [MSVC](https://ci.appveyor.com/project/briansmith/ring/build/1.0.1590/job/l6f5a4l12rp685xa) failed with `fatal error LNK1181: cannot open input file 'ring-asm.lib'

briansmith:
> MacOS: Linker complains about undefined Symbols from c++-stdlib

According to https://travis-ci.org/briansmith/ring/jobs/192323129#L310, it is trying to use libstdc++ instead of libc++. It should be using libc++ on Mac.

briansmith:
> WARNING: The following packages cannot be authenticated! qemu-user qemu-user-binfmt E: There are problems and -y was used without --force-yes

> Cannot add PPA: 'ppa:ubuntu-toolchain-r/test'. Please check that the PPA name or format is correct.

These are probably transient errors.

briansmith:
> TARGET_X=x86_64-unknown-linux-gnu CC_X=gcc-4.6, TARGET_X=i686-unknown-linux-gnu CC_X=gcc-4.6, TARGET_X=arm-unknown-linux-gnueabihf CC_X=arm-linux-gnueabihf-gcc and TARGET_X=aarch64-unknown-linux-gnu CC_X=aarch64-linux-gnu-gcc fails with: crypto/ec/ecp_nistz.h:106:20: error: inlining failed in call to ‘booth_recode.part.0’: call is unlikely and code size would grow [-Werror=inline]

There must be some difference in CPPFLAGS or CFLAGS between the new build system and the old build system. I'd compare the build log of a build using the old build system with the build log using the new build system to see which flags aren't the same.

briansmith:
> MSVC failed with `fatal error LNK1181: cannot open input file 'ring-asm.lib'

Looks like something maybe went wrong with copying the pregenerated file to the right place. In particular, maybe it's being copied to the wrong place?

briansmith:
I'm planning to re-review and merge this this week. If you have time to rebase it on top of the current master, that would be much appreciated.

As part of my review, I'll also do a prototype of porting the Windows MSBuild-based system to be purely based on build.rs. However, finishing the Windows work won't block your work from landing; I'm mostly going to be doing it to make sure I fully understand the code, and because I eventually want Windows and non-Windows builds to work as similarly as possible.

weiznich:
I've rebased to master, but I'm pretty sure that some ci builds will fail. (CI is not building…)

build.rs
```diff
@@ -130,6 +130,7 @@ const RING_X86_64_SRC: &'static [&'static str] =
       "crypto/bn/asm/x86_64-mont.pl",
       "crypto/bn/asm/x86_64-mont5.pl",
       "crypto/chacha/asm/chacha-x86_64.pl",
+      "crypto/cipher/asm/chacha20_poly1305_x86_64.pl",
```
briansmith: This file isn't actually part of the build yet. If having this file sitting around being unused causes trouble, then add a commit that deletes the file.

build.rs
```diff
@@ -342,13 +343,12 @@ fn build_msvc(target_info: &TargetInfo, disable_opt: bool, num_jobs: &str,
...
-        format!("/p:Configuration={}", configuration),
-        format!("/p:OutRootDir={}/", lib_path.to_str()
-                .expect("Invalid path")),
-    ];
+    let args = vec![format!("/m:{}", num_jobs),
+                    format!("/p:Platform={}", platform),
+                    format!("/p:Configuration={}", configuration),
+                    format!("/p:OutRootDir={}/",
+                            lib_path.to_str()
+                                .expect("Invalid path"))];
```
briansmith: I guess this should be undone? Looks like it's only reformatting, which shouldn't be in this PR.

build.rs
```diff
@@ -305,37 +306,14 @@ fn build_c_code(out_dir: PathBuf) {
 
 fn build_msvc(target_info: &TargetInfo, disable_opt: bool, num_jobs: &str,
               lib_path: PathBuf) {
-    let (platform, optional_amd64) = match target_info.target_arch() {
-        "x86" => ("Win32", None),
-        "x86_64" => ("x64", Some("amd64")),
+    use gcc::windows_registry;
```
briansmith: This seems reasonable, but why is it part of this PR? I think we should leave the Windows build alone for now.

briansmith: In particular, I suggest moving this to its own PR.

build.rs
```diff
@@ -399,15 +408,6 @@ fn build_unix(target_info: &TargetInfo, out_dir: PathBuf) {
         .map(Path::new)
         .any(|p| need_run(&p, test_target));
 
-    // copied from gcc
```
briansmith: Please fold this commit into the commit that it is modifying.

build.rs
```diff
@@ -438,7 +438,10 @@ fn build_unix(target_info: &TargetInfo, out_dir: PathBuf) {
                   out_dir.clone(),
                   test_header_change);
 
-    let libcxx = if target_info.target_vendor() == "apple" ||
+    let libcxx = if target_info.target_vendor()
+        .map(|v| v == "apple")
```
briansmith: I'd guess that the target vendor doesn't matter if we can detect target_os == "macos" or "ios". Conversely, I think that if we do this for "apple" then I think we wouldn't need to care what the target OS is. So either this logic is too complex or we need to have a comment about it as it is non-obvious.

Also, do you have a mac? If you don't, then don't worry about the Mac failures. I will fix the Mac build for you. Just concentrate on Linux.

build.rs
```diff
@@ -94,24 +94,18 @@ const RING_SRC: &'static [&'static str] =
...
       "crypto/ec/gfp_p384.c",
-      "crypto/ec/wnaf.c",
       "crypto/mem.c",
       "crypto/modes/gcm.c",
-      "crypto/poly1305/poly1305.c",
       "crypto/rand/sysrand.c",
-      "crypto/rsa/blinding.c",
-      "crypto/rsa/rsa.c",
-      "crypto/rsa/rsa_impl.c"];
+      "crypto/limbs/limbs.c"];
```
briansmith: Please merge all the changes in this commit into the preceding commits that it is fixing up. That is, please rebase the whole PR on top of the current master. I will prioritize getting this landed so you don't have to rebase again.

briansmith:
I removed the unused chacha20_poly1305_x86_64.pl file that is causing build errors on at least some platforms in a891756d8df3a82d56c15f5752244df7a80fc2e8.

Please make the changes in the comments I just made, rebase the whole series of commits on top of master, and make sure each commit contains the contributor's statement from https://github.com/briansmith/ring#contributing. Then I will do my best to get this merged ASAP.

build.rs
weiznich: The problem here is that target_vendor returns only a valid value if target_build_utils is build with a nightly rustc. See the corresponding [issue](https://github.com/nagisa/target_build_utils.rs/issues/1) for more information. Therefore this function returns an Option. If the value is there we could check if it is apple otherwise we need to check that the os is ios or macos. We could also use only the second check for all variants, if you prefere.

(I've only a linux machine to test things)

build.rs
weiznich: I will open a new pull request containing this, as soon as this one is merged.

weiznich:
I've rebased everything.

weiznich:
Any news on this?

Cargo.toml
```diff
@@ -266,6 +266,11 @@ name = "ring"
 libc = "0.2.20"
 untrusted = "0.3.2"
 
+[build-dependencies]
+gcc = "0.3"
+target_build_utils = "0.2"
```
briansmith: Please change the target_build_dependencies dependency to `{ version = "0.2", default-features = false }` to reduce the dependencies.

Cargo.toml
```diff
@@ -266,6 +266,11 @@ name = "ring"
 libc = "0.2.20"
 untrusted = "0.3.2"
 
+[build-dependencies]
+gcc = "0.3"
```
briansmith: We don't use `gcc = { version = "0.3", features = ["parallel"] }` because we do the parallelism ourselves, right? If that is the case, let's add a comment to that effect here.

Cargo.toml
briansmith: I filed #457 for removing the target_build_utils dependency after this PR is merged. The target_build_utils itself is about 10 additional transitive dependencies, which is too much long-term. It's OK for now, in order to make forward progress.

briansmith:
In travis.sh, we run "make --version". That should be removed.

build.rs
```diff
@@ -66,101 +66,266 @@
...
+#[cfg_attr(rustfmt, rustfmt_skip)]
+const C_FLAGS: &'static [&'static str] =
+    &["-std=c1x",
+      "-Wbad-function-cast",
+      "-Wmissing-field-initializers",
+      "-Wmissing-prototypes",
+      "-Wnested-externs",
+      "-Wstrict-prototypes"];
+
+const CXX_FLAGS: &'static [&'static str] = &["-std=c++0x"];
```
briansmith: Add a comment: `// GCC 4.6 requires "c++0x" instead of "c++11"`.

build.rs
```diff
@@ -66,101 +66,266 @@
...
+      "crypto/perlasm/x86asm.pl",
+      "crypto/perlasm/x86_64-xlate.pl",
+      "crypto/perlasm/arm-xlate.pl",
+      "crypto/perlasm/ppc-xlate.pl"];
+
+const RING_BUILD_FILE: &'static [&'static str] = &["build.rs"];
+
+#[cfg_attr(rustfmt, rustfmt_skip)]
+const C_FLAGS: &'static [&'static str] =
+    &["-std=c1x",
```
briansmith: Add a comment: `// GCC 4.6 requires "c1x" instead of "c11"`.

build.rs
```diff
@@ -218,6 +387,167 @@ fn build_msvc(target_triple: &[&str], disable_opt: bool, num_jobs: &str,
...
+                for f in C_FLAGS {
+                    let _ = c.flag(f);
+                }
+            },
+            Some(Some("cc")) => {
+                for f in CXX_FLAGS {
+                    let _ = c.flag(f);
+                }
+                if target_info.target_os() != "none" {
+                    let _ = c.flag("-fstack-protector");
```
briansmith: I think you confused "CPPFLAGS" and "CXXFLAGS". This should be done for both "c" and "cc".  Here's the logic you need to implement:
```
ifneq ($(filter-out none redox,$(TARGET_SYS)),)
	CPPFLAGS += -fstack-protector
endif
```

build.rs
```diff
@@ -218,6 +387,167 @@ fn build_msvc(target_triple: &[&str], disable_opt: bool, num_jobs: &str,
...
+                }
+                let _ = match (target_info.target_os(),
+                               target_info.target_arch()) {
+                    ("macos", _) => c.flag("-gfull"),
+                    _ => c.flag("-g3"),
```
briansmith: This needs to be done for both "c" and "cc". (Again, I think you confused CXXFLAGS and CPPFLAGS.)

build.rs
```diff
@@ -218,6 +387,167 @@ fn build_msvc(target_triple: &[&str], disable_opt: bool, num_jobs: &str,
...
+            Some(Some("cc")) => {
+                for f in CXX_FLAGS {
+                    let _ = c.flag(f);
+                }
+                if target_info.target_os() != "none" {
+                    let _ = c.flag("-fstack-protector");
+                }
+                let _ = match (target_info.target_os(),
+                               target_info.target_arch()) {
+                    ("macos", _) => c.flag("-gfull"),
```
briansmith: Copy the comment over: `// ``-gfull`` is required for Darwin's |-dead_strip|.`

mk/top_of_makefile.mk
```diff
@@ -1,245 +0,0 @@
...
-# libc++ is the default in newer toolchains, but libstdc++ was the default in
-# older toolchains.
-CXXFLAGS += -stdlib=libc++
-LDFLAGS += -fPIC -Wl,-dead_strip
-else
-CPPFLAGS += -g3
-LDFLAGS += -Wl,--gc-sections
-endif
-
-ASFLAGS += -Wa,--noexecstack,-gdwarf
```
briansmith: The `ASFLAGS` logic wasn't copied over.

build.rs
```diff
@@ -218,6 +387,167 @@ fn build_msvc(target_triple: &[&str], disable_opt: bool, num_jobs: &str,
...
+    let objs = additional.chain(lib_src.par_iter().map(|a| String::from(*a)))
+        .weight_max()
+        .map(|f| compile(&f, &target_info, out_dir.clone(), header_changed))
+        .map(|v| vec![v])
+        .reduce(Vec::new,
+                &|mut a: Vec<String>, b: Vec<String>| -> Vec<String> {
+                    a.extend(b.into_iter());
+                    a
+                });
+    if objs.par_iter()
```
briansmith: Please add a comment here about what this does. I think it would be something like "Rebuild the library if necessary."

build.rs
```diff
@@ -218,6 +387,167 @@ fn build_msvc(target_triple: &[&str], disable_opt: bool, num_jobs: &str,
...
+}
+
+
+fn build_library<P>(target: &Path, additional: P,
+                    lib_src: &'static [&'static str],
+                    target_info: &TargetInfo, out_dir: PathBuf,
+                    header_changed: bool)
+    where P: ParallelIterator<Item = String>
+{
+    let objs = additional.chain(lib_src.par_iter().map(|a| String::from(*a)))
```
briansmith: Let's add a comment here, e.g. "Compile all the (dirty) source files into object files."

briansmith:
When you make the next set of changes, please:

1. Rebase the entire series on top of the current master.
1. Make your changes that address the review comments as 1 or more separate commit(s), leaving the existing commits unchanged.

In particular, if you rebase on top of the current master then the build script output will be in the Travis CI logs because I just committed a change to run `cargo build` and `cargo test` with the `-vv` option.

briansmith:
Here's the process I'm using to review the changes related to compiler flags:

First I pick an arbitrary target, e.g. 64-bit Linux.

I run `cargo test -vv --features=rsa_signing` on current master to get a build log. Then I run `cargo test -vv --features=rsa_signing` on your branch to get a build log.

Then I find the command line for compiling an arbitrary C file, e.g. crypto.c, in each build log. I replace spaces with newlines, sort the lines, and then diff the two commands to see what flags are different between master and your PR.

Then I find a command line for compiling an arbitrary C++ file, e.g. bn_test.cc, and do the same diffing procedure. Similarly, tomorrow I will do the same for an arbitrary perl invocation, for an arbitrary assembly invocation, and for each linking/librarian invocation.

In addition to all the above, I've inspected the Makefiles, looking in particular for comments that explain important design decisions. Hopefully I've asked you to copy all the important comments in my current review comments. (There are a bunch of TODO comments that you don't need to worry about; I've captured them in a separate document.)

Then I will repeat the whole procedure for 32-bit Linux. And then 64-bit macOS, then 64-bit iOS, then 32-bit iOS, then 64-bit Android, then 32-bit Android.

Things have to be reviewed this way, instead of comparing the old Makefile logic to the new build.rs code, because (1) it is too hard to compare them exactly, and (2) the gcc crate adds its own additional command line parameters that aren't obvious.

So, it is slow going but it is getting done.

briansmith:
Oh, another thing I forgot in the previous comment: It is important not only to run `cargo test -vv --features=rsa_signing`, but also we must run the same commands with and without `--release`, because the release configuration is different. In particular, this is how I found the `NDEBUG` issue.

mk/top_of_makefile.mk
weiznich: ASFLAGS is not copied because it seems to be not used (As far as I can tell from the output of a Makefile based build)

mk/top_of_makefile.mk
```diff
@@ -1,245 +0,0 @@
...
-CMAKE_BUILD_TYPE ?= RELWITHDEBINFO
-
-# Although we don't use CMake, we use a variable CMAKE_BUILD_TYPE with similar
-# semantics to the CMake variable of that name.
-ifeq ($(CMAKE_BUILD_TYPE),MINSIZEREL)
-CPPFLAGS += -DNDEBUG -Os
-else ifeq ($(CMAKE_BUILD_TYPE),RELEASE)
-CPPFLAGS += -DNDEBUG -O3
-else ifeq ($(CMAKE_BUILD_TYPE),RELWITHDEBINFO)
-CPPFLAGS += -DNDEBUG -O3
```
weiznich: -DNDEBUG is now added if OPT_LEVEL != "0" otherwise -DDEBUG is added

weiznich:
I've rebased everything on master and addressed the comments. I had no time to do any in depth comparison between the output of the old and the new build scripts.

briansmith:
Still working on this. At this point there's no reason to update this PR any further. I've got a private branch that builds on this PR to get the macOS and Windows stuff working. Hopefully done in the next few days.

briansmith:
Thanks very much for contributing this! I landed a modified version of this in this series of commits:

86c24b9e6bef7a4d7312f0625ffdc54efc952100 Use only build.rs to build the native libraries for non-Windows builds.
fa55e87b31387d00da6daf0339973d4f78b18263 Pass `out_dir` by moving it in build.rs.
55ac79c5d92aabf7b9ffaeb302f693e425ec3fa7 Fix Windows build.
f5fc35a7e0abbce80912de250773b8b727af64a5 Build native libraries in parallel.
5a50c3c0a47b3a755aafa75a5ff809b4609d108f Remove Makefiles, which are now unused.
f72d7c4a063586a08fa9ed3bbf7bfdd7c55b9f9d Only run the build script if something changed.
65eeba2bc83315c1dbc168df33cacf7178ed75e0 Remove redundant `-fpic` C/C++ compiler option.

This seems to have been the fix for the macOS build failures:

5efd2ae4423528d74f8039dcf1d087459c3326ba Pass `-stdlib=c++` to Clang for Apple targets.

I also extended your PR to generate more output:
f66d6416572da00aac6562116846a5a8dce9f9cb Print command line for every command executed in build.rs like gcc-rs.

I also cleaned up some stuff (I actually pushed most of my cleanup stuff in commits earlier in the week):
5b4e29abdb8141a184b8d1c4291f97dae48aa0d0 Stop defining `BORINGSSL_IMPLEMENTATION`.

I also removed the target_build_utils dependency. It simply added too many dependencies and slowed down a clean build way too much. AFAICT all of its functionality is subsumed by recent versions of Cargo with the `CARGO_CFG_*` variables. That probably wasn't true when you originally wrote the PR though. Here's that commit:
5c77732c8233779928b4b289a9c04f9b1dbf4b8b Stop using target_build_utils.

Finally, because Windows is the primary platform I develop *ring* on, it is important to use the new build system on Windows right away, so I moved Windows over as well:
cfdc44ff79790596ede509c6861c7b9187fb6fae Generalize new PerlAsm logic in build.rs to work for -msvc, others.
b2ed2666e2de6425436bce8af2d64e631361e6df Stop using MSBuild on Windows.
0c8cb957d71aeb91fa1ab7bd92bc4b05dfceb753 Add more Visual-Studio-related stuff to .gitignore.

It's unlikely that things are perfect yet. I bet we broke some of the platforms we don't have on Travis CI like Redox (cc @jackpot51) and/or OpenBSD (cc @tatsuya6502) and/or iOS (cc @kali). My policy is to treat platforms that we don't have CI coverage for as "Tier 2" in that we don't back out changes that break them, but we accept bug fixes to get them working again, and we notify relevant people when we suspect we might have broken them.

And, I still need to go through the Windows build and add back some inessential compilation flags. I'll do that in the next few days.

Thanks again for the tremendous contribution here!

kali:
Wow, nice cleanup! I can confirm the tests are passing on ios on a real phone, compiled for armv7 and aarch64.

Building for the ios simulator is not workign though. I'll open a separate issue for that.

PR #324, opened 2016-10-29
Fix subkey order in chacha20-poly1305@openssh.com

kornholi:
[From the protocol docs:](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/PROTOCOL.chacha20poly1305?annotate=HEAD)

> 37: The first 256 bits consitute K_2 and the second 256 bits become
> 38: K_1.

Ran into this while moving away from libsodium in my SSH library. Any ideas why they chose this backwards order?

### DIFF
--- a/src/aead/chacha20_poly1305_openssh.rs
+++ b/src/aead/chacha20_poly1305_openssh.rs
@@ -137,14 +137,15 @@ struct Key {
 
 impl Key {
     pub fn new(key_material: &[u8; KEY_LEN]) -> Key {
+        // The first half becomes K_2 and the second half becomes K_1.
         Key {
             k_1: chacha::key_from_bytes(
                     slice_as_array_ref!(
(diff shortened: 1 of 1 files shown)


briansmith:
> Any ideas why they chose this backwards order?

I think the ordering in the spec makes some sense, because `K_2` is used like a normal AEAD and so one could use a normal AEAD API for it. It would be easier to understand if the names were swapped or if non-numeric names were used instead.

briansmith:
First, thanks!

Secondly, :sigh:. This is what I get for committing untested code.

/cc @pijul.

kornholi:
Did you close this by accident? :smile:

briansmith:
> Did you close this by accident? 😄

Yes. I seem to do it a lot. Apparently I'm not very compatible with the GitHub UI.

Thanks!

PR #325, opened 2016-10-29
Fix nonce encoding in chacha20-poly1305@openssh.com.

kornholi:
The sequence number is encoded in the SSH wire format, which is big-endian.

Together with the other PR, everything seems to be working fine now.


briansmith:
Thanks!

I've been working with @pijul on this and the interface is experimental and likely to change significantly, especially if/when we add new cipher suites. I'd love to hear your thoughts on the API: brian@briansmith.org.

PR #329, opened 2016-11-03
Use shared bytes-to-integer function in poly1305 (closes #133)

### LINKED ISSUE #133: Refactor `U8TO32_LITTLE` in ChaCha20 code
```
#define U8TO32_LITTLE(p)                              \
  (((uint32_t)((p)[0])) | ((uint32_t)((p)[1]) << 8) | \
   ((uint32_t)((p)[2]) << 16) | ((uint32_t)((p)[3]) << 24))
```

This should become `from_le_u32_ptr` in crypto/internal.h, alongside `from_be_u32_ptr` and friends. Also, we found that clang didn't optimize the generic code in `from_be_u32_ptr` well, so we should investigate whether we need to do the `memcpy` trick that we used for `from_be_u32_ptr`.

PeterReid:
This removes an endianness assumption.

### DIFF
--- a/crypto/poly1305/poly1305.c
+++ b/crypto/poly1305/poly1305.c
@@ -29,13 +29,6 @@
 #define POLY1305_BLOCK_STATE_SIZE 192
 
 
-/* We can assume little-endian. */
-static uint32_t U8TO32_LE(const uint8_t *m) {
-  uint32_t r;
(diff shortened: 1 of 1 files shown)


briansmith:
Thanks! This was committed as ace4d330ce92377d7526acbacb5391c0e15a5a96.

PR #344, opened 2016-11-14
Modifications to PR #262 made during review.

briansmith:
@samscott89 Could you please take a look at these changes I am proposing for your PR?

### DIFF
--- a/src/rsa/padding.rs
+++ b/src/rsa/padding.rs
@@ -13,6 +13,7 @@
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 use {bits, der, digest, error, polyfill};
+use super::PUBLIC_KEY_PUBLIC_MODULUS_MAX_LEN;
 use untrusted;
 
 #[cfg(feature = "rsa_signing")]
@@ -289,7 +290,7 @@ impl Verification for PSS {
         }
 
         // Step 7.
-        let mut db = [0u8; super::PUBLIC_MODULUS_MAX_LEN / 8];
+        let mut db = [0u8; PUBLIC_KEY_PUBLIC_MODULUS_MAX_LEN];
         let db = &mut db[..metrics.db_len];
 
         try!(mgf1(self.digest_alg, h_hash.as_slice_less_safe(), db));
--- a/src/rsa/verification.rs
+++ b/src/rsa/verification.rs
@@ -15,8 +15,8 @@
 /// RSA PKCS#1 1.5 signatures.
 
 use {bssl, c, error, private, signature};
-use super::{BIGNUM, PositiveInteger, PUBLIC_MODULUS_MAX_LEN, RSAParameters,
-            parse_public_key};
+use super::{BIGNUM, PositiveInteger, PUBLIC_KEY_PUBLIC_MODULUS_MAX_LEN,
+            RSAParameters, parse_public_key};
 use untrusted;
 
 
@@ -110,7 +110,7 @@ pub fn verify_rsa(params: &RSAParameters,
                   msg: untrusted::Input, signature: untrusted::Input)
                   -> Result<(), error::Unspecified> {
     let signature = signature.as_slice_less_safe();
-    let mut decoded = [0u8; (PUBLIC_MODULUS_MAX_LEN + 7) / 8];
+    let mut decoded = [0u8; PUBLIC_KEY_PUBLIC_MODULUS_MAX_LEN];
     if signature.len() > decoded.len() {
         return Err(error::Unspecified);
     }
@@ -121,7 +121,8 @@ pub fn verify_rsa(params: &RSAParameters,
     try!(bssl::map_result(unsafe {
         GFp_rsa_public_decrypt(decoded.as_mut_ptr(), decoded.len(), n.as_ref(),
                                e.as_ref(), signature.as_ptr(), signature.len(),
-                               params.min_bits, PUBLIC_MODULUS_MAX_LEN)
+                               params.min_bits,
+                               PUBLIC_KEY_PUBLIC_MODULUS_MAX_LEN * 8)
     }));
 
     untrusted::Input::from(decoded).read_all(
--- a/src/rsa/rsa.rs
+++ b/src/rsa/rsa.rs
@@ -24,8 +24,8 @@ pub use self::padding::{RSA_PKCS1_SHA256, RSA_PKCS1_SHA384, RSA_PKCS1_SHA512,
                         RSA_PSS_SHA256, RSA_PSS_SHA384, RSA_PSS_SHA512};
 
 
-// Maximum RSA modulus size supported for signature verification (in bits).
-const PUBLIC_MODULUS_MAX_LEN: usize = 8192;
+// Maximum RSA modulus size supported for signature verification (in bytes).
+const PUBLIC_KEY_PUBLIC_MODULUS_MAX_LEN: usize = 8192 / 8;
 
 /// Parameters for RSA verification.
 pub struct RSAParameters {


src/rsa/padding.rs
```diff
@@ -63,47 +63,48 @@ impl Encoding for PKCS1 {
...
 impl Verification for PKCS1 {
-    fn verify(&self, msg: untrusted::Input, m: untrusted::Input,
+    fn verify(&self, msg: untrusted::Input, m: &mut untrusted::Reader,
               _mod_bits: usize) -> Result<(), error::Unspecified> {
-        m.read_all(error::Unspecified, |em| {
-            if try!(em.read_byte()) != 0 ||
-               try!(em.read_byte()) != 1 {
-                return Err(error::Unspecified);
-            }
+        let em = m;
```
briansmith: :sigh: All of these changes are just supposed to be un-indenting and re-wrapping the code.

src/rsa/padding.rs
```diff
@@ -175,102 +176,100 @@ const PSS_PREFIX_ZEROS: [u8; 8] = [0u8; 8];
...
-            if (mod_bits - 1) % 8 == 0 {
-                if try!(m.read_byte()) != 0 {
-                    return Err(error::Unspecified);
-                }
-            };
-            let em = m;
-            let em_bits = mod_bits - 1;
-            let em_len = (em_bits + 7) / 8;
-            let top_byte_mask = 0xffu8 >> ((8 * em_len) - em_bits);
+        // RSASSA-PSS-VERIFY Step 2(c). The `m` this function is given is the
```
briansmith: Ditto. I'll verify that I didn't actually change the code in any significant way other than reformatting it.

src/rsa/pss_verify_tests.txt
```diff
@@ -0,0 +1,60 @@
+# Test vectors for RSA-PSS padding for variable public modulus lengths
```
samscott89: Outstanding tasks for this file:
 - Rename to `rsa_pss_verify_padding_tests.txt`.
 - Add failing cases.
 - Cover more digest types.

Would it be simpler to add these cases to the existing `rsa_pss_verify_tests.txt` file instead? The only reason I could see that these should have independent tests is if the padding functions were made public.

samscott89: I can make a start on adding failing cases.

src/rsa/verification.rs
```diff
@@ -103,10 +121,12 @@ pub fn verify_rsa(params: &RSAParameters,
...
         GFp_rsa_public_decrypt(decoded.as_mut_ptr(), decoded.len(), n.as_ref(),
                                e.as_ref(), signature.as_ptr(), signature.len(),
-                               params.min_bits, MAX_BITS)
+                               params.min_bits, PUBLIC_MODULUS_MAX_LEN)
     }));
 
-    params.padding_alg.verify(msg, untrusted::Input::from(decoded))
+    untrusted::Input::from(decoded).read_all(
+        error::Unspecified,
+        |m| params.padding_alg.verify(msg, m, n.length_in_bits()))
```
samscott89: Should the closing bracket be on the line below, at the same indentation level as line 127?

third-party/NIST/README.md
```diff
@@ -1,5 +1,8 @@
 # NIST Test Vectors
 
+[CAVP](CAVP) contains a number of test cases from the [NIST Individual Component
```
samscott89: These lines can be deleted since we not longer store the NIST tests here.

src/rsa/padding.rs
```diff
@@ -63,47 +68,48 @@ impl Encoding for PKCS1 {
...
-                    _ => {
-                        return Err(error::Unspecified);
-                    },
-                }
-            }
-            if ps_len < 8 {
-                return Err(error::Unspecified);
-            }
+        if try!(em.read_byte()) != 0 ||
+            try!(em.read_byte()) != 1 {
```
briansmith: I've now split it into two conditions so that the code coverage can see whether we're testing both checks.

src/rsa/padding.rs
```diff
@@ -159,3 +165,316 @@ pkcs1_digestinfo_prefix!(
...
+        try!(rng.fill(salt));
+
+        // Step 5 and 6.
+        let h_hash = pss_digest(self.digest_alg, msg, salt);
+
+        // Re-order steps 7,8, 9 and 10 so that we first output the db mask into
+        // the out buffer, and then XOR the value of db.
+
+        // Step 9. First output the mask into the out buffer.
+       let (mut masked_db, mut digest_terminator) =
```
briansmith: Fixed.

src/rsa/padding.rs
```diff
@@ -159,3 +165,316 @@ pkcs1_digestinfo_prefix!(
...
+            let b = try!(masked_bytes.read_byte());
+            if b & !metrics.top_byte_mask != 0 {
+                return Err(error::Unspecified);
+            }
+            db[0] ^= b;
+
+            // Step 8.
+            for i in 1..db.len() {
+                db[i] ^= try!(masked_bytes.read_byte());
+            }
```
briansmith: Yes, that's exactly what I want to happen.

src/rsa/padding.rs
```diff
@@ -159,3 +165,316 @@ pkcs1_digestinfo_prefix!(
...
+        if db[pad_len] != 1 {
+            return Err(error::Unspecified);
+        }
+
+        // Step 11.
+        let salt = &db[(db.len() - metrics.s_len)..];
+
+        // Step 12 and 13.
+        let h_prime = pss_digest(self.digest_alg, msg.as_slice_less_safe(),
+                                 salt);
```
briansmith: Fixed.

src/rsa/padding.rs
```diff
@@ -159,3 +165,316 @@ pkcs1_digestinfo_prefix!(
...
+    top_byte_mask: u8,
+}
+
+impl PSSMetrics {
+    fn new(digest_alg: &'static digest::Algorithm, mod_bits: usize)
+           -> Result<PSSMetrics, error::Unspecified> {
+        let em_bits = mod_bits - 1;
+        let em_len = (em_bits + 7) / 8;
+        let leading_zero_bits = (8 * em_len) - em_bits;
+        assert!(leading_zero_bits < 8);
```
briansmith: I changed it to a `debug_assert!`.

src/rsa/padding.rs
```diff
@@ -159,3 +165,316 @@ pkcs1_digestinfo_prefix!(
...
+    fn test_pss_verify() {
+
+        test::from_file("src/rsa/pss_verify_tests.txt",
+                        |section, test_case| {
+            assert_eq!(section, "");
+
+            let digest_name = test_case.consume_string("Digest");
+            let alg = match digest_name.as_ref() {
+                "SHA256" => &RSA_PSS_SHA256,
+                _ =>  { panic!("Unsupported digest: {}", digest_name) }
```
briansmith: I only wrote these patches to understand the code better and as penance for taking so long to review the code carefully. I'd be very happy if you would take these changes and build more improvements on top of them.

In particular, we do need to add (1) the negative test vectors, (2) padding tests for encoding, hopefully using the same test vectors as are used for padding verification, and (3) end-to-end tests for verification (in rsa_pss_verify_tests.txt) and encoding (in rsa_pss_sign_tests.txt) for non-even modulus lengths, particular {2048, 4096, 8192} ± {1, 7} bits.

src/rsa/signing.rs
```diff
@@ -212,11 +212,12 @@ impl RSASigningState {
...
                 rng: &rand::SecureRandom, msg: &[u8], signature: &mut [u8])
                 -> Result<(), error::Unspecified> {
-        if signature.len() != self.key_pair.public_modulus_len() {
+        let mod_len = self.key_pair.public_modulus_len();
+        if signature.len() != mod_len {
             return Err(error::Unspecified);
         }
 
-        try!(padding_alg.encode(msg, signature));
+        try!(padding_alg.encode(msg, signature, mod_len*8, rng));
```
briansmith: Fixed.

third-party/NIST/README.md
briansmith: Fixed.

briansmith:
@samscott89 Please check out the new commits I appended to this series to address your concerns. Let me know if you want me to squash them into the prior commits before or after you review them.

Pay special attention to the change I made to the PSS encoding, passing in the actual bit length of the public modulus instead of `8 * mod_len`. I think this was another bug, triggered by me saying that we only needed to support moduli that are even multiples of 8 for signing.

Also, do you want to take over the remaining tasks? I definitely don't want to steal this from you.

src/bits.rs
```diff
@@ -0,0 +1,60 @@
...
+// MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY
+// SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+// OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+use error;
+
+#[derive(Clone, Copy, Eq, PartialEq)]
+pub struct BitLength {
```
samscott89: Should this be called just `Length` since it will presumably always be used as `bits::Length` vs `bits::BitLength`?

Also, is this not practically the canonical example for using newtypes?

samscott89:
Yup, all looks good to me. The `BitLength` additional particularly seems like a good idea project-wide. Would it make sense to (in a separate PR) try and type all lengths? So that there is `ByteLength(usize)` and `BitLength(usize)` or something? Or is it okay to assume that all non-`BitLength` lengths are in bytes? 

> Also, do you want to take over the remaining tasks? I definitely don't want to steal this from you.

I'm more than happy to, and allow you to move onto something else. I think you've given me enough help at this point for me to get it across the finish line. I'm not worried about it being 'stolen' from me though, I'm just looking to learn/help as much as I can.

If you could squash it down to a reasonable number of commits (perhaps original commit + review or something?), that would be great. I'll then pull the `padding` branch and continue in the original PR.

briansmith:
> Yup, all looks good to me. The BitLength additional particularly seems like a good idea project-wide. 

I filed #350 for this.

> Would it make sense to (in a separate PR) try and type all lengths? So that there is ByteLength(usize) and BitLength(usize) or something? Or is it okay to assume that all non-BitLength lengths are in bytes?

I think we should just assume all non-BitLength lengths are in bytes. Otherwise we'd have to move ByteLength (at least) to untrusted.

> Also, do you want to take over the remaining tasks? I definitely don't want to steal this from you.
> I'm more than happy to, and allow you to move onto something else. I think you've given me enough help at this point for me to get it across the finish line. I'm not worried about it being 'stolen' from me though, I'm just looking to learn/help as much as I can.
> 
> If you could squash it down to a reasonable number of commits (perhaps original commit + review or something?), that would be great. I'll then pull the padding branch and continue in the original PR.

Will do.

src/bits.rs
briansmith: I agree it's weird to have `bits::BitLength` but I also think `bits::Length` or `bit::Length` is too clever. I found out that these names that only make sense with module prefixes don't work well in rustdoc (see `error::Unspecified` which rustdoc presents as just `Unspecified`, which is confusing).

I think we probably should create a "prelude" of common types that are used internally in *ring* and use those types (only) unprefixed. I think `BitLength` is a prime candidate for that.

briansmith: Regarding newtypes, see #350. The problem is that we have to constructors: from bits, and from bytes. A newtype would make the from-bits constructor implicit, but I'd rather have the constructors be used explicitly. However, we may want to switch to newtype style if it helps work around the lack of `const fn`.

briansmith:
I squashed and landed all these changes onto master. See discussion in #262.

PR #349, opened 2016-11-15
Expose `RSAEncoding` in the public API.

briansmith:
Expose `RSAEncoding` in the public API so that users of *ring* can use
it in type signatures.

/cc @bt


briansmith:
This landed with some changes in master as f0f9b00e1aa2bdc8b2fa9de1ff608487be170e88.

PR #351, opened 2016-11-15
Port C code supporting poly1305 to Rust

PeterReid:


src/poly1305.rs
```diff
@@ -16,31 +16,86 @@
...
 
 use {c, chacha, constant_time, error};
-
-pub struct SigningContext {
-    state: State,
-}
+use polyfill::slice::u32_from_le_u8;
+use std::cmp::min;
+use std::mem::align_of;
+use std::os::raw::c_void;
```
briansmith: 1. Please use `core` instead of `std` whenever possible.
2. `use core;` and then refer to these qualified as `core::cmp::min()`, etc.

crypto/poly1305/poly1305.c
```diff
@@ -1,142 +0,0 @@
...
-
-#if defined(OPENSSL_X86)
-/* See comment above |_poly1305_init_sse2| in poly1305-x86.pl. */
-OPENSSL_COMPILE_ASSERT(POLY1305_BLOCK_STATE_SIZE >= 4 * (5 + 1 + 4 + 2 + 4 * 9),
-                       poly1305_block_state_too_small);
-#elif defined(OPENSSL_X86_64)
-/* See comment above |__poly1305_block| in poly1305-x86_64.pl. */
-OPENSSL_COMPILE_ASSERT(POLY1305_BLOCK_STATE_SIZE >=
-                           4 * (5 + 1 + 2 * 2 + 2 + 4 * 9),
-                       poly1305_block_state_too_small);
```
briansmith: I would prefer to preserve these calculations/assertions of `BLOCK_STATE_SIZE` and the comments referencing how they are derived.

crypto/poly1305/poly1305.c
```diff
@@ -1,142 +0,0 @@
...
-#elif defined(OPENSSL_ARM)
-/* TODO(davidben): Figure out the layout of the struct. For now,
- * |POLY1305_BLOCK_STATE_SIZE| is taken from OpenSSL. */
-#elif defined(OPENSSL_AARCH64)
-/* TODO(davidben): Figure out the layout of the struct. For now,
- * |POLY1305_BLOCK_STATE_SIZE| is taken from OpenSSL. */
```
briansmith: Same with these TODO comments.

src/poly1305.rs
```diff
@@ -87,27 +142,69 @@ pub type Tag = [u8; TAG_LEN];
...
+             (self.emit_fn)(
+                 state.as_mut_ptr() as *mut c_void,
+                 tag.as_mut_ptr(),
+                 nonce.as_ptr()
+             );
+        }
+    }
+}
+
+#[repr(C)]
```
briansmith: I don't think this actually needs to be `#[repr(C)]`, does it? It seems like maybe only `Funcs` needs to be `#[repr(C)`]? If SigningContext doesn't need to be `#[repr(C)]` then you can use more native Rust types, such as `usize` instead of `c::int`, and in particular you can get rid of most/all casts in the code.

src/poly1305.rs
```diff
@@ -87,27 +142,69 @@ pub type Tag = [u8; TAG_LEN];
...
+    buf: [u8; 16],
+    buf_used: c::int,
+    func: Funcs
+}
 
 extern {
-    fn GFp_poly1305_init(state: &mut State, key: &KeyBytes);
-    fn GFp_poly1305_finish(state: &mut State, mac: &mut Tag);
-    fn GFp_poly1305_update(state: &State, in_: *const u8, in_len: c::size_t);
+    fn GFp_poly1305_init_asm(state: *mut c_void, key: *const u8, out_func: *mut c_void) -> c::int;
```
briansmith: if `state` needs to be 64-bit aligned then we should note it by defining a type named `State` as an alias for `[u64; (BLOCK_STATE_SIZE + 7) / 8]` with a comment explaining it needs to be aligned.

Also, does the key need to be 64-bit aligned?

I believe you can write this as:
```rust
    fn GFp_poly1305_init_asm(state: &mut State, key: &KeyBytes, out_func: &mut Func) -> c::int;
```

Please document what the return type signifies.

And similar for the other functions.

src/poly1305.rs
```diff
@@ -87,27 +142,69 @@ pub type Tag = [u8; TAG_LEN];
...
+                 tag.as_mut_ptr(),
+                 nonce.as_ptr()
+             );
+        }
+    }
+}
+
+#[repr(C)]
+pub struct SigningContext {
+    _align: [u64; 0],
```
PeterReid: Yup.

src/poly1305.rs
PeterReid: Yup, you are correct.

PeterReid:
It looks like assuming that u64s are 8-byte aligned is not a safe assumption on some platforms.

briansmith:
> It looks like assuming that u64s are 8-byte aligned is not a safe assumption on some platforms.

You might try using doubles (`f64`) instead. That's common in C code.

briansmith:
You use the constant `16`. Let's create a named constant named `BLOCK_LEN` and use that everywhere instead of `16`.

Please check out https://github.com/briansmith/ring/commits/poly1305-blocks. It shows that it is OK to call the `blocks` function with an empty input, at least on x86_64, and so probably on all platforms. If so, you can avoid the `input.len() >= 16` checks.

I also included a sketch of a different way of dealing with partial blocks, by changing the poly1305 interface such that it always operates on full blocks. Up to you if you want to switch to that way; the primary advantage short-term is some code simplification. Later, we may be able to simplify/optimize the assembly code to take advantage of the fact that all the inputs are full blocks.

PeterReid:
I updated the pull request to address your comments.

For alignment, rather than trying to trick the compiler into aligning the struct and hoping that worked, I made it do the pointer math to align the memory itself. The original C code memcpy'ed that block state onto the stack anyway, so that shouldn't be a performance regression.

I didn't roll your poly1305-blocks changes into this. I suspect it will be useful at some point to provide an interface that lets the user stream in the data to run poly1305 over, rather than having it all in memory at once, at which point that buffering would be needed again. (Also, the assembly already operates only on full blocks, unless I'm misunderstanding something.)

src/poly1305.rs
```diff
@@ -16,31 +16,115 @@
...
+    inner: &'a mut [u8; BLOCK_STATE_SIZE],
+}
+impl<'a> AlignedBlockState<'a> {
+    pub fn new(inner: &mut [u8; BLOCK_STATE_SIZE]) -> AlignedBlockState {
+        let mut st = AlignedBlockState {
+            buf: [0u8; BLOCK_STATE_SIZE + 7],
+            inner: inner,
+        };
+        AlignedBlockState::align(&mut st.buf).copy_from_slice(&st.inner[..]);
+        st
```
briansmith: This will construct an `AlignedBlockState` named `st`, calculate the offset, and then return a copy of it. But, the returned copy might not have the same alignment as `st`, so the offset might be wrong, right?

src/poly1305.rs
```diff
@@ -16,31 +16,115 @@
...
         ctx
     }
 
-    #[inline]
-    pub fn update(&mut self, input: &[u8]) {
-        unsafe {
-            GFp_poly1305_update(&mut self.state, input.as_ptr(), input.len());
+    pub fn update(&mut self, mut input: &[u8]) {
+        let mut block_state = AlignedBlockState::new(&mut self.opaque);
+        let block_state_buf = block_state.aligned_buf();
```
briansmith: I suggest that we do something like this:
```rust
let mut state = AlignedBlockState::new();
state.load_execute_save(&mut self.opaque, |state| {
    // use the state.
}
```

The first line will fix the address of `state` and the second line would then copy `self.opaque` into `state` at the appropriate offset to align it. Then it would execute the closure, passing it the aligned pointer. Then after the closure completes, it would copy the updated state back into `self.opaque`. This would be equivalent to the `Drop` stuff but IMO much more clear since it is explicit.

src/poly1305.rs
```diff
@@ -16,31 +16,115 @@
...
+impl<'a> AlignedBlockState<'a> {
+    pub fn new(inner: &mut [u8; BLOCK_STATE_SIZE]) -> AlignedBlockState {
+        let mut st = AlignedBlockState {
+            buf: [0u8; BLOCK_STATE_SIZE + 7],
+            inner: inner,
+        };
+        AlignedBlockState::align(&mut st.buf).copy_from_slice(&st.inner[..]);
+        st
+    }
+    fn align(xs: &mut [u8; BLOCK_STATE_SIZE + 7]) -> &mut [u8; BLOCK_STATE_SIZE] {
```
briansmith: please add a `debug_assert!(BLOCK_STATE_SIZE >= 7);` here, since this code only works when that condition is true

src/poly1305.rs
```diff
@@ -16,31 +16,115 @@
...
+        let aligned_start = (xs.as_ptr() as usize + 7) & !7;
+        let offset = aligned_start - (xs.as_ptr() as usize);
+        slice_as_array_ref_mut!(&mut xs[offset..offset+BLOCK_STATE_SIZE], BLOCK_STATE_SIZE).unwrap()
+    }
+    pub fn aligned_buf(&mut self) -> &mut [u8; BLOCK_STATE_SIZE] {
+        AlignedBlockState::align(&mut self.buf)
+    }
+}
+impl<'a> Drop for AlignedBlockState<'a> {
```
briansmith: I would rather just have the code explicitly copy the state from the temporarily aligned copy to the context's `opaque` when it is done. This `Drop` business is not explicit enough.

src/poly1305.rs
```diff
@@ -16,31 +16,115 @@
...
 }
 
 impl SigningContext {
     #[inline]
     pub fn from_key(key: Key) -> SigningContext {
-        let mut ctx = SigningContext {
-            state: [0u8; STATE_LEN],
+        let mut ctx = SigningContext{
+            opaque: [0; BLOCK_STATE_SIZE],
+            nonce: [0; 4],
```
briansmith: Let's initialize nonce here with the values from the `u32_from_le_u8(...)`.

crypto/poly1305/poly1305.c
```diff
@@ -1,142 +0,0 @@
...
-#endif
-
-int GFp_poly1305_init_asm(void *ctx, const uint8_t key[16], void *out_func);
-void GFp_poly1305_blocks(void *ctx, const uint8_t *in, size_t len,
-                         uint32_t padbit);
-void GFp_poly1305_emit(void *ctx, uint8_t mac[16], const uint32_t nonce[4]);
-
-struct poly1305_state_st {
-  alignas(8) uint8_t opaque[POLY1305_BLOCK_STATE_SIZE];
-  uint32_t nonce[4];
```
briansmith: Note that since `opaque` is 8-byte aligned and its size is a multiple of 8, `nonce` is implicitly 8-byte aligned too. Did you check the asm code to see if it expects `nonce` to be aligned?

src/poly1305.rs
```diff
@@ -16,31 +16,115 @@
...
+
+        if self.buf_used != 0 {
+            self.buf[self.buf_used] = 1;
+            for byte in &mut self.buf[self.buf_used+1..] {
+                *byte = 0;
+            }
+            self.func.blocks(block_state_buf, &self.buf[..], 0 /* already padded */);
+        }
+
+        self.func.emit(block_state_buf, tag_out, &self.nonce);
```
briansmith: This is what I'm wondering about: Does any implementation of `emit` require the nonce to be aligned?

src/poly1305.rs
```diff
@@ -16,31 +16,115 @@
...
+    }
 }
 
 impl SigningContext {
     #[inline]
     pub fn from_key(key: Key) -> SigningContext {
-        let mut ctx = SigningContext {
-            state: [0u8; STATE_LEN],
+        let mut ctx = SigningContext{
+            opaque: [0; BLOCK_STATE_SIZE],
```
briansmith: Almost all the occurrences of `[0; BLOCK_STATE_SIZE]` should be `BlockState`.

src/poly1305.rs
```diff
@@ -87,27 +191,62 @@ pub type Tag = [u8; TAG_LEN];
...
 pub const TAG_LEN: usize = 128 / 8;
 
-type State = [u8; STATE_LEN];
-const STATE_LEN: usize = 256;
+const BLOCK_STATE_SIZE: usize = 192;
+
+const BLOCK_LEN: usize = 16;
+
+/// The memory manipulated by the assembly. Using u64s ensures 8-byte alignment,
+/// which some of the assembly implementations need.
```
briansmith: The comment isn't accurate any more.

src/poly1305.rs
```diff
@@ -87,27 +191,62 @@ pub type Tag = [u8; TAG_LEN];
 /// The length of a `Tag`.
 pub const TAG_LEN: usize = 128 / 8;
 
-type State = [u8; STATE_LEN];
-const STATE_LEN: usize = 256;
+const BLOCK_STATE_SIZE: usize = 192;
```
briansmith: Let's use `_LEN` for lengths.

Question: Why do you prefix these names with `BLOCK_`? Why not `STATE_LEN` and `State`?

src/poly1305.rs
```diff
@@ -57,6 +141,26 @@ pub fn sign(key: Key, msg: &[u8], tag: &mut Tag) {
     ctx.sign(tag)
 }
 
+#[inline]
+fn check_block_state_layout() {
```
briansmith: This can be done as a test, right? I don't think we need to check this at runtime outside of tests.

crypto/poly1305/poly1305.c
PeterReid: It does not appear to expect the nonce to be aligned.

src/poly1305.rs
PeterReid: Yup that is a mistake! I am switching to the callback-based approach you described below.

PeterReid:
I've switched it to use function with a callback instead of that struct for aligning things.

PeterReid:
There were a few of failures, but I don't think any of them are related to poly1305:
- [aead::aes_gcm::tests::test_aes got the wrong answer](https://travis-ci.org/briansmith/ring/jobs/179320508) for stable arm-unknown-linux-gnueabihf.
- [aead::aes_gcm::tests::test_aes_gcm_256](https://travis-ci.org/briansmith/ring/jobs/179320479) timed out for aarch64-unknown-linux-gnu.
- Nightly on windows failed uniformly.
- Two rustc installs failed.

briansmith:
Thanks! I landed this as 2070f2a2e67d7be154d240a29c61bdadf22638aa.

PR #352, opened 2016-11-16
Expanded PSS padding test cases

samscott89:
This expands out the PSS *padding* tests.  This includes various public modulus lengths, and failing cases to cover the various checks that `verify` performs.

The valid PSS encodings are also used to test the PSS `encode` function.

These are complementary to the RSA-PSS signing/verification tests.


samscott89:
These tests cases were output by a custom program written using the implementation of the `emsaPSSEncode` function in [Go](https://github.com/golang/go/blob/master/src/crypto/rsa/pss.go#L20).

briansmith:
> These tests cases were output by a custom program written using the implementation of the emsaPSSEncode function in Go.

Can you please include the Go program in the submission?

I guess it is a good idea to test the message "" but the problem with using "" for all tests is that it doesn't test whether you actually remembed to call `Digest::update()` to digest the value. So I guess another non-empty value like "test" should be used for the test vectors that aren't testing that "" isn't problematic. (Sorry for suggesting to use "" in the first place.)

samscott89:
> Can you please include the Go program in the submission?

Can do. I didn't for a couple of reasons:
 - I've never written/looked at Go before, the code isn't very pretty
 - The encoding functions needed aren't publicly exposed by the Go library, so they are copy-pasted.

Unfortunately the last point seems common across most libraries. But I guess it raises licensing issues?
Alternatively, Go themselves use a few test vectors from various sources:
https://github.com/golang/go/blob/master/src/crypto/rsa/pss_test.go which could potentially be adapted (none of them seem to contain failing cases).

Anyway, the code I used is here:
https://gist.github.com/samscott89/b4ac7de8a7c8de77493dc14780dfeec8

briansmith:
> But I guess it raises licensing issues?

OK, understood. It is OK not to include it.

src/rsa/rsa_pss_padding_tests.txt
```diff
@@ -1,60 +1,703 @@
...
...5a24f2f0b388a5e1082babf37f7d7f60c566e4452dd03fe9153f009aafc0cda10a01cf9f72a8754f618fc0b67de8b2a0cd2ba9e6c241267dcee3b04f26f23fbfa3ace11e8f1a6ca6627efe30ad1973a42983a5e29c1646ebc3a50dd34e5f4b7a28a190cd8998d55ace1f46f8c23adf039ce5b9f0eea14b2d158076d194de8d15ac28d8b907002252e224e53cf3f6d4eda63f761395a9178d0d0c3d34e4db38d4574818ee596ac34bbcdf861f13ef86ad0e8e96cd5a7c6c693676ccc625e1208c59723631f216c1155eeac9205705aad83b01a11926d8ca477d047449d3fa2068660cdc61bef02439aea352bd32da7d2e5cf2ddaccdcd6a6eb7492f3b5af9a0bdf93427ef99218bef4d72588f943bd9b8f2a404ba061feb1d48c8430b3f2fe5de5c4fe9af5a48359bd91fa5d93a4d83d96b27177769f91e1d1cb1959790e607f9d30ee44339bc4da863eb6fe8b7457b533c5e22e0bc4855a20229c7ecd193e917cb695b26a84d721f4707dac53e9485234ebeed7bb505f0fa1393aff820462bdd1e26c14c541e551ab6ffb9d7b7e280e33f7b2ec30eb14f4d635a57fb568bf9c420f197673a5cb4d8b15f9e8c743af99c2c2899e7d15d877590501644aa863c20b392e72bdc330ad192147ef171516d815c02fe0e375317b76c9df2052bd2de1b116192785b2282119f875dd89ffaf5a5c36f0524b3a4376f0381184591473a7c131d5f58bbb00b921030a96f260ee78cf1c2c25337e69b4cc42fcbd85034bd41f2c62e990fdd298fd53b453a4232b61426ac0764a5d6b8c45fb1d3004e27a1afc5ed38cf2ec4ab70521e6af5c1a2cf22306c523f87100c14
```
briansmith: I suggest "F (PS is not all zeros)", since PS is the term used in the RFC. Or, even better would be to say "first byte of PS is not zero" or whatever the problem is.

I noticed there are no tests that the last byte of the padding is 0x01, from looking at the coverage report: https://coveralls.io/builds/8871076/source?filename=src%2Frsa%2Fpadding.rs#L342. 
It would be great to add such test vectors, since that would give padding.rs 100% code coverage by line!

briansmith: What's the difference between this test vector and the previous one? My guess is that the error in `PS` is in a different place. It would be great to have a more specific note as to the problem, that differentiates it from the previous test vector.

(It makes sense to verify that we check that the first byte of PS is zero and that the last byte of PS is zero in different tests, so I hope that's what these vectors are doing.)

briansmith:
@samscott89 Will you have time to respond to the review comments soon? I think most of the suggestions aren't hard to address and I could take care of them or defer some of them, depending on what you think. LMK.

samscott89:
I'll try to take care of them quickly now.

briansmith:
Awesome. Thanks a ton for doing this.

Landed as 7110b191c622daac245a8344fac9df2f478f9850 and 46b7a2b32e506a21709fa7096a6e392dbb652918.

Issue #353, opened 2016-11-17
Improve performance of `ring::rand` using getrandom() for 257-byte outputs


briansmith:
Currently we use the fact that `getrandom()` will never return a partial result if the requested output is 256 bytes or less, as specified in the getrandom man page. In particular, if we need more than 256 bytes of output then issue multiple syscalls for (up to) 256 bytes. I wrote that code on the assumption that we'd never need more than 256 bytes of output in a performance-sensitive situation.

However, the new RSA blinding code from #283 often needs 257 bytes of output, so the assumption is now wrong. Therefore, we should change the code to ask for all the random bytes at once in the getrandom syscall, handling the case where the syscall returns only a partial result by looping.

kimhyunkang:
I'm trying to start contributing to ring, and I'd like to take this.

briansmith:
> I'm trying to start contributing to ring, and I'd like to take this.

Awesome!

briansmith:
Done in #355.

PR #354, opened 2016-11-17
Allow calling `getrandom` with 257-byte buffer

kimhyunkang:
Fixed issue #353.

Now `rand::sysrand::fill` does not loop with fixed 256-byte chunks,
but calls `getrandom` to fill the whole buffer.

### DIFF
--- a/crypto/rand/sysrand.c
+++ b/crypto/rand/sysrand.c
@@ -33,6 +33,7 @@
 #include <openssl/base.h>
 
 #include <assert.h>
+#include <limits.h>
 #include <stddef.h>
 
 /* GFp_sysrand_chunk fills |len| bytes at |buf| with entropy from the
@@ -42,8 +43,6 @@
 int GFp_sysrand_chunk(void *buf, size_t len);
 
 #if defined(OPENSSL_WINDOWS)
-
-#include <limits.h>
 
 #pragma warning(push, 3)
 
@@ -59,7 +58,14 @@ int GFp_sysrand_chunk(void *buf, size_t len);
 #pragma warning(pop)
 
 int GFp_sysrand_chunk(void *out, size_t requested) {
-  return RtlGenRandom(out, (ULONG)requested) ? 1 : 0;
+  if (requested > ULONG_MAX) {
+    requested = ULONG_MAX;
+  }
+  if (requested > (size_t) INT_MAX) {
+    requested = INT_MAX;
+  }
+
+  return RtlGenRandom(out, (ULONG)requested) ? (int) requested : 0;
 }
 
 #elif defined(__linux__)
@@ -86,6 +92,13 @@ int GFp_sysrand_chunk(void *out, size_t requested) {
 #endif
 
 int GFp_sysrand_chunk(void *out, size_t requested) {
+  if (requested > ULONG_MAX) {
+    requested = ULONG_MAX;
+  }
+  if (requested > (size_t) INT_MAX) {
+    requested = INT_MAX;
+  }
+
   int r = syscall(SYS_getrandom, out, requested, 0u);
   if (r < 0) {
     // EINTR and EAGAIN are normal, and we can try again. Other error codes are fatal.
--- a/crypto/rand/sysrand.c
+++ b/crypto/rand/sysrand.c
@@ -59,7 +59,6 @@ int GFp_sysrand_chunk(void *buf, size_t len);
 #pragma warning(pop)
 
 int GFp_sysrand_chunk(void *out, size_t requested) {
-  assert(requested <= GFp_sysrand_chunk_max_len);
   return RtlGenRandom(out, (ULONG)requested) ? 1 : 0;
 }
 
--- a/crypto/rand/sysrand.c
+++ b/crypto/rand/sysrand.c
@@ -101,8 +101,8 @@ int GFp_sysrand_chunk(void *out, size_t requested) {
 
   int r = syscall(SYS_getrandom, out, requested, 0u);
   if (r < 0) {
-    // EINTR and EAGAIN are normal, and we can try again. Other error codes are fatal.
-    if (errno == EINTR || errno == EAGAIN) {
+    // EINTR is normal, and we can try again. Other error codes are fatal.
+    if (errno == EINTR) {
       return 0;
     }
     return -1;
--- a/crypto/rand/sysrand.c
+++ b/crypto/rand/sysrand.c
@@ -58,8 +58,6 @@ int GFp_sysrand_chunk(void *buf, size_t len);
 
 #pragma warning(pop)
 
-const size_t GFp_sysrand_chunk_max_len = ULONG_MAX;
-
 int GFp_sysrand_chunk(void *out, size_t requested) {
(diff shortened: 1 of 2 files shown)


crypto/rand/sysrand.c
```diff
@@ -88,21 +86,16 @@ int GFp_sysrand_chunk(void *out, size_t requested) {
...
-
 int GFp_sysrand_chunk(void *out, size_t requested) {
-  assert(requested <= GFp_sysrand_chunk_max_len);
-  if (syscall(SYS_getrandom, out, requested, 0u) < 0) {
-    if (errno == ENOSYS) {
-      return -1;
+  int r = syscall(SYS_getrandom, out, requested, 0u);
+  if (r < 0) {
+    // EINTR and EAGAIN are normal, and we can try again. Other error codes are fatal.
+    if (errno == EINTR || errno == EAGAIN) {
```
briansmith: We don't need to handle `EAGAIN` since we're not using the non-blocking flag, right? We can just treat it like other errors, I think.

src/rand.rs
```diff
@@ -119,14 +119,19 @@ use self::sysrand_or_urandom::fill as fill_impl;
...
-                super::GFp_sysrand_chunk(chunk.as_mut_ptr(), chunk.len())
-            }));
+        let mut read_len = 0;
+        while read_len < dest.len() {
+            let r = unsafe {
+                super::GFp_sysrand_chunk(dest[read_len ..].as_mut_ptr(), dest.len() - read_len)
+            };
+            if r < 0 {
+                return Err(error::Unspecified);
+            } else {
```
briansmith: The `else {` here isn't needed and should be removed.

Please add a comment here like this: "XXX: If r == 0 then this is a busy wait loop waiting for the
kernel entropy pool to be initialized."

src/rand.rs
```diff
@@ -327,18 +331,9 @@ mod tests {
             // case in the `SysRandOrDevURandom::Undecided` case. As above, we
             // only do this when there are at least 96 bytes after the first
             // chunk to avoid false positives.
-            if *len > 96 && *len - 96 > max_chunk_len() {
-                assert!(buf[max_chunk_len()..].iter().any(|x| *x != 0));
+            if *len > 256 + 96 {
+                assert!(buf[256..].iter().any(|x| *x != 0));
             }
```
briansmith: I guess this test just doesn't make sense any more, does it? Maybe we should just remove this

kimhyunkang: Maybe we could check the buffer is filled until the last byte. What if we check the last 96 bytes of the buffer has a non-zero byte?

crypto/rand/sysrand.c
```diff
@@ -58,11 +57,15 @@ int GFp_sysrand_chunk(void *buf, size_t len);
...
 
-const size_t GFp_sysrand_chunk_max_len = ULONG_MAX;
-
 int GFp_sysrand_chunk(void *out, size_t requested) {
-  assert(requested <= GFp_sysrand_chunk_max_len);
-  return RtlGenRandom(out, (ULONG)requested) ? 1 : 0;
+  if (requested > ULONG_MAX) {
+    requested = ULONG_MAX;
+  }
+  if (requested > (size_t) INT_MAX) {
```
briansmith: Nit: No space between the cast and the value, according to the BoringSSL style guide.

crypto/rand/sysrand.c
```diff
@@ -58,11 +57,15 @@ int GFp_sysrand_chunk(void *buf, size_t len);
...
+    requested = INT_MAX;
```
briansmith: NIT: A cast is required here.

crypto/rand/sysrand.c
```diff
@@ -58,11 +57,15 @@ int GFp_sysrand_chunk(void *buf, size_t len);
...
+  }
+
+  return RtlGenRandom(out, (ULONG)requested) ? (int) requested : 0;
```
briansmith: NIT: No space between the cast and the value.

crypto/rand/sysrand.c
```diff
@@ -88,21 +91,23 @@ int GFp_sysrand_chunk(void *out, size_t requested) {
...
-  assert(requested <= GFp_sysrand_chunk_max_len);
-  if (syscall(SYS_getrandom, out, requested, 0u) < 0) {
-    if (errno == ENOSYS) {
-      return -1;
+  if (requested > ULONG_MAX) {
+    requested = ULONG_MAX;
+  }
+  if (requested > (size_t) INT_MAX) {
+    requested = INT_MAX;
+  }
```
briansmith: I don't think the above clamping is useful for the Linux variant. The man page says that the length parameter is `size_t` and the function returns `int` so it must already do this clipping internally.

src/rand.rs
```diff
@@ -119,14 +119,20 @@ use self::sysrand_or_urandom::fill as fill_impl;
...
     pub fn fill(dest: &mut [u8]) -> Result<(), error::Unspecified> {
-        let chunk_len = unsafe { super::GFp_sysrand_chunk_max_len };
-        for mut chunk in dest.chunks_mut(chunk_len) {
-            try!(bssl::map_result(unsafe {
-                super::GFp_sysrand_chunk(chunk.as_mut_ptr(), chunk.len())
-            }));
+        let mut read_len = 0;
+        while read_len < dest.len() {
+            let r = unsafe {
+                super::GFp_sysrand_chunk(dest[read_len ..].as_mut_ptr(), dest.len() - read_len)
```
briansmith: NIT: wrap this at 80 characters like the rest of the file.

src/rand.rs
```diff
@@ -119,14 +119,20 @@ use self::sysrand_or_urandom::fill as fill_impl;
...
+            };
+            if r < 0 {
+                return Err(error::Unspecified);
+            }
+            // XXX: If r == 0 then this is a busy wait loop waiting for the
+            //      kernel entropy pool to be initialized
```
briansmith: NIT: Don't indent the second line of the comment by the extra width of "XXX:",
NIT: End the sentence with a period.

src/rand.rs
```diff
@@ -323,22 +328,13 @@ mod tests {
...
             // Make sure we didn't forget to finish filling in the rest of the
-            // buffer after we filled in the first chunk, especially in the
+            // buffer after we filled in the first chunk.
             // case in the `SysRandOrDevURandom::Undecided` case. As above, we
             // only do this when there are at least 96 bytes after the first
             // chunk to avoid false positives.
-            if *len > 96 && *len - 96 > max_chunk_len() {
-                assert!(buf[max_chunk_len()..].iter().any(|x| *x != 0));
+            if *len > 96 {
+                assert!(buf[buf.len() - 96 ..].iter().any(|x| *x != 0));
```
briansmith: I think we should just get rid of this. It really is about max_chunk_len() and it doesn't make sense now.

kimhyunkang:
@briansmith I'll close this PR and submit a new one with the squashed commits.

I'm contributing this as an individual, and I'm perfectly OK with your license.

PR #355, opened 2016-11-19
Allow calling `getrandom` with 257-byte buffer

kimhyunkang:
Fixes issue #353.

Now `rand::sysrand::fill` does not loop with fixed 256-byte chunks,
but calls `getrandom` to fill the whole buffer.


crypto/rand/sysrand.c
```diff
@@ -58,11 +58,15 @@ int GFp_sysrand_chunk(void *buf, size_t len);
 
 #pragma warning(pop)
 
-const size_t GFp_sysrand_chunk_max_len = ULONG_MAX;
-
 int GFp_sysrand_chunk(void *out, size_t requested) {
```
briansmith: See below. the return type needs to be `long` and we need to use `LONG_MAX` and `long` everywhere here.

crypto/rand/sysrand.c
```diff
@@ -88,21 +92,16 @@ int GFp_sysrand_chunk(void *out, size_t requested) {
...
- * getrandom() to read /dev/urandom for small values (<= 256) of buflen is
- * the preferred mode of usage." */
-const size_t GFp_sysrand_chunk_max_len = 256;
-
 int GFp_sysrand_chunk(void *out, size_t requested) {
-  assert(requested <= GFp_sysrand_chunk_max_len);
-  if (syscall(SYS_getrandom, out, requested, 0u) < 0) {
-    if (errno == ENOSYS) {
-      return -1;
+  int r = syscall(SYS_getrandom, out, requested, 0u);
```
briansmith: I noticed in the similar BoringSSL patch they are using `long`. And, according to http://man7.org/linux/man-pages/man2/syscall.2.html, `long` is the right type to use. That means that the function return type needs to be `long` and the FFI declaration in the Rust code needs to use `c::long`. I just added `c::long` in 5c4627a8499e3489a466e280bb606fbed6cc3e67.

src/rand.rs
```diff
@@ -217,7 +224,6 @@ pub unsafe extern fn RAND_bytes(rng: *mut RAND, dest: *mut u8,
 
 #[cfg(any(target_os = "linux", windows))]
 extern {
-    static GFp_sysrand_chunk_max_len: c::size_t;
     fn GFp_sysrand_chunk(buf: *mut u8, len: c::size_t) -> c::int;
```
briansmith: `-> c::long`.

crypto/rand/sysrand.c
briansmith: Also, according to http://man7.org/linux/man-pages/man2/getrandom.2.html, the flags argument type is `unsigned int`. I suggest that we write this:
```unsigned int NO_FLAGS = 0;
    syscall(..., NO_FLAGS);
```
This way, we are sure there's no funny business going on.

briansmith: I forgot that C's `long` is different on Windows than on other platforms. I pushed a fix for it: dcba8a845f645a4676eeaa1d57de827fce8aa94e

kimhyunkang:
@briansmith : I just updated the return types from `int` to `long`

briansmith:
Thanks for doing this! I merged this as 828e20fe713853b7587f7ea5493a680f6ae8c18e.

PR #358, opened 2016-11-22
Verify RSA PKCS#1 1.5 signatures by encode-then-check.

briansmith:
When we first implemented PKCS#1 1.5 signature verification we had not
implemented signature generation, so we implemented verification by
parsing the padding. Now that we have generation we can save some code
and arguably make verification safer. Also, this is the way RFC 3447
recommends to do it.

### DIFF
--- a/src/polyfill.rs
+++ b/src/polyfill.rs
@@ -17,6 +17,11 @@
 
 use core;
 
+// A better name for the `&*` idiom for removing the mutability from a
+// reference.
+#[inline(always)]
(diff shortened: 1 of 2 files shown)


briansmith:
@samscott89 WDYT?

samscott89:
Seems like a pretty clear win on all fronts.

briansmith:
Thanks! Landed in a38d9a4ef98f028dc939413e5a803a61e1f72e7b.

PR #360, opened 2016-11-28
Use Rust code to construct and check RSA private keys

briansmith:


briansmith:
The current version of this PR is a sketch of what the new RSA private key construction code should look like. It also moves *ring* towards doing the RSA signing operation in Rust (it is currently done in C by `GFp_rsa_private_transform`).

It contains a very lightweight static capability mechanism to help typecheck modular arithmetic. For example, the modulus for mod *n* arithmetic is of type `Modulus<N>`, the modulus for mod *p* arithmetic is of type `Modulus<P>`, etc. Similarly the multiplication is defined over elements of types  `Elem<N>`, `Elem<P>`, etc. so that the compiler only allows us to multiple two elements of the same field modulo the modulus for that field. It is neat, but I'm not sure how useful it is, so I might scrap it.

The naming of things is based on the naming for the similar field arithmetic code in src/ec/suite_b/ops. At some point in the future we might be able to define some traits that abstract over both the RSA and the ECC code.

There is still a substantial amount of cleanup to do, but it would be good to get some early feedback.

After this, and after `GFp_rsa_private_transform` and `GFp_rsa_public_decrypt` are replaced, we'll likely be able to make a substantial simplification (removal) of `BIGNUM` C code.

/cc @ranweiler @samscott89

briansmith:
Also, I'm looking for a proof than given:

* 0 < q < p < n, and
* ⌈lg p⌉ = ⌈lg q⌉ = ⌈(lg n)/2⌉, i.e. `p.bit_length() == q.bit_length() == n_bit_length().half_rounded_up()`

then checking p * q == 0 (mod n) is equivalent to checking p * q == n. (Our goal is to use only Montgomery modular multiplication, never non-modular arithmetic.)

ranweiler:
@briansmith Excellent! I'll check out the implementation tonight.

Unfortunately, I don't think that theorem holds. Consider q = 3, p = 4, n = 6. A counterexample for _prime_ q, p is given by (p, q) = (7, 5). I do have a proof if we strengthen the log-ceiling equalities to ⌈lg p⌉ = ⌈lg q⌉ = (lg n)/2, and essentially the same proof also works for max(⌈lg p⌉, ⌈lg q⌉) ≤ (lg n)/2.

briansmith:
> Unfortunately, I don't think that theorem holds. Consider q = 3, p = 4, n = 6.

OIC.

I think maybe we should use ⌈lg n⌉/2 instead of ⌈(lg n)/2⌉. Also, n is always odd, which may help.

> A counterexample for prime q, p is given by (p, q) = (7, 5).

I think you left out n here. It can't be 6 like in the previous example since it needs to be about twice the size of either p and q.

> I do have a proof if we strengthen the log-ceiling equalities to ⌈lg p⌉ = ⌈lg q⌉ = (lg n)/2, and essentially the same proof also works for max(⌈lg p⌉, ⌈lg q⌉) ≤ (lg n)/2.

Awesome! I think we should be able to figure it out.

ranweiler:
Ha, oops! You're right, my second counterexample is no good— I don't see any small counterexamples with both `p` and `q` prime. Also, it's true that we only actually care about prime RSA `p` and `q`, right?

ranweiler:
Btw, here's the proof I have, if you want to try to modify it for `⌈lg(n)⌉/2`.

**Proof.** Let `0 < q < p < n` and `m := max(⌈lg(q)⌉, ⌈lg(p)⌉)`.
Assume `m ≤ lg(n)/2`.

Suppose `pq = 0 (mod n)`, so `pq = nk` for some `k > 0`.
We want to show that `k = 1`, so `pq = n`.

Taking logarithms and applying the homomorphism identity, we have:
```
      lg pq = lg nk
lg p + lg q = lg n + lg k
```

Recall that for any `x`, we have `x ≤ ⌈x⌉`.
Then, applying our hypotheses, we obtain the bound:

```
lg n + lg k  = lg p + lg q
             ≤ ⌈lg p⌉ + ⌈lg q⌉
             ≤ m + m
             ≤ lg(n)/2 + lg(n)/2
             = lg(n)
```
Whence `lg n + lg k ≤ lg n`, and so `lg k ≤ 0.`
Since `0 ≤ lg k`, we have `lg k = 0`, and thus `k = 1`, as required.

briansmith:
I've landed some of the prerequisite changes from the first iteration. This second set of commits is closer to what the final form of this PR will be.

briansmith:
> Ha, oops! You're right, my second counterexample is no good— I don't see any small counterexamples with both p and q prime. Also, it's true that we only actually care about prime RSA p and q, right?

We are trying to implement the crt_pkv validation from NIST SP.800-56B rev1, section 6.4.1.4.3, RSA Key-Pair Validation (Intended Exponent-Creation Method Unknown) using a CRT-format key.

We don't know that `p` and `q` are prime. I guess we're allowed to presume so if (and only if?) all the checks pass. One of the checks is that p * q == n.

We also need to verify dmp1 · e mod (p – 1) == 1 and 1 == dmq1 * e mod (q - 1), which will also be non-trivial because I'm hoping to remove the ability to calculate modulos with even moduli.

briansmith:
I filed #362 about extending the checks to comply with the NIST recommendations.

samscott89:
Wasn't sure where to comment on the validation checks, so I put it in the other issue: https://github.com/briansmith/ring/issues/362#issuecomment-263552161

Bottom line was:
> Basically, it is fine to check whether `p*q = 0 (mod n)` so long as you also check that `p*q ≠ 0 (mod 2)` and `p*q ≠ 0 (mod 3)`.

samscott89:
I *think* that as long as you ensure p and q are not even you should be alright: https://github.com/briansmith/ring/issues/362#issuecomment-263623033.

briansmith:
> I think that as long as you ensure p and q are not even you should be alright: #362 (comment).

I think #362 is the best place to continue shoring up these checks.

My plan is to clean up the commits in this PR and land them soon. I am hoping that this PR will be enough of a model to help people to complete the oxidation of crypto/rsa.

briansmith:
OK, this is actually getting close for real now.

Most noteworthy is that `elem_mul_mixed` actually has the signature that one would expect:

```rust
pub fn elem_mul_mixed<F: Field>(a: &Elem<F>, b: &ElemDecoded<F>, m: &Modulus<F>)
                               -> Result<ElemDecoded<F>, error::Unspecified> {
```

This lets us use `Field` as a more sensible name for what was `Role`. If we were so inclined, it would be relatively straightforward to apply the "kernel" aspect of lightweight static capabilities. However, I'm not planning to do that unless/until we do multi-prime RSA.

I'll double-check the code tomorrow and hopefully land it.

src/rsa/bigint.rs
```diff
@@ -72,21 +110,217 @@ impl Positive {
...
+}
+
+impl core::cmp::PartialEq for Whole {
+    fn eq(&self, other: &Whole) -> bool {
+        self.partial_cmp(other) == Some(core::cmp::Ordering::Equal)
+    }
+}
+
+impl core::cmp::PartialOrd for Whole {
+    fn partial_cmp(&self, other: &Whole) -> Option<core::cmp::Ordering> {
```
ranweiler: **Update**: I see now that we just `derive` this elsewhere.

src/rsa/signing.rs
```diff
@@ -256,16 +392,13 @@ struct BN_BLINDING {
...
-                       p: &bigint::BIGNUM, q: &bigint::BIGNUM,
-                       iqmp: &bigint::BIGNUM) -> c::int;
 }
 
 #[allow(improper_ctypes)]
 extern {
-    fn GFp_rsa_private_transform(rsa: *const RSA, inout: *mut u8,
-                                 len: c::size_t, blinding: *mut BN_BLINDING,
-                                 rng: *mut rand::RAND) -> c::int;
+    fn GFp_rsa_private_transform(rsa: &RSA, base: &mut bigint::BIGNUM,
```
briansmith: `&mut` means “inout,” right? AFAICT `/*inout*/` only appears in C code for the declarations of *this* function.

src/rsa/bigint.rs
```diff
@@ -72,21 +110,217 @@ impl Positive {
...
+    }));
+    Ok(ElemDecoded {
+        value: r,
+        field: PhantomData
+    })
+}
+
+
+/// Whole integers: `Positive` ∪ {0}.
+struct Whole(*mut BIGNUM);
```
briansmith: I renamed it to `Nonnegative`.

src/rsa/bigint.rs
```diff
@@ -39,31 +39,69 @@ impl Positive {
...
+        }));
+        Ok(Elem {
+            value: self.0,
+            field: PhantomData,
+        })
+    }
+
+    pub fn into_elem_decoded<F: Field>(self, m: &Modulus<F>)
+            -> Result<ElemDecoded<F>, error::Unspecified> {
+        let cmp = unsafe {
```
briansmith: eventually `Modulus` will be defined as something like:

```rust
struct Modulus<M> {
    n0: Limb,
    rr: Elem<M>, // Not zero or one.
    value: OddPositive,
}
```
So eventually we'll do what you're suggesting.

src/rsa/bigint.rs
```diff
@@ -72,21 +110,217 @@ impl Positive {
...
+    pub unsafe fn as_mut_ref<'a>(&'a mut self) -> &'a mut BIGNUM {
+        self.value.as_mut_ref()
+    }
+
+    pub fn into_odd_positive(self) -> Result<OddPositive, error::Unspecified> {
+        self.value.into_odd_positive()
+    }
+}
+
+// `a` * `b` (mod `n`), assuming `a` and `b` are both less than `m`.
```
briansmith: Yes. I s/`n`/`m`/

briansmith: I also removed the "assuming.." {a, b} <= m is implied by the use of the `Elem[Decoded]` types.

src/rsa/bigint.rs
```diff
@@ -39,31 +39,69 @@ impl Positive {
...
         if untrusted::Reader::new(input).peek(0) {
             return Err(error::Unspecified);
         }
+        Self::from_be_bytes_padded(input)
+    }
+
+    pub fn from_be_bytes_padded(input: untrusted::Input)
+                                -> Result<Positive, error::Unspecified> {
+        // Reject empty inputs.
+        if input.len() == 0 {
```
briansmith: I added a commit to use `is_empty()` throughout the project.

src/rsa/bigint.rs
```diff
@@ -72,21 +110,217 @@ impl Positive {
...
+        Ok(r)
+    }
+
+    fn is_zero(&self) -> bool {
+        let is_zero = unsafe { GFp_BN_is_zero(self.as_ref()) };
+        is_zero != 0
+    }
+
+    fn is_one(&self) -> bool {
+        let is_zero = unsafe { GFp_BN_is_one(self.as_ref()) };
```
briansmith: I renamed is_zero to is_one as you suggested.

crypto/bn/montgomery.c
```diff
@@ -118,6 +118,10 @@
 #include "../internal.h"
 
 
+/* Avoid -Wmissing-prototypes warnings.*/
```
briansmith: Fixed. (No need for such polite qualifiers, BTW.)

briansmith:
Landed as:

0c2cffe33056791a91cf52aaf761049db3c44026 Check RSA public key parameters in Rust.
ea77b59dc879fe20d5ebf68e34dcfa55e234d93b Add extra check for minimum minimum RSA public modulus bit length.
f048caf2125cb287c38a34f91aedd36805a65f66 Check RSA private key parameter bounds in Rust.
cd221a9270c46a203b15089bddbf15b5d84be9c6 Construct most RSA private key Montgomery contexts in Rust.
3b188bbebf8a913599bb6b8ee8c2a0597b80709b Return boolean value instead of pointer from GFp_BN_copy.
80e6231cbd97c5c23a6c53099f842ab52980130f Check p * q == n in Rust.
1af1b4612375d4220b3500318f2fab0668725558 Convert iqmp to the Montgomery domain in Rust in RSA signing.
2906a4ce62a921f13918e771c5791a049414a5c3 Calculate q**2 (mod n) in Rust in RSA signing.
f1ad3efb3cbb63ec0b2d9dd44f36b39c25a5f7dd Remove rsa.c, including the checks on `dmp1` and `dmq1`.
5502beccd8ea94946a3db058ce4d775199bfd8e8 Manage `RSAKeyPair` attributes in Rust, not in the C `RSA` struct.
465bff90679888719774ba3c421d0e14295c4b4c Replace outermost steps of `GFp_BN_private_transform` with Rust code.
46f59b46ba2655073313ed98a3b44b200f09117a Use is_empty() instead of comparing len() to 0.

briansmith:
Also, thanks to both @samscott89 and @ranweiler for looking at this!

PR #368, opened 2016-12-04
Simplify poly1305 state handling.

briansmith:
Hi @PeterReid, could you look at this if you have time? I know you had some concerns about this change as far as future-proofing goes. I'm not sure if you know that the GCM code already works similarly. Even if we need it in the future, I'd rather not have each algorithm maintain its own separate I-U-F code.

WDYT?


PeterReid:
The only objection I had before was predicated on me thinking that poly1305 was part of the publicly-exposed API, which it turns out not to be. This change seems like an improvement to me.

briansmith:
> The only objection I had before was predicated on me thinking that poly1305 was part of the publicly-exposed API, which it turns out not to be. This change seems like an improvement to me.

Thanks.

briansmith:
Closing this for now. It may still be a good idea but it's not a high priority. Will revisit when we refactor the AEAD stuff to share more logic.

Issue #25, opened 2015-09-30
Add CCM AEAD


briansmith:
crypto/cmac is being removed because it depends on the non-AEAD cipher.h encryption interface, which is being removed. In order to implement the CCM AEAD, we need a new CMAC implementation. The new CMAC implementation can assume that AES is the only cipher that it will be used with, and that it will only be exposed by the ```ring::aead``` interface, which should simplify the new implementation.

a-dma:
Hi, quick question:

do you actually mean CMAC or did you instead mean CBC-MAC (which is used by AES-CCM) ?

If you did mean CBC-MAC, any plan to add CMAC? Should I open another issue for that?

briansmith:
> do you actually mean CMAC or did you instead mean CBC-MAC (which is used by AES-CCM) ?

I meant CBC-MAC, I guess, so we can eventually implement https://tools.ietf.org/html/rfc6655 and maybe eventually CCMP for 802.11 support.

> If you did mean CBC-MAC, any plan to add CMAC? Should I open another issue for that?

Let's discuss CMAC in another issue, along with a clear description of use cases motivation.

briansmith:
To add this to *ring*, I think we'd need to:

* Implement the CCM AEADs in `src/aead/aes_ccm.rs` using the same API style as AES-GCM and ChaCha20-Poly1305.
* Add the NIST test vectors from http://csrc.nist.gov/groups/STM/cavp/block-cipher-modes.html#test-vectors.
* Add any IETF test vectors.
* Add any additional tests.

briansmith:
This will be easier to do after #240 is done.

briansmith:
Now that we can use `#[repr(align)]` (we're already using it as of today, for ChacCha20-Poly1305), this seems a lot more practical. I cleaned up some of the C/asm code to simplify this work in case somebody wants to take this on.

jadamcrain:
@briansmith I'm interested in implementing this if a quality PR with test vectors, etc, would be reviewed. Is this still something that you're interested in supporting?

jadamcrain:
There's a PR [here](https://github.com/briansmith/ring/pull/1501) for this.

I couldn't find official test vectors for Tag ==16 bytes, Nonce = 12 bytes so I generated some using Rust Crypto's generic CCM implementation.

If anyone else knows of additional test vectors for this configuration of CCM, please point me towards them and I can add them.

[...]

PR #369, opened 2016-12-06
Prerequisite Exploration for Oxidizing AES-GCM.

briansmith:
Here are a bunch of changes, only about half of which we're likely to keep. The point of these changes is to see how close we can make the AES-GCM logic and the ChaCha20-Poly1305 logic so that we can eventually share even more code between them, and also have less C code.

Based on this, a lot of sharing of logic should be possible by factoring out lots of the code from src/aead/chacha20_poly1305.rs into src/aead/aead.rs and then adapting the remaining code in e_aes.c to expose an API that is like `GFp_ChaCha20_ctr32` and then adapting the remaining code in gcm.c to expose an API that is similar to the Poly1305 code.

One complication to factoring out the common logic is the `aesni_gcm_enabled`/`GFp_aesni_gcm_encrypt`/`GFp_aesni_gcm_decrypt` optimization.

Another complication to factoring out the common logic is the `GHASH_CHUNK` logic. Probably I'll just drop that for now, and then try to add it back in a way that can be used for ChaCha20-Poly1305 too. I'm not sure why OpenSSL chose not to have any `GHASH_CHUNK`-style logic in the ChaCha20-Poly1305 code we inherited from it.


briansmith:
I did quite a bit of work on this to move it further along. Unfortunately things get really ugly because of the alignment (to 16 bytes) requirements (probably just preferences) that the lower-level assembly language AES-GCM and low-level AES code has.  I implemented some workarounds, but they're ugly. Also, the primary purpose of switching to Rust is to improve static analysis of safety, improve readability, and improve maintainability, but the hacks currently needed to get things aligned are counter to all those goals.

Accordingly, I'm closing this w/o landing anything. I'll continue this in a different PR, using better code, once https://github.com/rust-lang/rust/issues/33626 is implemented, which could likely happen soon.

Note that after this is done, we'll be able to share a significant amount of logic between AES-GCM and ChaCha20-Poly1305. Probably that will also make it much easier to implement new AEADs, in particular [AES-CCM](https://github.com/briansmith/ring/issues/25) and maybe [AES-GCM-SIV](https://github.com/briansmith/ring/issues/412). So I'm postponing all that work until I can land something here.

PR #372, opened 2016-12-11
Do RSA Blinding in Rust.

briansmith:
This is the first draft of doing RSA blinding in Rust.


briansmith:
I think this is good to go, pending the CI results.

@samscott89 , @ranweiler, @djudd Do any of you have time to look at this? I think this is pretty much the key to making it straightforward to oxidize the rest of the RSA code and also to start oxidizing away `BIGNUM`.

The need to use `core::mem::replace` makes things a bit awkward. OTOH, it does prevent us from using a corrupted state where the blinding factor and its inverse are out of sync. That is a real bug in OpenSSL and LibreSSL (but not BoringSSL and *ring* because of an earlier commit).

I would have prefer to allocate all the memory for the `Blinding` in `Blinding::new()`, but it would have made `Blinding::new()` more complicated without making `blind()` less complicated, so I'll give up on that for now. That's something to revisit during oxidizing the `BIGNUM` code.

ranweiler:
@briansmith, I can look this over tonight!

src/rsa/blinding.rs
```diff
@@ -12,52 +12,127 @@
...
-    a: *mut bigint::BIGNUM,
-    ai: *mut bigint::BIGNUM,
-    counter: u32,
-}
+fn reset(arbitrary1: bigint::Elem<N>, arbitrary2: bigint::Elem<N>,
+         e: &bigint::OddPositive, n: &bigint::Modulus<N>,
+         rng: &rand::SecureRandom) -> Result<Contents, error::Unspecified> {
+    // Use `into_elem_decoded_montgomery_encoded` to grab the underling
+    // `BIGNUM` to avoid a superfluous `malloc()` & `free()`.
+    let mut random = arbitrary1.into_elem_decoded_montgomery_encoded();
```
djudd: I wonder if an `Arbitrary<N>` struct - with no-op conversions to `Elem` and `ElemDecoded` - would be a good complement to the `Elem*` structs and make this less awkward-looking

src/rsa/blinding.rs
```diff
@@ -12,52 +12,127 @@
...
 impl Blinding {
-    pub fn new() -> Result<Blinding, error::Unspecified> {
-        let r = unsafe { GFp_BN_BLINDING_new() };
-        if r.is_null() {
-            return Err(error::Unspecified);
-        }
-        Ok(Blinding(r))
+    pub fn new() -> Self { Blinding(None) }
+
+    pub fn blind<F>(&mut self, x: bigint::ElemDecoded<N>,
```
djudd: Is there a reason not to pass some of the arguments here which will be constant across invocations into the constructor, and hold on to a reference? If I understand correctly, we are relying on `e` and `n` to be the same for any given Blinding, and passing them into the constructor would make that clearer.

djudd:
Looks good! Left a couple of comments/questions, but not sure how much it's
worth massaging the code structure prior to oxidizing away more BIGNUM
stuff.

src/rsa/blinding.rs
briansmith: Thanks for the feedback.

The distinction between `RSASigningState` and `RSAKeyPair` is that `RSASigningState` is per-thread and `RSAKeyPair` is the state shared (immutably) across all threads by all the `RSASigningState` instances for the same key. The idea is that a high-performance server would likely create one `RSASigningState` per core (perhaps in a thread-local) per key. So, we generally want to make the `RSASigningState`s small.

Here's the definition of `RSASigningState`:
```rust
pub struct RSASigningState {
    key_pair: std::sync::Arc<RSAKeyPair>,
    blinding: blinding::Blinding,
}
```

We could instead define `RSASigningState` as a newtype:
```rust
pub struct RSASigningState(blinding::Blinding)
```
Then we'd move the `key_pair` field to the blinding, without increasing the size of `RSASigningState`.  After all, the idea of `RSASigningState` is to be the abstraction "a thing you need exclusive access to to sign with an RSA key, for some reason", and `Blinding` is the reason you need some extra thing. That would make it clear that the blinding values are specific to that key pair and that there's no mixing of key pairs for the same blinding. Then `RSASigningState::sign()` would borrow the `key_pair` out of the `Blinding`.

OTOH, it's a bit weird that the `Blinding` should know about the entire contents of the `RSAKeyPair` when only `n` and `e` are relevant to it. And it's also weird that it would need to know about `Arc` and threading behavior. What I'm hoping is that instead we'll be able to expand on the use of the `field: PhantomData<Field>` construct using the "lightweight static capabilities" idea to ensure that one can never mix two different tuples of (`N`, `P`, `Q`, ...). But we haven't implemented the important parts that actually enforce that yet, so maybe it is presumptuous to act like we did (or that we even know how to yet!).

Anyway, I'd love to hear what people think about this.

src/rsa/blinding.rs
briansmith: Perhaps `Storage` might be a better name than `Arbitrary`, since really it's entire purpose is to allow the `BIGNUM` to be reused.

Then we could have `elem_random<F: Field>(storage: Storage<F>, m: Modulus<F>) -> ElemDecoded<F>` and `elem_inverse<F>(storage: Storage<F>, a: &ElemDecoded<F>, m: &Modulus<F>) -> ElemDecoded<F>`. As a bonus, those functions would become pure functions,.

In contrast, we could use the name `Arbitrary` but it might be misleading to pass an "arbitrary" value to these functions. Especially the random generation taking an `Arbitrary` kind of implies that values is used for (re)seeding or something.

WDYT?

src/rsa/blinding.rs
briansmith: BTW, the entire `bigint` API already assumes that we can't mix different (`N`, `P`, `Q`, ...) values, and that assumption is made specifically to avoid needing to compare moduli for equality on every multi-operand function call. That doesn't mean that blinding must do things the same way, of course.

src/rsa/blinding.rs
```diff
@@ -12,52 +12,127 @@
...
+                let random = try!(random.into_elem(n));
+                let random_inv = try!(random_inv.into_elem(n));
+                return Ok(Contents {
+                    blinding_factor: random,
+                    blinding_factor_inv: random_inv,
+                    remaining: REMAINING_MAX - 1,
+                });
+            },
+            Err(bigint::InversionError::NoInverse) => {}, // continue
+            Err(_) => { return Err(error::Unspecified); }
```
briansmith: The coverage report shows that these two error cases are never hit. They aren't hit because:

* The coverage test, `test_signature_rsa_pkcs1_sign_blinding_creation_failure`, is fixing the output of the RNG to be all zeros.
* The inversion is done using `GFp_BN_mod_inverse_blinded`.
* That function uses `GFp_BN_rand_range_ex` to blind the inversion.
* `GFp_BN_rand_range_ex` fails because the RNG is consistently returning 0, which is the allowed out of range.

Thus, we never even get to the point where we call `GFp_BN_mod_inverse_odd`.

Further, even if we did, `GFp_BN_mod_inverse_odd` starts with:

```c
  if (!GFp_BN_is_odd(n)) {
    OPENSSL_PUT_ERROR(BN, BN_R_CALLED_WITH_EVEN_MODULUS);
    return 0;
  }
```

When `n` is zero, `GFp_BN_is_odd` will return false and so `GFp_BN_mod_inverse_odd` will fail *without setting `out_no_inverse` to 1*.

To fix the code coverage, I propose that we modify `GFp_BN_mod_inverse_odd` to require non-negative, non-positive inputs that are always in range, because those are the only values that `GFp_BN_mod_inverse_blinded` supplies to it, and `GFp_BN_mod_inverse_blinded` is the only non-test caller.

Then, we should extend the tests so that the RNG returns `p` or `q` for a composite `n == p * q`, since those would be the only non-invertible elements.

briansmith: Actually, we have:
```rust
        try!(bigint::elem_randomize(&mut random, n, rng));
        match bigint::elem_set_to_inverse_blinded(&mut random_inv, &random, n,
                                                  rng) {
```

So, we never even call `elem_set_to_inverse_blinded()`, since `elem_randomize()` will fail if the RNG only returns zeros.

briansmith:
I'm going to punt this to a slightly later release. I'll factor out the stuff that is needed by #373 and move it to #373 for 0.6 and leave the rest to 0.7.1 or so.

src/rsa/blinding.rs
djudd-stripe: I like `Storage` or `ElemStorage`.

briansmith:
> When n is zero, GFp_BN_is_odd will return false and so GFp_BN_mod_inverse_odd will fail without setting out_no_inverse to 1.

Sorry, that was a mistake on my part: `n` is the modulus, not the value being inverted.

The updated version of the PR does fix the above coverage problem, but there's coverage issues: We never test what happens when we fail to do the blinding 32 times and give up. This is unfortunate, but the code is clear enough that I'm not going to bother with trying to get 100% coverage at this time. Also we never test what happens when we give up blinding once and then try blinding again. But the code is pretty clear that the state is reset to `None` and `None` is the initial state, so this is not worth bothering with now. The effort trying to improve the test coverage here would probably better be spent on formal proofs of correctness.

briansmith:
> I wonder if an Arbitrary<N> struct - with no-op conversions to Elem and ElemDecoded - would be a good complement to the Elem* structs and make this less awkward-looking

David, it turns out that there is probably a straightforward way to achieve what you suggest here after about 50 other commits have landed. Conversely, it is quite awkward to do it the way the code is now. So, I'm going to punt this down the road.

briansmith:
This landed as d4204736cda4f354a0f61bd1f608bc768c7f9588.

PR #373, opened 2016-12-12
Use Rust for RSA signature verification.

briansmith:
@samscott89, @ranweiler Again, if you have some time to look at this, to make sure I didn't accidentally mess up any of the input validation checks in particular, I would appreciate it.

I actually wasn't intending to do this, but it ended up to be seemingly very simple since the RSA blinding code in an earlier PR already needed the same primitives.


briansmith:
Landed in e9149cfc9624856803c746ba047c174a29dad838.

Issue #379, opened 2016-12-14
Report ASN.1 decoding errors using a different error type than `Unspecified`


briansmith:
In https://github.com/briansmith/ring/pull/378#discussion_r92450802, @marshallpierce wrote:

> Returning Unspecified here just gets turned into BadDER elsewhere... maybe in the future it would be nice to have DER-related functions return a DER error for easier mapping?

Sounds like a good idea.

AFAICT there's no cases where we're parsing secret DER, so differentiating encoding errors from other errors isn't a security issue, and it would be a usability improvement. One way to do it would be to define an `strct ASN1Error;` type. Then we'd have to change ever DER-parsing function to return an enum of errors, one label of which is `ASN1Error`, e.g.

```rust
enum SignatureVerificationError { ASN1Error, SomeOtherError }
impl From<ASN1Error> for SignatureVerificationError {
    fn from(&ASN1Error) -> Self { SignatureVerificationError::ASN1Error }
}
```

This should allow `try!(...)` and friends to work reliably.

briansmith:
Closing this. Some of this was already done (e.g. for private key parsing). If more needs to be done, we can do that in separate issues/PRs.

PR #384, opened 2016-12-19
Use a more specific error instead of unspecified::Error for DER encoding errors, as per #379

marshallpierce:

### DIFF
--- a/src/der.rs
+++ b/src/der.rs
@@ -17,10 +17,18 @@
 //! This module contains the foundational parts of an ASN.1 DER parser.
 
 use untrusted;
-use error;
 
 pub const CONSTRUCTED: u8 = 1 << 5;
 pub const CONTEXT_SPECIFIC: u8 = 2 << 6;
+
+/// An error in ASN1 parsing. Normal hygiene applies: if parsing secret
+/// DER, do not differentiate this from any other error.
+#[derive(Clone, Copy)]
+pub struct ASN1Error;
+
+impl From<untrusted::EndOfInput> for ASN1Error {
+    fn from(_: untrusted::EndOfInput) -> Self { ASN1Error }
+}
 
 #[derive(Clone, Copy, PartialEq)]
 #[repr(u8)]
@@ -43,20 +51,20 @@ pub enum Tag {
 pub fn expect_tag_and_get_value<'a>(input: &mut untrusted::Reader<'a>,
                                     tag: Tag)
                                     -> Result<untrusted::Input<'a>,
-                                              error::Unspecified> {
+                                        ASN1Error> {
     let (actual_tag, inner) = try!(read_tag_and_get_value(input));
     if (tag as usize) != (actual_tag as usize) {
-        return Err(error::Unspecified);
+        return Err(ASN1Error);
     }
     Ok(inner)
 }
 
 pub fn read_tag_and_get_value<'a>(input: &mut untrusted::Reader<'a>)
                                   -> Result<(u8, untrusted::Input<'a>),
-                                            error::Unspecified> {
+                                      ASN1Error> {
     let tag = try!(input.read_byte());
     if (tag & 0x1F) == 0x1F {
-        return Err(error::Unspecified); // High tag number form is not allowed.
+        return Err(ASN1Error); // High tag number form is not allowed.
     }
 
     // If the high order bit of the first byte is set to zero then the length
@@ -67,7 +75,7 @@ pub fn read_tag_and_get_value<'a>(input: &mut untrusted::Reader<'a>)
         0x81 => {
             let second_byte = try!(input.read_byte());
             if second_byte < 128 {
-                return Err(error::Unspecified); // Not the canonical encoding.
+                return Err(ASN1Error); // Not the canonical encoding.
             }
             second_byte as usize
         },
@@ -76,12 +84,12 @@ pub fn read_tag_and_get_value<'a>(input: &mut untrusted::Reader<'a>)
             let third_byte = try!(input.read_byte()) as usize;
             let combined = (second_byte << 8) | third_byte;
             if combined < 256 {
-                return Err(error::Unspecified); // Not the canonical encoding.
+                return Err(ASN1Error); // Not the canonical encoding.
             }
             combined
         },
         _ => {
-            return Err(error::Unspecified); // We don't support longer lengths.
+            return Err(ASN1Error); // We don't support longer lengths.
         },
     };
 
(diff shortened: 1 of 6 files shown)


src/ec/suite_b/ops/ops.rs
```diff
@@ -338,7 +338,7 @@ pub struct PublicScalarOps {
 impl PublicScalarOps {
     pub fn scalar_parse(&self, input: &mut untrusted::Reader)
                         -> Result<Scalar, error::Unspecified> {
-        let encoded_value = try!(der::positive_integer(input));
+        let encoded_value = try!(der::positive_integer(input).map_err(|_| error::Unspecified));
```
marshallpierce: There were only a few places that needed to `map_err` so I chose to not implement `From<ASN1Error>`, but I can do that if you prefer.

src/der.rs
```diff
@@ -17,11 +17,19 @@
...
 use untrusted;
-use error;
 
 pub const CONSTRUCTED: u8 = 1 << 5;
 pub const CONTEXT_SPECIFIC: u8 = 2 << 6;
 
+/// An error in ASN1 parsing. Normal hygiene applies: if parsing secret
+/// DER, do not differentiate this from any other error.
+#[derive(Clone, Copy)]
+pub struct ASN1Error;
```
marshallpierce: `Asn1Error`? `DerError`?

src/ec/suite_b/ops/ops.rs
briansmith: It's OK to punt on that for now.

marshallpierce:
Does appveyor normally take this long? It's been a day and they still haven't run their builds.

marshallpierce:
I think appveyor is stuck; is there a way to trigger new builds without pushing a bogus commit?

src/der.rs
briansmith: I think `ASN1Error` is fine even because I can't stand `Asn1Error` or similar capitalization of acronyms.

src/der.rs
```diff
@@ -17,11 +17,19 @@
...
 
 use untrusted;
-use error;
 
 pub const CONSTRUCTED: u8 = 1 << 5;
 pub const CONTEXT_SPECIFIC: u8 = 2 << 6;
 
+/// An error in ASN1 parsing. Normal hygiene applies: if parsing secret
+/// DER, do not differentiate this from any other error.
+#[derive(Clone, Copy)]
```
briansmith: Should we derive Debug as well?

src/der.rs
```diff
@@ -17,11 +17,19 @@
...
+pub struct ASN1Error;
+
+impl From<untrusted::EndOfInput> for ASN1Error {
+    fn from(_: untrusted::EndOfInput) -> Self { ASN1Error }
+}
```
briansmith: I don't think this is correct. It is true that *in this module*, any time we encounter end-of-input, it's because of bad ASN.1. But, in generally it doesn't make sense to convert *any* or *all* `EndOfInput` errors to `ASN1Error`. Instead it has to be decided on a case-by-case basis.

Instead, we probably need to create wrapper functions around `Reader::read_byte()` and friends that map `EndOfInput` to `ASN1Error`, and use them here instead of using `Input`'s methods directly.

src/der.rs
```diff
@@ -117,26 +125,26 @@ fn nonnegative_integer<'a>(input: &mut untrusted::Reader<'a>, min_value: u8)
 
     let value = try!(expect_tag_and_get_value(input, Tag::Integer));
 
-    value.read_all(error::Unspecified, |input| {
+    value.read_all(ASN1Error, |input| {
```
briansmith: I don't think all the errors returned by this function are ASN.1 decoding errors. Some of them are value-out-of-range errors. Since the point of this PR is to allow us to differentiate different kinds of errors, it seems like we should differentiate out-of-range vs. bad-DER. WDYT?

The end-goal of this specific PR is to make it possible to rebase the webpki crate on top of the new version of *ring* that has #378 integrated, right? Presumably, in the future we'll change some public APIs in *ring* to return more specific errors, by extending the changes in this PR, but that's not a goal of the present PR, right?

Or is the purpose different? The intended purpose matters as far as this review is concerned because (1) this change may not be necessary for rebasing webpki on top of #378 and/or (2) this PR may not be sufficient for solving the intended problem if it doesn't differentiate bad-DER from other kinds of errors.

marshallpierce: > Since the point of this PR is to allow us to differentiate different kinds of errors, it seems like we should differentiate out-of-range vs. bad-DER. WDYT?

I would love to; the only reason I didn't do so was to hew more closely to the "expose as little as possible" goal. More on this below.

> The end-goal of this specific PR is to make it possible to rebase the webpki crate on top of the new version of ring that has #378 integrated, right? 

It would be possible, though a trifle code-smelly, to use this (plus #378) in webpki. The issue is the, ahem, *approximate* error handling like this, which is currently residing in #378:

```
            try!(der::expect_tag_and_get_value(input, der::Tag::Sequence)
                .map_err(|_| ParseSPKIError::BadDER)); 
```

It would be nice to not map `Unspecified` to `BadDER` just because it seems like that's probably what's going on... and sure enough, webpki has its own way of doing this:

```
#[inline(always)]
pub fn expect_tag_and_get_value<'a>(input: &mut untrusted::Reader<'a>,
                                    tag: Tag) ->
                                    Result<untrusted::Input<'a>, Error> {
    ring::der::expect_tag_and_get_value(input, tag).map_err(|_| Error::BadDER)
}
```

Mapping `Unspecified` to `BadDER` may be correct, but it doesn't feel very good.

Anyway, consider webpki's `parse_spki_value`:

```
fn parse_spki_value(input: untrusted::Input)
                    -> Result<SubjectPublicKeyInfo, Error> {
    input.read_all(Error::BadDER, |input| {
        let algorithm_id_value =
                try!(der::expect_tag_and_get_value(input, der::Tag::Sequence));
        let key_value = try!(der::bit_string_with_no_unused_bits(input));
...
```

In practice, the only `Error` variant this will return is `BadDER`. So, when webpki uses *ring* for its spki parsing needs and effectively replaces that function, it could quite neatly map `ParseSPKIError::BadDER` to `Error::BadDER` and call it a day, with or without a better breakdown of ASN.1 errors that would presumably all get mapped to `ParseSPKIError::BadDER` (maybe with a field in that variant to store which specific `ASN1Error` it was?).

In summary: no, I don't think we need more detailed errors to make it possible to use this in webpki. However, I wholeheartedly agree that it is not a very nice state of the world to report an value-out-of-bounds error the same way as an unexpected EOF, so I would happily make `ASN1Error` an enum with `UnexpectedEOF`, `ValueOutOfBounds`, etc variants if you're OK with that (or some other error scheme, though that's what came to mind). Based on your other comments, it sounds like parsing ASN.1 is not itself considered sensitive, so more detailed information about its parsing should not be a problem, and you know how much I love building to allow testability... :)

src/der.rs
```diff
@@ -43,49 +46,49 @@ pub enum Tag {
...
             combined
         },
         _ => {
-            return Err(error::Unspecified); // We don't support longer lengths.
+            return Err(ASN1Error); // We don't support longer lengths.
         },
     };
 
-    let inner = try!(input.skip_and_get_input(length));
+    let inner = try!(input.skip_and_get_input(length).map_err(|_| ASN1Error));
```
marshallpierce: There was only one usage of skip_and_get_input so I just mapped the error inline rather than with a wrapper function.

briansmith:
@marshallpierce Only now have I had time to return to this. Are you interested in bringing this PR up to date? If so, I would be able to review and merge the updated PR right away.

FWIW, I have decided to start using the `Asn1` vs. `ASN1` capitalization like everybody else. I already modified RSA key loading to report more specific errors in 9968baa55dd595d3120c5290d03341b69b3e9bfa.

briansmith:
Thanks @marshallpierce. I have a branch that does this that we can use instead. It was a mistake not to merge your PR before. Sorry!

Issue #285, opened 2016-08-26
Integrate rustfmt into the build system and/or Continuous Integration


briansmith:
[Edit: This is just about rustfmt now; Clippy integration is now #448]

#277 shows that [Clippy](https://github.com/Manishearth/rust-clippy) is useful. We should integrate it into the build system and/or the continuous integration so that we can catch problems before they get committed.

On Continue Integration, Clippy checks should run _AFTER_ the tests. I think it would be too annoying to have a push fail to even run the tests because of a formatting error.

It's probably best to use the [`cargo clippy`](https://github.com/Manishearth/rust-clippy#as-a-cargo-subcommand-cargo-clippy) subcommand instead of the compiler plugin, as I suspect it would be less brittle and more likely to work across stable/beta/nightly Rust.

briansmith:
See also http://johannh.me/blog/rustfmt-ci.html

pietro:
`rustfmt` currently does a lot of changes and breaks some code that has inline C-style block comments.

I guess the goal would be to take some of the changes, ignore some and add attributes to make `rustfmt` ignore some code blocks.

Can you give some examples of things that should not to be changed? This is the diff from running `rustfmt` on current master:
https://github.com/pietro/ring/tree/rust_fmt

briansmith:
The things I most strongly object to are mentioned in my review comments in https://github.com/briansmith/ring/pull/385#pullrequestreview-13896830. In general I very much like the idea of rustfmt, but I would like to see rustfmt fixed and/or made more flexible to output formatting that is optimized for minimizing the number of lines of code and optimized for the 80-column width, due mostly to limitations of my code reviewing system. Unless/until that happens I think it makes less sense to try to reformat more than what's already been done, because people won't generally be able to do `rustfmt` to automatically format their code until `rustfmt` is fixed.

One thing I'd also like to consider is reformatting everything to use 2 space indent, to compensate for the 80-column width rule. I've always wanted to use 2-space indent but when I started Rust programming I had the impression that nobody did that. Now I know I'm not the only one so i think it makes sense to go ahead with it.

Regarding C-style comments like `in_out`, we can probably just rename the parameters to have an `in_out_` prefix and get rid of the comment.

pietro:
[...]

PR #385, opened 2016-12-20
Apply rustfmt.

marshallpierce:
Various broken cases handled manually.
Also, add a Travis hook and a sample pre-commit hook.
(Addresses #285)

### DIFF
--- a/src/aead/aes_gcm.rs
+++ b/src/aead/aes_gcm.rs
@@ -41,7 +41,9 @@ pub static AES_256_GCM: aead::Algorithm = aead::Algorithm {
 fn aes_gcm_init(ctx_buf: &mut [u8], key: &[u8])
                 -> Result<(), error::Unspecified> {
     bssl::map_result(unsafe {
-        GFp_aes_gcm_init(ctx_buf.as_mut_ptr(), ctx_buf.len(), key.as_ptr(),
+        GFp_aes_gcm_init(ctx_buf.as_mut_ptr(),
+                         ctx_buf.len(),
+                         key.as_ptr(),
                          key.len())
     })
 }
@@ -52,20 +54,30 @@ fn aes_gcm_seal(ctx: &[u64; aead::KEY_CTX_BUF_ELEMS],
                 -> Result<(), error::Unspecified> {
     let ctx = polyfill::slice::u64_as_u8(ctx);
     bssl::map_result(unsafe {
-        GFp_aes_gcm_seal(ctx.as_ptr(), in_out.as_mut_ptr(), in_out.len(), tag,
-                         nonce, ad.as_ptr(), ad.len())
+        GFp_aes_gcm_seal(ctx.as_ptr(),
+                         in_out.as_mut_ptr(),
+                         in_out.len(),
+                         tag,
+                         nonce,
+                         ad.as_ptr(),
+                         ad.len())
     })
 }
 
 fn aes_gcm_open(ctx: &[u64; aead::KEY_CTX_BUF_ELEMS],
                 nonce: &[u8; aead::NONCE_LEN], in_out: &mut [u8],
                 in_prefix_len: usize, tag_out: &mut [u8; aead::TAG_LEN],
-                ad: &[u8]) -> Result<(), error::Unspecified> {
+                ad: &[u8])
+                -> Result<(), error::Unspecified> {
     let ctx = polyfill::slice::u64_as_u8(ctx);
     bssl::map_result(unsafe {
-        GFp_aes_gcm_open(ctx.as_ptr(), in_out.as_mut_ptr(),
-                         in_out.len() - in_prefix_len, tag_out, nonce,
-                         in_out[in_prefix_len..].as_ptr(), ad.as_ptr(),
+        GFp_aes_gcm_open(ctx.as_ptr(),
+                         in_out.as_mut_ptr(),
+                         in_out.len() - in_prefix_len,
+                         tag_out,
+                         nonce,
+                         in_out[in_prefix_len..].as_ptr(),
+                         ad.as_ptr(),
                          ad.len())
     })
 }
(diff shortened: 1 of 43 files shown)


marshallpierce:
Aesthetically I don't necessarily like some of the edits rustfmt applies, but if rustfmt checks via CI are the goal (and I do like consistent formatting), then I don't see a way around doing what it wants.

src/aead/chacha20_poly1305_openssh.rs
```diff
@@ -26,7 +26,8 @@
 //! `padding_length||payload||random padding`.
 //!
 //! [chacha20-poly1305@openssh.com]:
-//!    http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/PROTOCOL.chacha20poly1305?annotate=HEAD
+//!    http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/\
```
marshallpierce: This url-line-breaking is pretty unfortunate but I didn't see any other way to appease rustfmt.

src/der.rs
```diff
@@ -171,8 +172,9 @@ pub fn small_nonnegative_integer(input: &mut untrusted::Reader)
 /// Parses a positive DER integer, returning the big-endian-encoded value, sans
 /// any leading zero byte.
 #[inline]
-pub fn positive_integer<'a>(input: &mut untrusted::Reader<'a>)
-                            -> Result<untrusted::Input<'a>, error::Unspecified> {
+pub fn positive_integer<'a>
+    (input: &mut untrusted::Reader<'a>)
```
marshallpierce: This formatting is also very strange

src/digest/digest.rs
```diff
@@ -153,35 +153,43 @@ impl Context {
...
+                                                  1);
             }
             // We don't increase |self.completed_data_blocks| because the
             // padding isn't data, and so it isn't included in the data length.
             padding_pos = 0;
         }
 
-        polyfill::slice::fill(
-            &mut self.pending[padding_pos..(self.algorithm.block_len - 8)], 0);
+        let slice = self.pending[padding_pos..(self.algorithm.block_len - 8)];
```
marshallpierce: the inline calculation in the slice confused rustfmt mightily so I pulled out a local

src/ec/suite_b/ops/p256.rs
```diff
@@ -270,32 +270,56 @@ fn p256_scalar_inv_to_mont(a: &Scalar) -> ScalarMont {
...
-                                   g_scalar: *const Limb/*[COMMON_OPS.num_limbs]*/);
-
-    fn GFp_p256_scalar_mul_mont(r: *mut Limb/*[COMMON_OPS.num_limbs]*/,
-                                a: *const Limb/*[COMMON_OPS.num_limbs]*/,
-                                b: *const Limb/*[COMMON_OPS.num_limbs]*/);
-    fn GFp_p256_scalar_sqr_mont(r: *mut Limb/*[COMMON_OPS.num_limbs]*/,
-                                a: *const Limb/*[COMMON_OPS.num_limbs]*/);
-    fn GFp_p256_scalar_sqr_rep_mont(r: *mut Limb/*[COMMON_OPS.num_limbs]*/,
-                                    a: *const Limb/*[COMMON_OPS.num_limbs]*/,
+    fn GFp_nistz256_add(// [COMMON_OPS.num_limbs]
```
marshallpierce: Had to make these `//` comments so that it wouldn't try to put them on the same line, and then fall afoul of the line length limit

src/pbkdf2.rs
```diff
@@ -210,6 +211,7 @@ fn derive_block(secret: &hmac::SigningKey, iterations: usize, salt: &[u8],
 ///
 /// `derive` panics if `out.len()` is larger than (2**32 - 1) * the PRF digest
 /// length, per the PBKDF2 specification.
+#[cfg_attr(rustfmt, rustfmt_skip)]
```
marshallpierce: This function also made rustfmt mad.

src/poly1305.rs
```diff
@@ -30,9 +30,10 @@ fn with_aligned<F>(opaque: &mut Opaque, f: F)
     let mut buf = [0u8; OPAQUE_LEN + 7];
     let aligned_start = (buf.as_ptr() as usize + 7) & !7;
     let offset = aligned_start - (buf.as_ptr() as usize);
-    let aligned_buf =
-        slice_as_array_ref_mut!(
-            &mut buf[offset..(offset + OPAQUE_LEN)], OPAQUE_LEN).unwrap();
+    let aligned_buf = slice_as_array_ref_mut!(&mut buf[offset..(offset +
+                                                                OPAQUE_LEN)],
+                                              OPAQUE_LEN)
+        .unwrap();
```
marshallpierce: Yikes

src/aead/aes_gcm.rs
```diff
@@ -52,20 +54,30 @@ fn aes_gcm_seal(ctx: &[u64; aead::KEY_CTX_BUF_ELEMS],
...
     bssl::map_result(unsafe {
-        GFp_aes_gcm_seal(ctx.as_ptr(), in_out.as_mut_ptr(), in_out.len(), tag,
-                         nonce, ad.as_ptr(), ad.len())
+        GFp_aes_gcm_seal(ctx.as_ptr(),
+                         in_out.as_mut_ptr(),
+                         in_out.len(),
+                         tag,
+                         nonce,
+                         ad.as_ptr(),
+                         ad.len())
```
briansmith: This is most of the reason why we don't use rustfmt already. We need a formatting tool that it can format function calls in the compact format, and that generally optimizes for readability first and minimum number of lines second, like the Google clang-format style.

Alternatively, we might be able to use the default rustfmt style after we've oxidized more of *ring* so that there aren't so many function calls that take a large number of arguments. But not yet.

src/aead/aes_gcm.rs
```diff
@@ -52,20 +54,30 @@ fn aes_gcm_seal(ctx: &[u64; aead::KEY_CTX_BUF_ELEMS],
...
     })
 }
 
 fn aes_gcm_open(ctx: &[u64; aead::KEY_CTX_BUF_ELEMS],
                 nonce: &[u8; aead::NONCE_LEN], in_out: &mut [u8],
                 in_prefix_len: usize, tag_out: &mut [u8; aead::TAG_LEN],
-                ad: &[u8]) -> Result<(), error::Unspecified> {
+                ad: &[u8])
+                -> Result<(), error::Unspecified> {
```
briansmith: This is not such a big deal as the function call formatting, but still it seems unnecessary.

src/aead/chacha20_poly1305_openssh.rs
briansmith: Let's fix rustfmt instead. I think that keeping URLs intact should be an uncontroversial change to rustfmt, at least as an option.

src/der.rs
briansmith: Yes, it seems like maybe a rustfmt bug? When it does this to a `where` clause it puts the opening brace on a new line. Not sure why it doesn't do that in this case.

src/der.rs
```diff
@@ -221,7 +223,8 @@ mod tests {
           &[0x02, 0x00, 0x01],
           &[0x02, 0x01],
           &[0x02, 0x01, 0x00, 0x01],
-          &[0x02, 0x01, 0x01, 0x00], // Would be valid if last byte is ignored.
+          /* Would be valid if last byte is ignored. */
+          &[0x02, 0x01, 0x01, 0x00],
```
briansmith: Presumably this could be written with a C++ comment in that position instead of a C-style comment?

src/ec/suite_b/ops/p256.rs
briansmith: It seems better to just tell rustfmt to ignore this block.

marshallpierce:
I'm fine with punting. I'll close this (and submit some rustfmt bugs). I just picked it up because it looked like something I could get done while you looked at my other PR's. ;) It's not like this project had egregiously sloppy code formatting to begin with, so the win isn't very significant when it comes with so many annoying mis-formats.

src/aead/chacha20_poly1305_openssh.rs
marshallpierce: Already a bug: https://github.com/rust-lang-nursery/rustfmt/issues/506

PR #390, opened 2016-12-23
Add tests for maximum input length for digests

### LINKED ISSUE #341: Enforce maximum input length for digest functions
Each digest function has a maximum input length, but we don't enforce that.

The safe maximum input length may be shorter than when the digest function spec. was written, but at least we should enforce the limit from the spec.

If we use checked addition to increment the input length so far in `ring::digest::Context::update()` then we can probably do this check lazily in `ring::digest::Context::finish()`.

djudd:
Fixes #341 - kind of. The `completed_data_blocks` / `completed_data_bits` tracking appears to already effectively implement the strategy suggested in that issue, so this just adds tests. If we (~ @briansmith) aren't satisfied with a panic with an obscure message, it appears we need to change the `digest::digest` API, and it wasn't clear to me whether that's in scope.

(Also, if we do that, I wonder if we need to make the check less lazy, in case people end up relying on it for DOS-protection.)

### DIFF
--- a/src/digest/digest.rs
+++ b/src/digest/digest.rs
@@ -772,4 +772,53 @@ mod tests {
                             digest::digest(&digest::SHA512, b"hello, world")));
 
     }
+
+    mod max_input {
+        macro_rules! max_input_tests {
+            ( $algorithm_name:ident ) => {
+                #[allow(unused_results)]
+                #[allow(non_snake_case)]
+                mod $algorithm_name {
+                    use super::super::super::super::digest;
+
+                    fn nearly_full_context() -> digest::Context {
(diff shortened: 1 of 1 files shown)


briansmith:
Thanks. I rebased this and changed the formatting and landed it as f1e75fec17786978151074b46984012e8b9a72fd. In particular, instead of `#[allow(unused_result)]`, I used `let _= ....` to suppress the warning.

I also reorganized it a bit so that most of the logic is outside the macro in ff4ecfa4125acee1e334b4895c90e0cbde04097d.

I also added another variant of the test in 9e194a3a93ab2100817e38aaed38f811e2688bb2.

Thanks for doing this!

Issue #362, opened 2016-11-29
implement NIST SP-800-56B-rev1-compliant RSA private key validation.


briansmith:
See #360, where we've done some work on private key validation in Rust, replacing some similar code written in C.

We need to extend that work to comply with [NIST SP-800-56B-rev1](http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Br1.pdf). Specifically, since we don't know how the exponent is determined, we need to implement the crt_pkv validation mechanism for the RSA Key-Pair Validation (Intended Exponent-Creation Method Unknown) case. That is in section 6.4.1.4.3.

At a minimum, we need to:

* Annotate the code that does the private key validation with references to the steps given in the NIST document.
* Restrict RSA private keys (but not public keys), to having *e* >= 65536. Note that the document allows for *e* to be up to 256 bits, but we're only going to maintain our 33 bit limit for now. (This is assuming the NIST document allows us to be more strict than what it specifies.)
* Verify that (k<sup>e</sup>)<sup>d</sup>) == 1 for some 1 < *k* < `n` of our choosing.
* Document the proof that our check p * q == 0 (mod n) is equivalent mathematically to p * q == n, or if it isn't, then replace it with a compliant check.
* Add the checks on the consistency of *e* and *dmp1* and *dmq1* that we've skipped.

briansmith:
Note that computing e * dmp1 (mod p-1)  and e * dmq1 (mod q-1) requires doing modular multiplication with an even modulus. BoringSSL and OpenSSL use `BN_mod` (a macro around `BN_div`) for that, but we're hoping to remove `BN_div` completely. Accordingly, we can't do the reduction using the even modulus directly. However, we can use the CRT to do it; see http://www.people.vcu.edu/~jwang3/CMSC691/j34monex.pdf

This issue has many steps to it. I recommend that anybody taking it on should break it up into multiple commits and even multiple PRs.

briansmith:
Notes on the proof for pq == 0 mod n being equivalent to p * q == n. @ranweiler did some math in https://github.com/briansmith/ring/pull/360#issuecomment-263434916.

Also, it may be easier to think about the relationship between p and q if we realize that q < p < 2q since q < p and they have the same bit length, so we have 0 < q < p < 2q < n.

samscott89:
[...]

PR #391, opened 2016-12-23
Document NIST steps in RSA private key validation

djudd:
This is a start at #362 - just inline documentation of which NIST SP 800-56B validation steps are & are not currently implemented.


djudd:
Fixing at least the lower-hanging fruit before merging makes sense to me, as does restricting the input length to multiple of 1024 sooner rather than later.

I did make an effort at listing all of the missing checks, so to the extent I neglected to mention any, it's because I misunderstood something.

briansmith:
> Fixing at least the lower-hanging fruit before merging makes sense to me, as does restricting the input length to multiple of 1024 sooner rather than later.

OK. Can we do this in the reverse order? In particular, @djudd, can you submit your second commit in this PR as a separate PR based on the current master, i.e. basically reverse the order of the two commits?

More generally, I suggest we do this:

1. Recognize, through our communication in this PR and/or the related issue, which problems the code currently has.
2. Fix as many of those problems as we can.
3. Write the documentation about how we do and don't adhere to the NIST recommendations when we're closer to adhering to them.

> I did make an effort at listing all of the missing checks, so to the extent I neglected to mention any, it's because I misunderstood something.

The main issue, which I hadn't considered previously, is that we don't verify the public key (as claimed in a certificate or somewhere else) is consistent with the private key. Omitting that would lead to a lot of text, as you can see a in my comment above. I filed a separate issue for dealing with that since it seems like a big issue all on its own. For now, I'm fine with deferring 

The second major issue is that we don't accept acceptable public modulus lengths as inputs to this function. I suggest we do that in #401.

There are a bunch of smaller issues, such as the one that David fixed in his second commit in this PR. Another one is that the minimum public exponent for private key operations is 3 instead of 65537. I think we should clean up as many of these as possible before committing the documentation, to minimize the amount of nuanced text that we have to write, review, commit, and then later replace. We can use the first commit of this PR + my comments above + any extra findings to guide that work.

WDYT?

djudd:
Your suggested procedure seems reasonable. I'm happy to reverse the commit order.

The exponent minimum, `|p – q|`, and modulus length seem relatively straightforward to validate. The remaining missing validations--aside from the question of comparing with another representation of the public key, as you say--then involve `LCM(p – 1, q – 1)` or `GCD([p or q] – 1, e)`. Those seem less trivial. I see there is some GCD code inside [GFp_BN_mod_inverse_odd](https://github.com/briansmith/ring/blob/46f59b46ba2655073313ed98a3b44b200f09117a/crypto/bn/gcd.c#L116)... is that the only starting point? Unfortunately I feel a bit over my head dealing with the guts of the bignum arithmetic.

djudd:
What I see as the remaining low-hanging fruit has been picked off, aside from #401. Things that came up:
- I was able to add a test for too-small-`e`, but haven't yet figured out how to generate an otherwise-valid private key that violates the `|p-q|` or `d > sqrt(n)` conditions. Is that a blocker?
- If we're avoiding primality tests because of side channels, that suggests to me that subtracting `q` from `p` should be constant-time. I implemented it naively because I wasn't sure, but I can switch to using `limbs_reduce_once_constant_time` if that's something we need.

briansmith:
Because the stricter checks will break backward compatibility in rare cases (especially the e>=65537) cases, I'm going to wait until 0.7 to land this. That wait shouldn't be too long.

src/rsa/bigint.rs
```diff
@@ -132,6 +132,18 @@ impl Positive {
         let bits = unsafe { GFp_BN_num_bits(self.as_ref()) };
         bits::BitLength::from_usize_bits(bits)
     }
+
+    // TODO Does this need to be constant time?
+    pub fn subtract_smaller(&self, smaller: &Positive) -> Result<Positive, error::Unspecified> {
```
briansmith: We should define all the math in terms of `Elem`s, in particular using only modular arithmetic. In this case, we can use `elem_sub`. (We want to use only modular arithmetic because we want to minimize code size and code complexity.)

Eventually we'll change `elem_sub` and all the other modular arithmetic is constant-time. The existing use of `elem_sub` needs it to be constant-time too. But it isn't a priority now and the plan for doing so first requires us to simplify the `BIGNUM` code more. I'm working on it as I have time now.

briansmith: Also, since we know that `p` > `q` already, `elem_sub(copy_of_p, &q, &n)` is guaranteed to not to need any reduction because there won't be any underflow.

briansmith:
> haven't yet figured out how to generate an otherwise-valid private key that violates the |p-q|

Maybe search for somebody who has published the smallest two 1024-bit primes, or similar. They are probably too close together to pass that check.

briansmith:
> Maybe search for somebody who has published the smallest two 1024-bit primes, or similar. They are probably too close together to pass that check.

Also, search for "small prime difference" since that seems to be the attach that such checks guard against. Maybe you can find an example keypair from one of the papers.

Alternatively, you could modify the RSA key generation procedure of a crypto library (like BoringSSL) so that when it finds the first prime, it then generates the second prime by reusing the first (n-100) bits of the first prime and only generating random 100 bits, or similar.

briansmith:
> d > sqrt(n)

I think the check in the spec that you are referring to is “2<sup>nBits/2</sup>< d,” right? Can't we just check that the `d.bit_length()` is larger than the nBits/2 (equivalently, that `p.bit_length() < d.bit_length()`), and then write a proof in the comments that `nBits / 2 < d.bit_length() ⇒ 2<sup>nBits/2</sup>< d`?

Alternatively, maybe we can prove that if `e` is restricted to be no more than 33 bits, then `d` must be at least `nBits/2` because `e*d = 1 mod (...)`.

briansmith:
Also, you said you can't find an otherwise-valid private key where `d > sqrt(n)` doesn't hold. AFAICT, you'll never find a mathematically-valid private key like that. Instead, when that property fails to hold, that means that p, q, d, n, e are inconsistent. Thus, you might be able to forge such a private key by tampering with the value of `d`.

Also, you asked if testing these properties is a blocker. I'm not particularly worried about testing checks on `d` because we never use `d` in any actual computation any we don't even save it in the `RSAKeyPair`. It would be nice to have a check for the small prime difference check, but that also isn't required at this point in time. This function already has low test coverage.

briansmith:
> If we're avoiding primality tests because of side channels

We're avoiding the primality tests partially for side-channel avoidance, and partially for performance reasons. We'll probably add an alternate slow constructor that verifies primality after we've written the primality-verification/proving code.

briansmith:
> Maybe search for somebody who has published the smallest two 1024-bit primes, or similar. They are probably too close together to pass that check.

This was suggested to be on Twitter: https://gist.github.com/indutny/d2e6dadaab96f5cec770dc031e6e448a, via https://twitter.com/indutny/status/824351187881394176.

briansmith:
> This was suggested to be on Twitter: https://gist.github.com/indutny/d2e6dadaab96f5cec770dc031e6e448a, via https://twitter.com/indutny/status/824351187881394176.

Also https://gist.github.com/lgarron/44fb9af0e4ae922bce362b62de70e81b.

djudd:
> Can't we just check that the d.bit_length() is larger than the nBits/2

Yes, I think so, that's what I implemented. Sorry for the misleading description.

> we can use elem_sub

The only `elem_sub` I see is defined for p384 Elems. Should I be adapting that code? Or, what am I missing?

briansmith:
> The only elem_sub I see is defined for p384 Elems. Should I be adapting that code? Or, what am I missing?

There is one in src/rsa/bigint.rs: https://github.com/briansmith/ring/blob/9d6981f431e52a071c415f4db4e8373a7cab06cd/src/rsa/bigint.rs#L398-L409

djudd:
Ah I needed to rebase, sorry, though I'd done that already.

djudd:
@briansmith I've rebased and re-ordered the commits. I am happy to tackle #401 in this PR or another, as you prefer.

briansmith:
> I am happy to tackle #401 in this PR or another, as you prefer.

Thanks, that sounds great! I'd prefer to see #401 in a separate PR based on top of this one.

src/rsa/rsa_pkcs1_sign_tests.txt
```diff
@@ -460,3 +460,10 @@ Key = 308204a40201000282010100d3e94ceebdf732da0040d4ac1b3cc89b81718abfbb5661a516
 Msg = 0123
 Sig = b62bdd9c87f8f409178a5155768b536c1341f52a28c881494baa80b39f4a25a841f664eec49a464adf3cf5c920975cb83de4b06f0c7740685f0174627fcb4d0511bc6a25d473952c33fa5763a8e9bdd00bcdc1e4e5881d061193d77335d7e245987c9333455c9e1a2f0ff45242cff45cbd9ab8859d1c81dda62680ce388caba410be15bb40c0a66ea3a4ef07a3fb1f0f66c656c754dd3b15e694902ea79d1d0496a83408bdc2e94b7bdfef5883fe748ed50027ce0f35380dac8245e042ccdad69cd98440a2f77085a6d02d097f6ef650400b7eb469b9e6ee5284c7c7aee8e6365731114490dffadce428b2d16dd086ae0ef350f8e9dffe22d63203d891394322
 Result = Fail-Invalid-Key
+
+# Invalid key, p & q are too close
```
briansmith: If I calculated correctly, p - q == 348; i.e. they are extremely close together. Please include `p - q == 384` in the comment. Also, please add a comment that this is for a 4096-bit public modulus.

Please add a test vector for
https://gist.github.com/lgarron/44fb9af0e4ae922bce362b62de70e81b#file-2-nbits-div-2-100-apart. IIUC, this second test vector will test the case where they are almost far enough apart, but not quite.

Ideally, we'd also have a test vector for the case where they are just barely far enough apart.

src/rsa/signing.rs
```diff
@@ -77,8 +77,29 @@ impl RSAKeyPair {
...
+    ///   expected modulus length is provided as a parameter, the consistency
+    ///   check between these values and the private key's value of n isn't done.
+    /// - Section 6.4.1.2.1, Step 5: No primality tests are done, both for
+    ///   performance reasons and to avoid any side channels that such tests
+    ///   would provide.
+    ///
+    /// Steps 2 and 4 as described in 6.4.1.2.1 are omitted per 6.4.1.4.3.
+    ///
+    /// In addition to the NIST requirements, we require that p > q and that
+    /// e < 2^33.
```
briansmith: I suggest instead of "e < 2^33" that we either use "e < 2**33" or say that "e must be no more than 33 bits."

src/rsa/signing.rs
```diff
@@ -125,11 +125,34 @@ impl RSAKeyPair {
...
+                    let p_minus_q = {
+                        let p_minus_q = try!(bigint::elem_sub(
+                                p_mod_n, &q_mod_n_decoded, &n));
+                        try!(p_minus_q.into_positive())
+                    };
+                    let min_pq_bitlen_diff = try!(half_n_bits.try_sub(
+                            bits::BitLength::from_usize_bits(100)));
+                    if p_minus_q.bit_length() < min_pq_bitlen_diff {
+                        return Err(error::Unspecified);
+                    }
```
djudd: I think `n.bit_length()` returns `⌊lg(n)⌋+ 1`, actually: `1.bit_length()` = 1, `2.bit_length()` = `3.bit_length()` = 2, etc.

djudd: I think `n.bit_length()` returns `⌊lg(n)⌋+ 1`, actually: `1.bit_length()` = 1, `2.bit_length()` = `3.bit_length()` = 2, etc.

So we have:
`(p-q).bit_length() > lg(p-q) > (p-q).bit_length() - 1`, and `|(p-q).bit_length() - lg(p-q)| < 1`.

That means, if `lg(p-q) > x`, and x is an integer (together with all bit lengths), then `(p-q).bit_length() - 1 >= x`. And conversely if `(p-q).bit_length() - 1 < x`, then `lg(p-q) <= x`. So I think what we should be asserting is `p_minus_q.bit_length() - 1 >= x` or equivalently `p_minus_q.bit_length() > x`. Does that make sense?

Would mean that the expression in the if statement needs to replace a `<` with a `<=` or a `!( ... > ...)`.

src/rsa/rsa_pkcs1_sign_tests.txt
djudd: Ok. I just took this from the first Tweet you linked. I'll add the extra test vector.

(I don't think I updated since you started reviewing so it's weird GitHub thinks this comment is outdated.)

src/rsa/signing.rs
briansmith: Sorry, I could very well be wrong, but I think there's one small (and probably insignificant) edge case. Let's consider a simpler analogous problem: Let's say we want to accept x if and only if `lg x > 2`, i.e. reject x when `lg x ≤ 2`:

| `x` | `lg x`  | `⌊lg x⌋` |
|---|-------|--------|
| 1 | 0     | 0      |
| 2 | 1     | 1      |
| 3 | 1.58  | 1      |
| 4 | 2     | 2      |
| 5 | 2.32  | 2      |

Notice that in one case where `⌊lg x⌋ = 2` (`x = 4`, i.e. the case where `⌊lg x⌋ == lg x`), we should reject `x`, but in another case where `⌊lg x⌋ = 2` (`x = 5`) we accept `x`. IIUC, there's no way to handle this edge case perfectly correctly given only `⌊lg x⌋` but not whether or not `⌊lg x⌋ == lg x`, right?

Going back to our actual problem, IMO it doesn't matter what we do in the case where p - q is exactly equal to 2<sup>nBits/2 - 100</sup>. The important thing is that we minimize the amount of code we add for this check and that we document clearly something along the lines of "This code accepts p - q == 2**(nBits/2 - 100) even though the spec says that value is the largest value that should be rejected, because it's too much work to handle that one value the way the spec says." (I hope you can come up with better wording than I did.)

(See http://crypto.stackexchange.com/a/35096 for background.)

src/rsa/rsa.rs
```diff
@@ -102,13 +102,9 @@ fn check_public_modulus_and_exponent(
         return Err(error::Unspecified);
     }
 
-    // Step 2 / Step b. XXX: FIPS 186-4 seems to indicate that the minimum
-    // exponent value is 2**16 + 1, but it isn't clear if this is just for
-    // signing or also for verification. We support exponents of 3 and larger
-    // for compatibility with other commonly-used crypto libraries.
-    //
+    // Step 2 / Step b.
```
briansmith: I went ahead and added this comment when merging your commit.

src/rsa/signing.rs
```diff
@@ -125,11 +125,34 @@ impl RSAKeyPair {
...
                 // of CRT-based moduluar exponentiation used requires that
                 // |q > p|. (|p == q| is just wrong.)
-                let q = try!(q.into_odd_positive());
                 try!(bigint::verify_less_than(&q, &p));
+                {
+                    let p_mod_n = {
+                        let p = try!(p.try_clone());
+                        try!(p.into_elem_decoded(&n))
+                    };
+                    let p_minus_q = {
```
briansmith: I added this comment when merging this commit.

src/rsa/rsa.rs
briansmith: Please disregard the previous comment, since I'm not merging this yet, as I found a minor issue with the first commit.

src/rsa/signing.rs
briansmith: Please disregard the previous comment, since I'm not merging this yet, as I found a minor issue with the first commit.

src/rsa/signing.rs
```diff
@@ -176,10 +213,18 @@ impl RSAKeyPair {
                     return Err(error::Unspecified);
                 }
 
+                // [NIST SP-800-56B rev. 1]] 6.4.1.4.3 - Step 6
+                //
+                // TODO 6.a: d < LCM(p – 1, q – 1)
+                // TODO 6.b: 1 == (d · e) mod LCM(p – 1, q – 1)
+                //
+                // [NIST SP-800-56B rev. 1]] 6.4.1.4.3 - Step 6.a, partial
                 if !(d.bit_length() > half_n_bits) {
```
briansmith: I know I am the one that initially said the check should be `d.bit_length() > half_n_bits`, but is this actually mathematically consistent with what NIST says? Should it be `d.bit_length() >= half_n_bits` instead? It would be great to write the proof here regarding why the `bit_length()` check makes sense.

Note: It may be helpful to use related facts that that we know, such as `e.bit_length() <= 33` and/or that `d` is odd.

djudd:
Sorry I haven't been responsive to comments. Very busy the last couple of weeks. I hope I'll have some time to revisit by this weekend.

src/rsa/signing.rs
```diff
@@ -176,10 +213,18 @@ impl RSAKeyPair {
                     return Err(error::Unspecified);
                 }
 
+                // [NIST SP-800-56B rev. 1]] 6.4.1.4.3 - Step 6
+                //
+                // TODO 6.a: d < LCM(p – 1, q – 1)
+                // TODO 6.b: 1 == (d · e) mod LCM(p – 1, q – 1)
```
briansmith: During signing, we verify `(b<sup>d</sup>)<sup>e</sup> == b (mod n)`. Isn't this equivalent to checking `1 == (d * e) (mod LCM(p - 1, q - 1)`, assuming p and q are consistent with d. In other words, we verify that p, q, and e are all consistent with each other during signing.

Given that, I think that instead of actually doing these checks that require a LCM computation, we should just document in the public documentation for RSAKeyPair that:

* The validity of the mathematical relationship of `e`, `p`, `q`, and `n` are verified during signature verification, and not during construction. If a check of the validity of the consistency of these values with each other is necessary then that can be done by signing any message with the key pair.

* Some, but not all, validity checks on the value of `d` are skipped because because the checks would be expensive and/or would leak information through side channels. This is OK as far as *ring*'s usage of the key is concerned because *ring* never uses the value of `d` (*ring* always uses `p` and `q` via the Chinese Remainder Theorem, instead of using `d`). However, *ring*'s checks would not be sufficient for validating a key pair are not sufficient for validating a key for use by some other system; that other system must check the value of `d` itself if it uses it.

Then these two TODO comments can be replaced with something like "See the documentation above regarding why there are not checks for `d < LCM(p – 1, q – 1)` or `1 == (d · e) mod LCM(p – 1, q – 1)`.”

briansmith:
> Sorry I haven't been responsive to comments. Very busy the last couple of weeks. I hope I'll have some time to revisit by this weekend.

No worries. I was trying to land this ahead of some changes that I am making that conflict with your changes, but I think instead I'll have to merge my changes ahead of your changes. I think the conflicts will be easy to resolve so hopefully this won't be inconvenient for you.

src/rsa/signing.rs
djudd: We don't use `d`, but we do use `dP` and `dQ`. There's already a TODO in the code about validating `dP` and `dQ`, (predating this PR), which relies on the same argument about checking `(b<sup>d</sup>)<sup>e</sup> == b (mod n)` during signing. Meanwhile, we do validate up front that `pq = n`. 

So, perhaps the documentation should say instead:

* While some size checks of `d`, `dP` and `dQ` are performed at construction, the validity of the mathematical relationship of `dP`, `dQ`, `e` and `n` is verified only during signing, when we validate that `(b**d)**e == b mod n` (having calculated `b**d` using `dP` and `dQ`). If a preemptive check of the consistency of these values with each other is necessary then that can be done by signing any message with the key pair.
* `d` is fully validated neither at construction nor during signing. This is OK as far as _ring_'s usage of the key is concerned because _ring_ never uses the value of `d` (_ring_ always uses `p`, `q`, `dP` and `dQ` via the Chinese Remainder Theorem, instead of using `d`). However, _ring_'s checks would not be sufficient for validating a key pair for use by some other system; that other system must check the value of `d` itself if it uses it.

WDYT?

src/rsa/signing.rs
djudd: Ok. I didn't the exact translation from numbers to bits through very carefully the first time around because it didn't appear to me to matter in practice, so I'm +1 to just adding a comment.

src/rsa/signing.rs
djudd: The bit length of `2**n_bits/2` should be `half_n_bits + 1`, I think, so I think this is correct as-is: this gives us `d >= 2**(n_bits/2)` and knowing `d` is odd makes that `d > 2**(n_bits/2)`. But I'll add a comment.

src/rsa/signing.rs
briansmith: > d is fully validated neither at construction nor during signing.

Your suggested wording sounds good to me, except I don't like starting this sentence with "d is fully validated neither" since it could easily be misread as "d is fully validated". I'd rather start it saying "d is never fully validated, neither" or some such.

briansmith:
OK, I rebased this and landed the changes as:

185637c9e8e69687b84e7d3099221400c80793ad Document NIST steps we have decided to omit for RSAKeyPair validation
ae6946838a787f07f087a9e1a696d9d77ee3e68e Document NIST steps in RSA private key validation
cd49da689a7bf0ca4a923cd32d45a3b0b592f7f8 RSAKeyPair validation: require p - q > 2^(n_bits/2 - 100) per NIST
80270b3edb6bc41965f40c58e37368a23d67be51 RSAKeyPair validation: require e >= 65537 per NIST
2a586e12c0a97586aaa897912f945c4b5e4b1acc RSAKeyPair validation: replace p<d with NIST's stricter d>=half_n_bits

In addition, I made the following changes on top of your commits, to address minor issues:

3ff7e287ff8c26cd44eca70bd7e7cbb9040daaa4 Expand and clarify some of public docs for `RSAKeyPair` construction.
57fc07e266e6d1ecb49adaa02c6e1b11fed02d46 Move the loose upper bound check on `RSAKeyPair`'s `d` to Step 6.
bf5778fadc99f8710327fc980c001ae06efb15bd Clarify comments about steps 2 & 4 of `RSAKeyPair` construction.
a5e464b2d23ec3c85340c29d7d6fd7e3c5414d9d Reorganize the comments for step 1 of `RSAKeyPair` construction.
e9fbec44b644c70963aa58e6ad640b003a603d9c Reorganize the comments for step 3 of `RSAKeyPair` construction.
c4174a1a3635e84a72fb116b6ff95cff8d4ff0a1 Reorganize the comments for step 5 of `RSAKeyPair` construction.
686e24e19fd630b6627c4f54b29ef6ccd1c0fde8 Reorganize the comments for step 6 of `RSAKeyPair` construction.
2403746e3769698b832325e575e0c8901f7f923b Reorganize the comments for step 7 in `RSAKeyPair` construction.
433528d2607c9d5bdf1c9add3a2052ebec164f3d Use ASCII `-` and `*` instead of similar non-ASCII symbols in RSAKeyPair docs.

If/when you have time, please double-check I didn't make any mistakes with my additional changes.

Most, but not all, of my changes were stylistic. Note in particular that I tried to make a tighter correspondence between each NIST-documented (sub-)step and the line(s) of code that implement the (sub-)step, whereas some of your wording was looser. I think the tighter correspondence is better since the purpose of adding these comments is to help reviewers verify that *ring* does what it is required (by NIST or whoever) to do.

I also filed #467 for adding the tighter-bounded test case I mentioned above.

I have to say, this was *very* tedious! I appreciate your amazing patience in doing this!

Issue #380, opened 2016-12-15
Replace `ring::c` with use of the upcoming standard ctypes crate


briansmith:
See https://github.com/rust-lang/rfcs/pull/1783.

...if and only if the standard ctypes crate actually gets created in a way that's better than what libc offers now.

briansmith:
Looks like the Rust change isn't happening.

Issue #357, opened 2016-11-19
Replace C code for reading and parsing `/proc/cpuinfo` and `/proc/self/auxv` with Rust code


briansmith:
cpu-arm-linux.c contains code that determines the CPU capabilities using `getauxval()` on newer versions of Android and by reading from `/proc/cpuinfo` or `/proc/self/auxv` if `getauxval()` isn't available. This is the last bit of C code that does file I/O and one of the last remaining users of the C standard library in *ring*.

Regarding the part that parses /proc/cpuinfo, here's an example of the bad CPU that reports NEON is available but where NEON is actually broken, from https://bugs.chromium.org/p/chromium/issues/detail?id=606629#c12:
```
Processor	: ARMv7 Processor rev 0 (v7l)
processor	: 0
BogoMIPS	: 13.50

processor	: 1
BogoMIPS	: 13.50

Features	: swp half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 
CPU implementer	: 0x51
CPU architecture: 7
CPU variant	: 0x1
CPU part	: 0x04d
CPU revision	: 0

Hardware	: SAMSUNG M2
Revision	: 000e
```

Note that the spaces between the field name and the colon are actualy tabs.

We should have a test that verifies that the parser for the above output is correctly recognized as not supporting NEON, and probably other tests. We don't actually need to run the test on one of the buggy phones; instead we can have the test for this run on in all `#[cfg(all(test, feature = "use_heap"))]` builds on any CPU. In particular, this means that the code that opens the file and parses it needs to be callable from tests that pass in alternative file paths. Then we can put the above output into a test vector file.

briansmith:
See https://rbcommons.com/s/OpenH264/r/383/:

```
Do full register loads instead of single-lane loads in DeblockLumaEq4H_neon

> Instead of loading the registers one lane at a time, load full
> registers and then transpose them.
> 
> This is faster, reducing the runtime for the function from about
> 506 cycles to 434 cycles (tested on a Cortex A8).
> 
> This also avoids an issue which seems like a cpu bug, present
> on Sony Xperia T (cpu implementer 0x51 architecture 7 variant 0x1
> part 0x04d). On such a device, it seemed like the "vswp q9, q10"
> could start executing before the previous
> vld4.u8 {d20[x],d21[x],d22[x],d23[x]}, [r3], r1
> had finished and written back their result. Changing the
[...]

PR #392, opened 2016-12-24
Cpu feature detection

marshallpierce:
As per #357. It looked feasible to re-implement basically all of the offending C file, so I did so.

I've sprinkled TODOs where there are things still to resolve, but the main things I'd like your input on are:

- Linking `getauxval`: In C, it's linked weakly and used if it's available at runtime. I couldn't find how to do this in Rust. Is there a way, or is the best option to leave a weak linkage in C with a little helper function that we call from Rust to then invoke `getauxval` if available and otherwise return an error?
- How should I actually try running this? I don't have an ARM linux system lying around, so the stuff that actually wires the globals, etc in `cpu_feature.rs` is just a guess at how things might look.
- Searching for `GFp_has_broken_NEON` yields no usages. Can we ditch that function and its underlying `g_has_broken_neon` global? (`GFp_armcap_P` is used all over, so the logic is still used to populate that.)
- `getauxval` (and `/proc/self/auxv`) are defined in terms of `unsigned long`. To get things working I used `libc`'s `c_ulong`, but IIRC you're trying to stay `libc`-free. How should I represent `unsigned long` in its many varieties since #380 is still extant?


marshallpierce:
Oh, and I've had to liberally apply `#[allow(dead_code)]` because ultimately this is all invoked from an arm-only hook. Is that just how this has to be, or is there something cleverer?

src/cpu_feature/auxv.rs
```diff
@@ -0,0 +1,159 @@
...
+                Err(_) => return Err(AuxValError::IoError)
+            }
+        }
+
+        let mut reader = &buf[..];
+        // TODO determine length to read
+        let found_aux_type = reader.read_u64::<NativeEndian>()
+            .map_err(|_| AuxValError::InvalidFormat)?;
+        let aux_val = reader.read_u64::<NativeEndian>()
+            .map_err(|_| AuxValError::InvalidFormat)?;
```
briansmith: How terrible would it be to replace these two lines with lines that don't require the `byteorder` crate?

src/cpu_feature/cpu_feature.rs
```diff
@@ -0,0 +1,61 @@
...
+                HAS_BROKEN_NEON = broken_neon;
+            }
+        }
+    }
+}
+
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+#[allow(dead_code)] // TODO
+fn search_procfs_auxv(auxv_types: &[c_ulong]) -> Result<AuxVals, AuxValError> {
+    let mut auxv = File::open("/proc/self/auxv")
```
briansmith: I suggest that you merge all this logic into `auxv::search_auxv`. In particular, that function should take the path to the file to open, and it should open and read the file itself. Then that function would be used by all the tests, and the arm submodule would simply call `auxv::search_auxv("/proc/self/auxv", <auxv_types>)`.

Doing this would move all the platform-specific code from the cross-platform `cpu_features` module to the target-specific submodule.

src/cpu_feature/cpu_feature.rs
```diff
@@ -0,0 +1,61 @@
...
+extern "C" {
+    #[cfg(all(target_os = "linux", target_arch = "arm"))]
+    #[allow(non_upper_case_globals)]
+    pub static mut GFp_armcap_P: u32;
+}
+
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+#[allow(non_snake_case)]
+extern "C" fn GFp_cpuid_setup() {
+    if let Ok(r) = File::open("/proc/cpuinfo").map(|f| BufReader::new(f)) {
```
briansmith: See the suggestion below. All of this code is cross-platform except the reference to /proc/cpuinfo, so we should move it all to the cross-platform cpuinfo subcrate, possibly having `parse_cpuinfo` open the file and read it itself.

src/cpu_feature/cpu_feature.rs
```diff
@@ -0,0 +1,61 @@
...
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+use self::arm::arm_cpuid_setup;
+
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+static HAS_BROKEN_NEON: bool = false;
+
+// TODO is this used in practice? Its C equivalent is the only user of the
+// global, and the function name has no uses
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+#[allow(non_snake_case)]
```
briansmith: In BoringSSL, it is probably used by Google Chrome in order to help decide whether to prefer AES-GCM or ChaCha20-Poly1305 in TLS. We might do something similar later, but for now it is OK to remove it.

src/cpu_feature/cpu_feature.rs
```diff
@@ -0,0 +1,61 @@
...
+use self::libc::c_ulong;
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+use self::auxv::{AuxValError, AuxVals};
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+use self::cpuinfo::CpuInfo;
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+use self::arm::arm_cpuid_setup;
+
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+static HAS_BROKEN_NEON: bool = false;
```
briansmith: I think you're going to remove this (see below), but note that this would need to be `static mut` so that it can be properly set.

src/cpu_feature/cpu_feature.rs
```diff
@@ -0,0 +1,61 @@
...
+use std::fs::File;
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+use std::io::BufReader;
+// TODO represent unsigned long without libc
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+use self::libc::c_ulong;
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+use self::auxv::{AuxValError, AuxVals};
+#[cfg(all(target_os = "linux", target_arch = "arm"))]
+use self::cpuinfo::CpuInfo;
```
briansmith: I think all the above `use`s can be moved to the `arm` submodule.

src/cpu_feature/cpu_feature.rs
```diff
@@ -0,0 +1,61 @@
...
+    let mut auxv = File::open("/proc/self/auxv")
+        .map_err(|_| AuxValError::IoError)
+        .map(|f| BufReader::new(f))?;
+
+    auxv::search_auxv(&mut auxv, auxv_types)
+}
+
+mod auxv;
+mod cpuinfo;
+mod arm;
```
briansmith: It would be better to name this submodule `arm_linux`. We'll support many non-Linux ARM targets that won't have /proc/cpuinfo or /proc/self/aux, and we should reserve the name `arm` for those non-Linux platforms.

src/cpu_feature/arm.rs
```diff
@@ -0,0 +1,281 @@
+use std::collections::HashSet;
+use std::string::{String, ToString};
+
+use super::libc::c_ulong;
```
marshallpierce: If you're not worried about maintaining the platform/bitwidth mapping, sounds good to me!

src/cpu_feature/cpu_feature.rs
marshallpierce: Great, removed.

src/cpu_feature/cpu_feature.rs
marshallpierce: Done. Accepting a `Read` was a misguided attempt to aid testability; it turned out that I never wanted to stitch together blocks of 16 bytes by hand...

src/cpu_feature/cpu_feature.rs
marshallpierce: Done

src/cpu_feature/cpu_feature.rs
```diff
@@ -0,0 +1,61 @@
...
+fn search_procfs_auxv(auxv_types: &[c_ulong]) -> Result<AuxVals, AuxValError> {
+    let mut auxv = File::open("/proc/self/auxv")
+        .map_err(|_| AuxValError::IoError)
+        .map(|f| BufReader::new(f))?;
+
+    auxv::search_auxv(&mut auxv, auxv_types)
+}
+
+mod auxv;
+mod cpuinfo;
```
marshallpierce: Done, with a slightly different `cfg`; I think the parentheses didn't nest maybe the way you intended? The `target_os` bit ended up excluded from the `all`.

src/cpu_feature/cpu_feature.rs
marshallpierce: Done; `cpu_feature.rs` is now just a place to host `arm_linux`.

src/cpu_feature/auxv.rs
```diff
@@ -0,0 +1,159 @@
...
+                    }
+
+                    read_bytes += n;
+                }
+                Err(_) => return Err(AuxValError::IoError)
+            }
+        }
+
+        let mut reader = &buf[..];
+        // TODO determine length to read
```
marshallpierce: It's referring to the fact that I don't know at runtime whether to call read_u64 or read_u32.

According to libc's implementation of `c_ulong`, solaris is always 64 bits, so using pointer width won't work in that case. Same for `c_long`, by the way -- should we be adding this wrinkle to the declarations in `c.rs`? Redox OS is also always using 64 bits for longs, and that at least used to be 32-bit only OS.

Is there a way we could express via a feature what the width of `ulong` is? That way we could have two method declarations, one that invokes `read_u64` and one with `read_u32`, that are toggled with the feature flag.

src/cpu_feature/auxv.rs
marshallpierce: Depends how much it is problematic to depend on byteorder... :) 

At first glance, https://github.com/BurntSushi/byteorder/blob/master/src/lib.rs is not enormous, so we could strip that down to only support the 32 bit and 64 bit cases, and then do some manual buffering rather than use their `Read` extension. Several hundred lines in total, assuming we maintain the tests. It certainly can be done, but if we keep anything like the same API it will mean keeping most of that library as an internal copy. 

If avoiding this particular dependency is of non-trivial importance, I can trace through the byteorder code and see if there's a more svelte way to extract what we need that would be more svelte.

marshallpierce:
Thanks for the feedback; I think I've addressed everything in round 1. I originally had `auxv` and `cpuinfo` all in one, but it started feeling a little beefy as I was scrolling back and forth writing tests for this or that, so I split it up. If we put it all in `arm_linux.rs` it will be a ~600 line file. To my eyes it's a little more manageable mentally (and scroll-wise) in three files, but I don't feel strongly about it, so I don't mind coalescing them if you prefer.

I don't think I know of any other concrete cpuinfo scenarios that need tests, but I'm happy to add more if you know of other problematic cases.

src/cpu_feature/auxv.rs
briansmith: I see. The important thing is that the auxv code is only used in ARM Linux and we know the ARM Linux ABIs. Therefore, I suggest that instead of using/defining a generic "unsigned long" type in `ring::c`, we “simply” use something `#[cfg(all(target_pointer_width = "32", any(test, all(target_os = "linux", any(target_arch = "arm", target_arch = "aarch64")))))]`  and `#[cfg(all(target_pointer_width = "64", any(test, all(target_os = "linux", any(target_arch = "arm", target_arch = "aarch64")))))]` to define a type alias within this module. The idea is that when you run the tests on any 32-bit target, the 32-bit ARM variant gets used, and when you run the tests on a 64-bit target, the 64-bit AAarch64 variants get used, even when those types don't match the target platforms definition of "unsigned long". Does that make sense?

(The main reason we want to support running these tests on non-ARM-Linux platforms is to make them easy to maintain for people that don't have access to such targets.)

src/cpu_feature/arm_linux/arm_linux.rs
```diff
@@ -0,0 +1,307 @@
...
+    // |getauxval| is not available on Android until API level 20. If it is
+    // unavailable, read from /proc/self/auxv as a fallback. This is unreadable
+    // on some versions of Android, so further fall back to /proc/cpuinfo.
+    // See
+    // https://android.googlesource.com/platform/ndk/+/882ac8f3392858991a0e1af33b4b7387ec856bd2
+    // and b/13679666 (Google-internal) for details. */
+
+    // TODO link to getauxval weakly and use that if available. In particular,
+    // this is needed by the assumption in hwcap_from_cpuinfo that future architectures
+    // will have a working getauxval.
```
briansmith: For `getauxval()`, I recommend just defining a wrapper around it in C code that returns an error code if the function isn't available.

src/cpu_feature/auxv.rs
marshallpierce: It ended up being slightly more complicated because even on a 32-bit platform we need the tests to use 64 bits because that's what the test auxv files use. Fortunately, it only incurred a minor syntactic burden in a few places.

src/cpu_feature/arm_linux/arm_linux.rs
marshallpierce: Done. I'm not particularly confident if I put the C bits in the best possible place, though.

briansmith:
> It ended up being slightly more complicated because even on a 32-bit platform we need the tests to use 64 bits because that's what the test auxv files use. Fortunately, it only incurred a minor syntactic burden in a few places.

I'd prefer that we make some 32-bit ARM auxv files and use them on 32-bit platforms, and similarly get some 64-bit Aarch64 auxv files and use them on 64-bit platforms, and not test 64-bit files on 32-bit and vice versa. We can probably do a bit of crowdsourcing of you're having trouble getting such files.

marshallpierce:
Do you want me to scrap the `AuxvUnsignedLong` abstraction, then? Too many generics? :) I can virtualize 32-bit x86 linux and grab that auxv, but I don't have any arm hardware. I could try setting up qemu to emulate arm and run linux on that, and while that's better than nothing, it wouldn't really be representative of troublesome (or not) phone hardware. I searched a bit and did not find any sample auxv files, so if you have a way to get those bytes from users with the relevant phones, that would be great.

src/cpu_feature/arm_linux/arm_linux.rs
```diff
@@ -0,0 +1,514 @@
...
+    if hwcap & hwcap_features.ARM_HWCAP_NEON > AuxvUnsignedLongNative::from(0_u32) {
+        armcap |= ARMV7_NEON;
+
+        // Some ARMv8 Android devices don't expose AT_HWCAP2. Fall back to
+        // /proc/cpuinfo. See https://crbug.com/596156
+
+        let mut hwcap2 = AuxvUnsignedLongNative::from(0_u32);
+        if let Some(v) = getauxval_provider.getauxval(auxval_types.AT_HWCAP2) {
+            hwcap2 = v;
+        } else if let Some(v) = procfs_auxv.get(&auxval_types.AT_HWCAP2) {
```
marshallpierce: The original C did not check procfs auxv for `HWCAP2` if `getauxval` was unavailable. I can't imagine why we wouldn't check that, so I put it in, but maybe you can think of why that would be undesirable?

crypto/cpu.c
```diff
@@ -0,0 +1,19 @@
+#include <openssl/cpu.h>
+
+#ifdef __linux__
```
briansmith: Yes, that seems right to me. However this is also ARM-specific code. I'd prefer it to be in the cpu-arm-linux.c file that you deleted, to keep that clear.

marshallpierce:
Thanks for the detailed feedback. I certainly understand that this PR has become quite hefty, so I'll split it up. I'll leave this one open for a bit so I can continue to follow up on existing comments, but I'll close it once that's done.

crypto/cpu.c
```diff
@@ -0,0 +1,19 @@
...
+unsigned long getauxval_wrapper(unsigned long type, char *success) {
+    if (getauxval == NULL) {
+        *success = 0;
+        return 0;
+    }
+
+    unsigned long result = getauxval(type);
+    if (errno != 0) {
```
marshallpierce: I agree, this is weird, and felt wrong even with my pretty rusty C. Unfortunately, 0 is a valid value, so the original `getauxval` API was simply broken in that case... In the manpage, they have this:

> BUGS
       Before  the  addition  of  the ENOENT error in glibc 2.19, there was no way to unambiguously distinguish the case where type could not be found from the case where the value corresponding to type was zero.

This is the implementation in glibc's `misc/auxval.c`:
```
unsigned long int
__getauxval (unsigned long int type)
{
#ifdef HAVE_AUX_VECTOR
  ElfW(auxv_t) *p;
#endif

  if (type == AT_HWCAP)
    return GLRO(dl_hwcap);
  else if (type == AT_HWCAP2)
    return GLRO(dl_hwcap2);

#ifdef HAVE_AUX_VECTOR
  for (p = GLRO(dl_auxv); p->a_type != AT_NULL; p++)
    if (p->a_type == type)
      return p->a_un.a_val;
#endif

  __set_errno (ENOENT);
  return 0;
}
```

They also have `elf/tst-auxv.c`:
```
static int
do_test (int argc, char *argv[])
{
  errno = 0;
  const char *execfn = (const char *) getauxval (AT_NULL);

  if (errno != ENOENT)
    {
      printf ("errno is %d rather than %d (ENOENT) on failure\n", errno,
          ENOENT);
      return 1;
    }

  if (execfn != NULL)
    {
      printf ("getauxval return value is nonzero on failure\n");
      return 1;
    }

  errno = 0;
  execfn = (const char *) getauxval (AT_EXECFN);

  if (execfn == NULL)
    {
      printf ("No AT_EXECFN found, AT_EXECFN test skipped\n");
      return 0;
    }

  if (errno != 0)
    {
      printf ("errno erroneously set to %d on success\n", errno);
      return 1;
    }

  if (strcmp (argv[0], execfn) != 0)
    {
      printf ("Mismatch: argv[0]: %s vs. AT_EXECFN: %s\n", argv[0], execfn);
      return 1;
    }

  return 0;
}
```

I took this to mean we *should* check errno. https://www.securecoding.cert.org/confluence/pages/viewpage.action?pageId=6619179 seems to agree that while regrettable, this is how it has to be done when every possible return value is potentially valid. Thoughts?

src/cpu_feature/arm_linux/auxv.rs
```diff
@@ -0,0 +1,313 @@
...
+        assert_eq!(0, 1 << X86_ACPI & hwcap);
+
+        assert!(!vals.contains_key(&u64_auxv_types().AT_HWCAP2));
+
+        assert_eq!(&1000_u64, vals.get(&AT_UID_64).unwrap());
+    }
+
+    #[test]
+    fn test_parse_auxv_virtualbox_linux_32bit() {
+        let path = Path::new("src/cpu_feature/arm_linux/test-data/macos-virtualbox-linux-x86-4850HQ.auxv");
```
marshallpierce: I was unable to find 32-bit arm auxv files, so if you've got any, that would be great!

src/cpu_feature/arm_linux/arm_linux.rs
```diff
@@ -0,0 +1,514 @@
...
+        ret |= ARMV8_SHA1;
+    }
+    if hwcap2 & hwcap_features.ARM_HWCAP2_SHA2 > T::from(0) {
+        ret |= ARMV8_SHA256;
+    }
+
+    return ret;
+}
+
+fn hwcap_from_cpuinfo<T: AuxvUnsignedLong>(cpuinfo: &CpuInfo,
```
marshallpierce: Depends on what you expect as a reader from `cpuinfo`, I guess -- I kept that module completely hardware agnostic. At least in its current form, it only exists to put a pretty face on the format of `/proc/cpuinfo` (just as auxv.rs is only a nice way to access the aux vector), and all arm-specific decisions based on that data are elsewhere. For context, my instinct with this sort of thing is to try to make things general purpose so as to avoid leaking concerns across layers unless there are performance concerns or something like that that would preclude such a design. As it is, you could hoist cpuinfo.rs pretty much as is into its own crate and it wouldn't leak any of its arm-feature-detection heritage, and that feels right to me, but I'm not tied to that structure if it bugs you.

src/cpu_feature/arm_linux/cpuinfo.rs
```diff
@@ -0,0 +1,153 @@
...
+pub enum CpuInfoError {
+    IoError,
+    InvalidFormat
+}
+
+pub type CpuInfo = HashMap<String, String>;
+
+#[cfg(all(any(target_arch = "arm", target_arch = "aarch64"), target_os="linux"))]
+pub fn parse_cpuinfo() -> Result<CpuInfo, CpuInfoError> {
+    match File::open("/proc/cpuinfo").map(|f| BufReader::new(f)) {
```
marshallpierce: OK. BTW, do you object to `?` rather than the `try!` macro?

src/cpu_feature/arm_linux/cpuinfo.rs
```diff
@@ -0,0 +1,153 @@
...
+        Ok(mut r) => parse_cpuinfo_reader(&mut r),
+        Err(_) => Err(CpuInfoError::IoError)
+    }
+}
+
+/// parse the contents of /proc/cpuinfo into a map of field names to values.
+/// Keeps the first encountered value for each name.
+#[cfg(any(test, target_os="linux"))]
+pub fn parse_cpuinfo_reader(input: &mut BufRead)
```
marshallpierce: I could change it to take a Path instead of a BufRead and convert the tests that use string literals to all have their own files instead, but I'll still need *something* that's not just "read from exactly /proc/cpuinfo" to be able to test it.

The reason it's public is that it's used in testing the higher-level armcap logic with various cpuinfo files. Is there a better way to expose that to a test submodule of another module?

briansmith:
I'm closing this in favor of #408 and whatever follow-up PRs come from that.

src/cpu_feature/arm_linux/auxv.rs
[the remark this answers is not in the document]
gnzlbg: @briansmith did you manage to get any 32-bit ARM auxv files? We could use them in stdsimd :)

PR #395, opened 2016-12-28
WIP: Refactor bigint code to be more generic.

briansmith:
This PR will probably get broken up into multiple PRs, but I want to share an early draft of this.


briansmith:
@ranweiler Earlier you seemed interested in the way the `Elem` and `ElemDecoded` types encode information about the Montgomery factor. LMK if you have any thoughts on the refactoring of that idea in a0fbc89.

ranweiler:
@briansmith I _really_ like the approach in https://github.com/briansmith/ring/commit/a0fbc896af2c0f01bd35276a2fffc9787eb9279c.

It is subtle, but having the encoding status as a type parameter makes it a lot easier for me to read the resulting code. Even though a value being encoded makes sense as the "default", and thus worthy of being called an `Elem` (prior to these changes), the `ElemDecoded` naming felt inverted (and resulted in some function names that felt awkward in context). For example: `into_elem_decoded` vs. the new `into_unencoded`. I also like how this works out with polymorphic code.

Suggestion: for the relevant `PhantomData` field, consider a name other than `montgomery_factor`. With that naming, I'd expect the type that the field "uses" to represent a specific numeric choice of the Montgomery `R`, similar to how the `ring` field encodes a specific choice of ring. Instead, the type (and thus the struct field) tells us about the presence or absence of some unnamed, ambient factor of `R`. I would thus find a field name like `encoded` or `encoding` to be less surprising.

briansmith:
> Even though a value being encoded makes sense as the "default", and thus worthy of being called an Elem (prior to these changes), the ElemDecoded naming felt inverted (and resulted in some function names that felt awkward in context).

I agree. For historical context, in crypto/ec/suite_b I used the same naming convention, but it made more sense because in that code we're almost always dealing with values in the Montgomery domain and so any use of non-Montgomery forms were worth calling out.

> Suggestion: for the relevant PhantomData field, consider a name other than montgomery_factor. With that naming, I'd expect the type that the field "uses" to represent a specific numeric choice of the Montgomery R, similar to how the ring field encodes a specific choice of ring. Instead, the type (and thus the struct field) tells us about the presence or absence of some unnamed, ambient factor of R. I would thus find a field name like encoded or encoding to be less surprising.

That makes sense.

Now I'm actually thinking to have just one `ring: PhantomData<(M, F)>` field, instead of two separate ones. In particular, going forward I want to experiment with Almost Montgomery Multiplication, where we don't fully reduce elements during multiplication. If we were to do so, we'd need separate multiplication and addition operators for unencoded and encoded elements, and we probably wouldn't be able to use `elem_mul` as a shortcut for multiplying and changing the encoding, because the set of encoded elements is larger than the set of unencoded elements.

briansmith:
I'm closing this PR because I'm going to split up the changes into multiple PRs that are grouped more logically. Thanks for the feedback!

PR #396, opened 2016-12-29
Add Redox support

jackpot51:
This adds support for Redox. I will annotate changes in a GitHub review.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -260,7 +260,7 @@ name = "ring"
 [dependencies]
 untrusted = "0.3.2"
 
-[target.'cfg(unix)'.dependencies]
+[target.'cfg(any(target_os = "redox", unix))'.dependencies]
 lazy_static = "0.2.1"
 
 [features]
--- a/mk/top_of_makefile.mk
+++ b/mk/top_of_makefile.mk
@@ -41,11 +41,11 @@ else ifeq ($(findstring linux-android,$(TARGET_VENDOR)-$(TARGET_SYS)),linux-andr
 TARGET_ABI = $(TARGET_SYS)
 TARGET_VENDOR = unknown
 TARGET_SYS = linux
-else
+else ifeq ($(findstring redox,$(TARGET_SYS)),redox)
 
 # If we find `bsd` in the target system, we can assume a target triple is OK,
 # so skip the error here.
-ifeq (,$(findstring bsd,$(TARGET_SYS)))
+else ifeq (,$(findstring bsd,$(TARGET_SYS)))
 define NEWLINE
 
 
@@ -60,7 +60,6 @@ $(error TARGET must be of the form \
         Android example:	TARGET=arm-linux-androideabi $(NEWLINE)\
 \
         NOTE: Use "i586" instead of "x86")
-endif
 endif
 endif
 
@@ -231,7 +230,9 @@ CPPFLAGS += \
 # it's use seems questionable for that kind of target anyway.
 # The launchpad.net arm-none-eabi-gcc toolchain (at least) uses -fshort-enums.
 ifneq ($(TARGET_SYS),none)
-CPPFLAGS += -fstack-protector
+	ifneq ($(findstring redox,$(TARGET_SYS)),redox)
+		CPPFLAGS += -fstack-protector
+	endif
 endif
 
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -99,7 +99,8 @@
 #[cfg(feature = "internal_benches")]
 extern crate test as bench;
 
-#[cfg(any(all(unix,
+#[cfg(any(target_os = "redox",
+          all(unix,
               any(not(target_os = "linux"),
                   feature = "dev_urandom_fallback"))))]
 #[macro_use]
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -57,6 +57,9 @@ pub trait SecureRandom {
 /// implemented by reading from `/dev/urandom`. (This is something that should
 /// be improved, at least for platforms that offer something better.)
 ///
+/// On Redox, `fill()` is implemented by reading from `rand:`. In the future,
+/// a random primitive similar to `getrandom` may be provided.
+///
 /// On Linux, `fill()` will use the [`getrandom`] syscall. If the kernel is too
 /// old to support `getrandom` then by default `fill()` falls back to reading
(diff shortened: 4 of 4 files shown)


Cargo.toml
```diff
@@ -262,7 +262,7 @@ name = "ring"
 [dependencies]
 untrusted = "0.3.2"
 
-[target.'cfg(unix)'.dependencies]
+[target.'cfg(any(target_os = "redox", unix))'.dependencies]
```
jackpot51: Redox also needs lazy_static

mk/top_of_makefile.mk
```diff
@@ -41,11 +41,11 @@ else ifeq ($(findstring linux-android,$(TARGET_VENDOR)-$(TARGET_SYS)),linux-andr
 TARGET_ABI = $(TARGET_SYS)
 TARGET_VENDOR = unknown
 TARGET_SYS = linux
-else
+else ifeq ($(findstring redox,$(TARGET_SYS)),redox)
```
jackpot51: Redox is a supported system, so do not throw an error on finding it

mk/top_of_makefile.mk
```diff
@@ -231,7 +230,9 @@ CPPFLAGS += \
 # it's use seems questionable for that kind of target anyway.
 # The launchpad.net arm-none-eabi-gcc toolchain (at least) uses -fshort-enums.
 ifneq ($(TARGET_SYS),none)
-CPPFLAGS += -fstack-protector
+	ifneq ($(findstring redox,$(TARGET_SYS)),redox)
```
jackpot51: Redox does not support libssp at the moment

src/lib.rs
```diff
@@ -99,7 +99,8 @@
 #[cfg(feature = "internal_benches")]
 extern crate test as bench;
 
-#[cfg(any(all(unix,
+#[cfg(any(target_os = "redox",
```
jackpot51: Redox will need lazy_static (at line 196)

src/rand.rs
```diff
@@ -140,6 +140,28 @@ mod sysrand {
 }
 
 // Keep the `cfg` conditions in sync with the conditions in lib.rs.
+#[cfg(target_os = "redox")]
```
jackpot51: Redox has what Linux has at `/dev/urandom`, at `rand:`.

mk/top_of_makefile.mk
briansmith: Is there a Redox issue that we could link to in a comment? Or if not, could you create one and then link it?

src/rand.rs
briansmith: Rather than duplicate the code for the `urandom` submodule, let's abstract out the one difference, the path to the random file, and share the rest of the code.

briansmith:
Also, is the "C" calling convention in Redox the same as Linux or different in any way? This is important for understanding whether the assembly language stuff is correct.

mk/top_of_makefile.mk
jackpot51: Yes, there is an issue here: https://github.com/redox-os/redox/issues/815

src/rand.rs
jackpot51: I will see about that.

jackpot51:
@briansmith The calling conventions are identical to those in Linux - System V ABI

I will update the documentation.

There is no plan for a getrandom syscall. Redox is a microkernel, and randomness has been sourced through a userspace process, through a device node at `rand:`. A special case for randomness *might* be made to allow for centralized entropy collection, I have an issue here: https://github.com/redox-os/redox/issues/816

It would be possible but messy to run the tests on Redox. It would involve cross compiling the tests, running Redox in QEMU, and running a script in Redox to drive the automatic download and running of the tests.

src/rand.rs
```diff
@@ -140,17 +143,22 @@ mod sysrand {
...
-#[cfg(all(unix,
+#[cfg(all(any(target_os = "redox", unix),
           not(all(target_os = "linux",
                   not(feature = "dev_urandom_fallback")))))]
 mod urandom {
     extern crate std;
     use error;
 
     pub fn fill(dest: &mut [u8]) -> Result<(), error::Unspecified> {
+        #[cfg(target_os = "redox")]
```
jackpot51: I simplified the differences between Redox and Posix-like systems

src/rand.rs
```diff
@@ -57,6 +57,9 @@ pub trait SecureRandom {
 /// implemented by reading from `/dev/urandom`. (This is something that should
 /// be improved, at least for platforms that offer something better.)
 ///
+/// On Redox, `fill()` is implemented by reading from `rand:`. In the future,
```
jackpot51: I added the requested module-level documentation

jackpot51:
@briansmith What do you think of my updates?

mk/top_of_makefile.mk
briansmith: OK, please add a link to that issue as a comment here.

briansmith: Please add a comment with a link to that issue here.

jackpot51:
@briansmith I have squashed it, rebased, and added the contributors statement.

All I ask is that you please `cargo publish` as soon as possible, with the current 0.6* version, so that it can be used in the wild.

mk/top_of_makefile.mk
briansmith: When double-checking this before merging this, I realized that I don't understand this. What are the target triples for Redox, and in particular isn't the target triple always of the form $(architecture)-$(vendor)-$(os)-$(abi)?, e.g. "i686-unknown-redox-gnu"? If so, what's the point of this change? If not, what are the target triples for redox?

mk/top_of_makefile.mk
briansmith: Again, similar question. Why are we doing:
```make
ifneq ($(findstring redox,$(TARGET_SYS)),redox)
```
instead of:
```make
ifneq ($(TARGET_SYS),redox)
```

We need to know this so that we can do the right thing when we replace the makefile.

briansmith:
In 0.6.2:
8a0976b363e41ae9586391bd7da493b89862300d
429f5c25baecaccbbe649fa0ed906435bdda7754.

I tweaked your commit. Hopefully I didn't mess it up.

jackpot51:
Looking good - thanks @briansmith

PR #397, opened 2016-12-31
Add some tests of non-aligned digest inputs.

### LINKED ISSUE #178: Add tests of unaligned inputs for `ring::digest`
In particular, test on platforms that require 4-byte alignment for 32-bit integers and/or 8-byte alignment for 64-bit integers.

brian-pane:
(fixes #178)

This diff tests up to a 32-byte alignment, in case any current or future implementation uses vector instructions that require an alignment that large.


briansmith:
Brian, thanks for doing this!

Before I do the review, do you think it is worthwhile to factor out the common logic between this and the AEAD counterpart? (Maybe putting some abstraction in `ring::test`?)

brian-pane:
I like that idea; it will make the test code simpler overall.  I'll add an "AlignmentShiftByteVec" (or something like that), use it in the AEAD tests as well as here, and then update this pull request.

brian-pane:
My commits got duplicated due to a bad merge on my part, so I'm going to close this pull request and create a clean one.

PR #398, opened 2017-01-01
Use SecRandomCopyBytes on macOS

### LINKED ISSUE #149: Improve random number generation on Mac OS X and iOS
Avoid using `dev/urandom` on Mac OS X and instead do whatever is most like using the `getrandom` syscall on Linux (See #148).

References:
- http://stackoverflow.com/questions/9234686/generating-random-numbers-securely-in-objective-c
- https://stackoverflow.com/questions/21734909/is-os-xs-secrandomcopybytes-fork-safe
- https://developer.apple.com/library/prerelease/ios/documentation/Security/Reference/RandomizationReference/index.html#//apple_ref/c/func/SecRandomCopyBytes

Note, in particular, that `SecRandomCopyBytes` was added in OS X 10.7 (Lion). According to [this Rust issue](https://github.com/rust-lang/rust/issues/11927), 10.7 is the lowest version supported by Rust anyway, although we should verify that hasn't changed since then.

dirk:
Uses the [`SecRandomCopyBytes` function](https://developer.apple.com/reference/security/1399291-secrandomcopybytes?language=objc) from the `Security.framework` on macOS and iOS.

Fixes #149.

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -25,8 +25,10 @@
 //! (seccomp filters on Linux in particular). See `SystemRandom`'s
 //! documentation for more details.
 
-
-#[cfg(any(target_os = "linux", windows, test))]
+#[cfg(any(target_os = "linux",
+          target_os = "ios",
+          windows,
+          test))]
 use c;
 
 #[cfg(test)]
@@ -109,7 +111,10 @@ impl SecureRandom for SystemRandom {
     }
 }
 
-#[cfg(not(any(target_os = "linux", windows)))]
+#[cfg(not(any(target_os = "linux",
+              target_os = "macos",
+              target_os = "ios",
+              windows)))]
 use self::urandom::fill as fill_impl;
 
 #[cfg(any(all(target_os = "linux", not(feature = "dev_urandom_fallback")),
@@ -118,6 +123,9 @@ use self::sysrand::fill as fill_impl;
 
 #[cfg(all(target_os = "linux", feature = "dev_urandom_fallback"))]
 use self::sysrand_or_urandom::fill as fill_impl;
+
+#[cfg(any(target_os = "macos", target_os = "ios"))]
+use self::darwin::fill as fill_impl;
 
 #[cfg(any(target_os = "linux", windows))]
 mod sysrand {
@@ -143,6 +151,7 @@ mod sysrand {
 
 // Keep the `cfg` conditions in sync with the conditions in lib.rs.
 #[cfg(all(any(target_os = "redox", unix),
+          not(any(target_os = "macos", target_os = "ios")),
           not(all(target_os = "linux",
                   not(feature = "dev_urandom_fallback")))))]
 mod urandom {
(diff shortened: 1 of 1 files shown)


src/rand.rs
```diff
@@ -195,6 +205,21 @@ mod sysrand_or_urandom {
...
+
+    pub fn fill(dest: &mut [u8]) -> Result<(), error::Unspecified> {
+        let r = unsafe {
+            super::SecRandomCopyBytes(super::kSecRandomDefault, dest.len(), dest.as_mut_ptr())
+        };
+        match r {
+            0 => Ok(()),
+            _ => Err(error::Unspecified),
+        }
+    }
```
dirk: @briansmith: So I was originally doing something like this:

```rs
#[repr(c)]
struct SecRandomRef {}
```

But the compiler complained about empty struct layout (for good reason). I'm guessing `enum`—as you suggested—is the proper way to do this?

(Just satisfying my curiosity here. 🙃 )

src/rand.rs
```diff
@@ -227,6 +252,15 @@ extern {
     fn GFp_sysrand_chunk(buf: *mut u8, len: c::size_t) -> c::long;
 }
 
+#[cfg(any(target_os = "macos", target_os = "ios"))]
+#[link(name = "Security", kind = "framework")]
+extern {
```
dirk: @briansmith: I tried putting them in the sub-module, but doing so required another `use super::super::c` in the sub-module, which seemed to go against the prevailing style in this project. Would you prefer the sub-mobile-with-a-`use` approach?

briansmith: Instead of `use super::super::c` I think you can just `use c`;

briansmith:
> I'm guessing enum—as you suggested—is the proper way to do this?

It is the least bad way we have for now. I use this `enum` pattern for `BIGNUM` and `BN_MONT_CTX` in `ring::rsa::bigint`.

dirk:
@briansmith: Comments addressed! 😄

briansmith:
Great, thanks!

I squashed all your commits and put them in my own dirk-macos-use-secrandomcopybytes branch: https://github.com/briansmith/ring/commits/dirk-macos-use-secrandomcopybytes. I also reformatted it to the 80-column limit we use in *ring*.

Please amend the commit message to add the contributor's statement from https://github.com/briansmith/ring#contributing, and I'll check this in ASAP.

Thanks!

dirk:
@briansmith: Done! 😄

briansmith:
Thanks!

This landed in master as dd27044b8b360440d96f06556af23fad10b2ab98. I also landed a documentation update: caaa3f758492e15cf02986b0f18ccbeab442bd9f.

briansmith:
See the follow-up https://github.com/briansmith/ring/issues/416#issuecomment-274600719 and cca54a59c813dce60dcec040838b712da14a2fc7.

dirk:
@briansmith: Glad to see this land! 😄

briansmith:
Follow-up to fix a Rust Nightly build failure due to stricter unused-item checking in Nightl: 4410207a4ec399241209a69e979bcd3ceb93ee1d.

Thanks again for contributing this !

briansmith:
Another follow-up, mostly a no-op: b88b633179a464316dcf8749b932f1335e22d14b.

PR #399, opened 2017-01-02
Add tests of non-aligned digest inputs

### LINKED ISSUE #178: Add tests of unaligned inputs for `ring::digest`

brian-pane:
Add tests of non-aligned digest inputs, and reuse the supporting`PrefixedByteVec` class for the AEAD tests.
(fixes #178)

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -122,6 +122,7 @@ use bits;
 use {digest, error};
 
 use std;
+use std::ops::{Deref, DerefMut};
 use std::string::String;
 use std::vec::Vec;
 use std::io::BufRead;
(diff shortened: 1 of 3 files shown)


src/test.rs
```diff
@@ -370,6 +371,73 @@ fn parse_test_case(current_section: &mut String, lines: &mut FileLines)
...
+            prefix_len: prefix_len,
+            vec: Vec::with_capacity(prefix_len + data.len()),
+        };
+        v.replace_from(prefix_len, data);
+        v
+    }
+
+    /// Replace a PrefixedByteVec's contents with a copy of the provided
+    /// slice, preceded by `prefix_len` dummy bytes.
+    pub fn replace_from(&mut self, prefix_len: usize, data: &[u8]) {
```
briansmith: The problem that I see here is that we don't have any way of ensuring that things are aligned any particular way.

src/test.rs
```diff
@@ -370,6 +371,73 @@ fn parse_test_case(current_section: &mut String, lines: &mut FileLines)
...
+        self.vec.clear();
```
briansmith: `self.vec.clear()` may throw away the vector's storage...

src/test.rs
```diff
@@ -370,6 +371,73 @@ fn parse_test_case(current_section: &mut String, lines: &mut FileLines)
...
+        for _ in 0..prefix_len {
+            self.vec.push(123);
```
briansmith: Here we may allocate or re-allocate the storage one or more times. Potentially, these allocations could have differing alignments, such that when `alignment + prefix_len` results in duplicates, and such that some alignments wouldn't be tested.

src/test.rs
```diff
@@ -370,6 +371,73 @@ fn parse_test_case(current_section: &mut String, lines: &mut FileLines)
...
+        }
+        self.vec.extend(data.iter());
```
briansmith: ...and/or this could align the storage at potentially any alignment.

src/test.rs
```diff
@@ -370,6 +371,73 @@ fn parse_test_case(current_section: &mut String, lines: &mut FileLines)
...
+    pub fn with_capacity(size: usize) -> PrefixedByteVec {
+        PrefixedByteVec {
+            prefix_len: 0,
+            vec: Vec::with_capacity(size),
+        }
+    }
+
+    /// Create an PrefixedByteVec containing a copy of the provided slice,
+    /// preceded by `prefix_len` dummy bytes.
+    pub fn from(prefix_len: usize, data: &[u8]) -> PrefixedByteVec {
```
briansmith: This `from` function is dead code and needs to be removed.

briansmith:
Pushing to 0.7, which I am going to get started with ASAP. In particular, I'm removing the `in_prefix_len` logic from `ring::aead`. That will make it easier to test *alignment* stuff in ring::aead, which we're only kinda-sorta testing now, in addition to testing alignment stuff ring::digest. Hopefully we can share some of the logic between the two.

PR #400, opened 2017-01-03
Building on FreeBSD

paukstis:
Hi, clang is default on FreeBSD, would need to link with libc++, cargo test ok.


build.rs
```diff
@@ -137,7 +137,8 @@ fn build_c_code(out_dir: &str) -> Result<(), std::env::VarError> {
     // how to do that yet.
     println!("cargo:rustc-link-lib=static={}-test", LIB_NAME);
     if !use_msbuild {
-        let libcxx = if target_str.contains("-apple-") {
+        let libcxx = if target_str.contains("-apple-") ||
+                  target_triple[2].contains("freebsd") {
```
briansmith: Why did you use `target_triple[2].contains("freebsd")` instead of `target_triple[2] == "freebsd"`?

paukstis: Lazy to check the exact string, indeed, that is the exact match. I'll make a new PR.

PR #403, opened 2017-01-04
Target triple compare to exact string "freebsd"

paukstis:


briansmith:
Thanks! This landed in master as 095d1076d51708ca961db45ad32079e4b9c423b8.

When you submit PRs in the future, please add the contributor statement to the end of the commit message. See https://github.com/briansmith/ring#contributing. In this case, the change is small enough that it doesn't mater.

Also, this project uses a rebase-based worflow, not a merge-based workflow, so avoid merge commits in the PR's commit list. Basically, instead of merging, use `git rebase master`.

Thanks again!

PR #408, opened 2017-01-08
Detect cpu features on ARM with getauxval()

marshallpierce:
This is 1 and 2 of your suggested PR disassembly in #392. I elected to do them together because otherwise this would have failed to build on arm because there wouldn't be a replacement for `GFp_cpuid_setup`.

### DIFF
--- a/crypto/cpu-arm-linux.c
+++ b/crypto/cpu-arm-linux.c
@@ -16,6 +16,8 @@
 
 #ifdef __linux__
 
+#include <errno.h>
+
 /* |getauxval| is not available on Android until API level 20. Link it as a weak
  * symbol and use other methods as fallback. As of Rust 1.14 this weak linkage
  * isn't supported, so we do it in C.
@@ -24,24 +26,15 @@ unsigned long getauxval(unsigned long type) __attribute__((weak));
 
 /*
  * If getauxval is not available, or an error occurs, return 0.
- * Otherwise, return the value found.
+ * Otherwise, return the value found (which may be zero).
  */
 unsigned long getauxval_wrapper(unsigned long type);
-
-#include <errno.h>
 
 unsigned long getauxval_wrapper(unsigned long type) {
     if (getauxval == NULL) {
         return 0;
     }
 
-    unsigned long auxval = getauxval(type);
-    // map errors to a zero value
-    if (errno != 0) {
-        errno = 0;
-        return 0;
-    }
-
-    return auxval;
+    return getauxval(type);
 }
 #endif
--- a/crypto/cpu-arm-linux.c
+++ b/crypto/cpu-arm-linux.c
@@ -1,3 +1,17 @@
+/* Copyright (c) 2016, Google Inc.
+ *
+ * Permission to use, copy, modify, and/or distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
+ * SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+ * OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE. */
+
 #include <openssl/cpu.h>
 
 #ifdef __linux__
--- a/crypto/cpu-arm-linux.c
+++ b/crypto/cpu-arm-linux.c
@@ -13,10 +13,9 @@
  * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE. */
 
 #include <openssl/cpu.h>
+#include <errno.h>
 
 #ifdef __linux__
-
-#include <errno.h>
 
 /* |getauxval| is not available on Android until API level 20. Link it as a weak
  * symbol and use other methods as fallback. As of Rust 1.14 this weak linkage
--- a/crypto/cpu-arm-linux.c
+++ b/crypto/cpu-arm-linux.c
@@ -1,358 +1,45 @@
-/* Copyright (c) 2016, Google Inc.
- *
- * Permission to use, copy, modify, and/or distribute this software for any
- * purpose with or without fee is hereby granted, provided that the above
- * copyright notice and this permission notice appear in all copies.
- *
- * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
- * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
- * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
- * SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
- * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
- * OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
- * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE. */
-
 #include <openssl/cpu.h>
 
-#if defined(OPENSSL_ARM) && !defined(OPENSSL_STATIC_ARMCAP)
-
-#include <errno.h>
-#include <fcntl.h>
-#include <string.h>
-#include <sys/types.h>
-#include <unistd.h>
-
-#include <openssl/arm_arch.h>
-#include <openssl/mem.h>
-
-#include "internal.h"
-
-
-#define AT_HWCAP 16
-#define AT_HWCAP2 26
-
-#define HWCAP_NEON (1 << 12)
-
-/* See /usr/include/asm/hwcap.h on an ARM installation for the source of
- * these values. */
-#define HWCAP2_AES (1 << 0)
-#define HWCAP2_PMULL (1 << 1)
-#define HWCAP2_SHA1 (1 << 2)
-#define HWCAP2_SHA2 (1 << 3)
+#ifdef __linux__
 
 /* |getauxval| is not available on Android until API level 20. Link it as a weak
- * symbol and use other methods as fallback. */
+ * symbol and use other methods as fallback. As of Rust 1.14 this weak linkage
+ * isn't supported, so we do it in C.
+ */
 unsigned long getauxval(unsigned long type) __attribute__((weak));
 
-static int open_eintr(const char *path, int flags) {
-  int ret;
-  do {
-    ret = open(path, flags);
-  } while (ret < 0 && errno == EINTR);
-  return ret;
-}
+/*
+ * getauxval() may or may not be available. In addition, it may not find
+ * the type requested.
+ * - If getauxval() is not available, this returns -1.
+ * - If getauxval() is available but the requested type is not found or another
+ *   error occurs, this returns 0.
+ * - If getauxval() is available but a different error occurs, this returns -2.
+ * - If getauxval() is available and the requested type is found, this returns
+ *   1 and also writes to the result pointer param.
+ */
+int32_t getauxval_wrapper(unsigned long type, unsigned long *result);
 
-static ssize_t read_eintr(int fd, void *out, size_t len) {
-  ssize_t ret;
-  do {
-    ret = read(fd, out, len);
-  } while (ret < 0 && errno == EINTR);
-  return ret;
-}
+#include <errno.h>
 
-/* read_full reads exactly |len| bytes from |fd| to |out|. On error or end of
- * file, it returns zero. */
-static int read_full(int fd, void *out, size_t len) {
-  while (len > 0) {
-    ssize_t ret = read_eintr(fd, out, len);
-    if (ret <= 0) {
-      return 0;
-    }
-    out = (uint8_t *)out + ret;
-    len -= ret;
-  }
-  return 1;
-}
-
-/* read_file opens |path| and reads until end-of-file. On success, it returns
- * one and sets |*out_ptr| and |*out_len| to a newly-allocated buffer with the
- * contents. Otherwise, it returns zero. */
-static int read_file(char **out_ptr, size_t *out_len, const char *path) {
-  int fd = open_eintr(path, O_RDONLY);
-  if (fd < 0) {
-    return 0;
-  }
-
-  static const size_t kReadSize = 1024;
-  int ret = 0;
-  size_t cap = kReadSize, len = 0;
-  char *buf = OPENSSL_malloc(cap);
-  if (buf == NULL) {
-    goto err;
-  }
-
-  for (;;) {
-    if (cap - len < kReadSize) {
-      size_t new_cap = cap * 2;
-      if (new_cap < cap) {
-        goto err;
-      }
-      char *new_buf = OPENSSL_realloc(buf, new_cap);
-      if (new_buf == NULL) {
-        goto err;
-      }
-      buf = new_buf;
-      cap = new_cap;
+int32_t getauxval_wrapper(unsigned long type, unsigned long *result) {
+    if (getauxval == NULL) {
+        return -1;
     }
 
-    ssize_t bytes_read = read_eintr(fd, buf + len, kReadSize);
-    if (bytes_read < 0) {
-      goto err;
+    unsigned long auxval = getauxval(type);
+    if (errno == ENOENT) {
+        // as of glibc 2.19, errno is ENOENT if the type is not found.
+        errno = 0;
+        return 0;
+    } else if (errno != 0) {
+        // as of glibc 2.23 the only error is enoent, but more errors
+        // may be added
+        errno = 0;
+        return -2;
     }
-    if (bytes_read == 0) {
-      break;
-    }
-    len += bytes_read;
-  }
 
-  *out_ptr = buf;
-  *out_len = len;
-  ret = 1;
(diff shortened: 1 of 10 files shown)


mk/ring.mk
```diff
@@ -32,6 +32,7 @@ RING_SRCS = $(addprefix $(RING_PREFIX), \
   crypto/bn/random.c \
   crypto/bn/shift.c \
   crypto/cipher/e_aes.c \
+  crypto/cpu-arm-linux.c \
```
marshallpierce: This isn't ideal, and goes against the spirit of the filename...

Upon making this mini-PR I remembered why I put the C where I did: if the C bits are ARM-only with no provision for running tests, then so is all the Rust that uses it. So, to make the Rust (and Rust/C interaction) testable on average joe hardware (i.e. x86 Linux), I made cpu-arm-linux.c, uh, not ARM-specific. What's the right way to do this (where "this" is "make it testable on x86 linux, but in a production build only have this stuff present on ARM")?

crypto/cpu-arm-linux.c
```diff
@@ -1,358 +1,45 @@
...
- * copyright notice and this permission notice appear in all copies.
- *
- * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
- * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
- * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
- * SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
- * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
- * OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
- * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE. */
-
```
briansmith: The license must stay the same. If you think the Google copyright statement is not appropriate then suggest a different copyright statement.

crypto/cpu-arm-linux.c
```diff
@@ -1,358 +1,45 @@
...
+ * getauxval() may or may not be available. In addition, it may not find
+ * the type requested.
+ * - If getauxval() is not available, this returns -1.
+ * - If getauxval() is available but the requested type is not found or another
+ *   error occurs, this returns 0.
+ * - If getauxval() is available but a different error occurs, this returns -2.
+ * - If getauxval() is available and the requested type is found, this returns
+ *   1 and also writes to the result pointer param.
+ */
+int32_t getauxval_wrapper(unsigned long type, unsigned long *result);
```
briansmith: Our goal is to do as much in Rust as reasonable. That means doing as little in C as we can get away with. With that in mind, why not do this:

```c
/* getauxval_wrapper returns 0 if |getauxval| is available and 1 otherwise.
 * If |getauxval| is available then |*result| is set to the result of
 * |getauxval(type)| before |getauxval_wrapper| returns. */
int getauxval_wrapper(unsigned long type, unsigned long *result) {
  if (getauxval == NULL) {
    return 0;
  }
  *result = getauxval(type);
}
```

briansmith:
> What's the right way to do this (where "this" is "make it testable on x86 linux, but in a production build only have this stuff present on ARM")?

IMO, testing on x86 and x86-64 is a "nice to have" and IMO it's probably not worthwhile to do, once we've verified the code works on ARM and ARM64, because of all the complication it adds. I'd rather we find a way to make it easier to test this on ARM (e.g. make it easy to set up Android dev environments, perhaps via Docker) than to make the code even a little more complex or confusing to facilitate non-ARM testing.

briansmith:
Also, regarding licensing:

* Please add the contributor statement to the end of your commit message: https://github.com/briansmith/ring#contributing.

* If you add significant amounts of code to a file, feel free to add your own copyright line.

* Please state who the copyright holder of these changes is: Is it you, or some employer, or somebody else. Thanks!

src/cpu_feature/arm_linux/arm_linux.rs
```diff
@@ -0,0 +1,176 @@
+use self::auxv::AuxvUnsignedLong;
```
marshallpierce: Done

marshallpierce:
Do you want the contributor statement in every commit, or just one, or only the last or first one, or...? 

I'm the copyright holder; all this is done on my own time.

Also, please don't feel bad about telling me you don't like my approach on something. I'm here to learn how to build a crypto library, not to convince you that my way of building for testability or whatever is gospel. Obviously I choose an approach that seems best to me, and I wipe away a tiny tear when I delete code that enables testability or portability because those things are dear to me, but I'm perfectly happy to admit that my background is *not* in building low level crypto stuff, so my heuristics for what is worth the lines of code and what is not may well be not well-tuned for this arena.  Anyway, I appreciate your thoughtful reviews. You've clearly been doing this sort of code longer than I have, so I'm happy to incorporate your knowledge there into my personal stash.

crypto/cpu-arm-linux.c
```diff
@@ -14,345 +14,34 @@
...
     }
 
-    if (hwcap2 & HWCAP2_AES) {
-      GFp_armcap_P |= ARMV8_AES;
-    }
-    if (hwcap2 & HWCAP2_PMULL) {
-      GFp_armcap_P |= ARMV8_PMULL;
+    unsigned long auxval = getauxval(type);
+    // map errors to a zero value
+    if (errno != 0) {
```
briansmith: Given the above comment, this check of `errno` should just be removed. However, if we weren't going to remove it, then we'd need to change the check to `(auxval == 0 && errno != 0)`, as described in https://www.securecoding.cert.org/confluence/pages/viewpage.action?pageId=6619179#ERR10-C.Seterrnotozerobeforecallingalibraryfunctionknowntoseterrno,andcheckerrnoonlyafterthefunctionreturnsavalueindicatingfailure-LibraryFunctionsthatSeterrnoandReturnanIn-BandErrorIndicator.

src/cpu_feature/arm_linux/arm_linux.rs
```diff
@@ -0,0 +1,144 @@
...
+        let getauxv = StubGetauxval {
+            auxv: auxv
+        };
+
+        do_armcap_bits_test(getauxv, ARMV7_NEON | ARMV8_AES | ARMV8_SHA1 | ARMV8_SHA256 | ARMV8_PMULL);
+    }
+
+
+    #[test]
+    fn armcap_bits_arm_8_with_neon_only_getauxv_hwcap_and_aes_getauxv_hwcap2_yields_only_neon_aes_armcap() {
```
briansmith: Let's choose function names that are short enough that the `fn <name>() {` fits in 80 characters.

src/init.rs
```diff
@@ -12,13 +12,52 @@
...
+}
+
+extern "C" {
+    #[cfg(all(any(target_arch = "arm", target_arch = "aarch64"),
+        target_os="linux"))]
+    #[allow(non_upper_case_globals)]
+    pub static mut GFp_armcap_P: u32;
+}
+
+// On other platforms (excluding iOS aarch64), use C feature detection
```
briansmith: Use a period at the end of the sentence.

src/init.rs
```diff
@@ -12,13 +12,52 @@
...
     {
         extern crate std;
-        extern { fn GFp_cpuid_setup(); }
         static INIT: std::sync::Once = std::sync::ONCE_INIT;
-        INIT.call_once(|| unsafe { GFp_cpuid_setup() });
+        INIT.call_once(|| set_cpu_features() );
     }
 }
+
+// On arm linux, use pure rust
```
briansmith: Capitalize "ARM" and "Rust" and end with a period.

src/init.rs
```diff
@@ -12,13 +12,52 @@
...
+    target_os = "linux"))]
+fn set_cpu_features() {
+    let getauxval = auxv::NativeProvider {};
+
+    unsafe {
+        GFp_armcap_P |= arm_linux::armcap_from_features(getauxval);
+    }
+}
+
+extern "C" {
```
briansmith: We don't write "C" in *ring*, so just remove it.

src/init.rs
```diff
@@ -12,13 +12,52 @@
...
+
+// On arm linux, use pure rust
+
+#[cfg(all(any(target_arch = "arm", target_arch = "aarch64"),
+    target_os = "linux"))]
+fn set_cpu_features() {
+    let getauxval = auxv::NativeProvider {};
+
+    unsafe {
+        GFp_armcap_P |= arm_linux::armcap_from_features(getauxval);
```
briansmith: Shouldn't this be `=` instead of `|=`?

src/cpu_feature/arm_linux/auxv.rs
```diff
@@ -0,0 +1,40 @@
...
+        unsafe {
+            return getauxval_wrapper(auxv_type);
+        }
+    }
+}
+
+// from [linux]/include/uapi/linux/auxvec.h. First 32 bits of HWCAP
+// even on platforms where unsigned long is 64 bits.
+pub const AT_HWCAP: Type = 16;
+// currently only used by powerpc and arm64 AFAICT
```
briansmith: `AT_HWCAP2` is used for 32-bit ARM too, in particular for deteecting crypto instructions on ARMv8 CPUs. For example, consider a 32-bit Android app running on Aaarc64 Android, as is common.

src/cpu_feature/arm_linux/arm_linux.rs
```diff
@@ -0,0 +1,144 @@
...
+            auxv: auxv
+        };
+
+        do_armcap_bits_test(getauxv, ARMV7_NEON | ARMV8_AES);
+    }
+
+    #[test]
+    fn armcap_for_hwcap2_all_hwcap2_returns_all_armcap() {
+        assert_eq!(ARMV8_AES | ARMV8_PMULL | ARMV8_SHA1 | ARMV8_SHA256,
+            super::armcap_for_hwcap2(ARM_HWCAP2_AES
```
briansmith: NIT: The `|` should be the last character of the preceding line, instead of the first character of the next line.

src/cpu_feature/arm_linux/arm_linux.rs
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+
+    #[test]
+    fn armcap_bits_ok_neon_with_neon_getauxv_yields_neon_armcap() {
+        do_armcap_bits_test(hwcap_neon_getauxv(), ARMV7_NEON);
+    }
+
+    #[test]
+    fn armcap_bits_arm_8_with_neon_only_getauxv_hwcap_and_all_getauxv_hwcap2_yields_all_features_armcap() {
```
briansmith: Let's choose function names that are short enough that the `fn <name>() {` fits in 80 characters.

src/cpu_feature/arm_linux/arm_linux.rs
```diff
@@ -0,0 +1,144 @@
...
+        ret |= ARMV8_PMULL;
+    }
+    if hwcap2 & ARM_HWCAP2_SHA1 != 0 {
+        ret |= ARMV8_SHA1;
+    }
+    if hwcap2 & ARM_HWCAP2_SHA2 != 0 {
+        ret |= ARMV8_SHA256;
+    }
+
+    return ret;
```
briansmith: s/return ret;/ret/

src/cpu_feature/arm_linux/arm_linux.rs
```diff
@@ -0,0 +1,144 @@
...
+    let mut armcap: u32 = 0;
+    if hwcap & ARM_HWCAP_NEON != 0 {
+        armcap |= ARMV7_NEON;
+
+        let hwcap2 = getauxval.getauxval(auxv::AT_HWCAP2);
+
+        armcap |= armcap_for_hwcap2(hwcap2);
+    }
+
+    return armcap;
```
briansmith: s/return armcap;/armcap/

src/cpu_feature/arm_linux/auxv.rs
```diff
@@ -0,0 +1,40 @@
...
+impl Provider for NativeProvider {
+    fn getauxval(&self, auxv_type: Type) -> Type {
+        unsafe {
+            return getauxval_wrapper(auxv_type);
+        }
+    }
+}
+
+// from [linux]/include/uapi/linux/auxvec.h. First 32 bits of HWCAP
+// even on platforms where unsigned long is 64 bits.
```
marshallpierce: I can't seem to find the mailing list post where I saw this, but the gist is that to prevent people using the high 32 bits on a 64 bit system to express features that would then be hidden from 32 bit folks, they decided to only ever use the bottom 32.

src/init.rs
marshallpierce: The original C used `|=` so I stuck with it, but I don't know of any other code that writes to that global.

crypto/cpu-arm-linux.c
```diff
@@ -14,345 +14,34 @@
...
+        return 0;
     }
 
-    if (hwcap2 & HWCAP2_AES) {
-      GFp_armcap_P |= ARMV8_AES;
-    }
-    if (hwcap2 & HWCAP2_PMULL) {
-      GFp_armcap_P |= ARMV8_PMULL;
+    unsigned long auxval = getauxval(type);
+    // map errors to a zero value
```
marshallpierce: Apparently stopping writing Android code years ago wasn't enough to leave these problems behind me. ;)

src/c.rs
```diff
@@ -41,17 +41,6 @@ macro_rules! define_metrics_tests {
       $expected_align_factor:expr ) =>
     {
         #[cfg(test)]
-        extern {
```
marshallpierce: Leaving this in didn't work when adding `unsigned_long` because it uses the same size and alignment C globals, which then would get redefined.

src/c.rs
```diff
@@ -80,17 +69,57 @@ macro_rules! define_metrics_tests {
...
+    static GFp_uint32_t_size: u16;
+    static GFp_uint32_t_align: u16;
+    static GFp_int16_t_size: u16;
+    static GFp_int16_t_align: u16;
+    static GFp_uint16_t_size: u16;
+    static GFp_uint16_t_align: u16;
+    static GFp_int8_t_size: u16;
+    static GFp_int8_t_align: u16;
+    static GFp_uint8_t_size: u16;
+    static GFp_uint8_t_align: u16;
```
marshallpierce: So instead, I define them all beforehand. Maybe there are cleverer ways to conditionally define stuff in a macro?

src/c.rs
briansmith: Sorry, I'm not sure I understand. Did you try defining `GFp_unsigned_long_align` and `GFp_unsigned_long_size` in crypto/crypto.c and using them?

src/c.rs
```diff
@@ -80,17 +69,57 @@ macro_rules! define_metrics_tests {
...
 define_type!(int, i32, test_int_metrics, GFp_int_align, GFp_int_size,
              "The C `int` type. Equivalent to `libc::c_int`.");
 
 #[cfg(any(target_os = "windows", target_pointer_width = "32"))]
 define_type!(long, i32, test_long_metrics, GFp_long_align, GFp_long_size,
              "The C `long` type. Equivalent to `libc::c_long`.");
 
+#[cfg(any(target_os = "windows", target_pointer_width = "32"))]
+define_type!(unsigned_long, u32, test_unsigned_long_metrics,
+    GFp_long_align, GFp_long_size,
```
briansmith: This should be `GFp_unsigned_long_align`, `GFp_unsigned_long_size`.

crypto/crypto.c
```diff
@@ -82,23 +82,24 @@ uint32_t GFp_armcap_P = 0;
 /* These allow tests in other languages to verify that their understanding of
  * the C types matches the C compiler's understanding. */
 
-#define DEFINE_METRICS(ty) \
-  OPENSSL_EXPORT uint16_t GFp_##ty##_align = alignof(ty); \
-  OPENSSL_EXPORT uint16_t GFp_##ty##_size = sizeof(ty);
+#define DEFINE_METRICS(ty, ty_ident) \
+  OPENSSL_EXPORT uint16_t GFp_##ty_ident##_align = alignof(ty); \
+  OPENSSL_EXPORT uint16_t GFp_##ty_ident##_size = sizeof(ty);
```
marshallpierce: Needed a second param to have an identifier-safe version of the type.

src/c.rs
marshallpierce: Ah, gotcha. It didn't even occur to me to simply add more constants in crypto.c... Doh!

marshallpierce:
I'm trying to get an arm linux install set up via qemu so I can try out arm builds locally (rather than waiting for Travis). it's taking hours just to install, but anything is better than Travis's long queues...

briansmith:
> I'm trying to get an arm linux install set up via qemu so I can try out arm builds locally (rather than waiting for Travis). it's taking hours just to install, but anything is better than Travis's long queues...

One tip: In .travis.yml in your branch, delete all the configs for the targets you don't care about. Then Travis will only run the targets you do care about. And/or, set up your own travis account for your fork, so that your builds don't compete for resources with other PRs on this repo.

briansmith:
Some actual hardware to test this is on the way, and also I'm digging into the Android code as well.

marshallpierce:
Looks like the Android jobs time out and the aarch64 ones fail aes tests: https://travis-ci.org/marshallpierce/ring/builds/193520619. The last merged PR (my .gitignore lines) had one android job time out, but the aarch64 ones passed: https://travis-ci.org/briansmith/ring/builds/189857834

Is it possible that the arm hardware used is the broken hardware listed in #357?

briansmith:
Good news: I have some hardware that can be used to test this now. Hoping to get to the review soon.

I'm not sure what's up with that AAarch64 build. I will look into it. I doubt it's this PRs fault.

briansmith:
Sorry that so much time has passed on this, which seems to have gotten stuck on us not being able to debug some issue with failing tests a long, long time ago.

Now I have some time to work on this again, but a lot has changed. In particular, API level 28 is now the minimum for the Google Play store, so there's no reason to worry about whether `getauxval()` is missing, so we don't need to deal with weak symbol stuff. Also we don't need to worry, AFAICT, about `getauxval()` returning incomplete results anymore. I already wrote #748 to remove much of that.

I think then, on ARM Linux, we can use `libc::getauxval()` and avoid a lot of the complexity in this PR. Sorry that wasn't the case when you originally started this.

@marshallpierce Would you be interested in updating the PR?

marshallpierce:
Hypothetically, but I'm pretty maxed out right now with other projects. Give it another year and we'll see ;)

briansmith:
> Give it another year and we'll see ;)

:) I wish I could have predicted that time would make most of the complexity here irrelevant. I have a private branch that does this, which I will submit publicly soon, so I'm going to close this PR.

Thanks for the effort you put into this!

Issue #371, opened 2016-12-09
Remove shift-and-overwrite-prefix logic from `ring::aead::open_in_place`


briansmith:
`ring::aead::open_in_place` has a feature that lets one specify that the output should be in the same place as the input, except shifted forward `in_prefix_len` bytes. The cost of the feature is high because it is difficult to explain to people and it makes the API error-prone. That feature is useful for certain use cases, but those use cases can be accommodated in other ways. The marginal utility of the feature isn't sufficient for its cost, so we should just remove it.

briansmith:
PR #433 implements the removal of this feature. However, I'm experimenting with another approach that may let the feature remain.

briansmith:
Probably we can't remove this feature at least until we implement out-of-place opening (#436), since [Rustls](https://github.com/ctz/rustls) is currently using the shifting feature. Also, like I mentioned in the previous comment, I think we can improve the API in a way that lets the feature remain for those that want to use it, but which doesn't get in the way for people who don't benefit from it.

briansmith:
I'm going to close this. We have another issue/PR open for improving the API in general, and I don't think it's a net win to remove the feature (yet). We'll revisit naturally in the course of whatever improvements we make in the future.

Issue #291, opened 2016-08-28
aead::seal_in_place only accepts out_suffix_capacity exactly equal to max overhead


mcginty:
Tested against ring 0.3.1 from crates.io.

The `seal_in_place` documentation says "out_suffix_capacity must be at least `key.algorithm.max_overhead_len()`", however it looks like no value works except for exactly `key.algorithm().max_overhead_len()`. I'd prefer if the documentation's version were true since it would mean fewer allocations for me! Also having out_suffix_capacity as an argument isn't so useful if it can only be one thing ;).

**works**:

``` rust
let data = [0u8; 1350];
let out_suffix_capacity = ALGORITHM.max_overhead_len();
let mut in_out = vec![0u8; data.len() + out_suffix_capacity];
aead::seal_in_place(&key, &nonce_bytes,
                    &mut in_out, out_suffix_capacity,
                    &[]).unwrap();
```

**breaks**:

``` rust
let data = [0u8; 1350];
let out_suffix_capacity = ALGORITHM.max_overhead_len() + 1;
let mut in_out = vec![0u8; data.len() + out_suffix_capacity];
aead::seal_in_place(&key, &nonce_bytes,
                    &mut in_out, out_suffix_capacity,
                    &[]).unwrap();
```

briansmith:
I'm guessing the problem is that `slice_as_array_ref_mut!()` fails.

If you can build from the Git version, try this patch:

```
diff --git a/src/aead/aead.rs b/src/aead/aead.rs
index 22e914c..ad83625 100644
--- a/src/aead/aead.rs
+++ b/src/aead/aead.rs
@@ -178,7 +178,8 @@ pub fn seal_in_place(key: &SealingKey, nonce: &[u8], in_out: &mut [u8],
                          .ok_or(error::Unspecified));
     try!(check_per_nonce_max_bytes(in_out_len));
     let (in_out, tag_out) = in_out.split_at_mut(in_out_len);
-    let tag_out = try!(slice_as_array_ref_mut!(tag_out, TAG_LEN));
+    let tag_out =
+        try!(slice_as_array_ref_mut!(&mut tag_out[..TAG_LEN], TAG_LEN));
     try!((key.key.algorithm.seal)(&key.key.ctx_buf, nonce, in_out, tag_out, ad));
     Ok(in_out_len + TAG_LEN)
 }
```

briansmith:
Thanks for the bug report. I agree that things look wrong.

[...]

PR #421, opened 2017-01-24
Improve the usability of `ring::aead`.

briansmith:
This attempts to improve the API of `ring::aead` by at least a marginal amount by fixing #371 (Removing the `in_prefix_len` parameter from `open_in_place()`) and #291 (fixing the processing of `out_suffix_capacity` in seal_in_place()`), and by returning slices instead of lengths from these functions.

Feedback is much appreciated. I'd like to improve this API more, but we're constrained because these APIs must (IMO): 1. be in-place, 2. be non-allocating, and 3. enforce that the tag is serialized immediately after the ciphertext.

Note that I plan to offer a not-in-place API in addition to this API, which maybe will use `std::io::Write` or similar, which I suspect will be less error-prone.

/cc @Philipp91 @mcginty @frewsxcv

### DIFF
--- a/src/aead/aead.rs
+++ b/src/aead/aead.rs
@@ -119,8 +119,8 @@ impl OpeningKey {
 /// C analog: `EVP_AEAD_CTX_open`
 ///
 /// Go analog: [`AEAD.Open`](https://golang.org/pkg/crypto/cipher/#AEAD)
-pub fn open_in_place<'a>(key: &OpeningKey, nonce: &[u8], in_prefix_len: usize,
-                         in_out: &'a mut [u8], ad: &[u8])
+pub fn open_in_place<'a>(key: &OpeningKey, nonce: &[u8], ad: &[u8],
+                         in_prefix_len: usize, in_out: &'a mut [u8])
                          -> Result<&'a mut [u8], error::Unspecified> {
     let nonce = try!(slice_as_array_ref!(nonce, NONCE_LEN));
     let ciphertext_and_tag_len =
@@ -207,8 +207,8 @@ impl SealingKey {
 /// C analog: `EVP_AEAD_CTX_seal`.
 ///
 /// Go analog: [`AEAD.Seal`](https://golang.org/pkg/crypto/cipher/#AEAD)
-pub fn seal_in_place(key: &SealingKey, nonce: &[u8], in_out: &mut [u8],
-                     out_suffix_capacity: usize, ad: &[u8])
+pub fn seal_in_place(key: &SealingKey, nonce: &[u8], ad: &[u8],
+                     in_out: &mut [u8], out_suffix_capacity: usize)
                      -> Result<usize, error::Unspecified> {
     if out_suffix_capacity < key.key.algorithm.tag_len() {
         return Err(error::Unspecified);
@@ -270,8 +270,8 @@ pub struct Algorithm {
     init: fn(ctx_buf: &mut [u8], key: &[u8]) -> Result<(), error::Unspecified>,
 
     seal: fn(ctx: &[u64; KEY_CTX_BUF_ELEMS], nonce: &[u8; NONCE_LEN],
-              in_out: &mut [u8], tag_out: &mut [u8; TAG_LEN], ad: &[u8])
-              -> Result<(), error::Unspecified>,
+             in_out: &mut [u8], tag_out: &mut [u8; TAG_LEN], ad: &[u8])
+             -> Result<(), error::Unspecified>,
     open: fn(ctx: &[u64; KEY_CTX_BUF_ELEMS], nonce: &[u8; NONCE_LEN],
              in_out: &mut [u8], in_prefix_len: usize,
              tag_out: &mut [u8; TAG_LEN], ad: &[u8])
@@ -355,8 +355,8 @@ mod tests {
                 s_in_out.push(0);
             }
             let s_key = try!(aead::SealingKey::new(aead_alg, &key_bytes[..]));
-            let s_result = aead::seal_in_place(&s_key, &nonce[..],
-                                               &mut s_in_out[..], tag_len, &ad);
+            let s_result = aead::seal_in_place(&s_key, &nonce[..], &ad,
+                                               &mut s_in_out[..], tag_len);
             let o_key = try!(aead::OpeningKey::new(aead_alg, &key_bytes[..]));
 
             ct.extend(tag);
@@ -447,9 +447,9 @@ mod tests {
                     o_in_out.push(123);
                 }
                 o_in_out.extend_from_slice(&ct[..]);
-                let o_result = aead::open_in_place(&o_key, &nonce[..],
+                let o_result = aead::open_in_place(&o_key, &nonce[..], &ad,
                                                    *in_prefix_len,
-                                                   &mut o_in_out[..], &ad);
+                                                   &mut o_in_out[..]);
                 match error {
                     None => {
                         assert_eq!(Ok(ct.len()), s_result);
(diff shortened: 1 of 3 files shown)


src/aead/aead.rs
```diff
@@ -76,37 +76,24 @@ impl OpeningKey {
...
 ///
 /// C analog: `EVP_AEAD_CTX_open`
 ///
 /// Go analog: [`AEAD.Open`](https://golang.org/pkg/crypto/cipher/#AEAD)
-pub fn open_in_place(key: &OpeningKey, nonce: &[u8], in_prefix_len: usize,
-                     in_out: &mut [u8], ad: &[u8])
-                     -> Result<usize, error::Unspecified> {
+pub fn open_in_place<'a>(key: &OpeningKey, nonce: &[u8], in_out: &'a mut [u8],
+                         ad: &[u8])
+                         -> Result<&'a mut [u8], error::Unspecified> {
```
Philipp91: That's a neat idea! In-place, but the return value is exactly the thing you need.

At first I was slightly confused because the returned slice is mutable (am I supposed to write something into that return value?), but I guess this can come in handy in some use cases.

src/aead/aead.rs
```diff
@@ -163,36 +151,40 @@ impl SealingKey {
...
-    let tag_out = try!(slice_as_array_ref_mut!(tag_out, TAG_LEN));
-    try!((key.key.algorithm.seal)(&key.key.ctx_buf, nonce, in_out, tag_out, ad));
-    Ok(in_out_len + TAG_LEN)
+    let ciphertext_and_tag_len = in_out_len + TAG_LEN;
+    if ciphertext_and_tag_len > in_out.len() {
+        return Err(error::Unspecified);
+    }
+    {
+        let (in_out, tag_out) = in_out.split_at_mut(in_out_len);
+        if tag_out.len() < TAG_LEN {
```
Philipp91: What's that empty `if`?

frewsxcv:
I like these changes. 👍  This API looks simpler and more idiomatic than getting back a length and doing the slice indexing oneself.

briansmith:
I've refactored the API again, to make it even simpler and-hopefuly-more intuitive. In particular, `seal_in_place()` now takes a `plaintext_in_ciphertext: &mut [u8]` slice and a `tag_out: &mut [u8]` slice. This way, we eliminate the complexity of `out_suffix_capacity`. Also, the `ad` parameter was moved ahead of the ciphertext/plaintext parameter. Now the order is `ad`, then `plaintext/ciphertext`, then `tag_out`, which matches the order that the AAD, ciphertext, and tag appear on the wire (almost always).

Please check out the new code and let me know what you think.

src/aead/aead.rs
```diff
@@ -76,37 +76,24 @@ impl OpeningKey {
...
+/// `open_in_place` returns `Ok(plaintext)`, the decrypted output is
+/// `plaintext`, which is a slice of `in_out`.
 ///
 /// C analog: `EVP_AEAD_CTX_open`
 ///
 /// Go analog: [`AEAD.Open`](https://golang.org/pkg/crypto/cipher/#AEAD)
-pub fn open_in_place(key: &OpeningKey, nonce: &[u8], in_prefix_len: usize,
-                     in_out: &mut [u8], ad: &[u8])
-                     -> Result<usize, error::Unspecified> {
+pub fn open_in_place<'a>(key: &OpeningKey, nonce: &[u8], ad: &[u8],
```
mcginty: Is there a reason not to also break out the `tag` into its own argument here instead of requiring it be a suffix in `in_out`? Seems kinda asymmetric in its current state.

src/aead/aead.rs
```diff
@@ -76,37 +76,24 @@ impl OpeningKey {
...
+                         in_out: &'a mut [u8])
```
mcginty: On the same note, having `in_out` here and `plaintext_in_ciphertext_out` on `seal_in_place` is a bit confusing. Maybe have them both be just `buffer` or make this one the more descriptive `ciphertext_in_plaintext_out`?

I think the naming and arguments should be as symmetric as possible to be obvious about which outputs from one functions map to the inputs of the other.

Philipp91:
I like the direction.

Why does `seal_in_place` have `tag_out`, but `open_in_place` has no `tag_in`?
And why is `open_in_place`'s data parameter still called `in_out` rather than `ciphertext_in_plaintext_out`?
(Maybe I'm not seeing the right diff here.)

ctz:
> Note that I plan to offer a not-in-place API in addition to this API, which maybe will use std::io::Write or similar

rustls could use this. My input slice belongs to the caller and so I need to copy the whole thing to do an in-place encrypt/decrypt.

mcginty:
@ctz, @briansmith notes in the description that there are plans to add a not-in-place seal/open API as well, it's just out of the scope of this PR.

src/aead/aead.rs
```diff
@@ -240,13 +212,12 @@ impl Key {
...
 
-    seal: fn(ctx: &[u64; KEY_CTX_BUF_ELEMS], nonce: &[u8; NONCE_LEN],
-              in_out: &mut [u8], tag_out: &mut [u8; TAG_LEN], ad: &[u8])
-              -> Result<(), error::Unspecified>,
-    open: fn(ctx: &[u64; KEY_CTX_BUF_ELEMS], nonce: &[u8; NONCE_LEN],
-             in_out: &mut [u8], in_prefix_len: usize,
-             tag_out: &mut [u8; TAG_LEN], ad: &[u8])
-             -> Result<(), error::Unspecified>,
+    seal: fn(ctx: &[u64; KEY_CTX_BUF_ELEMS], nonce: &[u8; NONCE_LEN], ad: &[u8],
+              plaintext_in_ciphertext_out: &mut [u8],
```
frewsxcv: Nit: indentation is off here

Philipp91:
> My input slice belongs to the caller and so I need to copy the whole thing to do an in-place encrypt/decrypt.

My analysis found that 38% of all users of Rust crypto APIs are in the same position, that is, they don't have `&mut` access to the input data and someone has to copy it (and thus allocate another O(n) memory) at some point anyway. So a not-in-place API doesn't harm their performance. And an API with a `Vec<u8>` return value is more readable, plus it's hard for beginners to allocate a vector (!).

So a not-in-place API is definitely a good idea, whether it's in this PR or not. After the changes in this PR, I'm wondering (ahead of time) how the return value would work if it contains the ciphertext and the tag separately (maybe `-> (Vec<u8>, Vec<u8>)` ?). My impression (and maybe I can provide quantitative data on that, too) is that users don't need ciphertext and tag separately. Yes, it makes the `in_place` API easier, because it eliminates the suffix overhead, so it's good there as done in this PR. But for the not-in-place API, which allocates a return value for the caller anyway, it might as well include the tag in a single `Vec<u8>`.

This is with the idea in mind that the not-in-place API is at a higher level and doesn't have to suit absolutely everyone's use case. Maybe it's worth making this API even a bit more high-level by removing the `nonce` parameter, generating a nonce inside and also concatenating that with the ciphertext. Many beginners struggle to use proper random RNGs and if most callers have to write the same code to generate a nonce (or is there an easier way?):
```
use std::iter::repeat;
let mut nonce: Vec<u8> = repeat(0).take(aead::CHACHA20_POLY1305.nonce_len()).collect();
SystemRandom::new().fill(&mut nonce).unwrap();
```
then maybe the library should just do it for them.

Even less related to this PR: How about an `aead::CHACHA20_POLY1305.gen_nonce() -> Vec<u8>` function? Let's you omit the above three lines, generate the nonce in-line and prevents beginners from using insecure randomness.

briansmith:
I think this change paves the way for out-of-place `seal()` and `open()` so a PR adding that should be relatively easy for anybody to submit.

I don't recommend using random nonces with these AEADs (regardless of what crypto library you are using). We'll add nonce-misuse-resistant AEADs later that should be safer to use with random nonces. That said, nonce handling will get improved later so that people don't have to manage the nonces themselves.

src/aead/aead.rs
briansmith: > Is there a reason not to also break out the tag into its own argument here instead of requiring it be a suffix in in_out? Seems kinda asymmetric in its current state.

It is definitely asymmetric. But, I think that opening/sealing is conceptually asymmetric.

When sealing, you know how much plaintext you have, and you need to know that + the tag length to determine how much data to send.

When opening, you're almost always given the length of the length of `ciphertext + tag`, and you need to get the plaintext. You actually don't need to know how long the tag is, if you are given the plaintext length (or better, slice) as output.

Put another way, if we make `open_in_place()` symmetric with `seal_in_place()`, then it would actually require the caller to do more work, and that work would be error-prone. It would be more error-prone because the caller would need to split `ciphertext + tag`, being careful to handle the case where the input is shorter than the required tag length. Whereas, the way I've done it now seems less error-prone because this splitting is done internally.

OTOH, I do agree that the asymmetry is unfortunate. Maybe there's another solution I'm overlooking? Or maybe the benefit i described in this comment don't outweigh the pedagogical harm from having an asymmetric API? I do think some of the pedagogical harm of the asymmetry can be reduced with better documentation and/or better naming of the `in_out` parameter.

src/aead/aead.rs
briansmith: >  or make this one the more descriptive ciphertext_in_plaintext_out?

It would actually be `ciphertext_and_tag_in_plaintext_and_tag_out`. Totally doable.

src/aead/aead.rs
Philipp91: > because the caller would need to split `ciphertext + tag`

Why split? When they received it from the `seal_in_place()` function, it was already separate, wasn't it?

I think I can follow your argument, on the assumption that the caller still stores the `ciphertext + tag` in a single place, even though the API has now changed. All (4?) current users of ring's AEAD API would probably do that.
For new users, some might still do that. After all, ciphertext and tag somehow belong together.

But other users could find simpler ways to keep ciphertext and tag separate. Like a `struct Message { sender, time, ..., ciphertext: Vec<u8>, tag: Vec<u8>, ... }`, which they can even serialize to JSON or XML and read with their AEAD implementation in Javascript or elsewhere. For these users, it would be weird if they had to concatenate the two.
Then again, these users prefer the out-of-place API, because they wouldn't want their `message.ciphertext` field to be overwritten. Maybe my example isn't the best, but I hope you see my point.

src/aead/aead.rs
Philipp91: > It would actually be `ciphertext_and_tag_in_plaintext_and_tag_out`. Totally doable.

Do you mean `ciphertext_and_tag_in_plaintext_out` (no tag out)? The `ciphertext_in_plaintext_out` would be correct assuming that the parameter is split, as is being discussed in the comments above (line 86).

src/aead/aead.rs
briansmith: > Why split? When they received it from the seal_in_place() function, it was already separate, wasn't it?

In TLS and similar protocols, you get the length of `ciphertext+tag` as an input, and then you have to split them.

I see that JWE (JOSE in general, it seems) and probably XML security stuff doesn't work that way. See for example, https://tools.ietf.org/html/rfc7516#section-3.1 and https://tools.ietf.org/html/rfc7516#section-3.2.

> But other users could find simpler ways to keep ciphertext and tag separate. Like a struct Message { sender, time, ..., ciphertext: Vec<u8>, tag: Vec<u8>, ... }, which they can even serialize to JSON or XML and read with their AEAD implementation in Javascript or elsewhere. For these users, it would be weird if they had to concatenate the two.

That's what I'd like to discourage, because such applications have to keep two items (ciphertext and tag) in sync, instead of one item (ciphertext and tag). JOSE and WS-Security are extraordinarily terrible examples to follow in this respect.

So I guess we need to decide about whether we want to make it simpler to implement bad protocols (JOSE and WS-Security), or make it simpler to design protocols correctly.

src/aead/aead.rs
briansmith: > Do you mean ciphertext_and_tag_in_plaintext_out (no tag out)? The ciphertext_in_plaintext_out would be correct assuming that the parameter is split, as is being discussed in the comments above (line 86).

I'm not sure I understand what you mean. On input the parameter contains the ciphertext followed by the tag. The function replaces the ciphertext with the plaintext, so on output it would contain the plaintext followed by the tag. The slice in the return value is just the plaintext part. The user of the function should use the returned slice instead of using `ciphertext_and_tag_in_plaintext_and_tag_out` to extract the plaintext.

Philipp91: Yes, you're right, sorry. It's confusing, but thankfully the out-of-place API won't have such intricacies.

briansmith:
I rethought this PR and decided to scale back the changes to the ones that are uncontroversial. Thus, the API is basically the same except for some renaming and reordering of parameters, and except for the fact that `open_in_place()` returns a slice holding the plaintext instead of the plaintext length, and the expanded documentation for `open_in_place()`. In particular, the new documentation explains how to use `in_prefix_len` for its intended purpose.

#433 is a PR for removing `in_prefix_len`. #434 is a PR for replacing the `out_suffix_capacity` parameter of `seal_in_place()`.

briansmith:
> #433 is a PR for removing in_prefix_len.

Now it is #435, which is based on #434. (#433 is canceled.)

briansmith:
>  #434 is a PR for replacing the out_suffix_capacity

Sorry, I meant tha #434 is now #435, and #433 is now based on #435.

briansmith:
I have another proposal for reducing the chance of misuse of the current API, as well as supporting out-of-place operation, in #436.

PR #424, opened 2017-01-24
fix C4819 error in specific Windows build environment

itn3000:


crypto/libring.Windows.vcxproj
```diff
@@ -60,6 +60,7 @@
     <ClCompile>
       <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories);..\include</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <AdditionalOptions>/source-charset:utf-8</AdditionalOptions>
```
briansmith: Thanks! Should we use ` /utf-8` instead? In particular, it seems like we want to set the execution charset to UTF-8 to, if I'm understanding the documentation correctly. (The execution charset probably doesn't matter.)

briansmith:
Thanks for the PR! I fixed this a slightly different way in a67775a252a220578d642be175ee837ed1e3702e:
* I used `/utf-8` to set both the source charset and the execution charset.
* I moved the setting to Common.props so that it would be used for all C and C++ code.

Please let me know if I messed it up somehow!

itn3000:
It's OK because `/utf-8` is default behavior of cl.exe when input text is utf-8 with bom.
Thank you for fix.

PR #425, opened 2017-01-24
Allow static detection of NEON on non-Android ARMv7 targets.

briansmith:
This should allow armv7-apple-ios targets to use NEON without any
runtime checks.


briansmith:
Updated to remove the patch file and then landed as be172b196c7d52051e3ad68011ba2e7f60284622.

briansmith:
Follow-up to fix Android bustage: 7e648f43f821b72c110b0b00af7627fe665ca073

PR #429, opened 2017-01-26
external files in test support on iOS

kali:

### DIFF
--- a/src/bssl.rs
+++ b/src/bssl.rs
@@ -38,6 +38,7 @@ macro_rules! bssl_test {
             }
 
             init::init_once();
+            ::std::env::set_current_dir(::test::ring_src_path()).unwrap();
 
             let result = unsafe {
                 $bssl_test_main_fn_name()
@@ -61,6 +62,7 @@ macro_rules! bssl_test_rng {
             }
 
             init::init_once();
+            ::std::env::set_current_dir(::test::ring_src_path()).unwrap();
 
             let rng = rand::SystemRandom::new();
             let mut rng = rand::RAND { rng: &rng };
--- a/src/test.rs
+++ b/src/test.rs
@@ -217,6 +217,23 @@ impl TestCase {
     }
 }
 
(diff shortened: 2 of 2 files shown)


src/bssl.rs
```diff
@@ -38,6 +38,7 @@ macro_rules! bssl_test {
             }
 
             init::init_once();
+            ::std::env::set_current_dir(::test::ring_src_path()).unwrap();
```
briansmith: I'm not quite sure why this is needed on iOS but not Android. Here's how we do it on Android; can we do the same thing on iOS?
https://github.com/briansmith/ring/blob/c84e4118f8a956b306aa7c8a3eafa23159f05426/mk/travis.sh#L82-L90

Also, why is it important to switch the current directory, instead of just changing file accesses to use the full path?

kali: On a non-jailbroken iOS, the only (known) way to start a process is to make it an app, and lauching it as an app. It will only be started with the working directory set to `/`, I could not find a way around that.

So we have two ways to solve this: using absolute paths (as I did for the pure Rust tests) or cd-ing to the right place just before running tests and keep relative paths. On principle, the first approach feels better as you're not altering what is basically a process-wide global variable. I used the cd strategy for the bn tests because I felt the required changes would be too invasive (as we need to pass the path through the rust-to-c test machinery), but if you feel otherwise, I can give it a shot.

briansmith:
> I used the cd strategy for the bn tests because I felt the required changes would be too invasive (as we need to pass the path through the rust-to-c test machinery), but if you feel otherwise, I can give it a shot.

OK, I understand now. The strategy you chose makes sense. Note that we're goign to get rid of bn_test.cc soon so eventually this won't be a problem. We don't need a perfect solution for the interim state, just one that works.

briansmith:
Thanks. This landed as in master in e4443e8541d30201dd2d8b000645ba1e2fa83173. I changed the commit message to "Fix file-based tests on iOS."

PR #430, opened 2017-01-26
Fixing a compilation problem in current nightly

P-E-Meunier:
This problem completely forbids the use of ring in nightly.
It would be really awesome if this were updated soon on crates.io, because:

- I'm using ring + rustls + thrussh to write a webserver (soon to be merged with the new hyper).
- `impl trait` is kind of mandatory to write anything efficient with tokio+futures, and is available only on nightly.

I couldn't find the commit in the rust repository that introduced it.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -220,11 +220,11 @@ fn build_msvc(target_triple: &[&str], disable_opt: bool, num_jobs: &str,
     run_command_with_args(&msbuild, &test_args);
 }
 
-fn run_command_with_args<S>(command_name: S, args: &Vec<String>)
+fn run_command_with_args<S>(command_name: S, args: &[String])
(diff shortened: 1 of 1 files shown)


P-E-Meunier:
It actually fails with previous versions of rust.
Nevermind, sorry about that.

briansmith:
Thanks. Landed in 5beeff7e003ccd299512b253174e43c5c9cf91f5. I will do the 0.6.3 release today on crates.io.

P-E-Meunier:
Thanks! And thanks again for *ring*.

jplatte:
Nice to see this already fixed and almost released, was just hit by it when working on a bunch of ruma crates.

briansmith:
*ring* 0.6.3 is on crates.io now!

Thanks again for the PR!

PR #433, opened 2017-01-31
`ring::aead::open_in_place()`: Remove the `in_prefix_len` parameter.

briansmith:


briansmith:
This is a potential solution for #371. I'm also exploring an alternate solution.

briansmith:
Closing this in favor of the approach in #436.

PR #434, opened 2017-01-31
`ring::aead::seal_in_place()`: replace `out_suffix_capacity` parameter

briansmith:
This is a potential solution to #291, by simply removing the functionality. It may also enable some new ways of using the API more efficiently. The main, probably small, downside is that it would no longer encourage creating protocols where the ciphertext and tag are treated as a single entity.


Philipp91:
According to GitHub, this PR is empty?

briansmith:
It's #435 now.

PR #435, opened 2017-01-31
`ring::aead::seal_in_place()`: replace `out_suffix_capacity` parameter.

briansmith:
This is a potential solution to #291, by simply removing the functionality. It may also enable some new ways of using the API more efficiently. The main, probably small, downside is that it would no longer encourage creating protocols where the ciphertext and tag are treated as a single entity.

### DIFF
--- a/src/aead/aead.rs
+++ b/src/aead/aead.rs
@@ -194,7 +194,7 @@ impl SealingKey {
 
 /// Encrypts and signs (“seals”) data in place.
 ///
-/// `nonce` must be unique for every use of the key to seal data. It must be
+/// `nonce` must be unique for every use of the key to seal data; it must be
 /// exactly `key.algorithm().nonce_len()` bytes long. `ad` is the additional
 /// authenticated data, which won't be encrypted. `in_out` must contain the
 /// plaintext to encrypt on input and will contain the ciphertext on successful
--- a/src/aead/aead.rs
+++ b/src/aead/aead.rs
@@ -196,21 +196,22 @@ impl SealingKey {
 ///
 /// `nonce` must be unique for every use of the key to seal data; it must be
 /// exactly `key.algorithm().nonce_len()` bytes long. `ad` is the additional
-/// authenticated data, which won't be encrypted. `in_out` must contain the
-/// plaintext to encrypt on input and will contain the ciphertext on successful
-/// output; it may be empty. `tag_out` must be exactly
+/// authenticated data, which won't be encrypted. `plaintext_in_ciphertext_out`
+/// must contain the plaintext to encrypt on input and will contain the
+/// ciphertext on successful output; it may be empty. `tag_out` must be exactly
 /// `key.algorithm().tag_len()` bytes long and will contain the tag on output.
 ///
 /// C analog: `EVP_AEAD_CTX_seal`.
 ///
 /// Go analog: [`AEAD.Seal`](https://golang.org/pkg/crypto/cipher/#AEAD)
 pub fn seal_in_place(key: &SealingKey, nonce: &[u8], ad: &[u8],
-                     in_out: &mut [u8], tag_out: &mut [u8])
+                     plaintext_in_ciphertext_out: &mut [u8], tag_out: &mut [u8])
                      -> Result<(), error::Unspecified> {
     let nonce = try!(slice_as_array_ref!(nonce, NONCE_LEN));
     let tag_out = try!(slice_as_array_ref_mut!(tag_out, TAG_LEN));
-    try!(check_per_nonce_max_bytes(in_out.len()));
-    (key.key.algorithm.seal)(&key.key.ctx_buf, nonce, ad, in_out, tag_out)
+    try!(check_per_nonce_max_bytes(plaintext_in_ciphertext_out.len()));
+    (key.key.algorithm.seal)(&key.key.ctx_buf, nonce, ad,
+                             plaintext_in_ciphertext_out, tag_out)
 }
 
 /// `OpeningKey` and `SealingKey` are type-safety wrappers around `Key`, which
--- a/src/aead/aead.rs
+++ b/src/aead/aead.rs
@@ -77,11 +77,14 @@ impl OpeningKey {
 
 /// Authenticates and decrypts (“opens”) data in place. When
 ///
-/// The input may have a prefix that is `in_prefix_len` bytes long; any such
+/// `nonce` must be unique for every use of the key to seal data; it must be
+/// exactly `key.algorithm().nonce_len()` bytes long. `ad` is the additional
+/// authenticated data, which won't be encrypted; it may be empty. The input
+/// may have an optional prefix that is `in_prefix_len` bytes long; any such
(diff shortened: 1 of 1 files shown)


src/aead/aead.rs
```diff
@@ -77,11 +77,14 @@ impl OpeningKey {
 
 /// Authenticates and decrypts (“opens”) data in place. When
 ///
-/// The input may have a prefix that is `in_prefix_len` bytes long; any such
+/// `nonce` must be unique for every use of the key to seal data; it must be
```
Philipp91: Since this is the `open_in_place` method, maybe the nonce should be explained as "the same nonce as was used for sealing"?

briansmith: > Since this is the open_in_place method, maybe the nonce should be explained as "the same nonce as was used for sealing"?

The caller of `open_in_place()` generally cannot know the value of the nonce that was used for sealing; it basically can only know that it did the sealing itself. The nonce used in `open_in_place()` will be whatever nonce was sent by the peer if explicit nonces are used. In the good-but-rare cases where implicit nonces are used, the nonce will be whatever the protocol being implemented specifies it to be. In other words, the caller of `open_in_place()` doesn't have any real flexibility in choosing the nonce in any protocol.

When explicit nonces are used, the caller of `seal_in_place()` does have a lot more flexibility. Maybe we should advise them to ensure they always use a nonce that is unique for the given key, and to not to use random nonces since the nonce length is too small to satisfy the uniqueness requirement with random nonces. But providing more advice than that seems kind of out-of-scope since it's a protocol design issue.

Philipp91: > Maybe we should advise them to ensure they always use a nonce that is unique for the given key, and to not to use random nonces since the nonce length is too small to satisfy the uniqueness requirement with random nonces.

I was going to suggest that when you wrote "I don't recommend using random nonces with these AEADs" over here https://github.com/briansmith/ring/pull/421#issuecomment-276153893, which was a total surprise to me. As it's an actual security concern and I wouldn't expect anyone except absolute crypto experts to know about this, it's something that the docs should tell you.

> In the good-but-rare cases where implicit nonces are used, the nonce will be whatever the protocol being implemented specifies it to be.

Yes. Though my proposed "the same nonce as was used for sealing" doesn't contradict that.

> In other words, the caller of `open_in_place()` doesn't have any real flexibility in choosing the nonce in any protocol.

I get your point. It's from the viewpoint of someone who uses _ring_ to implement a given protocol.

As you may have noticed, my viewpoint is usually someone who just tries to encrypt something securely and send it accross the network. If you will, such users are designing their own "protocol" (albeit a tiny one) and for their protocol design, they have to understand the fundamental concept that the nonce needs to be the same on both ends. As the Rustdoc is now, it just tells the user how to choose nonces for sealing (that is, uniquely), but for all the user knows about opening, they could again specify a new, unique and random nonce.
That's especially the case when these users come from using AES-ECB encryption, and someone told them it's insecure and they have to use AEAD instead. They don't have to change that much, but they have to know about tags and nonces now.

briansmith: > I was going to suggest that when you wrote "I don't recommend using random nonces with these AEADs" over here #421 (comment), which was a total surprise to me.

See also https://boringssl-review.googlesource.com/c/13440/#message-e6184e6b129cf0ac487dbe18c0ae85c54174bed1 and https://www.blackhat.com/docs/us-16/materials/us-16-Devlin-Nonce-Disrespecting-Adversaries-Practical-Forgery-Attacks-On-GCM-In-TLS-wp.pdf.

> As it's an actual security concern and I wouldn't expect anyone except absolute crypto experts to know about this, it's something that the docs should tell you.

Maybe. Right now *ring*'s documentation isn't intended to teach people cryptography, just intended to teach people how to use *ring*.

> I get your point. It's from the viewpoint of someone who uses ring to implement a given protocol.

This is what *ring* is designed for.

> As you may have noticed, my viewpoint is usually someone who just tries to encrypt something securely and send it accross the network. If you will, such users are designing their own "protocol" (albeit a tiny one) and for their protocol design, they have to understand the fundamental concept that the nonce needs to be the same on both ends.

If they don't understand already that the nonce would need to be the same at both ends, they shouldn't be designing their own protocol.

> but for all the user knows about opening, they could again specify a new, unique and random nonce.

That is true, but then their thing won't work. And if they choose the wrong nonce then there isn't any security issue because AES-GCM is designed to fail safe when the wrong nonce is used during opening, and also because *ring* fails safe when things go wrong (e.g. overwriting the ciphertext/plaintext when authentication fails, which is what would happen with a wrong nonce or wrong key or wrong ciphertext or wrong AAD).

> That's especially the case when these users come from using AES-ECB encryption, and someone told them it's insecure and they have to use AEAD instead. They don't have to change that much, but they have to know about tags and nonces now.

They literally need to read a book. *ring*'s documentation isn't intended to replace a book or other educational tools about designing protocols. Rather, *ring* is designed to make it easy to verify that a *ring*-based implementation matches a specification of the higher-level protocol being implemented.

Philipp91: (Don't let this PR be blocked over this discussion.)

> And if they choose the wrong nonce then there isn't any security issue because AES-GCM is designed to fail safe ...

That's a good point and it settles the discussion whether `open_in_place` needs to explain its nonces better. You're right, it's not a misuse resistance issue, only a minor usability problem.

About whether _ring_ should teach cryptography: not really. APIs should be usable in a safe way without background knowledge (because, if they're not, bad things happen unfortunately, see below). The best way coincides with your goal to make _ring_'s API impossible to misuse. In some cases, the API itself can't do it and explaining stuff in the documentation (or at least pointing to the right explanations, making people aware that problems exist) is plan B.
For example, the documentation should state that random nonces are not perfectly secure (and maybe point to sites that explain it). Without that pointer, users won't know about this. I didn't know and I've been working on crypto and crypto APIs for half a year now. Normal developers have half a day before their encryption has to work. So the documentation doesn't replace a book, but it replaces the need to read an entire book and all recent security developments before beginnig a project by pointing to the relevant stuff.

> they shouldn't be designing their own protocol
> They literally need to read a book.

I agree, they should. But some won't read a book and others won't understand it (because of lacking time or interest) and they still need to send encrypted data. They don't intend to design a protocol and they don't approach it like that, they just encrypt and go ahead.

There is a good example from the SSL area. This may be anecdotal evidence, but I think it's somewhat representative: dev turns off certificate validation, researcher informs about the vulnerability, dev says its secure because, on Wireshark, the traffic is obviously encrypted.  [http://android-ssl.org/files/p49.pdf http://dl.acm.org/citation.cfm?id=2516655 section 4.2.1]
That reasoning also applies to random nonces for AEAD: "Clever people say that AE is secure, I'm using secure randomness, it's visibly encrypted, done."

> _ring_ is designed to make it easy to verify that a ring-based implementation matches a specification of the higher-level protocol being implemented.

Fair enough.
Like all other crypto libraries, _ring_ can't avoid being used by uninformed users, too. And in fact, _ring_ is closest to being recommendable for such users, and getting better with every commit.

briansmith:
Closing this in favor of the approach in #436.

Issue #299, opened 2016-09-11
Allow extraction of RSAKeyPair public key components


djc:
So I started implementing _ring_-based crypto in my project, and it turns out I forgot about this part. It needs what is basically a precursor to `signature::primitive::verify_rsa()`: a way to extract `n` and `e` out of an `RSAKeyPair`, similar to `EdDSA`'s `public_key_bytes()`. Does that sound like an okay API to add?

To avoid allocation, it could take a tuple of two `u8` slices as arguments, the first one of which should be exactly `public_modulus_len()` bytes long and the second one just 5 bytes, I guess. So:

`pub fn public_key_components((n, e): (&[u8], &[u8]) -> Result<(), Unspecified>`

It would return `Err` if the slices have incorrect sizes. Otherwise, the argument slices would contain `n` and `e` in big-endian form, similar to how `verify_rsa()` takes them. The zero padding for `e` is a bit awkward, I guess an alternative would be to just take `Vec`s for both?

What do you think? Hopefully this will be easier than the `verify_rsa()` API itself!

briansmith:
> To avoid allocation, it could take a tuple of two u8 slices as arguments, the first one of which should be exactly public_modulus_len() bytes long and the second one just 5 bytes, I guess.

The exponent isn't always 5 bytes. It could be one byte, e.g. when it is 3. So, with this design there would need to be a `public_exponent_len()` function added as well.

> pub fn public_key_components((n, e): (&mut [u8], &mut [u8]) -> Result<(), Unspecified>
> 
> It would return Err if the slices have incorrect sizes. Otherwise, the argument slices would contain n and e in big-endian form, similar to how verify_rsa() takes them.  The zero padding for e is a bit awkward, I guess an alternative would be to just take Vecs for both?

I think it should be the case that the output of `public_key_components()` should be able to be directly fed into `verify_rsa()`, which would mean no padding.

djc:
> I think it should be the case that the output of public_key_components() should be able to be directly fed into verify_rsa(), which would mean no padding.

That sounds right. So we pass in a slice of at least `public_modulus_len()` + 5, and it will return slices matching the lifetime of the argument?

briansmith:
I would expect something like this (which, I guess, should appear in a doc comment:

```
let n = vec![0, key_pair.public_modulus_len()];
let e = vec![0, key_pair.public_exponent_len()];
try!(key_pair.public_key_components(&mut n, &mut e));
[...]

PR #441, opened 2017-02-01
Allow exporting public components of `RSAKeyPair`.

stephank:
In [Portier], we load RSA private keys once, leveraging the DER parsing in Ring. However, we need to then serve the public keys as JWK to clients. These additions allow us to grab the public components of an `RSAKeyPair` to build the JWK.

 [Portier]: https://portier.github.io/

### DIFF
--- a/src/rsa/bigint.rs
+++ b/src/rsa/bigint.rs
@@ -38,7 +38,7 @@
 // XXX TODO: Remove this once RSA verification has been done in Rust.
 #![cfg_attr(not(feature = "rsa_signing"), allow(dead_code))]
 
-use {bits, bssl, c, der, error, rand, untrusted};
+use {bits, bssl, c, der, error, polyfill, rand, untrusted};
 use core;
 use core::marker::PhantomData;
 
@@ -221,6 +221,11 @@ impl<M> Modulus<M> {
     pub fn bit_length(&self) -> bits::BitLength {
         self.ctx.n().bit_length()
     }
+
+    pub fn fill_be_bytes(&self, out: &mut [u8])
+                         -> Result<(), error::Unspecified> {
+        self.ctx.n().fill_be_bytes(out)
+    }
 }
 
 // `Modulus` uniquely owns and references its contents.
@@ -303,10 +308,7 @@ impl<M> ElemDecoded<M> {
 
     pub fn fill_be_bytes(&self, out: &mut [u8])
                          -> Result<(), error::Unspecified> {
-        bssl::map_result(unsafe {
-            GFp_BN_bn2bin_padded(out.as_mut_ptr(), out.len(),
-                                 self.value.as_ref())
-        })
+        self.value.fill_be_bytes(out)
     }
 
     pub fn is_zero(&self) -> bool { self.value.is_zero() }
@@ -418,6 +420,33 @@ pub fn elem_sub<M>(a: ElemDecoded<M>, b: &ElemDecoded<M>, m: &Modulus<M>)
 /// [3, 2**PUBLIC_EXPONENT_MAX_BITS).
 #[derive(Clone, Copy)]
 pub struct PublicExponent(u64);
+
+impl PublicExponent {
+    pub fn bit_length(&self) -> bits::BitLength {
+        bits::BitLength::from_usize_bits(64 - self.0.leading_zeros() as usize)
+    }
+
+    pub fn fill_be_bytes(&self, out: &mut [u8])
+                         -> Result<(), error::Unspecified> {
+        let e_bytes = self.bit_length().as_usize_bytes_rounded_up();
+        if out.len() < e_bytes {
+            return Err(error::Unspecified);
+        }
+
+        let out_pad_bytes = out.len() - e_bytes;
+        for i in 0..out_pad_bytes {
+            out[i] = 0;
+        }
+
+        let src = polyfill::slice::be_u8_from_u64(self.0);
+        let src_pad_bytes = 8 - e_bytes;
+        for i in 0..e_bytes {
+            out[out_pad_bytes + i] = src[src_pad_bytes + i];
+        }
+
+        Ok(())
+    }
+}
 
 // This limit was chosen to bound the performance of the simple
 // exponentiation-by-squaring implementation in `elem_exp_vartime`. In
@@ -565,6 +594,11 @@ impl Nonnegative {
         self.0.bit_length()
     }
 
+    pub fn fill_be_bytes(&self, out: &mut [u8])
+                         -> Result<(), error::Unspecified> {
+        self.0.fill_be_bytes(out)
+    }
+
     pub fn try_clone(&self) -> Result<Nonnegative, error::Unspecified> {
         let mut r = try!(Nonnegative::zero());
         try!(bssl::map_result(unsafe {
@@ -580,7 +614,7 @@ impl Nonnegative {
 #[allow(non_snake_case)]
 mod repr_c {
     use core;
-    use {bits, c, limb};
+    use {bits, bssl, c, error, limb};
     use libc;
 
     /* Keep in sync with `bignum_st` in openss/bn.h. */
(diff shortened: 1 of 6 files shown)


stephank:
That is one huge CI setup. :)

Thinking some more about just what's in this PR, I have 2 questions myself:

 - Is the `export_` naming ok? I considered `read_`, but it also sounds weird. Something else?
 - While simpler, I think it's probably inconsistent and less useful to expose the public exponent as an `u64`? Now considering treating it like the modulus, adding `public_exponent_len` and `export_public_exponent` instead.

For the opposite direction (loading JWK), I want to propose an API change. This can be a separate PR just fine. My Rust is still not as fluent as I'd like it to be, but I think this should be possible:

 - Formalize the `(n, e)` pair we pass around as a public struct `RSAPublicKey`. (Some of the `src/rsa/rsa.rs` methods could probably be moved to this type.)
 - Replace `n` and `e` in `RSAKeyPair`, with an `RSAPublicKey` field.
 - Move current public component `RSAKeyPair` accessors to `RSAPublicKey`. (Can drop the `public_` prefix.)
 - Provide an accessor for `&RSAPublicKey` on `RSAKeyPair`.
 - Provide bigint accessors on `RSAKeyPair` for the private components, similar to what this PR does for the public components. (For completeness)
 - Provide constructors for DER and BE bigint byte arrays.
 - Add a public key type field to the trait `VerificationAlgorithm`, and a `verify_with_key` method as an alternative to `verify`, which operates on an instance of that type.
 - Apply same to Ed25519. Hopefully smooth. :)

Does that sound good?

src/rsa/signing.rs
```diff
@@ -219,6 +217,20 @@ impl RSAKeyPair {
     pub fn public_modulus_len(&self) -> usize {
         self.n_bits.as_usize_bytes_rounded_up()
     }
+
+    /// Exports the public modulus as big endian to the output array.
+    ///
+    /// The array should be of at least the size returned by
+    /// `public_modulus_len`.
```
briansmith: I suggest this (with fixed line wrapping to 80 characters):

```rust
/// Extracts the public modulus.
///
/// `out` must be exactly `public_modulus_len()` bytes long. It
/// will be filled with the big-endian-encoded bytes of the
/// public key, without any padding.

File: src/rsa/signing.rs [hunk 2]
```diff
@@ -219,6 +217,20 @@ impl RSAKeyPair {
...
+    pub fn export_public_modulus(&self, out: &mut [u8])
+                    -> Result<(), error::Unspecified> {
+        self.n.fill_be_bytes(out)
```
briansmith: `fill_be_bytes()` zero-pads the result. I think that JWT and similar things don't want zero-padded output, so there needs to be a check that `out` is exactly the right length. This check would make this symmetric with the `ring::signature::verify_rsa()` and other APIs exposed by *ring*.

Also, let's add unit tests that verify that we reject 1-byte-too-long and 1-byte-too-short `out`, and accept correct-length `out`.

src/rsa/signing.rs
```diff
@@ -219,6 +217,20 @@ impl RSAKeyPair {
...
+    }
+
+    /// Return the public exponent.
+    pub fn public_exponent(&self) -> u64 {
+        self.e.value()
```
briansmith: Maybe it's better to output the exponent as a big-endian byte slice, like the modulus? That would be symmetric with `ring::signature::verify_rsa` and `export_public_modulus()`, and it is probably what all the callers need anyway.

src/rsa/bigint.rs
```diff
@@ -218,6 +218,21 @@ pub struct Modulus<M> {
...
+        bits::BitLength::from_usize_bits(bits)
+    }
+
+    pub fn fill_be_bytes(&self, out: &mut [u8])
+                         -> Result<(), error::Unspecified> {
+        bssl::map_result(unsafe {
+            GFp_BN_bn2bin_padded(out.as_mut_ptr(), out.len(),
+                                 self.as_ref())
+        })
+    }
```
briansmith: Let's move this `unsafe` implementation of `fill_be_bytes()` to `Nonnegative`, and then make the `Modulus` and `Elem` implementations be simple wrappers around it.

src/rsa/bigint.rs
```diff
@@ -218,6 +218,21 @@ pub struct Modulus<M> {
     ring: PhantomData<M>,
 }
 
+impl<M> Modulus<M> {
```
briansmith: Please change this to (wrapping at 80 characters):

```rust
/// Exporting information about the value is only allowed for the public
/// modulus `N`, not for private moduli.
impl Modulus<super::N> {
```

briansmith:
> Is the export_ naming ok? I considered read_, but it also sounds weird. Something else?

It isn't perfect but I don't have a better suggestion either.

> While simpler, I think it's probably inconsistent and less useful to expose the public exponent as an u64? Now considering treating it like the modulus, adding public_exponent_len and export_public_exponent instead.

I agree.

> For the opposite direction (loading JWK), I want to propose an API change.

Your suggestion is now tracked in #445. Let's discuss it there. I'll need some time to process it.

briansmith:
This would fix #299.

src/rsa/signing.rs
stephank: Shouldn't we leave this to the user? The user must allocate enough space somehow based on `public_modulus_len`. If there's not enough space, `fill_be_bytes` errors, if there's more space, it's padded, but that may be the user's intent. Doesn't `verify_rsa` also accept padded input?

I'll look into the unit tests!

briansmith: Here's an excerpt from the documentation for `verify_rsa`:
```rust
/// `n` is the public key modulus and `e` is the public key exponent. Both are
/// interpreted as unsigned big-endian encoded values. Both must be positive
/// and neither may have any leading zeros.
```

These requirements come directly from JWA https://tools.ietf.org/html/rfc7518#section-6.3.1.1 (for n):
> Note that implementers have found that some cryptographic libraries prefix an extra zero-valued octet to the modulus representations they return, for instance, returning 257 octets for a 2048-bit key, rather than 256.  Implementations using such libraries will need to take care to omit the extra octet from the base64url-encoded representation.

and (for e):

> For instance, when representing the value 65537, the octet sequence to be base64url-encoded MUST consist of the three octets [1, 0, 1];

Thus supporting extra zero padding doesn't do anybody any good, AFAICT.

src/rsa/bigint.rs
stephank: Why would we not allow this? JWK also has a private key mapping: https://tools.ietf.org/html/rfc7517#appendix-A.2

src/rsa/bigint.rs
stephank: I'm not sure how to do this. `Modulus` contains a `BN_MONT_CTX`, which owns a `BIGNUM`, but `Nonnegative` also wants ownership of its `BIGNUM`.

`bit_length` has the same issue. I can try to add both to `BIGNUM`, then make the others wrappers.

stephank:
I've addressed feedback as far as I could.

I don't see a reason why we wouldn't allow exporting the private components too, though I'm also not sure what that API would look like. Doing the same for all RSA private parts would add a slew of methods. But I don't have anything else.

src/rsa/bigint.rs
briansmith: > Why would we not allow this? JWK also has a private key mapping: https://tools.ietf.org/html/rfc7517#appendix-A.2

In *ring*, when you construct a private or symmetric key object, there's no way to get the key material out. This way, you can safely pass an `&RSAKeyPair` or `&OpeningKey` to a function and know that that function won't ever leak the key.

If you look at `Ed25519KeyPair::generate_serializable`, you can see that when we generate a key pair, we also output the bytes needed to serialize the key, along with the generated key pair. When we add RSA key generation to *ring*, we'll do it the same way.

briansmith:
@stephank Sorry for the long delay on this. 

I independently did some foundational work for this. For example, I added the equivalent of the `RSAPublicKey` type here:
https://github.com/briansmith/ring/blob/0c0f7c47112f9ff3e1b8d8d4427d8246fcdc0794/src/rsa/verification.rs#L24-L25

Now there are a lot of conflicts. I'm going to close this PR since there's an issue open for getting RSA public keys from RSA private keys and I think we should explore a new approach to resolving that.

stephank:
@briansmith No worries! Thanks for all the effort you're putting into ring & co, and sorry I couldn't keep up. :)

Issue #409, opened 2017-01-09
Failure building release targeting x86_64-unknown-linux-musl


bruceadams:
I've only been able to produce the failure when building a binary. Libraries build just fine. _Using_ a library in a binary is when the problem appears. The key message from the failure appears to be this:

```
  = note: /Users/ba/rustls/target/x86_64-unknown-linux-musl/release/deps/libring-863fd845bbadbb87.rlib(bn.o): In function `memcpy':
/usr/include/x86_64-linux-gnu/bits/string3.h:53: undefined reference to `__memcpy_chk'
```

I tried four combinations; one fails. All of my attempts were using Ubuntu 16.04 in a Docker container.

C compiler | debug  | release
-----------|---------|--------
gcc            | success | _failure_
clang         | success | success

I'm more than happy to try other combinations. Ideally, we would find a way to see this issue in a build and/or test run within `ring`. I do not (yet?) understand ring's build and test combinations well enough.

## Demonstration of the problem

Building the `tlsclient` example in [rustls](https://github.com/ctz/rustls) using two, published, Docker images for the two C build chains: [gcc](https://hub.docker.com/r/bruceadams/rustup-musl/builds/b3hdxd3gumxn4wp8zah965g/) and [clang](https://hub.docker.com/r/bruceadams/rustup-musl/builds/bxuu7cderisp8cx68xsbbs3/). I use a function, `rmusl`, in `bash` for the complex `docker` command for running builds inside a Docker container with the current Rust project context.

```shell
function rmusl() {
    tag=$1
    shift
    docker run --interactive \
               --rm \
               --tty \
               --user "$(id -u):$(id -g)" \
               --volume $HOME/.cargo/git:/home/rust/.cargo/git \
               --volume $HOME/.cargo/registry:/home/rust/.cargo/registry \
               --volume $PWD:$PWD \
               --workdir $PWD \
               bruceadams/rustup-musl:$tag \
               "$@"
}
```
### `gcc` debug build
```shell
cargo clean
rmusl gcc cargo build --target x86_64-unknown-linux-musl --example tlsclient
   Compiling strsim v0.5.2
   Compiling regex-syntax v0.3.9
   Compiling base64 v0.2.1
   Compiling log v0.3.6
   Compiling libc v0.2.19
   Compiling winapi-build v0.1.1
   Compiling ring v0.6.1
   Compiling rustc-serialize v0.3.22
   Compiling utf8-ranges v0.1.3
   Compiling slab v0.1.3
[...]

PR #442, opened 2017-02-02
Disable _FORTIFY_SOURCE for Musl on Linux

bruceadams:
This _works_, but I don't really understand what I'm doing. I didn't manage to compile https://github.com/ctz/rustls with this change on top of ring's `master`. I could do a side-by-side comparison by basing by change on be1e17583af16f6261b336d4696dca88fbe792a0.

This is a workaround for #409.

### DIFF
--- a/mk/top_of_makefile.mk
+++ b/mk/top_of_makefile.mk
@@ -68,6 +68,13 @@ ifeq ($(TARGET_ABI),eabi)
 MABI = aapcs
 endif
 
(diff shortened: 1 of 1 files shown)


briansmith:
Thanks! I think we should merge #321 and then take another look at this.

briansmith:
Thanks. We just rewrote the build system and the new patch cannot be applied. To (over-)compensate, I did this in fb830c4de07bc91247d1aa925d8f3323c0fbb01b, hopefully correctly.

PR #453, opened 2017-02-10
Implement constant-time modular left shift by 1 bit.

briansmith:
This is the (blueprint for the) constant-time modular left shift designed to complete the removal of the timing side channel (inherited from OpenSSL and then BoringSSL) from RSA signing setup.

I suspect the performance of this will be approximately the same as what's currently done in the master branch. It would be great to confirm that. This PR isn't intended to address the performance issue; it's just intended to (be a blueprint for the) fix for the timing side channel. Still, please let me know whether it helps at all, or hurts, the performance. I know of ways to optimize this if necessary.

/cc @ctz


briansmith:
Landed in 77a1f935d9859abb355dd13101c25bae5f15491f.

PR #455, opened 2017-02-14
Allow Ring to be built by emscripten

ur0:
Rust now supports WA and asm.js but it was impossible to build ring since the CPU architecture wasn't defined in the OpenSSL headers. This defines emscripten to be a 32-bit CPU.

### DIFF
--- a/include/openssl/base.h
+++ b/include/openssl/base.h
@@ -93,6 +93,8 @@ extern "C" {
 #define OPENSSL_PNACL
 #elif defined(__myriad2__)
 #define OPENSSL_32_BIT
+#elif defined(__EMSCRIPTEN__)
+#define OPENSSL_32_BIT
 #else
 #error "Unknown target CPU"
 #endif


briansmith:
There's a lot of work needed for *ring* to be usable in WebAssembly. In particular, the core implementation of every algorithm in *ring* is in x86/x86-64/32-bit ARM/AAarch64 assembly language. We would need some WebAssembly-capable implementations of all those functions—i.e. most of *ring*.

Also, we need a way to run the test suite on a WebAssembly platform. I'd like to see at least this testing part—even if all the tests fail—before we merge anything specific to WebAssembly.

briansmith:
Also, what's the use case for Emscripten as a target for *ring*? Any web browser application will probably be better off using the Web Crypto API. Especially if Emscripten can only do 32-bit math, there's no way it can compete against the 64-bit math done by the native Web Crypto API implementations.

briansmith:
I'm going to close this because I think more work is needed to get any useful Emscripten support, and it doesn't make sense to do this part without doing more of the work. Thanks for the PR though!

ur0:
@briansmith sorry for the incredibly premature and stupid PR. I thought that this would be useful to implement crypto ops not currently supported by the WebCrypto spec. 

(If you're okay with this,) I'll probably submit a better and hopefully functional PR sometime next month, with atleast some baseline tests. Oh, and feel free to let me know if this doesn't make any sense.

briansmith:
> I thought that this would be useful to implement crypto ops not currently supported by the WebCrypto spec.

That could be true. However, maybe we need a new API for *ring* that is async? With such an async API, we could have *ring* call into the WebCrypto API for the operations that WebCrypto supports (note that WebCrypto ops are always async), and calculate directly the things that WebCrypto doesn't support (e.g. X25519). I want to add an async API anyway.

I suggest that you file an issue where you describe the design, what operations *ring* supports that aren't in WebCrypto, what should we do for operations that WebCrypto already supports, potential API changes/extensions to integrate better with the web, etc.

rmhrisk:
@briansmith I would gladly switch https://github.com/PeculiarVentures/webcrypto-liner to ring if it were an option.

briansmith:
I'm not planning to support Emscripten. Clang and Rust have native WebAssembly support now. wasm32-unknown-unknown will be tracked in #657.

PR #466, opened 2017-02-22
Adds test case

### LINKED ISSUE #23: Test cases where pbkdf2_hmac panics
Currently there are some test cases in pbkdf2_tests.txt that are commented out because they are invalid inputs for _ring_'s PBKDF2 module. In particular, they request an output longer than the output length of the the digest function being used. We should turn those test cases into the equivalent of `[should_panic]`. AFAICT, this means we need to:
1. Uncomment those test cases.
2. Annotate those test cases with "FAILS: OUTPUT_LARGER_THAN_DIGEST_OUTPUT".
3. Change the testing code so that for the cases marked "FAILS: OUTPUT_LARGER_THAN_DIGEST_OUTPUT", a task gets spawned that does the test case. We have to wait for the task to end and then verify that it was ended due to a `panic!`. In other words, we need to create the equivalent of a `[should_panic]` test.

Additionally, we should add more test cases, particularly a test case for `iterations == 0`, as BoringSSL has (had) a bug when `interations == 0`.

AFAICT, this is complicated by https://github.com/rust-lang/rust/issues/25869, as I believe we cannot do `[should_panic]` tests or create a data-driven alternative to `[should_panic]` on Win32 since Rust on Win32 doesn't implement `panic!` correctly; `panic!`` just kills the entire process. We should probably just work around that by not running these tests on Win32.

joshuata:
Closes #23

### DIFF
--- a/src/pbkdf2.rs
+++ b/src/pbkdf2.rs
@@ -304,4 +304,27 @@ mod tests {
             Ok(())
         });
     }
+
+    #[test]
+    #[should_panic]
+    pub fn pkbdf2_zero_iterations() {
+        let prf = &pbkdf2::HMAC_SHA256;
+        let secret = "ZeroIterationsTest".as_bytes();
+        let iterations: u32 = 0;
+        let salt = "salt".as_bytes();
+        let mut out = vec![0u8; 2];
+        pbkdf2::derive(prf, iterations, &salt, &secret, &mut out);
+    }
+
+    #[test]
+    // Control for pkbdf2_zero_iterations
+    pub fn pkbdf2_one_iteration() {
+        let prf = &pbkdf2::HMAC_SHA256;
+        let secret = "ZeroIterationsTest".as_bytes();
+        let iterations: u32 = 1;
+        let salt = "salt".as_bytes();
+        let mut out = vec![0u8; 2];
+        pbkdf2::derive(prf, iterations, &salt, &secret, &mut out);
+    }
+
 }


joshuata:
I made a mistake while committing and included changes from another pull request. I will revert them and we can squash the commits when merging.

src/pbkdf2.rs
```diff
@@ -304,4 +304,15 @@ mod tests {
...
+
+    #[test]
+    #[should_panic]
+    pub fn pkbdf2_zero_iterations() {
+        let prf = &pbkdf2::HMAC_SHA256;
+        let secret = "ZeroIterationsTest".as_bytes();
+        let iterations = 0;
+        let salt = "salt".as_bytes();
+        let mut out = vec![0u8; 2];
+        pbkdf2::derive(prf, iterations as u32, &salt, &secret, &mut out);
```
briansmith: Thanks!

NIT: don't use `as u32` here. If you need a type annotation then use `let iterations: u32 = 0;` instead.

Also, please make a variant of the test that has all the parameters the same, except iterations >= 1, and which doesn't panic. Add a comment like `// Control for `pkbdf2_zero_iterations()`.

Please add the contributor's statement to the end of your commit message; see https://github.com/briansmith/ring#contributing.

If you know how, please rebase this on master without the ASAN changes, e.g. using `git rebase -i master` and moving this commit to the top of the list. If you don't know how to do that, don't worry about it; I'll do it when I land the change.

briansmith:
Thanks! This landed as b141727eceb5e771a4689c2c291b3ff5e2d19050.

Issue #470, opened 2017-03-01
.der enocded RSA public key in test does not appear to be a valid key


heartsucker:
I noticed this when I was unable to get the `ring` library to verify an RSASSA-PSS signature.

```
heartsucker@pythagoras:~/code/misc/ring (master)$ openssl version
OpenSSL 1.1.0e  16 Feb 2017
heartsucker@pythagoras:~/code/misc/ring (master)$ openssl rsa -pubin -in ./src/rsa/signature_rsa_example_public_key.der -inform DER -outform PEM 
unable to load Public Key
139706476819648:error:0D0680A8:asn1 encoding routines:asn1_check_tlen:wrong tag:../crypto/asn1/tasn_dec.c:1112:
139706476819648:error:0D07803A:asn1 encoding routines:asn1_item_embed_d2i:nested asn1 error:../crypto/asn1/tasn_dec.c:274:Type=X509_ALGOR
139706476819648:error:0D08303A:asn1 encoding routines:asn1_template_noexp_d2i:nested asn1 error:../crypto/asn1/tasn_dec.c:609:Field=algor, Type=X509_PUBKEY
```

lawliet89:
Try it with `openssl rsa -RSAPublicKey_in -in ./src/rsa/signature_rsa_example_public_key.der -inform DER -outform PEM ` instead.

The `RSAPublicKey_in` is different.

briansmith:
Thanks to both of you for pointing this out. I'm going to close this in favor of landing #472, which will hopefully clear things up. I suspect #384 will also help people learn the right format to use. Also #378 will add support for parsing the RSAPublicKey format out of SubjectPublicKeyInfo format keys.

PR #472, opened 2017-03-06
Add some example commands to generate public key

lawliet89:
For signature verification

Should address #470

### DIFF
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -306,6 +306,27 @@ pub trait VerificationAlgorithm: Sync + private::Private {
 /// Verify the signature `signature` of message `msg` with the public key
 /// `public_key` using the algorithm `alg`.
 ///
+/// To generate the public key from your DER-encoded private key, use
+///
+/// ```sh
+/// openssl rsa -in private_key.der \
+///             -inform DER
+///             -RSAPublicKey_out \
+///             -outform DER \
+///             -out public_key.der
+/// ```
+///
+/// To convert a PEM formatted public key, use
+///
+/// ```sh
+/// openssl rsa -RSAPublicKey_in \
+///             -in public_key.pem \
+///             -inform PEM \
+///             -outform DER \
+///             -RSAPublicKey_out \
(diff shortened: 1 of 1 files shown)


src/signature.rs
```diff
@@ -306,6 +306,27 @@ pub trait VerificationAlgorithm: Sync + private::Private {
 /// Verify the signature `signature` of message `msg` with the public key
 /// `public_key` using the algorithm `alg`.
 ///
+/// To generate the public key from your DER-encoded private key, use
```
briansmith: Since this is RSA-specific, let's put this text in the RSA example below, after the sentence ”RSA signing (but not verification) requires the rsa_signing feature to be enabled” and before the example code block.

I suggest using the wording like this:

> Note that by default OpenSSL writes RSA public keys in SubjectPublicKeyInfo format, not RSAPublicKey format, and by default it Base64-encodes them (”PEM” format).
>
> To convert the PEM SubjectPublicKeyInfo format to the binary RSAPublicKey format needed by `verify()`, use:
> ```sh <example code #2>```
> To extract the RSAPublicKey-formatted public key from an ASN.1 (binary) DER-encoded RSAPrivateKey format private key file, use:
>```sh <example code #1>```

In particular, I think it makes more sense to document how to convert from the default OpenSSL format to the RSAPublicKey format first.

WDYT?

lawliet89: Sure. Will make the changes.

lawliet89: Done.

briansmith:
Thanks! I landed this as d48469d74acae8ab8ab794b0f980638d0e2ef231 with a couple of typo fixes.

If/when you contribute in the future, please read https://github.com/briansmith/ring#contributing, and in particular please add the contributor's statement to the end of each commit's commit message.

Thanks again!

lawliet89:
Oops. 

Thanks. Maybe you could add this information to the PR template. I didn't check it because I thought it's just a "simple" comment addition.

Issue #459, opened 2017-02-17
Rewrite `GFp_ed25519_sign()`, `GFp_ed25519_verify()`, and `GFp_ed25519_public_from_private()` in Rust


briansmith:
The most immediate goal is to get rid of the `GFp_SHA512_4()` function. That weird function calculates a SHA-512 hash over up to 4 byte arrays. In Rust code, it can be replaced with use of `digest::digest()` for the single-item case, or use of `digest::Context` for the multi-item case.

The longer-term goal is to move the Ed25519 code to Rust.

For this step, let's just replace `GFp_ed25519_verify()`, `GFp_ed25519_sign()`, and `GFp_ed25519_public_from_private()`, while keeping all the underlying functions in C.

This is easier than it sounds because it's basically just transliteration from C to Rust.

To demonstrate the approach that I think is best, I've done `GFp_ed25519_public_from_private()` in the [ed25519-rust](https://github.com/briansmith/ring/tree/ed25519-rust) branch.

The main issue that we need to keep track of is performance. We need to make sure the refactoring doesn't regress performance.

newpavlov:
How about using [`ed25519-dalek`](https://github.com/isislovecruft/ed25519-dalek) for it? (it could be published under `ed25519` later) If you don't want to depend on it, code can be copy-pasted with necessary changes (e.g. replace sha2 crate dependency with ring implementation). As I understand CC-0 license should allow it without any complications.

briansmith:
Using ed25519-dalek is a possibility I'm considering. However, I want to experiment with improving the *ring* implementation, such as doing things like this, first.

ranweiler:
@briansmith, my schedule is finally settling down again, so I'm picking this up.

ranweiler:
Opened https://github.com/briansmith/crypto-bench/pull/27 to support the performance side of this issue.

briansmith:
I'm closing this with the understanding that it is all done.

PR #479, opened 2017-03-08
Start Ed25519 port

ranweiler:
For #459.

Todo:
- [x] Try out a `guard!`/`check!` macro
- [x] Change private representation of `Ed25519KeyPair`
- [ ] Avoid unnecessary recomputations
- [ ] Add comments
- [ ] Gather updated benchmarks

### DIFF
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -30,7 +30,11 @@
 
 
 /* Prevent -Wmissing-prototypes warnings. */
+void GFp_ge_double_scalarmult_vartime(ge_p2 *r, const uint8_t *a,
+                                      const ge_p3 *A, const uint8_t *b);
 void GFp_ge_p3_tobytes(uint8_t *s, const ge_p3 *h);
+void GFp_x25519_ge_tobytes(uint8_t *s, const ge_p2 *h);
+int GFp_x25519_ge_frombytes_vartime(ge_p3 *h, const uint8_t *s);
 void GFp_x25519_ge_scalarmult_base(ge_p3 *h, const uint8_t a[32]);
 void GFp_x25519_sc_muladd(uint8_t *s, const uint8_t *a, const uint8_t *b,
                           const uint8_t *c);
@@ -956,7 +960,7 @@ static void fe_pow22523(fe out, const fe z) {
   fe_mul(out, t0, z);
 }
 
-static void x25519_ge_tobytes(uint8_t *s, const ge_p2 *h) {
+void GFp_x25519_ge_tobytes(uint8_t *s, const ge_p2 *h) {
   fe recip;
   fe x;
   fe y;
@@ -986,7 +990,7 @@ static const fe d = {-10913610, 13857413, -15372611, 6949391,   114729,
 static const fe sqrtm1 = {-32595792, -7943725,  9377950,  3500415, 12389472,
                           -272473,   -25146209, -2005654, 326686,  11406482};
 
-static int x25519_ge_frombytes_vartime(ge_p3 *h, const uint8_t *s) {
+int GFp_x25519_ge_frombytes_vartime(ge_p3 *h, const uint8_t *s) {
   fe u;
   fe v;
   fe v3;
@@ -3639,7 +3643,7 @@ static const ge_precomp Bi[8] = {
  * where a = a[0]+256*a[1]+...+256^31 a[31].
  * and b = b[0]+256*b[1]+...+256^31 b[31].
  * B is the Ed25519 base point (x,4/5) with x positive. */
-static void ge_double_scalarmult_vartime(ge_p2 *r, const uint8_t *a,
+void GFp_ge_double_scalarmult_vartime(ge_p2 *r, const uint8_t *a,
                                          const ge_p3 *A, const uint8_t *b) {
   signed char aslide[256];
   signed char bslide[256];
@@ -4558,7 +4562,7 @@ int GFp_ed25519_verify(const uint8_t *message, size_t message_len,
                        const uint8_t public_key[32]) {
   ge_p3 A;
   if ((signature[63] & 224) != 0 ||
-      x25519_ge_frombytes_vartime(&A, public_key) != 0) {
+      GFp_x25519_ge_frombytes_vartime(&A, public_key) != 0) {
     return 0;
   }
 
@@ -4579,10 +4583,10 @@ int GFp_ed25519_verify(const uint8_t *message, size_t message_len,
   GFp_x25519_sc_reduce(h);
 
   ge_p2 R;
-  ge_double_scalarmult_vartime(&R, h, &A, scopy);
+  GFp_ge_double_scalarmult_vartime(&R, h, &A, scopy);
 
   uint8_t rcheck[32];
-  x25519_ge_tobytes(rcheck, &R);
+  GFp_x25519_ge_tobytes(rcheck, &R);
 
   return GFp_memcmp(rcheck, rcopy, sizeof(rcheck)) == 0;
 }
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -1020,7 +1020,7 @@ int GFp_x25519_ge_frombytes_vartime(ge_p3 *h, const uint8_t *s) {
   if (fe_isnonzero(check)) {
     fe_add(check, vxx, u); /* vx^2+u */
     if (fe_isnonzero(check)) {
-      return -1;
+      return 0;
     }
     fe_mul(h->X, h->X, sqrtm1);
   }
@@ -1030,7 +1030,7 @@ int GFp_x25519_ge_frombytes_vartime(ge_p3 *h, const uint8_t *s) {
   }
 
   fe_mul(h->T, h->X, h->Y);
-  return 0;
+  return 1;
 }
 
 static void ge_p2_0(ge_p2 *h) {
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -3644,7 +3644,7 @@ static const ge_precomp Bi[8] = {
  * and b = b[0]+256*b[1]+...+256^31 b[31].
  * B is the Ed25519 base point (x,4/5) with x positive. */
 void GFp_ge_double_scalarmult_vartime(ge_p2 *r, const uint8_t *a,
-                                         const ge_p3 *A, const uint8_t *b) {
+                                      const ge_p3 *A, const uint8_t *b) {
   signed char aslide[256];
   signed char bslide[256];
   ge_cached Ai[8]; /* A,3A,5A,7A,9A,11A,13A,15A */
@@ -4547,50 +4547,12 @@ void GFp_x25519_sc_muladd(uint8_t *s, const uint8_t *a, const uint8_t *b,
 
 /* Prototype to avoid -Wmissing-prototypes warnings. */
 void GFp_ed25519_scalar_mask(uint8_t a[32]);
-int GFp_ed25519_verify(const uint8_t *message, size_t message_len,
-                       const uint8_t signature[64],
-                       const uint8_t public_key[32]);
 
 void GFp_ed25519_scalar_mask(uint8_t a[32]) {
   a[0] &= 248;
   a[31] &= 63;
   a[31] |= 64;
 }
-
-int GFp_ed25519_verify(const uint8_t *message, size_t message_len,
-                       const uint8_t signature[64],
-                       const uint8_t public_key[32]) {
-  ge_p3 A;
-  if ((signature[63] & 224) != 0 ||
-      GFp_x25519_ge_frombytes_vartime(&A, public_key) != 0) {
-    return 0;
-  }
-
-  fe_neg(A.X, A.X);
-  fe_neg(A.T, A.T);
-
-  uint8_t pkcopy[32];
-  memcpy(pkcopy, public_key, 32);
-  uint8_t rcopy[32];
-  memcpy(rcopy, signature, 32);
-  uint8_t scopy[32];
-  memcpy(scopy, signature + 32, 32);
-
-  uint8_t h[SHA512_DIGEST_LENGTH];
-  GFp_SHA512_4(h, sizeof(h), signature, 32, public_key, 32, message,
-               message_len, NULL, 0);
-
-  GFp_x25519_sc_reduce(h);
-
-  ge_p2 R;
-  GFp_ge_double_scalarmult_vartime(&R, h, &A, scopy);
-
-  uint8_t rcheck[32];
-  GFp_x25519_ge_tobytes(rcheck, &R);
-
-  return GFp_memcmp(rcheck, rcopy, sizeof(rcheck)) == 0;
-}
-
 
 #if defined(BORINGSSL_X25519_X86_64)
 
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -4543,8 +4543,6 @@ void GFp_x25519_sc_muladd(uint8_t *s, const uint8_t *a, const uint8_t *b,
 
 /* Prototype to avoid -Wmissing-prototypes warnings. */
 void GFp_ed25519_scalar_mask(uint8_t a[32]);
-void GFp_ed25519_sign(uint8_t *out_sig, const uint8_t *message,
-                      size_t message_len, const uint8_t private_key[64]);
 int GFp_ed25519_verify(const uint8_t *message, size_t message_len,
                        const uint8_t signature[64],
                        const uint8_t public_key[32]);
@@ -4553,30 +4551,6 @@ void GFp_ed25519_scalar_mask(uint8_t a[32]) {
   a[0] &= 248;
   a[31] &= 63;
   a[31] |= 64;
-}
-
-void GFp_ed25519_sign(uint8_t *out_sig, const uint8_t *message,
-                      size_t message_len, const uint8_t private_key[64]) {
-  uint8_t az[SHA512_DIGEST_LENGTH];
-  GFp_SHA512_4(az, sizeof(az), private_key, 32, NULL, 0, NULL, 0, NULL, 0);
-
-  GFp_ed25519_scalar_mask(az);
-
-  uint8_t nonce[SHA512_DIGEST_LENGTH];
-  GFp_SHA512_4(nonce, sizeof(nonce), az + 32, 32, message, message_len, NULL, 0,
-               NULL, 0);
-
-  GFp_x25519_sc_reduce(nonce);
-  ge_p3 R;
-  GFp_x25519_ge_scalarmult_base(&R, nonce);
-  GFp_ge_p3_tobytes(out_sig, &R);
-
-  uint8_t hram[SHA512_DIGEST_LENGTH];
-  GFp_SHA512_4(hram, sizeof(hram), out_sig, 32, private_key + 32, 32, message,
-               message_len, NULL, 0);
-
-  GFp_x25519_sc_reduce(hram);
-  GFp_x25519_sc_muladd(out_sig + 32, hram, az, nonce);
 }
 
 int GFp_ed25519_verify(const uint8_t *message, size_t message_len,
--- a/src/ec/eddsa.rs
+++ b/src/ec/eddsa.rs
@@ -14,7 +14,7 @@
 
 //! EdDSA Signatures.
 
-use {bssl, c, digest, error, private, rand, signature};
+use {c, digest, error, private, rand, signature};
 use untrusted;
 
 /// Parameters for EdDSA signing and verification.
@@ -168,16 +168,44 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
     fn verify(&self, public_key: untrusted::Input, msg: untrusted::Input,
               signature: untrusted::Input) -> Result<(), error::Unspecified> {
         let public_key = public_key.as_slice_less_safe();
-        if public_key.len() != PUBLIC_KEY_LEN ||
-           signature.len() != SIGNATURE_LEN {
+        if public_key.len() != PUBLIC_KEY_LEN {
             return Err(error::Unspecified);
         }
-        let msg = msg.as_slice_less_safe();
         let signature = signature.as_slice_less_safe();
-        bssl::map_result(unsafe {
-            GFp_ed25519_verify(msg.as_ptr(), msg.len(), signature.as_ptr(),
-                               public_key.as_ptr())
-        })
+        if signature.len() != SIGNATURE_LEN {
+            return Err(error::Unspecified);
+        }
+        if (signature[63] & 224) != 0 {
+            return Err(error::Unspecified);
+        }
+        let public_key = slice_as_array_ref!(public_key, SCALAR_LEN).unwrap();
+        let (signature_r, signature_s) = signature.split_at(SCALAR_LEN);
+        let msg = msg.as_slice_less_safe();
+        let mut a = ExtPoint::new_at_infinity();
+        let ret = unsafe {
+            GFp_x25519_ge_frombytes_vartime(&mut a, public_key)
+        };
+        if ret != 0 {
+            return Err(error::Unspecified)
+        }
+        for i in 0..ELEM_LIMBS {
+            a.x[i] = -a.x[i];
+            a.t[i] = -a.t[i];
+        }
+        let mut r_copy = [0u8; SCALAR_LEN];
+        r_copy.copy_from_slice(signature_r);
+        let mut s_copy = [0u8; SCALAR_LEN];
+        s_copy.copy_from_slice(signature_s);
+        let h_digest = eddsa_digest(signature_r, public_key, msg);
+        let h = digest_scalar(h_digest);
+        let mut r = Point::new_at_infinity();
+        unsafe { GFp_ge_double_scalarmult_vartime(&mut r, &h, &a, &s_copy) };
+        let mut r_check = [0u8; ELEM_LEN];
+        unsafe { GFp_x25519_ge_tobytes(&mut r_check, &r) };
+        if r_copy != r_check {
+            return Err(error::Unspecified);
+        }
+        Ok(())
     }
 }
 
@@ -219,11 +247,12 @@ fn public_from_private(seed: &Seed, out: &mut PublicKey) {
 
 extern  {
     fn GFp_ed25519_scalar_mask(a: &mut Scalar);
-    fn GFp_ed25519_verify(message: *const u8, message_len: c::size_t,
-                          signature: *const u8/*[64]*/,
-                          public_key: *const u8/*[32]*/) -> c::int;
+    fn GFp_ge_double_scalarmult_vartime(r: &mut Point, a_coeff: &Scalar,
+                                        a: &ExtPoint, b_coeff: &Scalar);
     fn GFp_ge_p3_tobytes(s: &mut [u8; ELEM_LEN], h: &ExtPoint);
+    fn GFp_x25519_ge_frombytes_vartime(h: &mut ExtPoint, s: &Scalar) -> c::int;
     fn GFp_x25519_ge_scalarmult_base(h: &mut ExtPoint, a: &Seed);
+    fn GFp_x25519_ge_tobytes(s: &mut Scalar, h: &Point);
     fn GFp_x25519_sc_muladd(s: &mut Scalar, a: &Scalar, b: &Scalar, c: &Scalar);
     fn GFp_x25519_sc_reduce(s: &mut UnreducedScalar);
 }
@@ -244,6 +273,24 @@ impl ExtPoint {
             y: [0; ELEM_LIMBS],
             z: [0; ELEM_LIMBS],
             t: [0; ELEM_LIMBS],
+        }
+    }
+}
+
+// Keep this in sync with `ge_p2` in curve25519/internal.h.
+#[repr(C)]
+struct Point {
+    x: Elem,
+    y: Elem,
+    z: Elem,
+}
+
+impl Point {
+    fn new_at_infinity() -> Self {
+        Point {
+            x: [0; ELEM_LIMBS],
+            y: [0; ELEM_LIMBS],
+            z: [0; ELEM_LIMBS],
         }
     }
 }
--- a/src/ec/eddsa.rs
+++ b/src/ec/eddsa.rs
@@ -109,6 +109,11 @@ impl<'a> Ed25519KeyPair {
         Ok(pair)
     }
 
+    // Returns a reference to the little-endian-encoded private key bytes.
+    fn private_key_bytes(&'a self) -> &'a [u8] {
+        &self.private_public[..SEED_LEN]
+    }
+
     /// Returns a reference to the little-endian-encoded public key bytes.
     pub fn public_key_bytes(&'a self) -> &'a [u8] {
         &self.private_public[SEED_LEN..]
@@ -117,9 +122,35 @@ impl<'a> Ed25519KeyPair {
     /// Returns the signature of the message `msg`.
     pub fn sign(&self, msg: &[u8]) -> signature::Signature {
         let mut signature_bytes = [0u8; SIGNATURE_LEN];
-        unsafe {
-            GFp_ed25519_sign(signature_bytes.as_mut_ptr(), msg.as_ptr(),
-                             msg.len(), self.private_public.as_ptr());
+        { // borrow signature_bytes;
+            let (signature_r, signature_s) =
+                signature_bytes.split_at_mut(SCALAR_LEN);
+            let signature_r =
+                slice_as_array_ref_mut!(signature_r, SCALAR_LEN).unwrap();
+            let signature_s =
+                slice_as_array_ref_mut!(signature_s, SCALAR_LEN).unwrap();
+            let az = digest::digest(&digest::SHA512, self.private_key_bytes());
+            let (a_encoded, z_encoded) = az.as_ref().split_at(SCALAR_LEN);
+            let mut a = [0; SCALAR_LEN];
+            a.copy_from_slice(a_encoded);
+            unsafe { GFp_ed25519_scalar_mask(&mut a) };
+            let mut nonce = digest_scalar({
+                let mut ctx = digest::Context::new(&digest::SHA512);
+                ctx.update(z_encoded);
+                ctx.update(msg);
+                ctx.finish()
+            });
+            let mut r = ExtPoint::new_at_infinity();
+            unsafe {
+                GFp_x25519_ge_scalarmult_base(&mut r, &nonce);
+                GFp_ge_p3_tobytes(signature_r, &r);
+            }
+            let hram_digest =
+                eddsa_digest(signature_r, self.public_key_bytes(), msg);
+            let hram = digest_scalar(hram_digest);
+            unsafe {
+                GFp_x25519_sc_muladd(signature_s, &hram, &a, &mut nonce);
+            }
         }
         signature::Signature::new(signature_bytes)
     }
@@ -152,6 +183,23 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
 
 impl private::Private for EdDSAParameters {}
 
+fn eddsa_digest(signature_r: &[u8], public_key: &[u8],
+                msg: &[u8]) -> digest::Digest {
+    let mut ctx = digest::Context::new(&digest::SHA512);
+    ctx.update(signature_r);
+    ctx.update(public_key);
+    ctx.update(msg);
+    ctx.finish()
+}
+
+fn digest_scalar(digest: digest::Digest) -> Scalar {
+    let mut unreduced = [0u8; digest::SHA512_OUTPUT_LEN];
+    unreduced.copy_from_slice(digest.as_ref());
+    unsafe { GFp_x25519_sc_reduce(&mut unreduced) };
+    let mut scalar = [0u8; SCALAR_LEN];
+    scalar.copy_from_slice(&unreduced[..SCALAR_LEN]);
+    scalar
+}
 
 fn public_from_private(seed: &Seed, out: &mut PublicKey) {
     let seed_sha512 = digest::digest(&digest::SHA512, seed);
@@ -169,19 +217,15 @@ fn public_from_private(seed: &Seed, out: &mut PublicKey) {
     }
 }
 
-
 extern  {
-    fn GFp_ed25519_scalar_mask(a: &mut [u8; SCALAR_LEN]);
-    fn GFp_ed25519_sign(out_sig: *mut u8/*[64]*/, message: *const u8,
-                        message_len: c::size_t, private_key: *const u8/*[64]*/);
-
+    fn GFp_ed25519_scalar_mask(a: &mut Scalar);
     fn GFp_ed25519_verify(message: *const u8, message_len: c::size_t,
                           signature: *const u8/*[64]*/,
                           public_key: *const u8/*[32]*/) -> c::int;
-
     fn GFp_ge_p3_tobytes(s: &mut [u8; ELEM_LEN], h: &ExtPoint);
     fn GFp_x25519_ge_scalarmult_base(h: &mut ExtPoint, a: &Seed);
-
+    fn GFp_x25519_sc_muladd(s: &mut Scalar, a: &Scalar, b: &Scalar, c: &Scalar);
+    fn GFp_x25519_sc_reduce(s: &mut UnreducedScalar);
 }
 
 // Keep this in sync with `ge_p3` in curve25519/internal.h.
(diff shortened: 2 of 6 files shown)


ranweiler:
FWIW, benchmarks so far, based on the following commits (with nicknames):
- "master": https://github.com/ranweiler/ring/commit/c97bce7ec83603762e09082257a61f7f239da5c1
- "transliterate": https://github.com/ranweiler/ring/commit/5326683de8fc0596c893790c31c4617f0f8ae2c6
- "digest": https://github.com/ranweiler/ring/commit/67420eaf25c55c37274e25587766a4d53830cc41

These capture the upstream `master` branch when I started this work, the transliteration of `sign` and `verify` into Rust (still using `GFp_SHA512_4`), and the transliteration after the removal of that function (replaced w/direct use of `digest`). I then used `cargo-benchcmp` to compare the output of `cargo bench ed25519` for the cases below.


"master"/"transliterate":
```
$ cargo benchcmp tmp/master.out tmp/transliterate.out 
 name                                   master.out ns/iter  transliterate.out ns/iter  diff ns/iter  diff % 
 signature::ed25519::generate_key_pair  50,467              50,898                              431   0.85% 
 signature::ed25519::sign_empty         48,890              49,752                              862   1.76% 
 signature::ed25519::verify             146,163             146,369                             206   0.14% 
```

"master"/"digest":
```
$ cargo benchcmp tmp/master.out tmp/digest.out 
 name                                   master.out ns/iter  digest.out ns/iter  diff ns/iter  diff % 
 signature::ed25519::generate_key_pair  50,467              50,389                       -78  -0.15% 
 signature::ed25519::sign_empty         48,890              48,874                       -16  -0.03% 
 signature::ed25519::verify             146,163             146,335                      172   0.12% 
```
The raw data for the "digest" case:
```
test signature::ed25519::generate_key_pair                    ... bench:      50,389 ns/iter (+/- 2,086)
test signature::ed25519::sign_empty                           ... bench:      48,874 ns/iter (+/- 1,415)
test signature::ed25519::verify                               ... bench:     146,335 ns/iter (+/- 9,116)
```
As you can see, the differences seem to be negligible.

ranweiler:
Okay, this is ready for a first look.

src/ec/eddsa.rs
```diff
@@ -112,9 +113,52 @@ impl<'a> Ed25519KeyPair {
...
+                             core::ptr::null(), 0,
+                             core::ptr::null(), 0);
         unsafe {
-            GFp_ed25519_sign(signature_bytes.as_mut_ptr(), msg.as_ptr(),
-                             msg.len(), self.private_public.as_ptr());
+            GFp_ed25519_scalar_mask(
+                slice_as_array_ref_mut!(&mut az[..SCALAR_LEN], SCALAR_LEN).unwrap())
+        };
+        let mut nonce = [0u8; digest::SHA512_OUTPUT_LEN];
+        let az_offset_ptr = unsafe { az.as_ptr().offset(32) };
```
briansmith: Instead of using `offset()`, prefer to use slicing: `az[32..].as_ptr()`. This way we get the Rust compiler to do the bounds checking (and hopefully optimize it away after verifying that the check is not necessary).

In this case, it seems you actually want to use `az.split_at_mut(32)`, since you use `&mut az[..SCALAR_LEN]` above.

Note also the inconsistency here in using `32` in one place and `SCALAR_LEN` in another place.

src/ec/eddsa.rs
```diff
@@ -112,9 +113,52 @@ impl<'a> Ed25519KeyPair {
...
+        digest::GFp_SHA512_4(nonce.as_mut_ptr(), nonce.len(),
+                             az_offset_ptr, 32,
+                             msg.as_ptr(), msg.len(),
+                             core::ptr::null(), 0,
+                             core::ptr::null(), 0);
```
briansmith: Ditto: let's jump directly to using the proper Digest API.

src/ec/eddsa.rs
```diff
@@ -112,9 +113,52 @@ impl<'a> Ed25519KeyPair {
...
+            GFp_x25519_ge_scalarmult_base(&mut r, nonce_half_ref_mut);
+            let signature_bytes_half_ref_mut =
+                slice_as_array_ref_mut!(&mut signature_bytes[..SCALAR_LEN], SCALAR_LEN).unwrap();
+
+            GFp_ge_p3_tobytes(signature_bytes_half_ref_mut, &r);
+        }
+        let mut hram = [0u8; digest::SHA512_OUTPUT_LEN];
+        let public_private_offset_ptr = unsafe {
+            self.private_public.as_ptr().offset(32)
+        };
```
briansmith: Same things apply here. Use `split_at_mut()` and/or slicing, use a named constant instead of `32`, use the regular Digest API.

src/ec/eddsa.rs
```diff
@@ -112,9 +113,52 @@ impl<'a> Ed25519KeyPair {
...
+        digest::GFp_SHA512_4(hram.as_mut_ptr(), hram.len(),
+                             signature_bytes.as_ptr(), 32,
+                             public_private_offset_ptr, 32,
+                             msg.as_ptr(), msg.len(),
+                             core::ptr::null(), 0);
+        unsafe {
+            GFp_x25519_sc_reduce(hram.as_mut_ptr());
+            GFp_x25519_sc_muladd(signature_bytes.as_mut_ptr().offset(32),
```
briansmith: Ditto.

src/ec/eddsa.rs
```diff
@@ -168,19 +212,16 @@ fn public_from_private(seed: &Seed, out: &mut PublicKey) {
...
-
     fn GFp_ed25519_verify(message: *const u8, message_len: c::size_t,
                           signature: *const u8/*[64]*/,
                           public_key: *const u8/*[32]*/) -> c::int;
-
     fn GFp_ge_p3_tobytes(s: &mut [u8; ELEM_LEN], h: &Point);
     fn GFp_x25519_ge_scalarmult_base(h: &mut Point, a: &Seed);
-
+    fn GFp_x25519_sc_muladd(s: *mut u8, a: *const u8, b: *const u8,
+                            c: *const u8);
```
briansmith: Should these parameters be of type `[u8; SCALAR_LEN]` and similar, instead of pointers?

src/ec/eddsa.rs
```diff
@@ -168,19 +212,16 @@ fn public_from_private(seed: &Seed, out: &mut PublicKey) {
...
+    fn GFp_x25519_sc_reduce(s: *mut u8);
```
briansmith: Ditto.

It may be a good idea to have `type Scalar = [u8; SCALAR_LEN];`.

crypto/curve25519/curve25519.c
```diff
@@ -4543,8 +4543,6 @@ void GFp_x25519_sc_muladd(uint8_t *s, const uint8_t *a, const uint8_t *b,
 
 /* Prototype to avoid -Wmissing-prototypes warnings. */
 void GFp_ed25519_scalar_mask(uint8_t a[32]);
-void GFp_ed25519_sign(uint8_t *out_sig, const uint8_t *message,
-                      size_t message_len, const uint8_t private_key[64]);
```
briansmith: In this kind of commit, where we replace (C) code with (Rust) code, let's just put the removal of the old code in the same commit as the addition of the new replacement code. That way the poor reviewer doesn't have to wonder if anything is still using the old version.

src/ec/eddsa.rs
```diff
@@ -131,12 +131,7 @@ impl<'a> Ed25519KeyPair {
                              core::ptr::null(), 0,
                              core::ptr::null(), 0);
         unsafe { GFp_x25519_sc_reduce(nonce.as_mut_ptr()) };
-        let mut r = Point {
-            x: [0; ELEM_LIMBS],
-            y: [0; ELEM_LIMBS],
-            z: [0; ELEM_LIMBS],
-            t: [0; ELEM_LIMBS],
-        };
+        let mut r = Point::new();
```
briansmith: Here is `r` at infinity? If so, I think `new_at_infinity()` or `at_infinity()` is a better name than `new()`. Note that the code in src/ec/suite_b/ops.rs defines `Point::new_at_infinity()` and it's good to be consistent with that. (You can imagine in the future we're likely to refactor things so that the P-{256, 384} code and the Ed25519 code share some of these data types.)

src/ec/eddsa.rs
```diff
@@ -234,6 +234,24 @@ impl Point {
     }
 }
 
+// Keep this in sync with `ge_p2` in curve25519/internal.h.
+#[repr(C)]
+struct ProjPoint {
```
briansmith: In `ring::ec::suite_b::ops` projective points are named just `Point`. I think it's better to name this `Point` too, renaming the four-component one.

src/ec/eddsa.rs
```diff
@@ -234,6 +234,24 @@ impl Point {
...
+  x: Elem,
+  y: Elem,
+  z: Elem,
+}
+
+impl ProjPoint {
+    fn new() -> Self {
```
briansmith: Again, `new_at_infinity()` seems like a better name.

src/ec/eddsa.rs
```diff
@@ -176,10 +176,56 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
...
-        })
+
+        let mut a = Point::new();
+        if (signature[63] & 224) != 0 {
+            return Err(error::Unspecified);
+        }
+        let ret = unsafe {
+            GFp_x25519_ge_frombytes_vartime(&mut a, public_key.as_ptr())
+        };
+        if ret != 0 { return Err(error::Unspecified) }
```
briansmith: IMO it makes more sense to change the return value convention of `GFp_x25519_ge_frombytes_vartime()` (in an earlier commit) to match the BoringSSL convention (1 is success, 0 is error) so that we can use `try!(bssl::map_result(...))` here instead.

src/ec/eddsa.rs
```diff
@@ -176,10 +176,56 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
...
+
+        for i in 0 .. ELEM_LIMBS {
+            a.x[i] = -a.x[i];
+            a.t[i] = -a.t[i];
+        }
+        let mut r_copy = [0u8; ELEM_LEN];
+        unsafe {
+            core::ptr::copy_nonoverlapping(signature.as_ptr(),
+                                           r_copy.as_mut_ptr(), ELEM_LEN)
+        };
```
briansmith: Again, I understand what you're trying to do regarding the direct transliteration and then improvement, but it's too much work to review the direct transliteration separately when it results in severely non-idiomatic Rust like this. In particular, we should use normal and safe array/slice copying here, instead of the unsafe `copy_nonoverlapping.

src/ec/eddsa.rs
```diff
@@ -204,11 +250,12 @@ fn public_from_private(seed: &Seed, out: &mut PublicKey) {
 
 extern  {
     fn GFp_ed25519_scalar_mask(a: &mut [u8; SCALAR_LEN]);
-    fn GFp_ed25519_verify(message: *const u8, message_len: c::size_t,
-                          signature: *const u8/*[64]*/,
-                          public_key: *const u8/*[32]*/) -> c::int;
+    fn GFp_ge_double_scalarmult_vartime(r: &mut ProjPoint, a_coeff: *const u8,
+                                        a: &Point, b_coeff: *const u8);
```
briansmith: Again, let's avoid pointer arguments when we can use array references instead.

src/ec/eddsa.rs
```diff
@@ -180,9 +180,10 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
         let ret = unsafe {
             GFp_x25519_ge_frombytes_vartime(&mut a, public_key.as_ptr())
         };
-        if ret != 0 { return Err(error::Unspecified) }
-
-        for i in 0 .. ELEM_LIMBS {
+        if ret != 0 {
+            return Err(error::Unspecified)
+        }
+        for i in 0..ELEM_LIMBS {
```
briansmith: Let's just fold this entire commit into the previous commits that it is fixing.

briansmith:
Sorry for the terse review comments. In general I'm happy that this is being done. Note that in your commit messages, you don't need to list things in such a detailed fashion; for the most part the bullets are stating the obvious. I'd rather the commit messages focus more on the "why" and "how" and things that would not be clear from reading the patch.

src/constant_time.rs
```diff
@@ -33,5 +33,5 @@ pub fn verify_slices_are_equal(a: &[u8], b: &[u8])
 }
 
 extern {
-    fn GFp_memcmp(a: *const u8, b: *const u8, len: c::size_t) -> c::int;
+    pub fn GFp_memcmp(a: *const u8, b: *const u8, len: c::size_t) -> c::int;
```
briansmith: This shouldn't be necessary....

src/ec/eddsa.rs
```diff
@@ -176,10 +176,56 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
...
+        unsafe {
+            GFp_ge_double_scalarmult_vartime(&mut r, h.as_ptr(), &a,
+                                             (&s_copy).as_ptr());
+        };
+        let mut r_check = [0u8; ELEM_LEN];
+        unsafe {
+            GFp_x25519_ge_tobytes(r_check.as_mut_ptr(), &r)
+        };
+        let cmp = unsafe {
+            constant_time::GFp_memcmp(r_check.as_ptr(), (&r_copy).as_ptr(), 32)
```
briansmith: This should just use the safe `verify_slices_are_equal()` function. Also maybe instead of serializing `r` to `r_check` and then comparing `r_check` to `r_copy`, we should just compare the deserialized values?

briansmith: > Also maybe instead of serializing r to r_check and then comparing r_check to r_copy, we should just compare the deserialized values?

Please ignore that suggestion.

src/ec/eddsa.rs
ranweiler: Fixed in 64b61234ff65dd222f4cc6497fad5ee58db6224d.

src/ec/eddsa.rs
ranweiler: Fixed in 64b61234ff65dd222f4cc6497fad5ee58db6224d.

src/ec/eddsa.rs
ranweiler: Done.

src/ec/eddsa.rs
ranweiler: Fixed in 64b61234ff65dd222f4cc6497fad5ee58db6224d.

src/ec/eddsa.rs
```diff
@@ -112,9 +113,52 @@ impl<'a> Ed25519KeyPair {
     /// Returns the signature of the message `msg`.
     pub fn sign(&self, msg: &[u8]) -> signature::Signature {
         let mut signature_bytes = [0u8; 64];
+        let mut az = [0u8; digest::SHA512_OUTPUT_LEN];
+        digest::GFp_SHA512_4(az.as_mut_ptr(), az.len(),
+                             self.private_public.as_ptr(), 32,
+                             core::ptr::null(), 0,
+                             core::ptr::null(), 0,
+                             core::ptr::null(), 0);
```
ranweiler: Done.

src/ec/eddsa.rs
ranweiler: Done.

crypto/curve25519/curve25519.c
ranweiler: Done.

src/constant_time.rs
ranweiler: It is, at least as of `rustc` 1.15.1. I would only expect it to be unnecessary if `constant_time` were a parent module of `ec::dsa`, but it is only an aunt(?).

ranweiler: Oh, you no doubt meant the direct usage of `GFp_memcmp` more generally, not just this specific change. Agreed.

src/ec/eddsa.rs
ranweiler: Done. Didn't end up using `split_at_mut` because it complicated borrowing unnecessarily in this case.

src/ec/eddsa.rs
ranweiler: Done.

src/constant_time.rs
ranweiler: Done, just used `verify_slices_are_equal` instead.

src/ec/eddsa.rs
ranweiler: Done (the first part).

src/ec/eddsa.rs
ranweiler: Done.

src/ec/eddsa.rs
ranweiler: Done.

briansmith:
Sorry, you'll probably have to rebase this on the msater branch. It seems that recently something (external?) changed that caused the Travis CI setup to break, which was causing your PR builds to fail. I just pushed a fix that seems to fix things.

ranweiler:
@briansmith, I've rebased onto `master` and responded to almost all of your comments, folding most of my changes back into the main "transliteration" commits for easier reviewing.

An except to this is the conversion of the extern function signatures. Before going back for an overall review, I suggest just taking a look at what I'm doing in bf1951f86ea273a54877d8cc8eb16a9aebae1118. After changing the signatures of the `extern` functions to accept array refs/the new `Scalar` alias, the polyfill usage ended up making call sites very noisy. To mitigate this, I added some functions which wrap the `slice_as_array_ref*!` macros. This helped, but if we reformat to wrap at 80 chars, it still makes things a bit harder to read.

There might be a nicer way to do this, and part of this is subjective, so I'd like your opinion on that commit. I've temporarily left those lines > 80 chars, so if you like, you can easily experiment with what formatting style we'd like to use in cases like this. Once we've settled on a style, and addressed anything else in that commit, I can split it up and fold the changes back into the earlier transliteration commits.

After that I'll respond to your last remaining comment about changing the return convention of `GFp_x25519_ge_frombytes_vartime` so we can use the `bssl::map_result`. At that point, the overall PR should be ready for second review.

src/ec/eddsa.rs
```diff
@@ -196,8 +198,12 @@ struct Point {
 type Elem = [i32; ELEM_LIMBS];
 const ELEM_LIMBS: usize = 10;
 
-type PublicKey = [u8; ELEM_LEN];
+type PublicKey = [u8; PUBLIC_KEY_LEN];
+const PUBLIC_KEY_LEN: usize = ELEM_LEN;
 
+const PRIVATE_KEY_LEN: usize = ELEM_LEN;
```
briansmith: This should be `SCALAR_LEN`, right?

src/ec/eddsa.rs
```diff
@@ -196,8 +198,12 @@ struct Point {
...
+const KEY_PAIR_LEN: usize = PUBLIC_KEY_LEN + PRIVATE_KEY_LEN;
+const SIGNATURE_LEN: usize = 64;
```
briansmith: The signature is, IIUC, one element concatenated with one scalar, so `ELEM_LEN + SCALAR_LEN`.

ranweiler: Yes, that's better.

briansmith: I will fix this and the other thing.

src/ec/eddsa.rs
ranweiler: Hmm, it actually seems like it should probably be `SEED_LEN`.

src/ec/eddsa.rs
ranweiler: Agreed, done in ac48786724ba97f863966bd85cf637aeb9215a62.

src/ec/eddsa.rs
ranweiler: Done.

src/ec/eddsa.rs
ranweiler: Yes, done.

src/ec/eddsa.rs
ranweiler: Okay, looking at it further, there seems to be a little bit of conceptual muddiness in the naming.  For example, consider the signature:
```rust
GFp_x25519_ge_scalarmult_base(h: &mut ExtPoint, a: &Seed)
```
It seems like the second argument here should actually be a `&Scalar`, based on the function name, and what it does (confirmed by the C comment). The only place `Seed` or the above function are used is in the module-private `public_from_private` helper, which in its name and usage accepts the bytes of a private key, but uses a different (more generic) type alias (`Seed`) for its first parameter.

Given that, I think we should get rid of `Seed`, add a `PrivateKey` type alias, and define `PUBLIC_KEY_LEN = 32`, since it isn't _really_ a scalar itself— it is just a bitstring which determines a scalar, and is determined by the Ed25519 `b` parameter, which we never explicitly mention in this code. What do you think?

src/ec/eddsa.rs
```diff
@@ -133,33 +174,144 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
...
+
+// Keep this in sync with `fe` in curve25519/internal.h.
+type Elem = [i32; ELEM_LIMBS];
+const ELEM_LIMBS: usize = 10;
+const ELEM_LEN: usize = 32;
+
+type PublicKey = [u8; PUBLIC_KEY_LEN];
+const PUBLIC_KEY_LEN: usize = ELEM_LEN;
+
+const PRIVATE_KEY_LEN: usize = ELEM_LEN;
```
briansmith: Regarding your issue comment regarding `PRIVATE_KEY_LEN` vs `SCALAR_LEN` vs `SEED_LEN`, I have two thoughts:

First, we could get rid of the idea of “private key” (and `PRIVATE_KEY_LEN`) and just have “seed” and “scalar”. If you think about it, the seed is the private key in terms of storage, but `a` is the private key in terms of ECC math, and `z` is another non-ECC private key component.

Second, note that currently we store the seed and the public key in the key pair. But, it would actually be more efficient to store `(a, z)` and the public key in the key pair, by transforming the seed to `(a, z)` during the construction of the key. Further note that we always mask `a` and we can do the masking of `a` during key construction and avoid having to do it every time it is used.

briansmith:
Also, I landed a few commits from the beginning of the PR, with some of the `LEN` stuff changed according to how I commented before.

briansmith:
I realized that I had set a bad example with one of the earlier commits that *I* wrote earlier. I tried to rectify this in 9f7d94da1feaecb41fe330322ff465073de90905 and b5922777d3df7671fc42fd4da8799bc3070237da. Note that the former commit attempts to address your concern regarding the `PRIVATE_KEY_LEN` vs. `SCALAR_LEN` vs. `SEED_LEN` confusion.

src/ec/eddsa.rs
ranweiler: > First, we could get rid of the idea of “private key” (and PRIVATE_KEY_LEN) and just have “seed” and “scalar”. 

I really like how this shakes out in 9f7d94da1feaecb41fe330322ff465073de90905 and b5922777d3df7671fc42fd4da8799bc3070237da.

> But, it would actually be more efficient to store (a, z) and the public key in the key pair, [...]

I like this idea. I'll experiment with it as I respond to the other feedback.

src/ec/eddsa.rs
```diff
@@ -80,42 +77,86 @@ impl<'a> Ed25519KeyPair {
...
             return Err(error::Unspecified);
         }
-        if public_key.len() != 32 {
+        if public_key.len() != PUBLIC_KEY_LEN {
             return Err(error::Unspecified);
         }
-        let mut pair = Ed25519KeyPair { private_public: [0; 64] };
-        for i in 0..32 {
+        let mut pair = Ed25519KeyPair { private_public: [0; KEY_PAIR_LEN] };
+        for i in 0..SCALAR_LEN {
```
ranweiler: Captured in b5922777d3df7671fc42fd4da8799bc3070237da.

src/ec/eddsa.rs
```diff
@@ -133,33 +174,144 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
...
+        let mut r = Point::new_at_infinity();
+        unsafe {
+            GFp_ge_double_scalarmult_vartime(&mut r,
+                                             slice_as_scalar_ref(&h).unwrap(),
+                                             &a,
+                                             slice_as_scalar_ref(&s_copy).unwrap())
+        };
+        let mut r_check = [0u8; ELEM_LEN];
+        unsafe { GFp_x25519_ge_tobytes(&mut r_check, &r) };
+        constant_time::verify_slices_are_equal(&r_check, &r_copy)
```
ranweiler: Good point— I'll benchmark this against a normal slice comparison, which would be even better.

src/ec/eddsa.rs
ranweiler: Addressed in 9f7d94da1feaecb41fe330322ff465073de90905, b5922777d3df7671fc42fd4da8799bc3070237da.

src/ec/eddsa.rs
briansmith: I think you don't need to benchmark it. Just switch to the normal slice comparison. I just noticed today we're already using the normal slice comparison in `EdDSAKeyPair::from_bytes()`.

src/ec/eddsa.rs
```diff
@@ -109,6 +109,11 @@ impl<'a> Ed25519KeyPair {
         Ok(pair)
     }
 
+    // Returns a reference to the little-endian-encoded private key bytes.
+    fn private_key_bytes(&'a self) -> &'a [u8] {
+        &self.private_public[..SEED_LEN]
```
briansmith: When working on the ECDSA signing stuff, I realized that eventually we should switch to `ec::PrivateKey` for holding the private scalar part of the key pair.

This code has the `private_public` member in the first place only because the original C code required the concatenated seed||public_key as an input. That's no longer the case after this PR.

So, keep in mind that in the future `private_public` will get split into multiple members, once of which is an `ec::PrivateKey`.

ranweiler: Noted, I'll make sure we're anticipating that.

src/ec/eddsa.rs
```diff
@@ -245,6 +245,24 @@ impl ExtPoint {
...
+// Keep this in sync with `ge_p2` in curve25519/internal.h.
+#[repr(C)]
+struct Point {
+    x: Elem,
+    y: Elem,
+    z: Elem,
+}
+
+impl Point {
+    fn new_at_infinity() -> Self {
```
ranweiler: Of the current commits in this PR, this is the only one that won't build, because `Point` is dead code. So, let me know if you'd rather fold this into the next (verification) commit.

src/ec/eddsa.rs
ranweiler: Done.

src/ec/eddsa.rs
```diff
@@ -80,42 +77,86 @@ impl<'a> Ed25519KeyPair {
...
     /// Returns the signature of the message `msg`.
     pub fn sign(&self, msg: &[u8]) -> signature::Signature {
-        let mut signature_bytes = [0u8; 64];
+        let mut signature_bytes = [0u8; SIGNATURE_LEN];
+        let mut az = [0u8; digest::SHA512_OUTPUT_LEN];
+        az.copy_from_slice(digest::digest(&digest::SHA512,
+                                          &self.private_key_bytes()).as_ref());
+        unsafe {
+            GFp_ed25519_scalar_mask(slice_as_scalar_ref_mut(&mut az).unwrap())
+        };
```
ranweiler: Done. This pattern ended up being _much_ nicer.

src/ec/eddsa.rs
```diff
@@ -80,42 +77,86 @@ impl<'a> Ed25519KeyPair {
...
+        let mut nonce = [0u8; digest::SHA512_OUTPUT_LEN];
+        nonce.copy_from_slice({
+            let mut ctx = digest::Context::new(&digest::SHA512);
+            ctx.update(&az[SCALAR_LEN..]);
+            ctx.update(&msg);
+            ctx.finish().as_ref()
+        });
+        unsafe {
+            GFp_x25519_sc_reduce(slice_as_scalar_ref_mut(&mut nonce).unwrap())
+        };
```
ranweiler: This was just wrong, because I misinterpreted `_sc_reduce`. Now we just pass `&mut nonce`, where `nonce` itself is an array of appropriate size (64 bytes).

src/ec/eddsa.rs
```diff
@@ -133,33 +174,144 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
...
-                                       in_: *const u8/*[32]*/);
+    fn GFp_ed25519_scalar_mask(a: &mut Scalar);
+    fn GFp_ge_double_scalarmult_vartime(r: &mut Point, a_coeff: &Scalar,
+                                        a: &ExtPoint, b_coeff: &Scalar);
+    fn GFp_ge_p3_tobytes(s: &mut [u8; ELEM_LEN], h: &ExtPoint);
+    fn GFp_x25519_ge_frombytes_vartime(h: &mut ExtPoint, s: &Scalar) -> c::int;
+    fn GFp_x25519_ge_scalarmult_base(h: &mut ExtPoint, a: &Seed);
+    fn GFp_x25519_ge_tobytes(s: &mut Scalar, h: &Point);
+    fn GFp_x25519_sc_muladd(s: &mut Scalar, a: &Scalar, b: &Scalar, c: &Scalar);
+    fn GFp_x25519_sc_reduce(s: &mut Scalar);
```
ranweiler: Fixed.

ranweiler:
@briansmith, I just did a _whole_ lot of rebasing, and now the intermediate `sign` and `verify` port commits reflect your initial feedback, and are in a much less egregiously un-idiomatic state.

I'm going to respond to the remaining feedback tomorrow evening. Let me know if you'd prefer that I fold tomorrow's changes back into the relevant `sign`/`verify` port commits for the sake of easier review. Otherwise, I'll just append them as usual.

briansmith:
I suggest combining all the signing PRs and all the verification PRs together, except split them enough that the “Use idiomatic return value checking instead of custom handling” commit is on its own. Basically, imagine that each commit would land on its own, in order. If you want to split them up more than I suggest, that's fine with me, as long as each one builds.

briansmith:
Two high-level comments:
1. Should we say the point (0, 0, 0, 0) is at infinity, or that it is the identity point? I noticed EdDSA stuff seems to use “identity”.
2. Is there a better name than `ExtPoint` for the 4-tuple points? “Ext” seems quite vague, but I haven't read the X25519/EdDSA stuff in a while, so I can't remember any better names.

ranweiler:
> Should we say the point (0, 0, 0, 0) is at infinity, or that it is the identity point? I noticed EdDSA stuff seems to use “identity”.

Hmm, actually, is there an implementation convention I'm not aware of? AFAIK, (0, ..., 0) does not ever denote a point in projective space, which excludes it by definition. The identity/point at infinity, though, _does_ live in projective space, and has concrete representations. In the different coordinate systems for Twisted Edwards curves, the coordinates of the identity/point at infinity are nonzero field element multiples of:
- Projective:`(X, Y ,Z) = (0, 1, 0)`
- Extended: `(X, Y, Z, T) = (0, 1, 1, 0)`

See [here](https://www.iacr.org/archive/asiacrypt2008/53500329/53500329.pdf).

If this is correct, then we should actually rename the `suite_b::ops::Point` constructor too, right?

So, _is_ there an encoding convention/computational shortcut being used here that I'm unaware of? If so, should we call it out in a comment?

Assuming there is some convention that I'm ignorant of, then I lean towards `new_at_infinity`. I think this is why: being "at infinity" describes a point in projective space generally. Then, for the subset of points which meet a precondition (they are on the curve) we _use_ the desired point _as_ an "identity" when we go one step further and endow the curve with an algebraic structure.

IMHO, it would feel most appropriate to use `new_at_identity` for a struct or trait whose name referenced its group interpretation more explicitly. Though, with elliptic curves (and most groups we'd compute with), we're constantly slurring the distinction, so either seems fine. Certainly in this context overall, we don't care about points that aren't on the curve, and thus in the group, so I could go either way.

> Is there a better name than ExtPoint for the 4-tuple points? “Ext” seems quite vague, but I haven't read the X25519/EdDSA stuff in a while, so I can't remember any better names.

I think the most specific name for such coordinates is "Extended Twisted Edwards Coordinates" (see the above paper). I'm not sure if it is worth cranking up the verbosity very much. The only viable alternative I can think of is "ExtendedPoint".

briansmith:
Thanks for the detailed explanation. It matches my understanding.
> If this is correct, then we should actually rename the suite_b::ops::Point constructor too, right?

In that code, we consider every triple (x, y, z) where z == 0 to be at infinity and an (additive) group identity element. I think this is basically what you're saying, that we're using it as *an* identity.

Given what you said, I think the names we have now are OK.

src/ec/eddsa.rs
ranweiler: I squashed this.

src/ec/eddsa.rs
```diff
@@ -80,42 +77,86 @@ impl<'a> Ed25519KeyPair {
...
     pub fn sign(&self, msg: &[u8]) -> signature::Signature {
-        let mut signature_bytes = [0u8; 64];
+        let mut signature_bytes = [0u8; SIGNATURE_LEN];
+        let mut az = [0u8; digest::SHA512_OUTPUT_LEN];
+        az.copy_from_slice(digest::digest(&digest::SHA512,
+                                          &self.private_key_bytes()).as_ref());
+        unsafe {
+            GFp_ed25519_scalar_mask(slice_as_scalar_ref_mut(&mut az).unwrap())
+        };
+        let mut nonce = [0u8; digest::SHA512_OUTPUT_LEN];
```
ranweiler: I didn't rename this. Instead, here and elsewhere, I repeatedly used a pattern where we start with some initial datum, get a large (often 64-byte) array-oriented view of it, do some stuff to initialize the large array, then rebind the name with an array reference to one half of the original array.

On one hand, I dislike shadowing/rebinding names. On the other hand, this lets us hide the original sense of the data from later code, without us having to e.g. introduce a block for just this purpose. It also made it much easier to introduce new names for array halves and make fewer calls to `slice_as_array_ref*!`, which massively cleaned up the unsafe FFI blocks.

src/ec/eddsa.rs
```diff
@@ -80,42 +77,86 @@ impl<'a> Ed25519KeyPair {
...
+        };
+        let mut r = ExtPoint::new_at_infinity();
+        unsafe {
+            GFp_x25519_ge_scalarmult_base(&mut r, slice_as_scalar_ref_mut(&mut nonce).unwrap());
+            GFp_ge_p3_tobytes(slice_as_scalar_ref_mut(&mut signature_bytes).unwrap(), &r);
+        }
+        let mut hram = [0u8; digest::SHA512_OUTPUT_LEN];
+        hram.copy_from_slice({
+            let mut ctx = digest::Context::new(&digest::SHA512);
+            ctx.update(&signature_bytes[..SCALAR_LEN]);
```
ranweiler: I made this more explicit by using `split_at_mut` and the `array_ref` macros to get refs to the signature halves named `signature_r` and `_s`.

src/ec/eddsa.rs
```diff
@@ -80,42 +77,86 @@ impl<'a> Ed25519KeyPair {
...
+            ctx.update(&self.public_key_bytes());
+            ctx.update(&msg);
```
ranweiler: Fixed throughout.

src/ec/eddsa.rs
```diff
@@ -80,42 +77,86 @@ impl<'a> Ed25519KeyPair {
...
+            ctx.finish().as_ref()
+        });
         unsafe {
-            GFp_ed25519_sign(signature_bytes.as_mut_ptr(), msg.as_ptr(),
-                             msg.len(), self.private_public.as_ptr());
+            GFp_x25519_sc_reduce(slice_as_scalar_ref_mut(&mut hram).unwrap());
+            GFp_x25519_sc_muladd(slice_as_scalar_ref_mut(&mut signature_bytes[SCALAR_LEN..]).unwrap(),
+                                 slice_as_scalar_ref(&hram).unwrap(),
+                                 slice_as_scalar_ref(&az).unwrap(),
```
ranweiler: Done.

src/ec/eddsa.rs
ranweiler: Done.

src/ec/eddsa.rs
```diff
@@ -133,33 +174,144 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
...
+    }
+    let mut a = ExtPoint::new_at_infinity();
+    unsafe {
+        GFp_x25519_ge_scalarmult_base(&mut a, &a_bytes);
+        GFp_ge_p3_tobytes(out, &a);
+    }
+}
+
+fn slice_as_scalar_ref(slice: &[u8]) -> Result<&Scalar, error::Unspecified> {
+    slice_as_array_ref!(&slice[..SCALAR_LEN], SCALAR_LEN)
```
ranweiler: Done. I've removed all use of the new functions in `sign` and `verify`, and will remove the final occurrence in `public_from_private` shortly.

src/ec/eddsa.rs
```diff
@@ -80,42 +77,86 @@ impl<'a> Ed25519KeyPair {
...
+            let mut ctx = digest::Context::new(&digest::SHA512);
+            ctx.update(&signature_bytes[..SCALAR_LEN]);
+            ctx.update(&self.public_key_bytes());
+            ctx.update(&msg);
+            ctx.finish().as_ref()
+        });
         unsafe {
-            GFp_ed25519_sign(signature_bytes.as_mut_ptr(), msg.as_ptr(),
-                             msg.len(), self.private_public.as_ptr());
+            GFp_x25519_sc_reduce(slice_as_scalar_ref_mut(&mut hram).unwrap());
```
ranweiler: This worked well! I feel like the name could be improved, but I stuck with the one you suggested to be consistent with the other code.

src/ec/eddsa.rs
```diff
@@ -133,33 +174,144 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
...
+        }));
+        for i in 0..ELEM_LIMBS {
+            a.x[i] = -a.x[i];
+            a.t[i] = -a.t[i];
+        }
+        let mut r_copy = [0u8; SCALAR_LEN];
+        r_copy.copy_from_slice(&signature[..SCALAR_LEN]);
+        let mut s_copy = [0u8; SCALAR_LEN];
+        s_copy.copy_from_slice(&signature[SCALAR_LEN..]);
+        let mut h = [0u8; digest::SHA512_OUTPUT_LEN];
```
ranweiler: Done.

src/ec/eddsa.rs
ranweiler: Did some prep work for this in 397de12b4db1dda7924d1c766524eaef463993d3.

src/polyfill.rs
```diff
@@ -178,3 +178,20 @@ macro_rules! slice_as_array_ref_mut {
         }
     }
 }
+
+/// Converts a boolean into a `Result<(), err::Unspecified>` in the
+/// obvious way. By composing this with `try!`, we can compactly
+/// represent `Result`-valued early returns that guard on a condition.
+macro_rules! check {
```
briansmith: This doesn't need to be a macro, does it? I think it can be an inline function.

IMO it would be better to do this as a separate PR that replaced`if x { return Err(...); }` throughout the whole codebase so we can better see the full impact it would have.

ranweiler: > This doesn't need to be a macro, does it? I think it can be an inline function.

It doesn't— it only started life as a macro so I could do an early return, then later realized it made more sense to just compose it with `try!`.

Looking at it again, I don't think `polyfill` is the best module for this. Maybe `error`, instead?

> IMO it would be better to do this as a separate PR

Yes, that makes sense to me. Will try that tonight.

briansmith:
@ranweiler Check out https://github.com/briansmith/ring/commits/start-ed25519-port, in particular the commits I added. This is how I'd prefer to merge it, if it looks good to you.

ranweiler:
@briansmith your branch looks good to me, with two nits: the main commit messages for 7d9dd8fdba42c2cb18b383b274d146d585219fcd and 652666f678a54b0bf67564ee6d42244383009217 have typos (`ED`/`Ed`, and `EdDSAKeyPair` isn't the type name).

I have some other local WIP, based on your suggestions, but we can close out this PR and deal with that separately. The local WIP is about caching the secret scalar and prefix (using the terms in RFC 8032, `a` and `z` in existing code). Let's open a new issue for that optimization?

Also, curious what you think about 397de12b4db1dda7924d1c766524eaef463993d3.

briansmith:
OK, I landed these:

33fa6de736183d3670af8448cde695a636232118 Rename 4-coordinate `Point` to `ExtPoint`.
3814df2a0db764dec8e7c304209d48b2f117a14c Add `new_at_infinity` constructor to `ExtPoint`.
470896abc1f9e7590e6e3dcda47ffac90b2763fd Port Ed25519 signing to Rust.
1a5f57bb348958c121c9b438d0f41d989dedccdf Prepare C functions used in verification for Rust usage.
5f53f18d8ccad83f9ce9cb8152426d358b1c27b9 Port Ed25519 verification to Rust.
9f021136d9739e925dece0cd1366ad839fbbe1b3 Use idiomatic return value checking instead of custom handling.
a2572916a19ba735714bb54e847ea2b8d92ae2d5 Fix some stylistic nits in `Ed25519KeyPair::sign()`.
ee2a47ea26688ed492ee459da938a37ec94b8596 Fix some stylistic nits in `ED25519`'s `verify()`.
1e7be53f58c0b70d48ca7bdb1d780f3dc18e81b0 Factor out Twisted Edwards extended point inversion.
0aca3336e4126b786e3d1b743a631e56a7acdae8 Rename `ExtPoint::invert()` to `ExtPoint::invert_vartime()`.
4d8baa9bbb7b75c74080661aa1eadf22349d199c Tweak formatting in `eddsa_digest()`.

Let's do the rest in another PR.

Good work on this! Thanks for submitting these!

Issue #473, opened 2017-03-06
Switch from GCC to clang for android builds


briansmith:
Currently we use `arm-linux-androideabi-gcc` and `arm-linux-androideabi-g++` for Android builds. According to [the NDK 14 changelog](https://github.com/android-ndk/ndk/wiki/Changelog-r14), support for GCC is deprecated in the NDK. So, let's switch to clang.

briansmith:
Fixed by #480.

briansmith:
Reopining because I rolled this back in 63691d4245866c6081b9cc89aa5a0a94f90532ae since switching to clang seems to have broke things. Thanks to @pietro for pointing out the bustage.

pietro:
I got AArch64 working when using clang by defining `BORINGSSL_CLANG_SUPPORTS_DOT_ARCH` and passing the flag `-fno-integrated-as`.

I haven't found the right flags to get 32 bit arm working with clang.

briansmith:
For the record, this was done in 72d3b9d3ba8dcab5da235b10fc67c69b33ff3591 (PR #652) back on May 18, 2018.

PR #480, opened 2017-03-09
Use clang on android.

pietro:
I stopped defining `__ANDROID_API__` which means we drop support for gcc on android. This is a PR for issue #473.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -547,11 +547,8 @@ fn cc(file: &str, ext: &str, target: &Target, out_dir: &Path) -> Command {
         // http://www.openwall.com/lists/musl/2015/06/17/1
         let _ = c.flag("-U_FORTIFY_SOURCE");
     }
-    if target.os() == "android" {
-        // Define __ANDROID_API__ to the Android API level we want.
-        // Needed for Android NDK Unified Headers, see:
-        // https://android.googlesource.com/platform/ndk/+/master/docs/UnifiedHeaders.md#Supporting-Unified-Headers-in-Your-Build-System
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -547,11 +547,9 @@ fn cc(file: &str, ext: &str, target: &Target, out_dir: &Path) -> Command {
         // http://www.openwall.com/lists/musl/2015/06/17/1
         let _ = c.flag("-U_FORTIFY_SOURCE");
     }
-    if target.os() == "android" {
-        // Define __ANDROID_API__ to the Android API level we want.
-        // Needed for Android NDK Unified Headers, see:
-        // https://android.googlesource.com/platform/ndk/+/master/docs/UnifiedHeaders.md#Supporting-Unified-Headers-in-Your-Build-System
-        let _ = c.define("__ANDROID_API__", Some("18"));
+    if target.os() == "android" && target.arch() == "arm" {
+            let _ = c.flag("-march=armv7-a");
```
pietro: `crypto/curve25519/asm/x25519-asm-arm.S` fails to compile if I don't pass this flag to clang. It looks like it needs to be done anyway:  https://developer.android.com/ndk/guides/standalone_toolchain.html#working_with_clang

briansmith:
This looks reasonable to me.

briansmith:
Thank you!

briansmith:
I rolled this back in 63691d4245866c6081b9cc89aa5a0a94f90532ae since switching to clang seems to have broke things. Thanks to @pietro for pointing out the bustage.

PR #483, opened 2017-03-10
Fix Android NDK caching. This is a directory.

pietro:
I just realized every build was reinstalling the ndk.


briansmith:
This landed as  2ca8cba8795091c2bcd6945f99b2036807a6883b. Thanks!

briansmith:
Reverted in 39982f81c761595b57945cd9b75181b687251e4f because I had thought this change was part of the switch to clang on Android. Re-landed in e6f97fa1f543fed1059c40e5a714edd3769336c5 after realizing that wasn't the case.

PR #484, opened 2017-03-10
Use travis_wait only on ARM release builds.

pietro:
The other builds don't seem to timeout. This way we have faster/better feedback for the arm builds.


briansmith:
This landed as 7a20a09745b0d5153a31be028f0e481915d6e99d. Thanks for doing this!

PR #487, opened 2017-03-14
Remove option that breaks newer android emulators.

pietro:
This gets the android builds running on the emulator and shows that the switch to clang unfortunately broke ring on android. I won't have time until later in the week to try to find the cause. I tried passing a lot of different flags to clang to try to fix this but I've had no luck so far.


briansmith:
Thanks! Merged in 70ca967455cb313931d7c4d8d53d23d0a7fa9c3d.

PR #488, opened 2017-03-16
Use armv7 for android build on travis.

pietro:
I noticed that just by changing the target these flags, `-march=armv7-a` and `mfpu=vfpv3-d16`, get passed to gcc. According to [this](https://developer.android.com/ndk/guides/standalone_toolchain.html#abi_compatibility) we should also pass `-mfloat-abi=softfp`. We may also want to pass `-thumb` and `-mfpu=neon` and lean on on the runtime detection for arm cpu features. There's also the recommended flags to the linker that we may want to pass to both gcc and rustc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ matrix:
       os: osx
       osx_image: xcode8.2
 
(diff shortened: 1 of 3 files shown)


briansmith:
Thanks! I landed this in master as  d048e1e81847cc555388269c6c3063fc19994624. I also filed #490 to track your other suggestions.

Issue #461, opened 2017-02-18
Improve consistency of `Algorithm` attributes & accessors


briansmith:
`ring::digest::Algorithm` uses public attributes, not functions, to expose algorithm properties like `output_len`, `chaining_len`, etc. But, `ring::aead::Algorithm` uses accessor functions like `tag_len()` instead. This is confusing. The original motivation for doing things the current way might no longer be valid.

ocrickard:
What's the status of these commits?

Zeromika:
What's the status of this issue, I am interested in picking it up.

briansmith:
0.17.0 addressed this for `digest::Algorithm`.

PR #491, opened 2017-03-19
Replace digest Algorithm attribute accesses with getter methods. #461

46bit:
I've made the attributes private, added attribute getters for `output_len`, `chaining_len` and `block_len`, and updated all uses throughout the codebase. I've also implemented a test that the values are correct, and checked it works.

> The original motivation for doing things the current way might no longer be valid.

At present nothing in the codebase tries to set these parameters. Beyond that I don't have enough context to comment.

### DIFF
--- a/src/digest/digest.rs
+++ b/src/digest/digest.rs
@@ -274,7 +274,7 @@ impl core::fmt::Debug for Digest {
 /// C analog: `EVP_MD`
 pub struct Algorithm {
     /// C analog: `EVP_MD_size`
-    pub output_len: usize,
+    output_len: usize,
 
     /// The size of the chaining value of the digest function, in bytes. For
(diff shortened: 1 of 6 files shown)


briansmith:
Thanks for the PR. I think we probably want to switch the other way: make the AEAD API more like the digest API, replacing functions with attributes. In particular, I'm hoping that eventually we'll be able to use expressions like `aead::AES_128_GCM.tag_len` as constant expressions, e.g. to fix array lengths: `[u8; aead::AES_128_GCM.tag_len]`.

WDYT?

46bit:
That makes a lot of sense - the array lengths will be very useful to have. I don't know if I'll get to implement that anytime soon; things are rather busy on my end. I'll close this for now and try to get it done sometime soon.

PR #495, opened 2017-03-25
allow implementing PBKDF2 for other digest algorithms

Geal:
To implement authentication in systems like [MongoDB](https://docs.mongodb.com/manual/core/security-scram-sha-1/), we need other variants of PBKDF2, like HMAC-SHA1.

Since you might not want to commit the API to provide weaker variants by default, I thought exposing the digest algorithm member of PRF would allow it discreetly if someone needs it.

### DIFF
--- a/src/pbkdf2.rs
+++ b/src/pbkdf2.rs
@@ -254,7 +254,8 @@ pub fn verify(prf: &'static PRF, iterations: u32, salt: &[u8],
 
 /// A PRF algorithm for use with `derive` and `verify`.
 pub struct PRF {
-    digest_alg: &'static digest::Algorithm,
+    /// digest algorithm usable as a PBKDF2 PRF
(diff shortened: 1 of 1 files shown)


briansmith:
Thanks for the PR.

If we want to support PBKDF2-SHA1 then we should do so using the interface we already had. We removed the PBKDF2-SHA1 interface because we replaced the fast SHA1 code with slower, simpler, smaller code. PBKDF2 needs to be fast so that you can maximize the number of rounds. For example, iOS 10 does 10 million (yes, ten million) rounds of PBKDF2 when unlocking the phone. IIUC, such limits would be way too slow for our current SHA-1 implementation. Do we want to bring back the fast SHA-1 implementations? I'm not sure.

Regardless, I'm going to close this PR because we're definitely not going to do it this way. Thanks again, though!

PR #496, opened 2017-03-29
Create Android AVD after test build and use more RAM on the emulator.

pietro:
This way we'll get to build issues faster.

### DIFF
--- a/mk/travis.sh
+++ b/mk/travis.sh
@@ -88,7 +88,9 @@ fi
 case $TARGET_X in
 armv7-linux-androideabi)
   cargo test -vv -j2 --no-run ${mode-} ${FEATURES_X-} --target=$TARGET_X
-  emulator @arm-18 -no-skin -no-boot-anim -no-window &
+  echo no | android create avd --name arm-18 --target android-18 --abi armeabi-v7a
+  android list avd
+  emulator @arm-18 -memory 2048 -no-skin -no-boot-anim -no-window &
   adb wait-for-device
   adb push $target_dir/ring-* /data/ring-test
   for testfile in `find src crypto -name "*_test*.txt"`; do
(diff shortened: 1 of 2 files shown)


mk/travis.sh
```diff
@@ -88,7 +88,9 @@ fi
 case $TARGET_X in
 armv7-linux-androideabi)
   cargo test -vv -j2 --no-run ${mode-} ${FEATURES_X-} --target=$TARGET_X
```
briansmith: Please add a comment about why we're doing this here.

Also, would it be possible/beneficial for us to create the AVD once, have Travis CI cache it, and then make a copy of the cached AVD that we then run the tests on?

pietro: I'll add the comment.

I thought about caching the AVD but that would slow down all builds. And it will get worse when we add AArch64 Android to Travis.

briansmith:
Thanks for pointing out the potential issue with caching the AVD. I landed your change as 4a1360853697a6d34d0c569a682aa8f5d4e980b4 in master and I added efb4b0e1ee3bb2ff4dfdc01f1fe42a76c0f26422 to document why we do things in this order.

Thanks a lot!

pietro:
I think I'm wrong about the AVD caching issue. I think it would be ok to cache `$HOME/.android` or at least `$HOME/.android/avd` and check if `/home/travis/.android/avd/arm-18.avd` exists before creating the AVD for ARMv7a.

According to Travis CI caching [docs](https://docs.travis-ci.com/user/caching/#Caches-and-build-matrices) the ENV vars we export on each build is taken in consideration when creating/using the cache so non-android builds won't have the AVD cache.

briansmith:
@pietro Thanks for pointing that out. I filed #518 to track that so we don't forget.

PR #497, opened 2017-03-29
Allow C++ tests to be disabled

steveatinfincia:
This allows the C++ tests to be optionally disabled but still run by default, as mentioned in #477. 

Fixes build issues on musl target (like #494), though I have only tested `x86_64-unknown-linux-musl` and not `i686-unknown-linux-musl`.

Locally this seems to work as intended, g++ is not called, no c++ std lib is linked, and the test does not run:

```
$ cargo test --no-default-features --features use_heap
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running target/x86_64-unknown-linux-musl/debug/deps/ring-c6fb18900b937232

running 112 tests
<snip>
```

Default still runs the additional test:

```
$ cargo test                                          
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running target/debug/deps/ring-73d758db34b545a4

running 113 tests
<snip>
test tests::test_bn ... ok
```


briansmith:
@steveatinfincia for your use case, do you use RSA signing? I noticed that all the things that bn_test.cc still tests are only used for RSA signing. So another approach might be to build/run bn_test only when the rsa_signing feature is enabled. WDYT?

steveatinfincia:
@briansmith I'm using ring downstream in rustls, so I suppose the answer is yes, as [they enable](https://github.com/ctz/rustls/blob/master/Cargo.toml#L17) the rsa_signing feature.

steveatinfincia:
@briansmith that being the case, if rsa_signing is off by default, perhaps bn_tests can be as well?

briansmith:
> @briansmith that being the case, if rsa_signing is off by default, perhaps bn_tests can be as well?

Yes, I think that would be an improvement.

Cargo.toml
```diff
@@ -287,6 +287,7 @@ rsa_signing = []
 slow_tests = []
 test_logging = []
 use_heap = []
+bn_tests = []
```
briansmith: The way this ended up, `bn_tests` won't be enabled unless somebody specifically asks for it. That's not what we want, right?

I suggest that we do this instead:
1. Make the `rsa_signing` feature depend on the `bn_tests` feature. That way, whenever RSA signing is enabled, the `bn_tests` will be built and run. But when RSA signing isn't enabled, the `bn_tests` will be skipped.

2. You can see that we already have disabled bn_tests for macOS/iOS targets for similar reasons (trouble with C++) even when RSA signing is enabled. I suggest we add `target_env = "musl"` to the list of conditions where the bn_tests are skipped.

WDYT?

steveatinfincia: Yea you're right, that did seem like a breaking change to the intent at least, I'll fix it :)

briansmith:
I squashed your commits and landed them as f674a7e9454d2e2f8ef040343d3fd2fffe121f98, which will be part of *ring* 0.7.4.

Thanks!

PR #499, opened 2017-04-01
Fix windows gnu build.

crlf0710:
Hi, i'm using windows gnu x64 rust build. I'm trying to build Rocket.rs but found that `ring` does not compile. So i downloaded it and fixed it. Now all the `ring` tests pass (one ignored?) with `cargo test`. I think this might be useful to others.

I haven't investigated on the "precompiled" conditional branch though. I'm not really sure it's working. Will take another PR if i found another problem.

### DIFF
--- a/crypto/cpu-intel.c
+++ b/crypto/cpu-intel.c
@@ -62,10 +62,14 @@
 #include <inttypes.h>
 
 #if defined(OPENSSL_WINDOWS)
+#ifdef _MSC_VER
 #pragma warning(push, 3)
+#endif
 #include <immintrin.h>
 #include <intrin.h>
+#ifdef _MSC_VER
 #pragma warning(pop)
+#endif
 #endif
 
 #include "internal.h"
--- a/crypto/curve25519/asm/x25519-asm-x86_64.S
+++ b/crypto/curve25519/asm/x25519-asm-x86_64.S
@@ -27,6 +27,10 @@
 /* OS X's C ABI prefixes functions with underscore. */
 #define C_ABI(x) _ ## x
 #define HIDDEN .private_extern
+#elif defined(__MINGW32__) || defined(__MINGW64__)
+/* MinGW build support */
+#define C_ABI(x) x
+#define HIDDEN .globl
 #else
 #define C_ABI(x) x
 #define HIDDEN .hidden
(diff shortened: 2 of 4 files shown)


crypto/cpu-intel.c
```diff
@@ -62,11 +62,15 @@
 #include <inttypes.h>
 
 #if defined(OPENSSL_WINDOWS)
+#ifdef _MSC_VER
```
briansmith: NIT: we always use `#if defined(...)` in the C code. This is a convention from BoringSSL.

I'll fix this for you.

crypto/curve25519/asm/x25519-asm-x86_64.S
```diff
@@ -27,6 +27,10 @@
 /* OS X's C ABI prefixes functions with underscore. */
 #define C_ABI(x) _ ## x
 #define HIDDEN .private_extern
+#elif defined(__MINGW32__) || defined(__MINGW64__)
+/* MinGW build support */
+#define C_ABI(x) x
+#define HIDDEN .globl
```
briansmith: I think this should be just `define HIDDEN`, based on what perlasm does. I'll make this change. Please test it and let me know if it breaks things.

briansmith:
Landed in master with the changes I mentioned above in a004549b970eab4b26a53ecbb3a86d7772e13591. Thanks a lot for this!

crypto/curve25519/asm/x25519-asm-x86_64.S
crlf0710: @briansmith Hi, no. I tried it and if we just #define the HIDDEN macro into nothing then  it doesn't compile.

Looking in another win32 port project i found that they are defining HIDDEN into .globl too. So i wrote it this way.
https://github.com/git-for-windows/MINGW-packages/blob/2a4f9271d7b17498b46b73c11f249ca56b93dfc3/mingw-w64-clang/0061-libunwind-add-support-for-mingw-w64.patch#L92

I don't have any better idea yet.

crypto/cpu-intel.c
crlf0710: Great, thanks.

briansmith:
OK, thanks for letting me know. I misunderstood how the macro was used in the code below. I fixed it in 3aed44eeaace0294b7559a1f298493478b7f75ef. Thanks again!

Issue #502, opened 2017-04-04
"ignoring unknown option '/utf-8'" causing the build to fail


r3c0d3x:
`cargo` is attempting to build `ring 0.7.3` as a dependency for a crate I'm using, but `build.rs` appears to be [misinterpreting](https://github.com/briansmith/ring/blob/8a46eb57b861973c934f6df034ac9c4c24e425c5/build.rs#L712) a warning from the compiler as an actual error (because it outputs warnings using stderr).

```
--- stderr
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
thread '<unnamed>' panicked at 'execution failed', C:\Users\r3c0d3x\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.7.3\build.rs:712
note: Run with `RUST_BACKTRACE=1` for a backtrace.
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
cl : Command line warning D9002 : ignoring unknown option '/utf-8'
```

It appears to be caused by [this](https://github.com/briansmith/ring/blob/8a46eb57b861973c934f6df034ac9c4c24e425c5/build.rs#L291). I'm using whatever build tools are provided with Visual Studio 2015 and it apparently doesn't need the command line option, but I get that older toolchains might.

briansmith:
I think you may not be using the newest version of the MSVC 2015 build tools. Can you do `cl /version`? And/or, try removing your MSVC 2015 build tools and then downloading and installing the newest (pre-2017) build.

[...]

PR #503, opened 2017-04-06
remove non-ascii-character for msvc compilation

itn3000:
relate to #502


briansmith:
Thanks! Not sure why my earlier search didn't find these. I merged this, with very minor changes, as 2e499aa9167d78319b1fb8d7993b313a667a86ff.

PR #504, opened 2017-04-07
Update to rayon 0.7

cuviper:

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -72,8 +72,8 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::fs::{self, DirEntry};
 use std::time::SystemTime;
-use rayon::par_iter::{ParallelIterator, IntoParallelIterator,
-                      IntoParallelRefIterator};
+use rayon::iter::{ParallelIterator, IndexedParallelIterator,
+                  IntoParallelIterator, IntoParallelRefIterator};
 
 const X86: &'static str = "x86";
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -429,7 +429,7 @@ impl Target {
 
 fn build_c_code(target: &Target, out_dir: &Path) {
     let includes_modified = RING_INCLUDES.par_iter()
-        .weight_max()
+        .with_max_len(1)
```
briansmith: Could you give me a pointer to the documentation for this? It's not clear to me that these are the same thing or how they are different. Thanks!

cuviper: `weight_max()` was the same as `weight(INFINITY)`, which effectively meant to split the jobs into the smallest possible size -- a single item.

[`with_max_len(n)`](https://docs.rs/rayon/0.7.0/rayon/iter/trait.IndexedParallelIterator.html#method.with_max_len) basically means to process no more than `n` items in each job.  So `with_max_len(1)` splits it completely, same as `weight_max` did.

In general, we hope that `with_min_len`/`with_max_len` are more intuitive than the rather arbitrary floating point "weight" thresholds were.  `INFINITY` is easy, but for something like `1000.0` it was hard to figure out what rayon would do.

briansmith:
Thanks!

landed in 9f0e0286b375b4382cb226c2e440be82887c8835. Will be in *ring* 0.7.4.

PR #505, opened 2017-04-09
Add utility to map boolean checks to `Result<(), error::Unspecified>`

ranweiler:
Picking up from be4e25ad7b43d27e37082fc84b95e042535c9a44.

I'm experimenting a bit with whether this should be a function that simply does the mapping from `bool` to `Result<(), error::Unspecified>`, or if it should actually be a _macro_ that always performs an early return itself. The latter might result in nicer looking code, but I don't want to bikeshed it too much, so I'm starting with the function implementation, as it seems a bit more idiomatic to use `try!`.

FYI, for now, I'm going to push a bunch of separate commits that I expect to be squashed/dropped.

### DIFF
--- a/src/constant_time.rs
+++ b/src/constant_time.rs
@@ -22,9 +22,7 @@ use {c, error};
 /// `a` and `b`.
 pub fn verify_slices_are_equal(a: &[u8], b: &[u8])
                                -> Result<(), error::Unspecified> {
-    if a.len() != b.len() {
-        return Err(error::Unspecified);
-    }
+    try!(error::check(a.len() == b.len()));
     let result = unsafe { GFp_memcmp(a.as_ptr(), b.as_ptr(), a.len()) };
     match result {
         0 => Ok(()),
--- a/src/constant_time.rs
+++ b/src/constant_time.rs
@@ -22,7 +22,7 @@ use {c, error};
 /// `a` and `b`.
 pub fn verify_slices_are_equal(a: &[u8], b: &[u8])
                                -> Result<(), error::Unspecified> {
-    try!(error::check(a.len() == b.len()));
+    check!(a.len() == b.len());
     let result = unsafe { GFp_memcmp(a.as_ptr(), b.as_ptr(), a.len()) };
     match result {
         0 => Ok(()),
--- a/src/aead/aead.rs
+++ b/src/aead/aead.rs
@@ -216,7 +216,7 @@ impl SealingKey {
 pub fn seal_in_place(key: &SealingKey, nonce: &[u8], ad: &[u8],
                      in_out: &mut [u8], out_suffix_capacity: usize)
                      -> Result<usize, error::Unspecified> {
-    try!(error::check(out_suffix_capacity >= key.key.algorithm.tag_len()));
+    check!(out_suffix_capacity >= key.key.algorithm.tag_len());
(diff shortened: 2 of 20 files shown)


ranweiler:
@briansmith, my thoughts:

1. On average, I think function version makes things _less_ readable in practice. The fact that it doesn't return early means the net effect is just compacting 3 lines of code into very identifier-dense line. I'm not sure that's worth the change. In some cases, it seems to make intent clearer, and in some it obscures it. The fact that we always qualify the usage as `error::check` also introduces new, annoying line continuations in some places. If we went with this, the name should probably change, too, as it isn't "checking" anything. A more appropriate name is probably something like `map_result`, `as_result`.
1. I think the macro version _is_ a net improvement. Unlike the function implementation, the macro performs an early return and does _not_ return `Ok(())` from its block. This means we continue to explicitly return `Ok(())` in e.g. some verification functions— I think this is slightly preferable. Unlike the function version, the resulting one-liner is much less dense with "boilerplate" identifiers, so the condition in question is easier to focus on. Also, as a macro which performs special control flow, the name `check!` makes much more sense.

In short, I'd suggest we use the early-return macro version, if the cost of adding a new macro seems worth it, or just scrap it altogether. I think it _does_ seem worth it, especially as we port more C to Rust. One normal early return in a function is fine, but in functions with lots of intermediate checks, the macro helps a lot. This is best seen in the RSA stuff.

(FYI, I expect to squash this all into one commit if we want to go with one of these approaches, but I'm leaving things split up right now for the sake of easy comparison)

src/constant_time.rs
```diff
@@ -22,9 +22,7 @@ use {c, error};
 /// `a` and `b`.
 pub fn verify_slices_are_equal(a: &[u8], b: &[u8])
                                -> Result<(), error::Unspecified> {
-    if a.len() != b.len() {
-        return Err(error::Unspecified);
-    }
+    check!(a.len() == b.len());
     let result = unsafe { GFp_memcmp(a.as_ptr(), b.as_ptr(), a.len()) };
     match result {
```
briansmith: This `match` would also become a `check!()`, no?

briansmith:
Definitely there are some pros to doing this. But, I see some cons too.

In particular, it's going to be harder to use line-based coverage tools usefully with this style, right? It's already a problem with `try!`, but at least `try!` (or `?`) is idiomatic Rust, where this would be our own thing.

Also, IMO the real problem is that there are lots of branches in the code. We want to have simpler code, with fewer branches. But, it's harder to see the problematic branchiness of the code when it's hidden behind `check!()`. In the case of the huge amount of `try!()` in `ring::rsa`, most of that is actually going to go away! I'd like to find ways to get most of the other branchiness to go away too.

In general, I use macros as a last resort, `try!` being the main exception.

Still, again, I admit that the code *does* look nicer with this. I'm just not sure the code actually *is* nicer with it. So I'm leaning towards just dropping this.

WDYT?

src/constant_time.rs
ranweiler: It definitely could (I was trying to be conservative, pending more general discussion).

ranweiler:
Let's pass on this. I think it was worth experimenting with, but I've mulled it over, and I'd definitely rather have somewhat less-nice code that is more amenable to automated analysis. I also agree that it is better to have syntactically-explicit branchiness, and that we should favor simplifying refactors over band-aids. Thanks for looking it over!

PR #506, opened 2017-04-10
Split internal storage of `Ed25519KeyPair`

ranweiler:
Followup on #479, for #459.

Before, we stored the bytes of the private and public key together in one contiguous array to support using the foreign `GFp_ed25519_sign` function. Now EdDSA signing is implemented in Rust, and we no longer use that function, so we can switch to a more direct internal representation.


ranweiler:
Note: if we land anything from #505 before this, we should make sure we use that we don't revert the change to the early return in `Ed25519KeyPair::from_bytes()`.

briansmith:
Thanks!

PR #507, opened 2017-04-10
WIP: Cache private scalar and prefix

ranweiler:
NOTE: currently based on the work in #506.

- [x] Add private scalar and prefix fields
- [x] Use cached values in `sign`
- [ ] Use cached values in `public_from_private`
- [ ] Benchmark

### DIFF
--- a/src/ec/eddsa.rs
+++ b/src/ec/eddsa.rs
@@ -156,16 +156,9 @@ impl<'a> Ed25519KeyPair {
             let signature_s =
                 slice_as_array_ref_mut!(signature_s, SCALAR_LEN).unwrap();
 
-            let az = digest::digest(&digest::SHA512, &self.private_key);
-            let (a_encoded, z_encoded) = az.as_ref().split_at(SCALAR_LEN);
-
-            let mut a = [0; SCALAR_LEN];
-            a.copy_from_slice(a_encoded);
-            unsafe { GFp_ed25519_scalar_mask(&mut a) };
-
             let nonce = {
                 let mut ctx = digest::Context::new(&digest::SHA512);
-                ctx.update(z_encoded);
+                ctx.update(&self.private_prefix);
                 ctx.update(msg);
                 ctx.finish()
             };
@@ -180,7 +173,7 @@ impl<'a> Ed25519KeyPair {
             let hram_digest = eddsa_digest(signature_r, &self.public_key, msg);
             let hram = digest_scalar(hram_digest);
             unsafe {
-                GFp_x25519_sc_muladd(signature_s, &hram, &a, &nonce);
+                GFp_x25519_sc_muladd(signature_s, &hram, &self.private_scalar, &nonce);
             }
         }
         signature::Signature::new(signature_bytes)
--- a/src/ec/eddsa.rs
+++ b/src/ec/eddsa.rs
@@ -23,6 +23,16 @@ pub struct EdDSAParameters;
 /// An Ed25519 key pair, for signing.
 pub struct Ed25519KeyPair {
     private_key: [u8; SEED_LEN],
+
+    // As described in RFC 8032 Section 5.1.6, Step 1, we compute and cache
+    // these values when generating key pairs, since they are used in every
+    // signature. They are defined by:
+    //
+    //     SHA-512(private_key) = private_scalar || private_prefix
+    //
+    private_scalar: [u8; SCALAR_LEN],
+    private_prefix: [u8; PREFIX_LEN],
+
     public_key: [u8; PUBLIC_KEY_LEN],
 }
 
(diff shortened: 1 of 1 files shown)


src/ec/eddsa.rs
```diff
@@ -22,18 +22,50 @@ pub struct EdDSAParameters;
...
+    public_key: [u8; PUBLIC_KEY_LEN],
 }
 
 /// The raw bytes of the Ed25519 key pair, for serialization.
 pub struct Ed25519KeyPairBytes {
     /// Private key bytes.
     pub private_key: [u8; SEED_LEN],
 
+    private_scalar: [u8; SCALAR_LEN],
+    private_prefix: [u8; PREFIX_LEN],
```
briansmith: We don't need to output these. We just need to output the seed and public key, since that's the standard form of the private key serialization.

src/ec/eddsa.rs
```diff
@@ -22,18 +22,50 @@ pub struct EdDSAParameters;
 
 /// An Ed25519 key pair, for signing.
 pub struct Ed25519KeyPair {
-    private_public: [u8; KEY_PAIR_LEN],
+    private_key: [u8; SEED_LEN],
```
briansmith: We can just remove this private_key field, right?

src/ec/eddsa.rs
```diff
@@ -22,18 +22,50 @@ pub struct EdDSAParameters;
...
+
+    // As described in RFC 8032 Section 5.1.6, Step 1, we compute and cache
+    // these values when generating key pairs, since they are used in every
+    // signature. They are defined by:
+    //
+    //     SHA-512(private_key) = private_scalar || private_prefix
+    //
```
briansmith: I think `private_scalar` should be masked before being stored here so that it is a true scalar.

I think we should either note that `private_scalar` is “the secret scalar *s*` in the RFC and that `private_prefix` is *prefix* from the RFC, or we should just name these `s` and `prefix` respectively and note that the names come from the RFC.

I'd rather not document how they are calculated since that will be evident from the code below.

src/ec/eddsa.rs
```diff
@@ -22,18 +22,50 @@ pub struct EdDSAParameters;
...
 
+    private_scalar: [u8; SCALAR_LEN],
+    private_prefix: [u8; PREFIX_LEN],
+
     /// Public key bytes.
     pub public_key: [u8; PUBLIC_KEY_LEN],
 }
 
+impl Ed25519KeyPairBytes {
+    fn from_key_pair(key_pair: &Ed25519KeyPair) -> Ed25519KeyPairBytes {
```
briansmith: I think it's better to just do this inline in `generate_serializable()`. Doing it inline in that function makes it clearer that there's no way to extract the private key from the key pair, by making it clear that there's no other place where this logic could be invoked.

src/ec/eddsa.rs
```diff
@@ -256,6 +274,20 @@ fn public_from_private(seed: &Seed, out: &mut PublicKey) {
     }
 }
 
+fn private_scalar_prefix_from_seed(seed: &Seed) -> (Scalar, Prefix) {
```
briansmith: I suggest `scalar_and_prefix_from_seed` as the name.

src/ec/eddsa.rs
```diff
@@ -256,6 +274,20 @@ fn public_from_private(seed: &Seed, out: &mut PublicKey) {
...
+    let h = digest::digest(&digest::SHA512, seed);
+    let (scalar_encoded, prefix_encoded) = h.as_ref().split_at(SCALAR_LEN);
+
+    let mut prefix = [0u8; PREFIX_LEN];
+    prefix.copy_from_slice(prefix_encoded);
```
briansmith: Style nit: Build `prefix` after building`scalar`; i.e. move these two lines down.

src/ec/eddsa.rs
```diff
@@ -151,11 +170,10 @@ impl<'a> Ed25519KeyPair {
                 GFp_ge_p3_tobytes(signature_r, &r);
             }
 
-            let hram_digest =
-                eddsa_digest(signature_r, self.public_key_bytes(), msg);
+            let hram_digest = eddsa_digest(signature_r, &self.public_key, msg);
             let hram = digest_scalar(hram_digest);
             unsafe {
-                GFp_x25519_sc_muladd(signature_s, &hram, &a, &nonce);
+                GFp_x25519_sc_muladd(signature_s, &hram, &self.private_scalar, &nonce);
```
briansmith: This line is too long now.

briansmith:
With this PR, we have some uses of `Scalar` that assume the scalar is fully reduced, and we have some other uses that assume it isn't. IMO, it would be better to have separate types for the not-yet-reduced and is-fully-reduced cases. Also, IMO we should start moving using `struct Scalar([u8; SCALAR_LEN])` instead of `type Scalar = [u8; SCALAR_LEN]`. Maybe this PR, or one to immediately precede this one, is a good time to do that.

Finally, I noticed the `copy_from_slice()` stuff is really uglifying this. I wonder if we can do avoid this.

briansmith:
Thanks for doing this! I landed this on master as:
f7bd7eb0fcd4a808b63c7adfa4ac1a2fdf1aa191 Add fields for cached Ed25519 private key computations.
c77d36332e760a61657bf73866b2f2fa853b6e05 Use cached private scalar and prefix in Ed25519 signing.

I also landed these follow-up commits. Note that there are minor stylistic changes from the version you reviewed earlier today:
984f026f90f5355f6eea75736bcef6e840c17551 Fix formatting nits in EdDSA.
de2ad8828fcb7e119a70b67d7255f90095311517 Simplify EdDSA private key construction.
85bb49af49ff37378e353fe1740cb42a16c0963b Fix naming-related nits in EdDSA.
d33bced4f9a22b966548630a8c342a719ddadf63 Simplify EdDSA construction further.
bd208547ddfda5d840816d61506c41f7855a38a9 Simplify public key length check in `Ed25519KeyPair::from_bytes()`.
7edfad5ba44de6e30479b87709420f8dd52cc6ed Inline `private_scalar_prefix_from_seed()` into `Ed25519KeyPair::from_seed()`.
9ed9255043d5b81d9924d3af6f15962c8f5d3492 Simplify public key length check in Ed25519 verification.

Thanks again for doing this!

PR #512, opened 2017-04-15
Convert random chunk generation from C to Rust

PeterReid:
This removes the remaining C wrappers around the system random functions.


src/rand.rs
```diff
@@ -123,23 +120,82 @@ use self::sysrand_or_urandom::fill as fill_impl;
...
+    const GETRANDOM: c::long = 355;
+    #[cfg(target_arch = "x86_64")]
+    const GETRANDOM: c::long = 318;
+
+    const EINTR: c::long = 4;
+
+    #[inline]
+    pub fn chunk(dest: &mut [u8]) -> Result<usize, error::Unspecified> {
+        let chunk_len: c::size_t = dest.len();
+        let flags: c::int = 0;
```
briansmith: The flags argument of getrandom is `unsigned int`.

src/rand.rs
```diff
@@ -123,23 +120,82 @@ use self::sysrand_or_urandom::fill as fill_impl;
...
+        let r = unsafe {
+            syscall(GETRANDOM, dest.as_mut_ptr(), chunk_len, flags)
+        };
+        if r < 0 {
+            if r == EINTR {
```
briansmith: This can never be true because `r < 0` and `EINTR == 4`. It should be comparing `errno` instead.

src/rand.rs
```diff
@@ -123,23 +120,82 @@ use self::sysrand_or_urandom::fill as fill_impl;
...
+                // If an interrupt occurs while getrandom() is blocking
+                // to wait for the entropy pool, then EINTR is returned.
+                // Returning 0 will cause the caller to try again.
+                return Ok( 0 )
```
briansmith: NIT: formatting. Remove the spaces around `0` and terminate with a semicolon.

src/rand.rs
```diff
@@ -123,23 +120,82 @@ use self::sysrand_or_urandom::fill as fill_impl;
...
+            }
+            return Err(error::Unspecified);
+        }
+        Ok( r as usize )
```
briansmith: NIT: remove spaces around `r as usize`.

src/rand.rs
```diff
@@ -123,23 +120,82 @@ use self::sysrand_or_urandom::fill as fill_impl;
...
+    }
+
+    #[inline]
+    pub fn chunk(dest: &mut [u8]) -> Result<usize, error::Unspecified> {
+        let len = min(dest.len(), i32::MAX as usize);
+
+        if unsafe { SystemFunction036(dest.as_mut_ptr(), len as u32) } != 0 {
+            Ok( len )
+        } else {
+            Err( error::Unspecified)
```
briansmith: NIT: remove extra spaces in both branches.

src/rand.rs
```diff
@@ -123,23 +120,82 @@ use self::sysrand_or_urandom::fill as fill_impl;
 #[cfg(any(target_os = "macos", target_os = "ios"))]
 use self::darwin::fill as fill_impl;
 
+#[cfg(target_os = "linux")]
+mod sysrand_chunk {
+    use error;
+    use c;
+
+    extern {
+        fn syscall(number: c::long, ...) -> c::long;
```
briansmith: I did not realize we could call vararg functions from Rust! TIL.

src/rand.rs
```diff
@@ -123,23 +120,82 @@ use self::sysrand_or_urandom::fill as fill_impl;
...
+
+#[cfg(windows)]
+mod sysrand_chunk {
+    use std::cmp::min;
+    use std::i32;
+    use error;
+
+    #[link(name = "Advapi32")]
+    extern "system" {
+        fn SystemFunction036( // also known as RtlGenRandom
```
briansmith: Please reformat this in the style that the rest of *ring* uses, something like:

```rust
extern "system" {
    fn SystemFunction036(random_buffer: *mut u8,
                         random_buffer_length: c::win32::ULONG)
                         -> c::win32::BOOLEAN;
}
```

Let's add a submodule `win32` to `ring::c` that defines the type aliases for ULONG and BOOLEAN. Note in particular that the infrastructure in `ring::c` allows us to test that our mapping of `ULONG` to `u32` and `BOOLEAN` to `u8` is correct on both 32-bit and 64-bit Windows.

src/rand.rs
```diff
@@ -123,23 +120,82 @@ use self::sysrand_or_urandom::fill as fill_impl;
...
+            random_buffer: *mut u8,
+            random_buffer_length: u32
+        ) -> u8;
+    }
+
+    #[inline]
+    pub fn chunk(dest: &mut [u8]) -> Result<usize, error::Unspecified> {
+        let len = min(dest.len(), i32::MAX as usize);
```
briansmith: Let's use `core::meme::size_of::<c::win32::ULONG>()` instead.

briansmith:
Peter, I remember why we decided not to do this before on Linux: We want to to remove the `syscall(2)` dependency. See #514 and https://github.com/japaric/steed/issues/62.

The Windows change is good to go though.

briansmith:
Thinking about this some more...I'm OK with merging this now, w/ the nits addressed. In the future we might do something different to avoid the varargs function as part of our static analysis effort. But, if so, we can deal with that at that time.

Thanks again for submitting this!

PeterReid:
All changes are addressed. The only wrinkle was that errno is a thread-local variable, which rust has not yet stabilized, so I had to write a C wrapper for _that_.

src/rand.rs
```diff
@@ -123,23 +120,87 @@ use self::sysrand_or_urandom::fill as fill_impl;
...
+        fn SystemFunction036(random_buffer: *mut u8,
+                             random_buffer_length: ULONG)
+                             -> BOOLEAN;
+    }
+
+    #[inline]
+    pub fn chunk(dest: &mut [u8]) -> Result<usize, error::Unspecified> {
+        assert!(size_of::<usize>() >= size_of::<ULONG>());
+        let max_chunk_len = (1 << (size_of::<ULONG>()*8 - 1)) - 1;
+        assert_eq!(max_chunk_len, 2147483647);
```
briansmith: Why not 0xffff_ffff, i.e. 4294967295?

PeterReid: I did that because the old code enforced that bound: https://github.com/briansmith/ring/blob/48f6b36582a0743d4c9678949efee000c6209d34/crypto/rand/sysrand.c#L69 

I could not find any documentation to support needing to have that limit, though.

PeterReid:
Thanks for the thorough review again!

briansmith:
> I did that because the old code enforced that bound:

Thanks for pointing that out. The old code returned the actual length in a signed value, where negative values indicated errors. We don't need to do that now, so we can use the full range of the unsigned value.

briansmith:
@PeterReid I thought about this some more. Please check out the new commits I added to my own oxidize-random branch: https://github.com/briansmith/ring/tree/oxidize-random.

Basically, if we're going to use libc's `syscall()` from Rust, we might as well use the libc crate's definition of it and EINTR. Similarly, we can use the libc crate's workaround for errno. WDYT?

PeterReid:
That all looks good to me! I did not think to look in libc for these things, but that seems lots better than including it in this library.

briansmith:
Landed in master as 63069777b5c96e21482f46b2d768af753176b291.

I also landed these follow-ups:
3896e21bbaee86627f632b610b4f51cadd2f8a37 Use libc's `__errno_location()` and `EINTR`
91d73acbeeed92a6e70f1bc28875ab1d4d29113f Use Linux kernel headers' value of SYS_getrandom when available.
11bcedea2d843221151355bd0f95efd5e6d2fe30 Style and formatting nits.

Thanks for doing this!

PR #515, opened 2017-04-15
Use OpenSSL's assembly implementations of memcmp

cmr:
I've only tested these lightly, and haven't read them at all. I assume
they work.


briansmith:
The main reason I didn't do this before is because I've not found time to review the code. In general I'm not so eager to import code from OpenSSL that was explicitly excluded in BoringSSL, without carefully reviewing it. So, this review will likely have high latency.

In general I'd prefer a PR like this to be split into two commits:
1. Import the files unmodified from OpenSSL, except the OpenSSL license is copied inline instead of referenced as LICENSE.txt. I will use a diff tool to verify that the files exactly match the OpenSSL version except for the license block.
2. Rename the symbols to use the `GFp_` prefix.

Note that the OpenSSL license block references LICENSE.txt and which has different contents between *ring* (and BoringSSL) and OpenSSL. Also it references a URL that's not so relevant for *ring*.

cmr:
I'd like to use GFp_memcmp as a case study in the feasibility of asm verification, so I'm happy to put this on the table completely while I work on that.

briansmith:
@cmr I have a new idea for how to get the side-channel-resistance properties from the C code without having to import in the OpenSSL stuff, so I'm going to close this and update the related issue. Thanks for attempting this though!

Issue #500, opened 2017-04-03
Replace Curve25519 `fe_invert()` C function with a Rust equivalent


briansmith:
See `p256_elem_inv()` and `p384_elem_inv()` for the model to copy. Basically, leave the field multiplication and squaring operations in C, but do higher-level part of the Fermat's Littler Theorem implementation in Rust.

ranweiler:
Picking this up!

briansmith:
Note to self for the review:
```
  b1 = z
  b10 = sqr(b1)
  b1001 = sqr_mul(b10, 1 + 1, b1)
  b1011 = mul(b1001, b10)
  acc = sqr_mul(b1011, 1, b1001)
  t1 = sqr_mul(acc, 5, acc)
  t2 = sqr_mul(t1, 10, t1)
  t2 = sqr_mul(t2, 20, t2)
  t1 = sqr_mul(t2, 10, t1)
  t2 = sqr_mul(t1, 50, t1)
  t2 = sqr_mul(t2, 100, t2)
  t1 = sqr_mul(t2, 50, t1)
  out = sqr_mul(t1, 5, 0b1011)
```

PR #519, opened 2017-04-20
Port Ed25519 element inversion to Rust.

ranweiler:
For #500.

### DIFF
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -30,6 +30,8 @@
 
 
 /* Prevent -Wmissing-prototypes warnings. */
+void GFp_fe_mul(fe h, const fe f, const fe g);
+void GFp_fe_sq(fe h, const fe f);
 void GFp_ge_double_scalarmult_vartime(ge_p2 *r, const uint8_t *a,
                                       const ge_p3 *A, const uint8_t *b);
 void GFp_ge_p3_tobytes(uint8_t *s, const ge_p3 *h);
@@ -284,7 +286,7 @@ static void fe_sub(fe h, const fe f, const fe g) {
  * Can get away with 11 carries, but then data flow is much deeper.
  *
  * With tighter constraints on inputs can squeeze carries into int32. */
-static void fe_mul(fe h, const fe f, const fe g) {
+void GFp_fe_mul(fe h, const fe f, const fe g) {
   int32_t f0 = f[0];
   int32_t f1 = f[1];
   int32_t f2 = f[2];
@@ -507,10 +509,8 @@ static void fe_mul(fe h, const fe f, const fe g) {
  *    |f| bounded by 1.65*2^26,1.65*2^25,1.65*2^26,1.65*2^25,etc.
  *
  * Postconditions:
- *    |h| bounded by 1.01*2^25,1.01*2^24,1.01*2^25,1.01*2^24,etc.
- *
- * See fe_mul.c for discussion of implementation strategy. */
-static void fe_sq(fe h, const fe f) {
+ *    |h| bounded by 1.01*2^25,1.01*2^24,1.01*2^25,1.01*2^24,etc. */
+void GFp_fe_sq(fe h, const fe f) {
   int32_t f0 = f[0];
   int32_t f1 = f[1];
   int32_t f2 = f[2];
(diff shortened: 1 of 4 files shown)


ranweiler:
@briansmith the later commits should all be squashed into the first before merging, but I'm leaving them separate for the moment so we can easily take or leave the change in b6eeaeb75f4a39b41f7e5dc203e20caa473d8521.

src/ec/eddsa.rs
```diff
@@ -297,60 +273,80 @@ fn elem_sq_in_place(a: &mut Elem) {
 #[doc(hidden)]
 #[no_mangle]
 pub unsafe extern fn eddsa_elem_invert(out: &mut Elem, a: &Elem) {
+    // Sets |out| to be |a| * |b| (mod p).
```
briansmith: Let's use the Rust style \`a\` instead of BoringSSL style |a|` in the comments.

More generally, we should make this more consistent with the P-256/P-384 code. In particular:

* In the P-256/P-384 code we usually to return a `Elem` instead of having an `Elem` out parameter. If that is less efficient then we should switch that code to use the style you used style. However, if it is just as efficient we should do things the way the P-256/P-384 code does.

* I am not wedded to the names I used in the P-256 and P-384 code but it would be good to use the same names here and there. For example, we have `elem_product` in P-256/P-384 which is like your `mul` and we have `elem_mul` which is like your `mul_in_place`. I would suggest that, given more experience with Rust convention sand things like the `MulAssign` trait, `elem_mul_assign` seems better than `elem_mul` or `mul_in_place`. (Note that here you could actually implement `Mul`, `MulAssign`, etc., but I wouldn't bother since it'd likely be undone in the not-too-distant future.)

src/ec/eddsa.rs
```diff
@@ -297,60 +273,80 @@ fn elem_sq_in_place(a: &mut Elem) {
...
     for _ in 1..2 {
-        elem_sq_in_place(&mut t1);
+        sqr_in_place(&mut t1);
     }
-    elem_mul_in_place(&mut t1, a);
-    elem_mul_in_place(&mut t0, &t1);
-    elem_sq(&mut t2, &t0);
-    elem_mul_in_place(&mut t1, &t2);
-    elem_sq(&mut t2, &t1);
+    mul_in_place(&mut t1, a);
```
briansmith: Notice that every loop of `sqr_in_place()` is followed by `mul_in_place()` (Including the last one, because P is odd and so P - 2 is odd). That, and improved readability are why I did this:

https://github.com/briansmith/ring/blob/86a45d046e92640463923b2e35c2a94387109789/src/ec/suite_b/ops/p256.rs#L86-L94
https://github.com/briansmith/ring/blob/86a45d046e92640463923b2e35c2a94387109789/src/ec/suite_b/ops/ops.rs#L354-L373

Let's do the same here. (Note I guess I used `acc` instead of `r` there; I guess I could go either way as long as everything is made consistent.)

src/ec/eddsa.rs
```diff
@@ -248,6 +250,104 @@ extern  {
...
+///
+/// This implementation of the modular exponentiation is ultimately derived
+/// from Daniel J. Bernstein's reference implementation in qhasm, via
+/// BoringSSL, and referenced in [Curve15519]. The overall strategy is to
+/// square and multiply against the fixed base |a| to double and add to the
+/// exponent, reusing intermediate computations as we are able.
+///
+/// [Curve25519]: https://cr.yp.to/ecdh.html#curve25519-paper
+#[doc(hidden)]
+#[no_mangle]
```
briansmith: In the end, this function should be renamed `elem_inverse`, it should have the type `Elem -> Elem` or `&Elem -> Elem`, it shouldn't be public, and it shouldn't be no_mangle.

(Aside: any `no_mangle` function in *ring* should have the ugly `GFp_` prefix. The ugliness is a feature.)

crypto/curve25519/curve25519.c
```diff
@@ -909,65 +849,65 @@ static void fe_pow22523(fe out, const fe z) {
...
 
 void GFp_x25519_ge_tobytes(uint8_t *s, const ge_p2 *h) {
   fe recip;
   fe x;
   fe y;
 
-  fe_invert(recip, h->Z);
-  fe_mul(x, h->X, recip);
-  fe_mul(y, h->Y, recip);
+  eddsa_elem_invert(recip, h->Z);
```
briansmith: I suggest that, as the very first commit, we find a way to pull out the call to eddsa_elem_invert from C to Rust. e.g. we could add `recip` as a parameter to `GFp_x25519_ge_tobytes`. That way we can avoid the no_mangle and naming stuff.

briansmith:
Note: I left a bunch more comments but they are hidden in the wonderful "outdated diff" UI.

briansmith:
Another thing we'll need to check: Does exporting `GFp_fe_mul` and `GFp_fe_sqr` affect performance in any way, especially for platforms where we don't have assembly language implementations?

src/ec/eddsa.rs
```diff
@@ -248,6 +250,105 @@ extern  {
...
+    for _ in 1..50 {
+        sqr_in_place(&mut t2);
+    }
+    mul_in_place(&mut t2, &t1);
+    sqr(&mut t3, &t2);
+    for _ in 1..100 {
+        sqr_in_place(&mut t3);
+    }
+    mul_in_place(&mut t2, &t3);
+    sqr_in_place(&mut t2);
```
briansmith: Why square `t2` once here, and then 49 times below, instead of just 50 times below?

src/ec/eddsa.rs
```diff
@@ -248,6 +250,105 @@ extern  {
...
+    for _ in 1..50 {
+        sqr_in_place(&mut t2);
+    }
+    mul_in_place(&mut t1, &t2);
+    sqr_in_place(&mut t1);
```
briansmith: Ditto.

ranweiler:
> This inversion function isn't specific to Ed25519, right? I think we probably should have a `ring::ec::curve25519::{curve25519, ed25519, x25519}` set of modules, and this should go in the curve25519 module.

Agreed!

> Was there a reason why the callers of `fe_invert()` weren't oxidized first? I don't think it's a giant deal either way, but in general we try to minimize Rust -> C -> Rust code paths. If it's problematic to oxidize the callers first then it would be good to understand why.

No blocking reason, I was mostly just keeping my nose close to the original issue to start with. `fe_invert()` is called in `x25519_scalar_mult_generic()`, which calls many other `fe_*` C functions, so I also avoided it to keep the first move small.

You've actually raised and answered a bunch of questions I had re larger-scale code organization and naming preferences. I'll loop back on this, rewriting the commit history so we start with oxidizing the `fe_invert()` callers, reorganize the modules, then implement an oxidization that more closely mirrors the P-256/384 approach.

briansmith:
> You've actually raised and answered a bunch of questions I had re larger-scale code organization and naming preferences. I'll loop back on this, rewriting the commit history so we start with oxidizing the fe_invert() callers, reorganize the modules, then implement an oxidization that more closely mirrors the P-256/384 approach.

Like I said, I'm not wedded to the P-256/P-384 naming. Those names were chosen when I was more focused on restructuring the code, and also we've learned a lot since I did it. The important thing is that we should strive for ~100% consistency between the Suite-B code and the Curve25519 code, except for areas where they differ by nature (e.g. point-on-the-curve check vs. no such check, Montgomery ladder vs. traditional scalar mulitplication). However, I imagine it will take some non-trivial effort to get to 100% consistency, so for now we should just focus on avoiding gratuitous inconsistencies.

crypto/curve25519/curve25519.c
```diff
@@ -909,65 +849,65 @@ static void fe_pow22523(fe out, const fe z) {
   fe t2;
   int i;
 
-  fe_sq(t0, z);
-  fe_sq(t1, t0);
+  GFp_fe_sq(t0, z);
```
briansmith: Since the time you wrote this PR, I changed the strategy for exporting the `fe_*` functions from C to Rust. Instead of renaming the original functions and making them `extern` (non-`static`), I have created an extern `GFp_fe_*()` wrapper around each static `fe_*()` function. Thus, the Rust code goes through the wrapper while the C code doesn't. Thus, we don't need to change any `fe_sq`/`fe_mul` callers in the C code any more

(I imagine that shortly after this PR, we will replace this function with Rust code anyway.)

briansmith:
@ranweiler I'm going to close this PR. If/when you are interested in picking it back up, please submit a new PR. Thanks for your efforts here!

PR #522, opened 2017-04-22
Port Curve25519 encoding operations

ranweiler:
In support of #500, we want to oxidize the callers of `fe_invert()`. This work is in support of that intermediate step, as well as inching the Curve25519 arithmetic operations into alignment with the Suite B EC code.

### DIFF
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -30,10 +30,12 @@
 
 
 /* Prevent -Wmissing-prototypes warnings. */
+void GFp_fe_invert(fe out, const fe z);
+uint8_t GFp_fe_isnegative(const fe f);
+void GFp_fe_mul(fe h, const fe f, const fe g);
+void GFp_fe_tobytes(uint8_t *s, const fe h);
 void GFp_ge_double_scalarmult_vartime(ge_p2 *r, const uint8_t *a,
                                       const ge_p3 *A, const uint8_t *b);
-void GFp_ge_p3_tobytes(uint8_t *s, const ge_p3 *h);
-void GFp_x25519_ge_tobytes(uint8_t *s, const ge_p2 *h);
 int GFp_x25519_ge_frombytes_vartime(ge_p3 *h, const uint8_t *s);
 void GFp_x25519_ge_scalarmult_base(ge_p3 *h, const uint8_t a[32]);
 void GFp_x25519_sc_muladd(uint8_t *s, const uint8_t *a, const uint8_t *b,
@@ -131,7 +133,7 @@ static void fe_frombytes(fe h, const uint8_t *s) {
  *
  *   Have q+2^(-255)x = 2^(-255)(h + 19 2^(-25) h9 + 2^(-1))
  *   so floor(2^(-255)(h + 19 2^(-25) h9 + 2^(-1))) = q. */
-static void fe_tobytes(uint8_t *s, const fe h) {
+void GFp_fe_tobytes(uint8_t *s, const fe h) {
   int32_t h0 = h[0];
   int32_t h1 = h[1];
   int32_t h2 = h[2];
@@ -284,7 +286,7 @@ static void fe_sub(fe h, const fe f, const fe g) {
  * Can get away with 11 carries, but then data flow is much deeper.
  *
  * With tighter constraints on inputs can squeeze carries into int32. */
-static void fe_mul(fe h, const fe f, const fe g) {
+void GFp_fe_mul(fe h, const fe f, const fe g) {
   int32_t f0 = f[0];
   int32_t f1 = f[1];
   int32_t f2 = f[2];
@@ -507,9 +509,7 @@ static void fe_mul(fe h, const fe f, const fe g) {
  *    |f| bounded by 1.65*2^26,1.65*2^25,1.65*2^26,1.65*2^25,etc.
  *
  * Postconditions:
- *    |h| bounded by 1.01*2^25,1.01*2^24,1.01*2^25,1.01*2^24,etc.
- *
- * See fe_mul.c for discussion of implementation strategy. */
+ *    |h| bounded by 1.01*2^25,1.01*2^24,1.01*2^25,1.01*2^24,etc. */
 static void fe_sq(fe h, const fe f) {
   int32_t f0 = f[0];
   int32_t f1 = f[1];
@@ -641,7 +641,7 @@ static void fe_sq(fe h, const fe f) {
   h[9] = (int32_t)h9;
 }
 
-static void fe_invert(fe out, const fe z) {
+void GFp_fe_invert(fe out, const fe z) {
   fe t0;
   fe t1;
   fe t2;
@@ -653,50 +653,50 @@ static void fe_invert(fe out, const fe z) {
   for (i = 1; i < 2; ++i) {
     fe_sq(t1, t1);
   }
-  fe_mul(t1, z, t1);
-  fe_mul(t0, t0, t1);
+  GFp_fe_mul(t1, z, t1);
+  GFp_fe_mul(t0, t0, t1);
   fe_sq(t2, t0);
-  fe_mul(t1, t1, t2);
+  GFp_fe_mul(t1, t1, t2);
   fe_sq(t2, t1);
   for (i = 1; i < 5; ++i) {
     fe_sq(t2, t2);
   }
-  fe_mul(t1, t2, t1);
+  GFp_fe_mul(t1, t2, t1);
   fe_sq(t2, t1);
   for (i = 1; i < 10; ++i) {
     fe_sq(t2, t2);
   }
-  fe_mul(t2, t2, t1);
+  GFp_fe_mul(t2, t2, t1);
   fe_sq(t3, t2);
   for (i = 1; i < 20; ++i) {
     fe_sq(t3, t3);
   }
-  fe_mul(t2, t3, t2);
+  GFp_fe_mul(t2, t3, t2);
   fe_sq(t2, t2);
   for (i = 1; i < 10; ++i) {
     fe_sq(t2, t2);
   }
-  fe_mul(t1, t2, t1);
+  GFp_fe_mul(t1, t2, t1);
   fe_sq(t2, t1);
   for (i = 1; i < 50; ++i) {
     fe_sq(t2, t2);
   }
-  fe_mul(t2, t2, t1);
+  GFp_fe_mul(t2, t2, t1);
   fe_sq(t3, t2);
   for (i = 1; i < 100; ++i) {
     fe_sq(t3, t3);
   }
-  fe_mul(t2, t3, t2);
+  GFp_fe_mul(t2, t3, t2);
   fe_sq(t2, t2);
   for (i = 1; i < 50; ++i) {
     fe_sq(t2, t2);
   }
-  fe_mul(t1, t2, t1);
+  GFp_fe_mul(t1, t2, t1);
   fe_sq(t1, t1);
   for (i = 1; i < 5; ++i) {
     fe_sq(t1, t1);
   }
-  fe_mul(out, t1, t0);
+  GFp_fe_mul(out, t1, t0);
 }
 
 /* h = -f
@@ -734,7 +734,7 @@ static void fe_cmov(fe f, const fe g, unsigned b) {
  *    |f| bounded by 1.1*2^26,1.1*2^25,1.1*2^26,1.1*2^25,etc. */
 static int fe_isnonzero(const fe f) {
   uint8_t s[32];
-  fe_tobytes(s, f);
+  GFp_fe_tobytes(s, f);
 
   static const uint8_t zero[32] = {0};
   return GFp_memcmp(s, zero, sizeof(zero)) != 0;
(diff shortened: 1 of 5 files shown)


ranweiler:
At first, for `ExtPoint`, I implemented an `assign_bytes()` method which avoided a `copy_from_slice()` call introduced [here](https://github.com/briansmith/ring/pull/522/commits/7f26e1d560beaf543664c8d3d661d532a83c5e52#diff-3b0b79572e588435f30a2c0a0aa1cff2R148), and instead updated a mutable array reference.

I benchmarked the difference with and without it, and in comparison to upstream master, and the difference seemed negligible, seen here:

```
# Upstream `master`
test signature::ed25519::generate_key_pair                    ... bench:      49,632 ns/iter (+/- 869)
test signature::ed25519::sign_empty                           ... bench:      47,642 ns/iter (+/- 7,259)

# With `assign_bytes()`
test signature::ed25519::generate_key_pair                    ... bench:      49,761 ns/iter (+/- 694)
test signature::ed25519::sign_empty                           ... bench:      47,642 ns/iter (+/- 4,951)

# Without `assign_bytes()`
test signature::ed25519::generate_key_pair                    ... bench:      49,574 ns/iter (+/- 1,020)
test signature::ed25519::sign_empty                           ... bench:      47,670 ns/iter (+/- 4,617)
```

src/ec/curve25519/ops.rs
```diff
@@ -63,4 +82,18 @@ impl Point {
...
+    pub fn to_bytes(&self) -> [u8; ELEM_LEN] {
+        let mut bytes = [0u8; ELEM_LEN];
+        unsafe { GFp_x25519_ge_tobytes(&mut bytes, self); }
+
+        bytes
+    }
+}
+
+extern {
+    fn GFp_ge_p3_tobytes(s: &mut [u8; ELEM_LEN], h: &ExtPoint);
```
ranweiler: Note the change of signature, explained in the commit message.

These functions will be oxidized next (they are callers of `fe_invert()`, which is the motivation). Maybe we should add a separate type alias for encoded/compressed points, not just their byte length?

ranweiler: Update: I did this in subsequent commits.

crypto/curve25519/curve25519.c
```diff
@@ -653,50 +653,50 @@ static void fe_invert(fe out, const fe z) {
   for (i = 1; i < 2; ++i) {
     fe_sq(t1, t1);
   }
-  fe_mul(t1, z, t1);
-  fe_mul(t0, t0, t1);
+  GFp_fe_mul(t1, z, t1);
```
briansmith: Does this change the code generation in any significant way? In other words, is the C compiler already generating regular function falls for this, or was it doing fancy inlining stuff?

src/ec/curve25519/curve25519.rs
```diff
@@ -12,7 +12,9 @@
 // OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
-//! Elliptic curve operations on Curve25519.
+//! Elliptic curve operations and schemes using Curve25519.
```
briansmith: I don't understand this clarification.

ranweiler: This was an attempt to distinguish between the sorts of things in `ops` (elliptic curve arithmetic operations) and `ed25519` (a signature scheme).

ranweiler: IOW, the analogous line in `curve25519::ops` is:
```
//! Elliptic curve operations on Curve25519.
```

src/ec/curve25519/ops.rs
```diff
@@ -84,16 +81,32 @@ impl Point {
...
+    let mut y_over_z = [0; ELEM_LIMBS];
+    let mut bytes = [0; ELEM_LEN];
+
+    unsafe {
+        GFp_fe_invert(&mut recip, z);
+        GFp_fe_mul(&mut x_over_z, x, &recip);
+        GFp_fe_mul(&mut y_over_z, y, &recip);
+        GFp_fe_tobytes(&mut bytes, &y_over_z);
     }
+    bytes[31] ^= unsafe { GFp_fe_isnegative(&x_over_z) << 7 };
```
briansmith: Determining whether x_over_z is negative needs to be done in constant time, but merging its result into `bytes[31]` doesn't need to be constant time, right?

Maybe it would be better to do the constant-time operation in a separate statement from the non-constant-time operation? Then every statement would be constant-time or not? Then we could also document that it isn't necessary for the `bytes ^= ...` to be constant time even though it is necessary to calculate both inputs of the operation in constant time.

crypto/curve25519/curve25519.c
ranweiler: Hmm, not obviously, but I'll double-check. I'm about to benchmark at the latest commit, as well.

ranweiler:
Benchmark at 56980ceaebba88315ff4ffcec60bba86ecaf527b: 
```
test signature::ed25519::generate_key_pair                    ... bench:      49,597 ns/iter (+/- 888)
test signature::ed25519::sign_empty                           ... bench:      47,620 ns/iter (+/- 4,508)
```

src/ec/curve25519/curve25519.rs
briansmith: Gotcha.

src/ec/curve25519/ops.rs
```diff
@@ -0,0 +1,120 @@
...
+            z: [0; ELEM_LIMBS],
+        }
+    }
+
+    pub fn encode(&self) -> EncodedPoint {
+        encode_point(&self.x, &self.y, &self.z)
+    }
+}
+
+fn encode_point(x: &Elem, y: &Elem, z: &Elem) -> [u8; ELEM_LEN] {
```
briansmith: As I noted in another comment, I think we should give some descriptive type alias to this `[u8; ELEM_LEN]`.

ranweiler: That's exactly what `EncodedPoint` is: I just somehow forgot to apply it throughout.

src/ec/curve25519/ed25519.rs
```diff
@@ -147,9 +144,8 @@ impl<'a> Ed25519KeyPair {
             let mut r = ExtPoint::new_at_infinity();
             unsafe {
                 GFp_x25519_ge_scalarmult_base(&mut r, &nonce);
-                GFp_ge_p3_tobytes(signature_r, &r);
             }
-
+            signature_r.copy_from_slice(&r.encode());
```
briansmith: We don't need to worry about performance changes regarding adding a few *one-time* 32-byte or 64-byte copies. The things that are likely to significantly affect performance are: 
 
1. Making some C function `extern` that was `static` in a way that causes the C compiler to optimize it (and probably the whole curve25519.c file) worse.

2. When passing arrays from one function to another, when we change the calls from C -> C to Rust -> C, worse code could get generated because this would inhibit autovectorization and/or inhibit the compiler's ability to rely on arrays being aligned and/or similar things.

src/ec/curve25519/ops.rs
ranweiler: Amended.

src/ec/curve25519/ops.rs
```diff
@@ -0,0 +1,66 @@
+// Copyright 2017 Brian Smith.
```
briansmith: I will change this to 2015-2017. (Not sure it matters.)

briansmith:
Please fold the "Use descriptive names for Curve25519 point encoding." change into "Add methods for Curve25519 point encoding."

src/ec/curve25519/ops.rs
ranweiler: I suggest checking the other `curve25519` modules while you're in there— the top-level `curve25519` module date should probably be changed.

src/ec/curve25519/ops.rs
ranweiler: I like this idea of keeping constant/vartime lines distinct.

ranweiler: I folded this suggestion into the "Port [...]" commit [here](https://github.com/briansmith/ring/pull/522/commits/4c27993ade28a99ab498fc8fcdedf9dea0803cb8#diff-d26b986b38fde7051f0848091dad3a9bR105).

ranweiler:
> Please fold the "Use descriptive names for Curve25519 point encoding." change into "Add methods for Curve25519 point encoding."

Done.

src/ec/curve25519/ops.rs
```diff
@@ -39,6 +40,24 @@ impl ExtPoint {
...
+        let mut point = Self::new_at_infinity();
+
+        try!(bssl::map_result(unsafe {
+            GFp_x25519_ge_frombytes_vartime(&mut point, bytes)
+        }));
+
+        Ok(point)
+    }
+
+    pub fn to_bytes(&self) -> [u8; ELEM_LEN] {
```
ranweiler: It looks like we _could_ move `self`. Grepping a bit, it seems we do that with similar methods in the `rsa::bigint` code. Since it is more idiomatic Rust to prefer refs to self, I would want to keep it as a ref unless we think it would prevent or discourage specific errors.

This seems like a gray area to me. On one hand, an owner of `self` calling this method multiple times needn't be incorrect per the meaning of the function on its own. In the actual usage, though, this would always be an error. This method is also private, so we don't have any unforeseen use cases to worry about, so the ref preference is less important.

I'm fine either way. If we do change it, we should change the name from `encode()` (which is what we ended up with) to `into_encoded()` or similar. What do you prefer?

src/ec/curve25519/ops.rs
ranweiler: I also got rid of the magic number.

src/ec/curve25519/ops.rs
ranweiler: Done.

crypto/curve25519/curve25519.c
ranweiler: Looks like regular function calls, at least on macOS.

briansmith:
> It looks like we could move self. Grepping a bit, it seems we do that with similar methods in the rsa::bigint code.

> Since it is more idiomatic Rust to prefer refs to self, I would want to keep it as a ref unless we think it would prevent or discourage specific errors.

I don't think refs to self are more (or less) idiomatic in Rust than moving self.

> This seems like a gray area to me. On one hand, an owner of self calling this method multiple times needn't be incorrect per the meaning of the function on its own.

I agree that this is a good rule of thumb for *public* functions, if qualified with "unless there would be a significant performance impact."

> In the actual usage, though, this would always be an error. This method is also private, so we don't have any unforeseen use cases to worry about, so the ref preference is less important.

Exactly.

I don't think it would make any (significant) difference here.  But, in general, in private functions I slightly prefer move over ref because potentially the compiler would have an easier time optimizing the move version. In contrast, move vs. ref makes a big difference in the RSA code because it determines whether we can reuse the inner `BIGNUM` of the moved argument; i.e. using moves helps us avoid mallocs there. So if we're aiming for ultimate consistency across the RSA, Suite B, and Curve25519 code, it probably makes sense to use moves when it's practical and performant to do so, since it is important for the RSA case. (I'm planning to change the Suite B code to be more like the RSA code, when I have time.)

> I'm fine either way. If we do change it, we should change the name from encode() (which is what we ended up with) to into_encoded() or similar. What do you prefer?

I guess `into_encoded_point()` would be most consistent with the RSA code.

I guess if we want to do things the "Idiomatic Rust way" we'll eventually replace all these `into_xxx_xxx` functions with implementations of `Into<XxxXxx>` some day.

ranweiler:
> But, in general, in private functions I slightly prefer move over ref because potentially the compiler would have an easier time optimizing the move version.

Yes, I suspected that was another reason. Thinking about your other comments about e.g. auto-vectorization, "telling the compiler more" seems preferable, even though I have only a vague sense of what it might get us (I don't feel like I could make a strong concrete argument in this case).

I'll go ahead and rename the methods to `into_encoded_point` and `from_encoded_point` in the commit where they are added.

ranweiler:
> I'll go ahead and rename the methods to `into_encoded_point` and `from_encoded_point` in the commit where they are added.

Done.

briansmith:
Thanks! I landed these with minor changes (where my chjanges are seperate commits, except the copyright line change) as:

39b35d3fda039712454ec3bc50362e83ac4c5cae Move Curve25519 operations into their own module.
88579cf0c4e4073515d64abbac9541ec4b49f070 Add methods for Curve25519 point encoding.
49c6a48738a66e1e3bf8a25dc25032940ed2a351 Use array assignment instead of `copy_from_slice()` in one place in ed25519.rs.
6d6aeb6a4fa814b7278565ed2e9036fdcc962edb Port Curve25519 point encoding to Rust.
bc5bd4a0bdd151c3cab2e33b797efa65333c5fb1 Clarify `unsafe` and constant-timedness of Curve25519 `encode_point()`.
7d46aad3e2aaad9cd6f5530fa34c586f9914716f Move all non-protocol Curve25519 types into `ops` module.

Please take some time to review my changes, which were all minor (and hopefully correct).

briansmith:
BTW, I think this is all really good. The added clarity from the logical refactoring you've done has made things a lot clearer.

Issue #418, opened 2017-01-20
Add examples in doctests for `ring::aead` API


briansmith:
The AEAD API is one of the few that doesn't have example code. Let's add the example code, in a similar style as the existing examples for other features.

We're likely to remove the `in_prefix_len` parameter, so if we only have resources to write two examples, one with `in_prefix_len` zero and one where it is non-zero, it's best to write one where `in_prefix_len` is zero.

A good example would be TLS records, which have a 5 byte AAD prefix of `{ type: u8, ProtocolVersion: u16, length: u16 }`.

It would also be good to have an example where there's no ciphertext, and there's only AAD.

Philipp91:
Proper nonce generation should be explained somewhere. Especially if random nonces aren't secure, as that would be surprising to most users.

briansmith:
Thanks for doing this! I landed it as d48469d74acae8ab8ab794b0f980638d0e2ef231.

I hope you continue to contribute in the future. If/when you do, please read https://github.com/briansmith/ring#contributing and add the contributor's statement to the end of each commit's commit message.

Thanks!

Philipp91:
Are you sure you closed the right issue?

briansmith:
Thanks @Philipp91. Reopened.

dsprenkels:
If possible, I would like to take this issue. :smile:

briansmith:
My understanding is that there is at least one PR and maybe two different PRs that attempt to fix this issue. Additionally, there are some other patches that completely change the API. So basically I need to decide which direction to go. I imagine I'll just replace the API with a better one. I'll be discussing this with some users of the API soon and I'll update this issue when I've made a decision.

dsprenkels:
Ok. Thanks for your quick reply! Do you have any good suggestions on which issues are ready to be tackled by a beginner like me?

briansmith:
See #805 where the beginning of the refactoring of the AEAD API has started. Feedback appreciated.

artosan:
Any news on this issue? I am having problems using the AEAD API and it would be super nice to have some working examples which show correct nonce use and so on.

xliiv:
Any update?

xliiv:
Btw i found this
https://gist.github.com/jaysonsantos/117c1f7623a2fb2c39e47f6cf87860a3
but it's outdated

rsdy:
[...]

PR #524, opened 2017-04-25
Add a simple Drop for aead::Key

jaysonsantos:
I saw this TODO and decided to implement it. Just let me know if that is what you were thinking about.

### DIFF
--- a/src/aead/aead.rs
+++ b/src/aead/aead.rs
@@ -239,7 +239,15 @@ struct Key {
     algorithm: &'static Algorithm,
 }
 
-// TODO: Implement Drop for Key that zeroizes the key data?
+impl Drop for Key {
+    /// Reset key's buffer to zero to avoid someone reading the data after the software
+    /// finishes running.
+    fn drop(&mut self) {
+        for idx in self.ctx_buf.iter_mut() {
+            *idx = 0u64;
+        }
+    }
+}
 
 const KEY_CTX_BUF_ELEMS: usize = (KEY_CTX_BUF_LEN + 7) / 8;
 


briansmith:
Hi. Thank you very much for looking at the code and submitting this PR.

 I wrote that TODO very early in the *ring* project and I've made a lot of progress in our thinking regarding zeroization since then. Off-and-on I am working on documenting if/why/how we should do zeroization in *ring*, but it will be a while before that write-up is ready. In the meantime we're going to hold off on zeroization while we focus on the oxidation effort and other code cleanup. With that in mind, I committed d020ba9511156c1abe9971d98af708de502f5131 to remove the TODO comment so nobody else gets misled by it.

Thanks again for the PR though. Feel free to email me (brian@briansmith.org) if you want to chat about other changes that would be appropriate for the current efforts.

jaysonsantos:
Hey @briansmith thanks for your response. Do you have plans to add memory lock? Should I file an issue for wishes?

jaysonsantos:
Ah, taking advantage of this channel for communication. Reading the docs/code I found a bit hard to know how to do simple encryption/decryption stuff because I don't know much about encryption. If you have plans to add more descriptive docs let me know, maybe I could also help with what I figured out reading/debugging the code trying to figure out which Unspecified error happened :D.

briansmith:
>  Do you have plans to add memory lock? 

I think it should probably be done as a custom allocator. Then we can consider adding support for custom allocators and then get that for free.

briansmith:
> Reading the docs/code I found a bit hard to know how to do simple encryption/decryption stuff because I don't know much about encryption. If you have plans to add more descriptive docs let me know, maybe I could also help with what I figured out reading/debugging the code.

That would be great. See https://github.com/briansmith/ring/issues/418.

Issue #509, opened 2017-04-10
Avoid fe_frombytes/fe_tobytes conversions (and similar ones for scalars) in curve25519.c


briansmith:
In intermediate functions, instead of converting inputs from bytes to the internal representation, and then converting the results back to bytes, we should just pass the inputs in the internal (Limb) representations. 

Ideally the only time we should convert from bytes to internal representation is when we're parsing the peer's public key or when we're parsing a signature. Similarly, ideally the only time we should convert from the internal representation to bytes is when we're serializing the output(s) of the calculations.

Note that there is a `fe_frombytes()`/`fe_tobytes()` pair of functions, but there isn't a similar pair of functions for deserializing/serializing scalars. Note that the start and end of functions like `GFp_x25519_sc_muladd()` inline the deserialization/serialization logic instead.

This could maybe be a step towards unifying the P-256/P-384 code with the Curve25519 code, if we can reuse `Elem` and `Scalar` from the P-256/P-384 code.

/cc @ranweiler

PR #525, opened 2017-04-26
Port `X25519_scalar mult()` callers

ranweiler:
Towards #500.

### DIFF
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -31,7 +31,9 @@
 
 /* Prevent -Wmissing-prototypes warnings. */
 void GFp_curve25519_scalar_mask(uint8_t a[32]);
+void GFp_fe_add(fe h, const fe f, const fe g);
 void GFp_fe_invert(fe out, const fe z);
+void GFp_fe_sub(fe h, const fe f, const fe g);
 uint8_t GFp_fe_isnegative(const fe f);
 void GFp_fe_mul(fe h, const fe f, const fe g);
 void GFp_fe_tobytes(uint8_t *s, const fe h);
@@ -246,6 +248,11 @@ static void fe_add(fe h, const fe f, const fe g) {
   }
 }
 
+/* Wrapper to enable Rust FFI without impacting C compiler optimization. */
+void GFp_fe_add(fe h, const fe f, const fe g) {
+    fe_add(h, f, g);
+}
+
 /* h = f - g
  * Can overlap h with f or g.
  *
@@ -260,6 +267,11 @@ static void fe_sub(fe h, const fe f, const fe g) {
   for (i = 0; i < 10; i++) {
     h[i] = f[i] - g[i];
   }
+}
+
+/* Wrapper to enable Rust FFI without impacting C compiler optimization. */
+void GFp_fe_sub(fe h, const fe f, const fe g) {
+    fe_sub(h, f, g);
 }
 
 /* h = f * g
@@ -4674,54 +4686,6 @@ void GFp_x25519_scalar_mult(uint8_t out[32], const uint8_t scalar[32],
 #endif
 
   x25519_scalar_mult_generic(out, scalar, point);
-}
-
-#endif  /* BORINGSSL_X25519_X86_64 */
-
-
-/* Prototypes to avoid -Wmissing-prototypes warnings. */
-void GFp_x25519_public_from_private(uint8_t out_public_value[32],
-                                    const uint8_t private_key[32]);
-
-#if defined(BORINGSSL_X25519_X86_64)
-
-/* When |BORINGSSL_X25519_X86_64| is set, base point multiplication is done with
- * the Montgomery ladder because it's faster. Otherwise it's done using the
- * Ed25519 tables. */
-
-void GFp_x25519_public_from_private(uint8_t out_public_value[32],
-                                    const uint8_t private_key[32]) {
-  static const uint8_t kMongomeryBasePoint[32] = {9};
-  GFp_x25519_scalar_mult(out_public_value, private_key, kMongomeryBasePoint);
-}
-
-#else
-
-void GFp_x25519_public_from_private(uint8_t out_public_value[32],
-                                    const uint8_t private_key[32]) {
-#if defined(BORINGSSL_X25519_NEON)
-  if (GFp_is_NEON_capable()) {
-    static const uint8_t kMongomeryBasePoint[32] = {9};
-    GFp_x25519_NEON(out_public_value, private_key, kMongomeryBasePoint);
-    return;
-  }
-#endif
-
-  uint8_t e[32];
-  memcpy(e, private_key, 32);
-  GFp_curve25519_scalar_mask(e);
-
-  ge_p3 A;
-  GFp_x25519_ge_scalarmult_base(&A, e);
-
-  /* We only need the u-coordinate of the curve25519 point. The map is
-   * u=(y+1)/(1-y). Since y=Y/Z, this gives u=(Z+Y)/(Z-Y). */
-  fe zplusy, zminusy, zminusy_inv;
-  fe_add(zplusy, A.Z, A.Y);
-  fe_sub(zminusy, A.Z, A.Y);
-  GFp_fe_invert(zminusy_inv, zminusy);
-  GFp_fe_mul(zplusy, zplusy, zminusy_inv);
-  GFp_fe_tobytes(out_public_value, zplusy);
 }
 
 #endif  /* BORINGSSL_X25519_X86_64 */
--- a/src/ec/curve25519/ops.rs
+++ b/src/ec/curve25519/ops.rs
@@ -121,6 +121,79 @@ fn encode_point(x: Elem, y: Elem, z: Elem) -> EncodedPoint {
     bytes
 }
 
+// If the target architecture is non-Windows x86-64, then it is faster to
+// compute base point scalar multiplication using the target-optimized general
+// scalar multiplication.
+#[cfg(all(target_arch = "x86_64", not(windows)))]
+pub fn scalar_mult_with_base_point(scalar: &Scalar) -> EncodedPoint {
+    let mut point = [0u8; ELEM_LEN];
+    unsafe { GFp_x25519_scalar_mult(&mut point, scalar, &BASE_POINT); }
+    point
+}
+
+// If the target architecture is NEON-capable ARM, then it is faster to compute
+// base point scalar multiplication using the target-optimized general scalar
+// multiplication. Otherwise, if NEON is not available, we fall back on the
+// architecture-generic base point multiplication.
+#[cfg(target_arch = "arm")]
+pub fn scalar_mult_with_base_point(scalar: &Scalar) -> EncodedPoint {
+    if is_neon_capable() {
+        let mut point = [0u8; ELEM_LEN];
+
+        // We could use `GFp_x25519_scalar_mult()` here, instead. When compiled
+        // for ARM, that function performs the same NEON capability check we
+        // just did, then calls the following function when it succeeds. Since
+        // we already know we're in a branch where it is appropriate, we can
+        // skip the extra check and call the NEON-optimized implementation.
+        unsafe { GFp_x25519_NEON(&mut point, scalar, &BASE_POINT); }
+
+        return point;
+    }
+
+    scalar_mult_with_base_point_generic(scalar)
+}
+
+#[cfg(any(all(target_arch = "x86_64", not(windows)), target_arch = "arm"))]
+static BASE_POINT: EncodedPoint = [
+    9, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0,
+];
+
+#[cfg(not(any(all(target_arch = "x86_64", not(windows)), target_arch = "arm")))]
+pub fn scalar_mult_with_base_point(scalar: &Scalar) -> EncodedPoint {
+    scalar_mult_with_base_point_generic(scalar)
+}
+
+#[cfg(not(all(target_arch = "x86_64", not(windows))))]
+fn scalar_mult_with_base_point_generic(scalar: &Scalar) -> EncodedPoint {
+    let mut point = ExtPoint::new_at_infinity();
+
+    unsafe { GFp_x25519_ge_scalarmult_base(&mut point, scalar); }
+
+    let mut z_plus_y = [0i32; ELEM_LIMBS];
+    let mut z_minus_y = [0i32; ELEM_LIMBS];
+    let mut z_minus_y_inv = [0i32; ELEM_LIMBS];
+    let mut u_coord = [0i32; ELEM_LIMBS];
+    unsafe {
+        GFp_fe_add(&mut z_plus_y, &point.z, &point.y);
+        GFp_fe_sub(&mut z_minus_y, &point.z, &point.y);
+        GFp_fe_invert(&mut z_minus_y_inv, &z_minus_y);
+        GFp_fe_mul(&mut u_coord, &z_plus_y, &z_minus_y_inv);
+    }
+
+    let mut encoded_point = [0u8; ELEM_LEN];
+    unsafe { GFp_fe_tobytes(&mut encoded_point, &u_coord); }
+
+    encoded_point
+}
+
+#[cfg(target_arch = "arm")]
+fn is_neon_capable() -> bool {
+    unsafe { GFp_is_NEON_capable_at_runtime() == 1 }
+}
+
 extern {
     fn GFp_fe_invert(out: &mut Elem, z: &Elem);
     fn GFp_fe_isnegative(elem: &Elem) -> u8;
@@ -128,4 +201,25 @@ extern {
     fn GFp_fe_tobytes(bytes: &mut EncodedPoint, elem: &Elem);
     fn GFp_x25519_ge_frombytes_vartime(h: &mut ExtPoint, s: &EncodedPoint)
                                        -> c::int;
+}
+
+// Externs needed for the generic base point multiplication impl.
+#[cfg(not(all(target_arch = "x86_64", not(windows))))]
+extern {
+    fn GFp_fe_add(out: &mut Elem, f: &Elem, g: &Elem);
+    fn GFp_fe_sub(out: &mut Elem, f: &Elem, g: &Elem);
+    fn GFp_x25519_ge_scalarmult_base(point: &mut ExtPoint, scalar: &Scalar);
+}
+
+#[cfg(target_arch = "arm")]
+extern {
+    fn GFp_is_NEON_capable_at_runtime() -> u8;
+    fn GFp_x25519_NEON(out: &mut EncodedPoint, scalar: &Scalar,
+                       point: &EncodedPoint);
+}
+
+#[cfg(all(target_arch = "x86_64", not(windows)))]
+extern {
+    fn GFp_x25519_scalar_mult(out: &mut EncodedPoint, scalar: &Scalar,
+                              point: &EncodedPoint);
 }
--- a/src/ec/curve25519/x25519.rs
+++ b/src/ec/curve25519/x25519.rs
@@ -61,14 +61,12 @@ fn x25519_public_from_private(public_out: &mut [u8],
                               -> Result<(), error::Unspecified> {
     let public_out = try!(slice_as_array_ref_mut!(public_out, PUBLIC_KEY_LEN));
 
-    // XXX: This shouldn't require dynamic checks, but rustc can't slice an
-    // array reference to a shorter array reference. TODO(perf): Fix this.
-    let private_key =
-        try!(slice_as_array_ref!(&private_key.bytes[..PRIVATE_KEY_LEN],
-                                 PRIVATE_KEY_LEN));
-    unsafe {
-        GFp_x25519_public_from_private(public_out, private_key);
-    }
+    let mut scalar = [0u8; ops::SCALAR_LEN];
+    scalar.copy_from_slice(&private_key.bytes[..PRIVATE_KEY_LEN]);
+    unsafe { GFp_curve25519_scalar_mask(&mut scalar); }
+
+    *public_out = ops::scalar_mult_with_base_point(&scalar);
+
     Ok(())
 }
 
@@ -100,13 +98,7 @@ fn x25519_ecdh(out: &mut [u8], my_private_key: &ec::PrivateKey,
 }
 
 const ELEM_AND_SCALAR_LEN: usize = ops::ELEM_LEN;
-
-// An X25519 private key as an unmasked scalar.
-type PrivateKey = [u8; PRIVATE_KEY_LEN];
 const PRIVATE_KEY_LEN: usize = ELEM_AND_SCALAR_LEN;
-
-// An X25519 public key as an encoded Curve25519 point.
-type PublicKey = [u8; PUBLIC_KEY_LEN];
 const PUBLIC_KEY_LEN: usize = ELEM_AND_SCALAR_LEN;
 
 // An X25519 shared secret as an encoded Curve25519 point.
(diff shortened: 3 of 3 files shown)


ranweiler:
@briansmith, I'm looking at the NEON part of this now. In reading up on the SIMD/`target_feature` issues, I'm seeing you pop up a bunch, so I'm going to try to save some time by asking: do you have any thoughts on the best approach here?

From what I've seen so far, it seems like we really want to use `#cfg(target_feature = "neon")`, but that needs stabilization. In meantime, for `x25519_public_from_private`, we can do a runtime check using `GFp_is_NEON_capable`, then conditionally call `GFp_x25519_NEON`. That's the approach I'm going to take, unless you had an idea for a deeper solution.

briansmith:
At some point, when we know NEON is available or definitely not available (via the target feature stuff, when that becomes available), we should skip the runtime check. But otherwise we should do the runtime check so that NEON is used when NEON is available, and otherwise fall back to the non-NEON code path.

src/ec/curve25519/ops.rs
```diff
@@ -121,6 +121,72 @@ fn encode_point(x: Elem, y: Elem, z: Elem) -> EncodedPoint {
...
+        GFp_fe_mul(&mut coord_elem, &z_plus_y, &z_minus_y_inv);
+    }
+    let encoded_point = [0u8; ELEM_LEN];
+    encoded_point.copy_from_slice(&coord_elem);
+    encoded_point
+}
+
+#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+fn is_neon_capable() -> bool {
+    unsafe { GFp_is_NEON_capable_at_runtime() == 1 }
```
briansmith: Isn't NEON always available on AAarch64?

src/ec/curve25519/ops.rs
```diff
@@ -121,6 +121,72 @@ fn encode_point(x: Elem, y: Elem, z: Elem) -> EncodedPoint {
...
+    let mut z_minus_y_inv = [0i32; ELEM_LIMBS];
+    let mut coord_elem = [0i32; ELEM_LIMBS];
+    unsafe {
+        GFp_fe_add(&mut z_plus_y, &a.z, &a.y);
+        GFp_fe_sub(&mut z_minus_y, &a.z, &a.y);
+        GFp_fe_invert(&mut z_minus_y_inv, &z_minus_y);
+        GFp_fe_mul(&mut coord_elem, &z_plus_y, &z_minus_y_inv);
+    }
+    let encoded_point = [0u8; ELEM_LEN];
+    encoded_point.copy_from_slice(&coord_elem);
```
briansmith: Can't we use `encoded_point = coord_elem;`?

And, can't we even avoid the copy at all? I guess maybe the C code did this kind of copying as a way of trying to guarantee alignment for all the internal `static` functions and/or to allow the compiler to pass the arguments using a special calling convention. It doesn't seem to make sense here since those things can't happen in Rust->C calling.

src/ec/curve25519/ops.rs
```diff
@@ -121,6 +121,72 @@ fn encode_point(x: Elem, y: Elem, z: Elem) -> EncodedPoint {
...
+        unsafe { GFp_x25519_NEON(&mut result, scalar, point); }
+        return result;
+    }
+
+    scalar_mult_with_base_point_generic(scalar)
+}
+
+#[cfg(any(all(target_arch = "x86_64", not(windows)),
+          any(target_arch = "arm", target_arch = "aarch64")))]
+const BASE_POINT: EncodedPoint = [
```
briansmith: Maybe `static` makes more sense here?

src/ec/curve25519/ops.rs
```diff
@@ -121,6 +121,72 @@ fn encode_point(x: Elem, y: Elem, z: Elem) -> EncodedPoint {
     bytes
 }
 
+#[cfg(all(target_arch = "x86_64", not(windows)))]
+pub fn scalar_mult_with_base_point(scalar: &Scalar) -> EncodedPoint {
+    let mut point = [0u8; ELEM_LEN];
+
+    unsafe { GFp_x25519_scalar_mult(&mut point, scalar, &BASE_POINT); }
+
+    point
```
briansmith: I suggest getting rid of the blank lines here.

src/ec/curve25519/x25519.rs
```diff
@@ -69,33 +68,39 @@ fn x25519_public_from_private(public_out: &mut [u8],
...
+                                 PUBLIC_KEY_LEN));
     bssl::map_result(unsafe {
         GFp_x25519_ecdh(out, my_private_key, peer_public_key)
     })
 }
 
+const ELEM_AND_SCALAR_LEN: usize = ops::ELEM_LEN;
+
+type PrivateKey = [u8; PRIVATE_KEY_LEN];
+const PRIVATE_KEY_LEN: usize = ELEM_AND_SCALAR_LEN;
```
briansmith: I'm not sure we should use “private key” in the Curve25519 code because IMO it is too ambiguous between “seed” and “scalar”/“private scalar”.

ranweiler:
Hi @briansmith, this isn't quite ready for review, particularly the last commit. Feel free to look at if you like, but the last commit will change a lot (I'm just trying to let CI check that I have build stuff set up correctly, atm).

src/ec/curve25519/x25519.rs
```diff
@@ -78,9 +78,18 @@ fn x25519_ecdh(out: &mut [u8], my_private_key: &ec::PrivateKey,
...
-    bssl::map_result(unsafe {
-        GFp_x25519_ecdh(out, my_private_key, peer_public_key)
-    })
+
+    unsafe {
+        GFp_x25519_scalar_mult(out, my_private_key, peer_public_key);
+    }
+
+    let zeros: SharedSecret = [0; SHARED_SECRET_LEN];
+    if constant_time::verify_slices_are_equal(out, &zeros).is_ok() {
```
briansmith: Gross! :) In the future I think we'll be able to eventually replace this with `limbs_are_zero_constant_time`, before we convert the limbs to bytes. OK for now.

briansmith:
Regarding CI: Please set up Travis CI on your own GitHub and push to a branch that isn't associated with this PR. That way, your personal CI testing won't use the resources for this account. (Only one CI build can run at a time; i.e. one PR or one commit, and each one takes a few hours of wall time.)

ranweiler:
> Regarding CI: Please set up Travis CI on your own GitHub and push to a branch that isn't associated with this PR.
Done!

crypto/curve25519/curve25519.c
```diff
@@ -4618,7 +4624,10 @@ static void GFp_fe_mul121666(fe h, fe f) {
   h[9] = (int32_t)h9;
 }
 
-static void x25519_scalar_mask(uint8_t e[32]) {
+/* Prevent -Wmissing-prototypes warnings. */
+void GFp_x25519_scalar_mask(uint8_t e[32]);
```
ranweiler: This function is only compiled if `BORINGSSL_X25519_X86_64` is not defined, which is why I kept it here. Would you like the prototype to be unconditionally declared?

I think this is _untrue_ for `GFp_ed25519_scalar_mask()`, though, so I'll fix that either way.

src/ec/curve25519/ops.rs
ranweiler: Done.

src/ec/curve25519/ops.rs
```diff
@@ -121,6 +121,72 @@ fn encode_point(x: Elem, y: Elem, z: Elem) -> EncodedPoint {
...
+}
+
+#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
```
ranweiler: Hmm, yes. But, looking back at it, the ASM `GFp_x25519_NEON()` does not get compiled for AArch64, only for `__arm__`. This ASM function comes from the DJB implementation in SUPERCOP, which (in its `architectures` file) only lists `arm` and `armeabi`, with `aarch64` appearing in other implementations (so no equivocation is occurring).

So, this specialized implementation is only (possibly) appropriate for 32-bit ARM, and right now, we should actually use the generic base point implementation for AArch64, independent of the NEON support of that architecture.

src/ec/curve25519/ops.rs
ranweiler: I believe so, but this is going away anyway, as mentioned above.

src/ec/curve25519/ops.rs
ranweiler: I ended up adding a comment about what we're calling, so I added the newlines back in.

src/ec/curve25519/ops.rs
ranweiler: Agreed, and that is most consistent with the source implementation. Changed.

src/ec/curve25519/ops.rs
ranweiler: This was wrong as-written. I fixed it, but we still need `encoded_point` because its type differs (`[i32; ELEM_LIMBS]` vs. `[u8; ELEM_LEN]`). In my mind, when we address #509, it will take care of this as well.

src/ec/curve25519/x25519.rs
ranweiler: Yeah, looking forward to doing something better here. I started out adding a new `constant_time` function to test for zero, but thought it best to start with the simplest thing that was obviously correct.

src/ec/curve25519/x25519.rs
ranweiler: Hmm. Well, we got rid of the type alias either way, but the constant remains.

I think, even though we're in a 25519 setting, the naming situation is a little bit different from the Ed25519 case. Here, we aren't holding on to any derived values, and there is no notion of "seed" (since there's no hashing) or explicit "private scalar" in the language of RFC 7748. In my mind, as long as we keep the `private_*` concepts scoped to the relevant protocol modules, it isn't ambiguous— the ambiguity is particular to Ed25519.

In the X25519 case, I don't know what else you would call it— maybe `private_bytes`? That would be accurate, since it isn't a "seed" (we don't hash it), and it isn't sure to be a "scalar" (until masked).

I'm inclined to keep it as-is. What do you think?

ranweiler:
> > Regarding CI: Please set up Travis CI on your own GitHub and push to a branch that isn't associated with this PR.

> Done!

@briansmith, there might be a little CI noise from my last pushes, sorry about that. Since I now have CI set up on my fork, I'll make a point of leaning on that in the future, and avoid clogging the upstream CI with WIP PR's.

src/ec/curve25519/ops.rs
ranweiler: The thing I'd really like to improve here is the treatment of scalars. I'd like `Scalar` to be a _non-alias_ type with named curve phantom data that encodes whether it's been masked for the Montgomery or twisted Edwards operations.

In fact, stepping back, the `scalar_mult_with_base_point` functions should either live in the `x25519` module, or be named such that their fixed relationship to the Montgomery curve is explicit (since they are _not_ valid for each curve form, due to the masking). Or, we can pull the masking out into their caller.

I think right now, I'd like to keep those functions in `ops`, but move the copy and masking into the caller, `x25519_public_from_private()`. How does that sound?

ranweiler: I tried out the above here: https://github.com/ranweiler/ring/commit/7201e6f3bb2fa1d298d6dd257dfd8e316a062a70

Note that the above is also rebased on `master`.

If that looks good, I'll fold it into the "Port `GFp_x25519_public_from_private()` commit (and rebase _this_ branch on `master` while I'm at it).

ranweiler: Yeah, so actually looking into them and comparing against the RFCs, there only needs to be one *25519 scalar mask function, because they accomplish exactly the same thing (which is actually what I'd expect). I'm going to unify these.

briansmith: You are saying that one implementation clears the second-highest bit and then sets it again, whereas the other implementation leaves the second-highest bit unchanged and then sets it, and these steps are equivalent, right? Or am I missing something?

briansmith: > n fact, stepping back, the scalar_mult_with_base_point functions should either live in the x25519 module, 

Shouldn't the stuff shared between X25519 and Ed25519 be put in the containing module, `ring::curve25519`? I think `ring::curve25519::x25519` should just be for the ECDH algorithm and `ring::curve25519::ed25519` should be just for the signature algorithm.

ranweiler: > You are saying that one implementation clears the second-highest bit and then sets it again, whereas the other implementation leaves the second-highest bit unchanged and then sets it, and these steps are equivalent, right? Or am I missing something?
 
Right.

ranweiler: > Shouldn't the stuff shared between X25519 and Ed25519 be put in the containing module, ring::curve25519? I think ring::curve25519::x25519 should just be for the ECDH algorithm and ring::curve25519::ed25519 should be just for the signature algorithm.

Agreed, or `ops`, which is what I'm currently doing in these commits.

ranweiler:
@briansmith, just FYI, you can now consider this ready for review, when you have time.

crypto/curve25519/curve25519.c
```diff
@@ -4520,13 +4521,9 @@ void GFp_x25519_sc_muladd(uint8_t *s, const uint8_t *a, const uint8_t *b,
   s[31] = (uint8_t)(s11 >> 17);
 }
 
-
-/* Prototype to avoid -Wmissing-prototypes warnings. */
-void GFp_ed25519_scalar_mask(uint8_t a[32]);
-
-void GFp_ed25519_scalar_mask(uint8_t a[32]) {
+void GFp_x25519_scalar_mask(uint8_t a[32]) {
```
briansmith: Instead of renaming and modifying the Ed25519 one, could you instead just rename the X25519 one and keep it as-is?

briansmith: Also, I suggest we name it `GFp_curve25519_scalar_mask` since it works for both Ed25519 and X25519.

ranweiler: I think this is just an artifact of the `x25519_scalar_mask()` function only being compiled in some cases, whereas the Ed25519 one was already _unconditionally_ compiled.

The reason I kept the impl with `&= 127` is because it most closely tracked the language in RFC 8032 (step 2 [here](https://tools.ietf.org/html/rfc8032#section-5.1.5)) and the code in RFC 7748 (the `decodeScalar25519` function [here](https://tools.ietf.org/html/rfc7748#section-5)).

Anyway, that was my reasoning. Would you still like to change it to use the `&= 63`, or was that not what you meant?

ranweiler: > Also, I suggest we name it `GFp_curve25519_scalar_mask` since it works for both Ed25519 and X25519.

Agreed— there are some other functions with a `x25519` prefix that are used in the Ed25519 code. Would it be best to change these names too, in one separate commit?

briansmith:
> Anyway, that was my reasoning. Would you still like to change it to use the &= 63, or was that not what you meant?

OK, I see. Never mind my comment.

crypto/curve25519/curve25519.c
```diff
@@ -239,13 +242,16 @@ static void fe_1(fe h) {
...
-static void fe_add(fe h, const fe f, const fe g) {
+void GFp_fe_add(fe h, const fe f, const fe g) {
   unsigned i;
   for (i = 0; i < 10; i++) {
     h[i] = f[i] + g[i];
   }
 }
 
+/* Prototype to avoid -Wmissing-prototypes warnings. */
+void GFp_fe_sub(fe h, const fe f, const fe g);
```
briansmith: Ditto.

crypto/curve25519/curve25519.c
```diff
@@ -1069,12 +1075,12 @@ static void ge_p2_dbl(ge_p1p1 *r, const ge_p2 *p) {
...
+  GFp_fe_add(r->Y, p->X, p->Y);
   fe_sq(t0, r->Y);
-  fe_add(r->Y, r->Z, r->X);
-  fe_sub(r->Z, r->Z, r->X);
-  fe_sub(r->X, t0, r->Y);
-  fe_sub(r->T, r->T, r->Z);
+  GFp_fe_add(r->Y, r->Z, r->X);
+  GFp_fe_sub(r->Z, r->Z, r->X);
+  GFp_fe_sub(r->X, t0, r->Y);
+  GFp_fe_sub(r->T, r->T, r->Z);
```
briansmith: This is a really performance-sensitive function. Does this change the code generation or performance in any substantial way?

Maybe a more conservative change would be to have  `GFp_fe_sub()` and `GFp_fe_add()` be `extern` wrappers around a `static` `fe_sub` and `fe_add`. Then we don't have to worry about potential performance implications as much (especially on ARM). We could then clean it up if/when we have a good story for verifying performance on non-X86_64 hardware. WDYT?

src/ec/curve25519/ops.rs
```diff
@@ -121,6 +121,79 @@ fn encode_point(x: Elem, y: Elem, z: Elem) -> EncodedPoint {
     bytes
 }
 
+// If the target architecture is non-Windows x86-64 or NEON-capable ARM, it is
+// faster to do scalar base point multiplication using the generic scalar
+// multiplication implementation.
+#[cfg(all(target_arch = "x86_64", not(windows)))]
+pub fn scalar_mult_with_base_point(scalar: &Scalar) -> EncodedPoint {
```
briansmith: This is an `EncodedPoint` in the X25519 sense, not the Ed25519 sense, right? It would be nice to find a way to make this clearer.

src/ec/curve25519/ops.rs
```diff
@@ -121,6 +121,79 @@ fn encode_point(x: Elem, y: Elem, z: Elem) -> EncodedPoint {
...
+    let mut point = [0u8; ELEM_LEN];
+    unsafe { GFp_x25519_scalar_mult(&mut point, scalar, &BASE_POINT); }
+    point
+}
+
+// If the target architecture is NEON-capable 32-bit ARM, then base point
+// scalar multiplication can be done faster using the architecture-optimized
+// generic scalar multiplication implementation. When the `target_feature`
+// compilation attribute is stabilized, we can remove the runtime check.
```
briansmith: Let's remove “When the `target_feature` compilation attribute is stabilized, we can remove the runtime check.” More likely, we'll add a feature that says "replace runtime CPU capability checks with compile-time CPU capability checks" and that will allow for the removal of the runtime check when that bit is flipped. But there are always going to be platforms (e.g. some 32-bit Android) where we will do a runtime check, indefinitely.

crypto/curve25519/curve25519.c
ranweiler: >  I do think it would be good to (eventually) change the shared function's names to `curve25519` instead of `x25519`.

Agreed, I've made a note to do this.

src/ec/curve25519/ops.rs
ranweiler: > This is an EncodedPoint in the X25519 sense, not the Ed25519 sense, right?

Exactly.

> It would be nice to find a way to make this clearer.

My plan has been, after the existing changes were approved, to add more commits which add a statically-checked type-level distinction between the different kinds of encoded points, and between cases where we pass true `Scalars` (masked) and byte arrays that _will become_ scalars (still need to be masked). My understanding is that there was a little bit of uncertainty still about how much of a win some of those tactics were in the RSA/Montgomery arithmetic work, so I held off to avoid dumping a ton of commits on you for review.

I'll make a run at that _after_ I loop back and address your other comments. At a minimum, we can improve the use of aliases.

crypto/curve25519/curve25519.c
ranweiler: > Maybe a more conservative change would be [...]

Yes, I want to do the cautious thing here. I have a few ideas for enabling non-x86-64 benchmarking. For now, I'm going to adopt your `GFp_` wrapper suggestion.

src/ec/curve25519/x25519.rs
ranweiler: Done in #538.

src/ec/curve25519/x25519.rs
```diff
@@ -69,33 +68,39 @@ fn x25519_public_from_private(public_out: &mut [u8],
...
     })
 }
 
+const ELEM_AND_SCALAR_LEN: usize = ops::ELEM_LEN;
+
+type PrivateKey = [u8; PRIVATE_KEY_LEN];
+const PRIVATE_KEY_LEN: usize = ELEM_AND_SCALAR_LEN;
+
+type PublicKey = [u8; PUBLIC_KEY_LEN];
+const PUBLIC_KEY_LEN: usize = ELEM_AND_SCALAR_LEN;
```
ranweiler: Done in #538, also for `SharedSecret`.

crypto/curve25519/curve25519.c
ranweiler: > Agreed, I've made a note to do this.

That is: I changed the `GFp_x25519_scalar_mask()` to `GFp_curve25519_scalar_mask()`, and will change the others separately.

ranweiler:
> The first commits that are doing cleanup seem really straightforward and I think we may be better off splitting them off into a separate PR.

I split those commits out as #538. I'm responding to your other feedback in a non-PR branch on my account, and will reset this PR's branch against that after #538 has been merged.

src/ec/curve25519/ops.rs
ranweiler: Done.

crypto/curve25519/curve25519.c
```diff
@@ -230,6 +230,9 @@ static void fe_1(fe h) {
   h[0] = 1;
 }
 
+/* Prototype to avoid -Wmissing-prototypes warnings. */
+void GFp_fe_add(fe h, const fe f, const fe g);
+
```
ranweiler: Fixed (and now the prototype is for a _wrapper_ for the original `static` function).

crypto/curve25519/curve25519.c
ranweiler: Fixed.

ranweiler:
@briansmith:

> This is an EncodedPoint in the X25519 sense, not the Ed25519 sense, right? It would be nice to find a way to make this clearer.

> `type EncodedPoint = [u8; ELEM_LEN];`
> Is there a way to use two different Rust types, one for X25519 and one for Ed25519?

These are the only comments I haven't addressed in my update of this branch. My current thinking is: we have the `EncodedPoint` alias overloading _prior_ to this PR. I'd like to improve the typing on the Rust side of this code more pervasively: addressing the overall distinction between Edwards and Montgomery forms, and improving the `Scalar` types. Unless you prefer I fold a specific, narrower type improvement into one of the commits already in this PR, I suggest I do that in a fresh PR, after this has been merged. Happy to go about it however you prefer!

briansmith:
> I suggest I do that in a fresh PR, after this has been merged. Happy to go about it however you prefer!

That sounds fine with me.

I merged the first three commits from this PR already, and I cherry-picked the typo fix as well. Please rebase the one remaining commit on top of the current master.

Do you have a concrete idea for the ARM performance testing already? If so, it would be good to discuss it somewhere. In particular, I can see now that making `GFp_x25519_ge_scalarmult_base()` extern like we did before already runs the same risks as we discussed above w.r.t. possible performance impact. Also it's going to be difficult to do the rest of the oxidation work here without resolving the perf impact concern. I'll see if I can find a way to start measuring this.

ranweiler:
Rebased! Will follow up via email re ARM perf testing.

briansmith:
@ranweiler I am not sure if you are still interested in this, but I have an update

Background: Around the time you submitted this, other people were working on a formally-verified curve25519 implementation to be integrated into BoringSSL, and I deferred merging this until I could learn more about the scope of that work. Now that work has been merged into *ring* and it turns out that, IIUC, the formally-verified portions of the code are mutually exclusive with the changes in this PR. That is to say, AFAICT this PR isn't replacing any formally-verified C code with unverified Rust code. So, I do think it is a good idea to proceed with this.

Please LMK, here or privately, what the best way too proceed is. Thanks.

ranweiler:
@briansmith thanks for the update/context! I'll pick this back up (along with #519).

briansmith:
@ranweiler I'm going to close this PR. If/when you are interested in picking it back up, please submit a new PR. Thanks for your efforts here!

Issue #413, opened 2017-01-13
Implement ChaCha20-Poly1305-SIV(-ish) Nonce-Misuse-Resistant AEAD


briansmith:
[I've not thought deeply about whether this makes sense. I'm just recording it so we can start making progress on nonce-misuse-resistant AEADs.]

See #412 about adding AES-GCM-SIV. The same reasons that make ChaCha20-Poly1305 useful (vs. AES-GCM) seem to imply that a nonce-misuse-resistant variant of ChaCha20-Poly1305 is warranted.

It seems like we can generalize AES-GCM-SIV and then instantiate the generalized form with ChaCha20-Poly1305 to get a Nonce-Misuse-Resistant AEAD with similar properties, which seem like they would be better than what XChaCha20-Poly1305 (#411) would offer. Interestingly, our ChaCha20-Poly1305 code already computes ChaCha20 and Poly1305 serially, unlike (some variants of) our AES-GCM code, which interleaves them. Thus it seems like ChaCha20-Poly1305-SIV should perform at about the same speed as our ChaCha20-Poly1305 implementation. (This differs from AES-GCM-SIV vs AES-GCM on some architectures, because some AES-GCM implementations exploit the ability to compute AES-CTR and GCM in parallel.)

The [AES-GCM-SIV paper](https://eprint.iacr.org/2015/102.pdf) already defines a generalized Cipher-Authenticator-SIV construct similar to what we need. However, the instantiation for AES-GCM doesn't map 1-to-1 to what the instantiation for ChaCha20-Poly1305 would be due to inherent differences, IIUC. However, having just read the AES-GCM-SIV draft IETF RFC and the AES-GCM-SIV paper again, it doesn't seem too hard. However, I could very well be missing something really important.

briansmith:
[...]

PR #526, opened 2017-04-28
Add a simple example for aead

jaysonsantos:
Hi, I've added this to try to help newcomers (like me) when using a cryptography library to avoid common errors and to have something to run ASAP. 
Let me know if I wrote some BS and if this is the rigt place to put it.


src/aead/aead.rs
```diff
@@ -25,6 +25,82 @@
...
+//!
+//!    // Usually the salt has some random data and something that relates to the user
+//!    // like an username and it must be 8 bytes (64 bits)
+//!    let salt = [0, 1, 2, 3, 4, 5, 6, 7];
+//!
+//!    // Keys must have 32 bytes and if you want to use an user password for it you have
+//!    // to use an algorithm to derive it (in this case pbkdf2).
+//!    let mut key = [0; 32];
+//!    derive(&HMAC_SHA256, 100, &salt, &password[..], &mut key);
+//!
```
Philipp91: I'd cut off the example here (move the part above elsewhere) and show how to generate a random key here, instead. Many users of other libraries (not ring) fail to generate random keys and hard-code something instead.

Next to the line where the generation happens, you can then explain that the key can also be derived from the user's password or sth. like that and then point to the part of the code above (moved to a second code-block here or to the `pbkdf2` module).

Added benefit: this code sample becomes shorter (it's a bit too long and users might skip it for that reason).

src/aead/aead.rs
```diff
@@ -25,6 +25,82 @@
...
+//!    // Additional data that you would like to send and it would not be encrypted but it
+//!    // would be signed
+//!    let additional_data: [u8; 0] = [];
+//!
+//!    // Ring uses the same input variable as output
+//!    let mut in_out = content.clone();
+//!
+//!    // The input/output variable need some space to store the tag which will have the
+//!    // signature
+//!    println!("Tag len {}", CHACHA20_POLY1305.tag_len());
```
Philipp91: Remove this line?

src/aead/aead.rs
```diff
@@ -25,6 +25,82 @@
...
+//!    for _ in 0..CHACHA20_POLY1305.tag_len() {
+//!        in_out.push(0);
+//!    }
+//!
+//!    // Opening key used to decrypt data
+//!    let opening_key = OpeningKey::new(&CHACHA20_POLY1305, &key).unwrap();
```
Philipp91: It is counter-intuitive to initialize the `OpeningKey` here -- it should be moved behind all the encryption code. In real-world code, this would be in a different function.

src/aead/aead.rs
```diff
@@ -25,6 +25,82 @@
...
+//!    // would be signed
+//!    let additional_data: [u8; 0] = [];
+//!
+//!    // Ring uses the same input variable as output
+//!    let mut in_out = content.clone();
+//!
+//!    // The input/output variable need some space to store the tag which will have the
+//!    // signature
+//!    println!("Tag len {}", CHACHA20_POLY1305.tag_len());
+//!    for _ in 0..CHACHA20_POLY1305.tag_len() {
```
Philipp91: `in_out.resize(in_out.len() + CHACHA20_POLY1305.tag_len(), 0u8);

src/aead/aead.rs
```diff
@@ -25,6 +25,82 @@
...
+//!    let mut nonce = vec![0; 12];
+//!
+//!    // Fill nonce with random data
+//!    let rand = SystemRandom::new();
+//!    rand.fill(&mut nonce).unwrap();
+//!
+//!    // Encrypt data into in_out variable
+//!    let output_size = seal_in_place(&sealing_key, &nonce, &additional_data, &mut in_out,
+//!                                    CHACHA20_POLY1305.tag_len()).unwrap();
+//!
```
Philipp91: You should use `output_size` here to `.truncate()` the ciphertext, which is now in `in_out`.

src/aead/aead.rs
```diff
@@ -25,6 +25,82 @@
...
+//!    // Opening key used to decrypt data
+//!    let opening_key = OpeningKey::new(&CHACHA20_POLY1305, &key).unwrap();
+//!
+//!    // Sealing key used to encrypt data
+//!    let sealing_key = SealingKey::new(&CHACHA20_POLY1305, &key).unwrap();
+//!
+//!    // Random data must be used only once per encryption and ideally if you encrypt blocks
+//!    // it should be generated on per block see an example of poorly generated nonces
+//!    // where a new nonce is generated at every few pixels at
+//!    // https://github.com/jaysonsantos/bad-encryption-example
```
Philipp91: This may be true (I'm not a crypto expert), but the explanation is too hard to understand.

I reckon average users wouldn't encrypt every pixel of an image individually, though (as you seem to do here https://github.com/jaysonsantos/bad-encryption-example/blob/master/src/main.rs#L69), but rather the whole thing at once. Similarly, they would encrypt an entire text at once, and not the individual characters. And average users don't have "blocks", they just have their data. So the only explanation needed here is: "Never reuse a nonce, that is, always generate a fresh nonce (e.g. with `rand.fill()`) for *every* `seal_in_place` call you make."

jaysonsantos: The idea of encrypting pixel by pixel is to have a visual analogy of what would happen if you encrypt many files/something else using the same nonce, I think it illustrates well what could happen because if you are not using with cryptography stuff you simply think that as it is not the secret itself you could keep reusing it

src/aead/aead.rs
jaysonsantos: That was a problem when I started playing with crypto algos and would be nice to say that the password can be converted into keys using derive. Do you think it is worth it to remove thinking that it is a common problem users find when using cryptography for the first time?

src/aead/aead.rs
jaysonsantos: Wouldn't I lose the tag? I thought it had the signed data.

src/aead/aead.rs
```diff
@@ -25,6 +25,82 @@
 //! [AEAD]: http://www-cse.ucsd.edu/~mihir/papers/oem.html
 //! [`crypto.cipher.AEAD`]: https://golang.org/pkg/crypto/cipher/#AEAD
 
+//!
+//! **Example (if you don't know what you are doing, read this first)**
```
jaysonsantos: Do you have a suggestion for a start line?

src/aead/aead.rs
jaysonsantos: I think I didn't follow here. Do you think I should move it even being an example thing?

src/aead/aead.rs
Philipp91: Brian wrote:

> I don't think we should teach people how to derive a key from a password in this example. This example scheme is actually not as secure as we'd like it to be w.r.t. the nonce handling.

I can't find that comment anymore here. Without understaind his exact reasons (again, I'm not a crypto expert, so I can't judge), I'd say the example code shouldn't demonstrate the derivation if it's not safe.

OTOH: Yes, it is a common problem, and it's common to derive a secret from a password and then encrypt something with it. So eventually, this (upper) part of the sample code will be relevant -- don't throw it away. I hope that *ring* will soon offer a safe solution to do that? Maybe it has to do with nonce-misuse-resistant ciphers? https://github.com/briansmith/ring/issues/413

Once the key derivation sample becomes relevant, it can be debated whether it should be placed here or elsewhere. Either way, I think it should not be in this same *block*  of code.

src/aead/aead.rs
Philipp91: I think it should be in line 95, because after line 94 you can (mentally or actually) draw a line -- that's where the encryption is complete. The `OpeningKey` is not needed for the encryption operation, so it should be below that "line".

src/aead/aead.rs
briansmith: Don't try to explain what an AEAD is in this example. We already try to do that and link to a paper about it above.

src/aead/aead.rs
Philipp91: The tag's length is not guaranteed to be `out_suffix_capacity`. So it might be that you `resize()` the vector from 80 to 100 to make room for the tag, but then it turns out the tag is only 15 long, so your ciphertext is 95. That's the number returned from `seal_in_place`.

I don't know if that difference actually ever occurs -- it might well be that `output_size` always matches the size of the vector, so a `.truncate()` call wouldn't do anything at all. Nevertheless, that's how the API is designed (if I understand it correctly, that is).

So now, AFAIK you wouldn't loose the tag, it's included in that size. Just try out, if the decryption fails, something went wrong, otherwise it's correct.

briansmith:
First, thanks for doing this!

I accidentally submitted some of my draft feedback earlier than I intended to, so then I had to submit more before I intended to. I will review this carefully in a few days.

src/aead/aead.rs
Philipp91: > This module is what you need when you want to encrypt something symmetrically (using the same secret key for encryption and decryption). Instead of using unauthenticated encryption (AES-ECB, AES-CBC, AES-CTR and the like), you probably want to use authenticated encryption (aka. AEAD) as implemented in this module -- unless you know what you are doing.

The text deliberately mentions the things that it replaces (ECB, CBC) because experiments I conducted (with Rust/crypto beginners) have showed that people search for those (on the web and in the documentation) and find other implementations, so mentioning these helps potential *ring* users find what they thought they needed.

briansmith: The scope of this PR is to add an example. If we want to add documentation beyond the example then let's do it as a separate PR. In particular, I don't want to expand the scope of what's required of @jaysonsantos way beyond what he originally was intending.

jaysonsantos:
@briansmith 
> You can't use PBKDF2 to derive a key from a password and then use that key for AEAD operations without having a secure way to generate nonces. (I don't want to encourage people to use random nonces with these algorithms.)
I changed it in my code but could you explain why this is a problem? If you derive from a password and generate nonces with SystemRandom isn't it enough? And it seems it is a common thing to do when creating keys from passwords.

src/aead/aead.rs
jaysonsantos: done in https://github.com/briansmith/ring/pull/526/files#diff-049cb6a3c2f0c7b0a6d0ce40f58461ddR85

jaysonsantos:
I also changed the header to `# Example` but it will render it literally instead of a <hN> and I couldn't not find the difference from this to signature's or pbkdf2's docs

briansmith:
> generate nonces with SystemRandom isn't it enoug

The nonces here are only 96 bits, which means random nonce collision probability is 1/2**48 assuming a perfect PRNG. Some people may be OK with that, but I don't want to advocate for something like that in the *ring* examples.

src/aead/aead.rs
```diff
@@ -24,6 +24,75 @@
...
+//!
+//!use ring::aead::*;
+//!use ring::rand::SystemRandom;
+//!
+//!fn main() {
+//!    let rand = SystemRandom::new();
+//!
+//!    // Usually the salt has some random data and something that relates to the user
+//!    // like a username and it must be 8 bytes (64 bits)
+//!    let salt = [0, 1, 2, 3, 4, 5, 6, 7];
```
isonmad: This variable is unused since 20c9835364bdbfce80abe3f002037e336e4d13e0.

src/aead/aead.rs
```diff
@@ -24,6 +24,71 @@
...
 //! [AEAD]: http://www-cse.ucsd.edu/~mihir/papers/oem.html
 //! [`crypto.cipher.AEAD`]: https://golang.org/pkg/crypto/cipher/#AEAD
+//!
+//! # Examples
+//!
+//!```rust
+//!extern crate ring;
+//!
+//!use ring::aead::*;
+//!use ring::rand::SystemRandom;
```
isonmad: Oops, I missed this before. You need something like `use ring::rand::{SystemRandom, SecureRandom};` to avoid build failures.

jaysonsantos: That is true, i saw some buildings that passed and I thought it was working.

jaysonsantos: And actually, cargo test works on my machine, I will try to figure out before changing it.

briansmith: You probably will need the thing @isonmad suggested if/when you rebase it on the current master.

jaysonsantos:
I just rebased it with upstream's master

cedenday:
Sorry, I didn't see this, and submitted #536.

cedenday:
Why are you using CHACHA20_POLY1305 not one of the AES_GCM variants? Are they suspected to be compromised?

jaysonsantos:
@cedenday 
> Why are you using CHACHA20_POLY1305 not one of the AES_GCM variants? Are they suspected to be compromised?
Because is the one I was using at the time and it is faster on systems that don't have processor instructions for AES.

cedenday:
@briansmith What is blocking the acceptance of this pull request?

briansmith:
> @briansmith What is blocking the acceptance of this pull request?

The AEAD API is getting replaced real soon now, which will obsolete this. Sorry.

PR #534, opened 2017-05-08
add failing elem_inverse_invertible test

kcking:
test case generated using https://play.golang.org/p/d6QemIaeK8


briansmith:
Thanks for the PR! The test fails because the input values are too small. This function is only guaranteed to work with (composite) moduli of ~2047 bits or larger, since that is what RSA needs. Some of the test vectors are smaller than 2047 bits, but that's because I was lazy and they happen to currently work.

kcking:
Ah thank you, I missed that.

Do you know any rust libraries that handle smaller modular arithmetic?

PR #539, opened 2017-05-16
Remove stray spaces in comment text

ranweiler:
This is a very small nit I noticed when examining the type-level programming happening in this module. FWIW, I grepped through the other modules for similar typos, and only found occurrences here.


briansmith:
Thanks. Cherry-picked as 9030b8d27ad73c02a10a476f6a7809c609aee4cf.

PR #549, opened 2017-05-27
build: use -shared when building objects with musl

### LINKED ISSUE #544: musl/amd64: relocation R_X86_64_32 against hidden symbol can not be used when making a shared object
Trying to build a project which depends on ring (current release, 0.9.7) for musl-amd64 target fails at linktime with a relocation error.

The environment is an up-to-date debian stretch (next stable), with musl 1.1.16, ld.bfd 2.28 and rustc 1.17.0.

This can be reproduced by building any of the examples from this [wip branch](https://github.com/lucab/dkregistry-rs/tree/ups/hyper-rustls):
```
$ cargo build --target x86_64-unknown-linux-musl --release --example trace
[...]
error: linking with `cc` failed: exit code: 1
  |
[...]
  = note: /usr/bin/ld: dkregistry-rs/target/x86_64-unknown-linux-musl/release/deps/libring-4ab2a2cec4828006.rlib(e_aes.o): relocation R_X86_64_32 against hidden symbol `GFp_AES_encrypt' can not be used when making a shared object
          /usr/bin/ld: dkregistry-rs/target/x86_64-unknown-linux-musl/release/deps/libring-4ab2a2cec4828006.rlib(curve25519.o): relocation R_X86_64_32 against `.rodata.kMongomeryBasePoint.3027' can not be used when making a shared object; recompile with -fPIC
          /usr/bin/ld: dkregistry-rs/target/x86_64-unknown-linux-musl/release/deps/libring-4ab2a2cec4828006.rlib(ecp_nistz256.o): relocation R_X86_64_32 against `.rodata.GFp_nistz256_precomputed' can not be used when making a shared object; recompile with -fPIC
          /usr/bin/ld: dkregistry-rs/target/x86_64-unknown-linux-musl/release/deps/libring-4ab2a2cec4828006.rlib(gfp_p384.o): relocation R_X86_64_32 against `.rodata.Q' can not be used when making a shared object; recompile with -fPIC
          /usr/bin/ld: dkregistry-rs/target/x86_64-unknown-linux-musl/release/deps/libring-4ab2a2cec4828006.rlib(gcm.o): relocation R_X86_64_32 against `.rodata.ZEROS.2281' can not be used when making a shared object; recompile with -fPIC
          /usr/bin/ld: final link failed: Nonrepresentable section on output
          collect2: error: ld returned 1 exit status
```

lucab:
rustc musl targets have an hardcoded `-static` flag, which results
in ring *.o artifacts not being properly relocatable despite the -fPIC
flag.
This adds a later `-shared` flag to override this behavior when building
library objects.

Fixes #544

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -610,6 +610,7 @@ fn cc(file: &Path, ext: &str, target: &Target, warnings_are_errors: bool,
         // http://www.openwall.com/lists/musl/2015/02/04/3
         // http://www.openwall.com/lists/musl/2015/06/17/1
         let _ = c.flag("-U_FORTIFY_SOURCE");
+        let _ = c.flag("-shared");
     }
     if target.os() == "android" {
         // Define __ANDROID_API__ to the Android API level we want.


isonmad:
What about filing a bug upstream about fixing it at the source https://github.com/alexcrichton/gcc-rs/commit/f308ef2121a812c07718e15e689cbb1c90778513 ?

lucab:
@isonmad thanks, I didn't realize this was coming from gcc-rs, I thought it was directly from target definition. I reported it at https://github.com/alexcrichton/gcc-rs/issues/165. However, I'm not sure that this is going to be reverted there, as the target assumes static linking and here we are just hitting a corner-case while building intermediate artifacts.

lucab:
Bump, I'd need this to avoid build failure in a dependent crate, can we perhaps get this merged as a stop-gap? Even if the ticket at gcc-rs gets resolved removing the hardcoded static flag, ring would still need some logic here to opt-in shared linking.

briansmith:
@lucab I have to admit I am hesitating mostly because I don't fully understand what `-shared` does. It seems like it benign in cases where it isn't helpful? One description I read said that it would only affect which static libraries are forced to be linked, and didn't affect the generated code, but didn't say anything about relocations.

Nobody has needed it until now. This makes me wonder if there's another part of your project that is doing things differently with `-fpic` and/or `-fPIC` than gcc-rs and/or the rest of the infrastructure expects? But also, again it seems like it would not hurt anything. But I don't like doing things I don't fully understand.

So, it would be good to get a link or to to a description of what `-shared` really does. Especially in this kind of situation where, IIUC, no shared libraries are actually involved?

lucab:
> This makes me wonder if there's another part of your project that is doing things differently with -fpic and/or -fPIC than gcc-rs and/or the rest of the infrastructure expects

The source project is linked in the original issue, there are no other non-pure-rust dependencies (that I'm aware of). Any other flag mismatch would be between ring/gcc-rs and the musl target itself.

> So, it would be good to get a link or to to a description of what -shared really does. Especially in this kind of situation where, IIUC, no shared libraries are actually involved?

According to gcc manpage: `Produce a shared object which can then be linked with other objects to form an executable.  Not all systems support this option.  For predictable results, you must also specify the same set of options used for compilation (-fpic, -fPIC, or model suboptions) when you specify this linker option`.

I'm not a linker expert, but this seems to be case: ring is building several objects (e.g. e_aes.o) which are then bundled into a rust lib (i.e. libring-[hash].rlib) and finally linked together into an executable (i.e. my project).

PR #550, opened 2017-05-29
Updated a comment to no longer refer to an old method

alex:
`generate_serializable()` doesn't exist anymore as far as I can tell.


briansmith:
Thanks for noticing this! I ended up just removing the whole function in 313116de6e9df6c35123563bed49b8531b570e0d.

Issue #542, opened 2017-05-18
Implement `PartialEq` and `Eq` for all public Algorithm types


briansmith:
I suggest that for each Algorithm type, we define a simple enum, like this:

```rust
#[derive(Eq)]
pub struct WhateverAlgorithm {
    // existing fields
    id: WhateverAlgorithmID, // Add this field
}

#[derive(PartialEq, Eq)]
#[allow(non_snake_case_types)]
enum WhateverAlgorithmID { // Not public
     FIRST_ALGORITHM,
     SECOND_ALGORITHM,
     // etc.
}

impl PartialEq for WhateverAlgorithm {
    fn eq(&self, other: &Self) -> bool { self.id == other.id }
}
```

The use case for this to make it easier to embed references to algorithms into other (external user) types that want to implement `PartialEq` and `Eq`.

mrwhythat:
Just submitted PR #553 for this.

jeffgran:
Looks like this should be closed?

briansmith:
> Looks like this should be closed?

Some of the work has been done but there's more work to do.

briansmith:
I think this is done. If there's anything missing, please submit a PR.

PR #553, opened 2017-06-23
Implement PartialEq and Eq for algorithm types

mrwhythat:
PR for #542 

`#[derive(Eq)]` seems to not work for algorithm types because of other fields (with `fn` types for example), so explicit `impl Eq for Type {}` were added.

Also, some unintended diff in `aead/chacha20_poly1305.rs` and `agreement.rs`, because I use `rustfmt` tool in my emacs. If this is a problem, I will restore it.

### DIFF
--- a/src/aead/aead.rs
+++ b/src/aead/aead.rs
@@ -279,6 +279,7 @@ pub struct Algorithm {
              tag_out: &mut [u8; TAG_LEN]) -> Result<(), error::Unspecified>,
 
     key_len: usize,
+    id: AlgorithmID,
 }
 
 impl Algorithm {
@@ -309,6 +310,19 @@ impl Algorithm {
     pub fn nonce_len(&self) -> usize { NONCE_LEN }
 }
 
+#[allow(non_camel_case_types)]
+#[derive(Eq, PartialEq)]
+enum AlgorithmID {
+    AES_128_GCM,
+    AES_256_GCM,
+    CHACHA20_POLY1305,
+}
+
+impl PartialEq for Algorithm {
+    fn eq(&self, other: &Self) -> bool { self.id == other.id }
+}
+
+impl Eq for Algorithm {}
 
 /// The maximum length of a tag for the algorithms in this module.
 pub const MAX_TAG_LEN: usize = TAG_LEN;
--- a/src/aead/chacha20_poly1305.rs
+++ b/src/aead/chacha20_poly1305.rs
@@ -24,6 +24,7 @@ pub static CHACHA20_POLY1305: aead::Algorithm = aead::Algorithm {
     init: chacha20_poly1305_init,
     seal: chacha20_poly1305_seal,
     open: chacha20_poly1305_open,
+    id: aead::AlgorithmID::CHACHA20_POLY1305,
 };
 
 /// Copies |key| into |ctx_buf|.
--- a/src/agreement.rs
+++ b/src/agreement.rs
@@ -96,6 +96,12 @@ pub struct Algorithm {
     pub i: ec::AgreementAlgorithmImpl,
 }
 
+impl PartialEq for Algorithm {
+    fn eq(&self, other: &Self) -> bool { self.i.curve.id == other.i.curve.id }
+}
+
+impl Eq for Algorithm {}
+
 /// An ephemeral private key for use (only) with `agree_ephemeral`. The
 /// signature of `agree_ephemeral` ensures that an `EphemeralPrivateKey` can be
 /// used for at most one key agreement.
--- a/src/ec/suite_b/ecdsa.rs
+++ b/src/ec/suite_b/ecdsa.rs
@@ -26,8 +26,23 @@ use untrusted;
 pub struct ECDSASigningAlgorithm {
     curve: &'static ec::Curve,
     pkcs8_template: &'static pkcs8::Template,
+    id: ECDSASigningAlgorithmID
 }
 
+#[allow(non_camel_case_types)]
+#[derive(PartialEq, Eq)]
+enum ECDSASigningAlgorithmID {
+    ECDSA_P256_SHA256_FIXED_SIGNING,
+    ECDSA_P384_SHA384_FIXED_SIGNING,
+    ECDSA_P256_SHA256_ASN1_SIGNING,
+    ECDSA_P384_SHA384_ASN1_SIGNING,
+}
+
+impl PartialEq for ECDSASigningAlgorithm {
+    fn eq(&self, other: &Self) -> bool { self.id == other.id }
+}
+
+impl Eq for ECDSASigningAlgorithm {}
 
 /// An ECDSA verification algorithm.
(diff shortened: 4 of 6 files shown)


src/aead/aead.rs
```diff
@@ -309,6 +310,19 @@ impl Algorithm {
...
+    AES_128_GCM,
+    AES_256_GCM,
+    CHACHA20_POLY1305,
+}
+
+impl PartialEq for Algorithm {
+    fn eq(&self, other: &Self) -> bool { self.id == other.id }
+}
+
+impl Eq for Algorithm {}
```
briansmith: I think `Eq` can be derived and only `PartialEq` needs to be explicitly implemented, right? Assuming so, let's always derive `Eq` instead of doing `impl Eq for ...`.

src/aead/chacha20_poly1305.rs
```diff
@@ -24,19 +24,20 @@ pub static CHACHA20_POLY1305: aead::Algorithm = aead::Algorithm {
     init: chacha20_poly1305_init,
     seal: chacha20_poly1305_seal,
     open: chacha20_poly1305_open,
+    id: aead::AlgorithmID::CHACHA20_POLY1305,
 };
 
 /// Copies |key| into |ctx_buf|.
 pub fn chacha20_poly1305_init(ctx_buf: &mut [u8], key: &[u8])
-                              -> Result<(), error::Unspecified> {
+    -> Result<(), error::Unspecified> {
```
briansmith: Probably we should reformat the code, but not in this PR. Please remove all the unrelated formatting changes.

src/agreement.rs
```diff
@@ -94,8 +94,28 @@ pub struct Algorithm {
...
+
+/// Key agreement algorithm types.
+#[derive(PartialEq, Eq)]
+#[allow(non_camel_case_types)]
+pub enum AlgorithmID {
+    /// Defined in 'ec/suite_b/ecdh.rs'
+    ECDH_P256,
+    /// Defined in 'ec/suite_b/ecdh.rs'
+    ECDH_P384,
+    /// Defined in 'ec/curve25519/x25519.rs'
```
briansmith: Let's drop all the "Defined in" comments.

src/agreement.rs
```diff
@@ -94,8 +94,28 @@ pub struct Algorithm {
     // `ec::AgreementAlgorithmImpl` isn't public.
     #[doc(hidden)]
     pub i: ec::AgreementAlgorithmImpl,
+    #[doc(hidden)]
+    pub id: AlgorithmID,
+}
+
+/// Key agreement algorithm types.
+#[derive(PartialEq, Eq)]
+#[allow(non_camel_case_types)]
```
briansmith: Let's put the `allow(...)` ahead of `derive(...)` in this and the other ones. No order seems better than any other, so let's things in alphabetical order so we don't have to decide another order.

src/ec/suite_b/ecdsa.rs
```diff
@@ -50,6 +66,12 @@ enum ECDSAVerificationAlgorithmID {
     ECDSA_P384_SHA384_FIXED,
 }
 
+impl PartialEq for ECDSAVerificationAlgorithm {
+    fn eq(&self, other: &Self) -> bool { self.id == other.id }
+}
```
briansmith: I think we should not implement PartialEq for this type. Instead we should implement PartialEq for `signature::VerificationAlgorithm` in some way. Maybe we need to make `signature::VerificationAlgorithm` a struct that can hold the id field first. WDYT?

src/aead/aead.rs
mrwhythat: Well, I tried deriving `Eq` first, but it broke the build. It seems that non-`Eq` fields make it impossible to derive. Or maybe it's the ordering. I'll research.

mrwhythat:
@briansmith , just committed requested changes

src/aead/aead.rs
mrwhythat: Did some research and it seems that `derive(Eq)` simply checks for every field, it does not look for existing required impls, like Haskell does.

src/ec/suite_b/ecdsa.rs
mrwhythat: Well, there is `PartialEq` implementation for `ECDSASigningAlgorithm`, so implementing it for `ECDSAVerificationAlgorithm` only benefits the symmetricity. Also, `signature::VerificationAlgorithm` is a very elegant single-method protocol, which I'm a big fan of, and which modern cryptographic libraries lack, so I think coercing it to struct is a bad idea, but it may just be my irrational thinking:)

src/aead/aead.rs
briansmith: I looked into it. It used to work, but now it doesn't, and that is intentional: https://github.com/rust-lang/rust/issues/36830.

src/agreement.rs
```diff
@@ -94,8 +94,25 @@ pub struct Algorithm {
...
+}
+
+/// Key agreement algorithm types.
+#[allow(non_camel_case_types)]
+#[derive(PartialEq, Eq)]
+pub enum AlgorithmID {
+    ECDH_P256,
+    ECDH_P384,
+    X25519,
 }
```
briansmith: `CurveID` already exists, so we don't need this. I'll remove this in favor of an approach using `CurveID`.

src/ec/suite_b/ecdsa.rs
briansmith: > Well, there is PartialEq implementation for ECDSASigningAlgorithm

Yes, but that's because we've not yet found a solution for creating a unifying `ring::signature::SigningAlgorithm` interface.

However, we already have `ring::signature::VerificationAlgorithm`, and ultimatenly for these comparison operators to be most useful, we need to be able to compare two `VerificationAlgorithm`s, including comparing an RSA algorithm to an ECDSA algorithm. That means that `ring::signature::VerificationAlgorithm` is what ultimately needs to implement `Eq`/`PartialEq`. Let's try to find a way to do that. For now, I'll remove this part.

src/agreement.rs
mrwhythat: Removed it in favor of `algorithm.i.curve.id` comparison.

mrwhythat:
Does it require any additional changes?

ctz:
This will let me delete some code that computes `digest::Algorithm` equality the dumb way 👍

briansmith:
This landed with modifications and was released in *ring* 0.12.1. The commits are:

d8134d64fa5a1109c938ebb0258b6e7aeb1b3439
eb668d5d02256ae35f13bbb80f7f0f60071126d5
b3cd49cd7f41e31faef35a29f688c20e6290d016

Thanks for submitting this and sorry for the long delay in getting it landed.

@ctz wrote:
> This will let me delete some code that computes digest::Algorithm equality the dumb way 👍

I just noticed that, and just sent you a PR: https://github.com/ctz/rustls/pull/117

PR #557, opened 2017-07-13
fix mingw cross compile advapi32 linking

rustysec:
Linking against upper case library names breaks cross compiling from Linux to Windows using mingw. This change doesn't affect native Windows builds (case-insensitive) or native Linux builds (don't use advapi32). All Travis tests passed.


briansmith:
Thanks. I landed this as 77c07cfe7b5ec6d216d78c3563d74bdf6bd259cd in master.

Issue #128, opened 2016-02-29
Don't use `[0; T]` to construct not-yet-initialized values


briansmith:
It turns out that some test vectors are using all-zero values as inputs, which then get used as expected outputs when we verify,for example, that decrypting a ciphertext results in the original plaintext. Accordingly, we should avoid using zero as the initial value for arrays/vectors/slices and indead using something like `[0xDE; T]` instead. This will make it clearer that zeroed-but-otherwise-unwritten memory is being read.

PR #564, opened 2017-07-27
Don't initialize with [0; ...] in tests.

nick70:
#128

### DIFF
--- a/tests/aead_tests.rs
+++ b/tests/aead_tests.rs
@@ -128,7 +128,7 @@ fn test_aead(aead_alg: &'static aead::Algorithm, file_path: &str) {
             257, // 16 AES blocks or 8 ChaCha20 blocks, plus 1.
         ];
 
-        let mut more_comprehensive_in_prefix_lengths = [0; 4096];
+        let mut more_comprehensive_in_prefix_lengths = [0xDE; 4096];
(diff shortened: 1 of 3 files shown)


tests/aead_tests.rs
```diff
@@ -128,7 +128,7 @@ fn test_aead(aead_alg: &'static aead::Algorithm, file_path: &str) {
             257, // 16 AES blocks or 8 ChaCha20 blocks, plus 1.
         ];
 
-        let mut more_comprehensive_in_prefix_lengths = [0; 4096];
+        let mut more_comprehensive_in_prefix_lengths = [0xDE; 4096];
```
briansmith: Instead of hard-coding the `0xDE` constant here, please give it a name with `const UNINITIALIZED: u8 = 0xDE;` and then use `UNINITIALIZED` everywhere.

briansmith: Oh, wait. I see that you chose a different value for each test. I can see the value in that. However, it's also more confusing for the reader because they need to understand the significance of that constant. Maybe it's better to use the same non-zero named and well-documented constant in all tests?

nick70:
I applied those changes here: https://github.com/briansmith/ring/pull/600 because I deleted my fork for some reason.

PR #567, opened 2017-08-01
Bump rayon

Manishearth:


briansmith:
Thanks for pointing this out! We use `rayon = "0.8.2"` as of abc0c47957ce5e22311ae6094ce6d5b54644075e, so we don't need this PR any more.

Issue #579, opened 2017-10-11
Exporting components of secret keys


P-E-Meunier:
Following up on #545, I'd like to get access to the components of RSA and Ed25519 secret keys, in order to send them to an SSH agent.

If there is a strong reason not to have them public, I would like to know more about that: if I really need secret key components (which I really do), I can use another library.

briansmith:
> If there is a strong reason not to have them public, I would like to know more about that: if I really need secret key components (which I really do), I can use another library.

We intentionally don't allow extracting the private/secret components of key objects, because we want to promise that if you pass a `&RSAKeyPair` or `&Ed25519KeyPair` to a function, that function won't be able to break the abstraction and leak the private key (without doing nasty `transmute()`s and similar).

I suggest that we expose an API that lets you do the parsing of a private key into the raw key material fields. That is, parse the `&[u8]` into a struct that consists of `&[u8]` components of the key. These new public functions should live in `ring::signature::primitive`. The code is already available; it just needs to be extracted out of the `*KeyPair` constructor functions. The code that constructs `RSAKeyPair` and `Ed25519KeyPair`s and `ECDSAKeyPair`s should be updated to the new functions that were extracted out of the existing code..

P-E-Meunier:
> I suggest that we expose an API that lets you do the parsing of a private key into the raw key material fields.

That would be amazing!

jprider63:
I need to serialize keys to a different format, so I need this functionality as well. It looks like #582 is a work in progress implementing this. If it's helpful at all, I can help with the implementation. 

One thing I noticed is that `RSAKeyPairComponents` is made of `Input`'s. Would it make more sense to use big integers or `Vec<u8>`'s, so that users can more easily inspect the fields? I suppose users can always call `bigint::Positive::from_be_bytes` themselves. 

In the meantime to get things working, it seems like the first 16 bytes of pkcs8 ed25519 keys are constant, followed by 32 bytes of the private key, followed by a 5 byte constant, followed by 32 bytes of the public key. Do you think it's safe to (temporarily) pull the private and public keys out of these positions from the output of `from_pkcs8`?

rphmeier:
A strong API which prevents leakage of the private key is great. But there should also be an API which does allow inspecting the private key. Maybe expose a `KeyPair`/`OpaqueKeyPair` duality where `KeyPair: Into<OpaqueKeyPair>` but not vice-versa?
Another alternative is to do signing via traits and not inherent implementations.

[...]

PR #582, opened 2017-10-18
Exporting all components of RSA and ED25519 secret keys

P-E-Meunier:
Solves #579 for my RSA keys.
This also makes it possible to derive a public key from a secret key (which is also something I really needed for RSA keys).

### DIFF
--- a/src/ec/curve25519/ed25519.rs
+++ b/src/ec/curve25519/ed25519.rs
@@ -247,30 +247,17 @@ impl<'a> Ed25519KeyPairComponents {
         // This implicitly verifies that `public_key` is the right length.
         // XXX: This rejects ~18 keys when they are partially reduced, though
         // those keys are virtually impossible to find.
-        if public_key != pair.public_key_bytes() {
+        if public_key != &pair.public_key[..] {
             return Err(error::Unspecified);
         }
 
         Ok(pair)
     }
-    /// Constructs a Ed25519 key pair from the private key seed `seed`.
-    ///
-    /// It is recommended to use `Ed25519KeyPair::from_pkcs8()` instead. When
-    /// that is not practical, it is recommended to use
-    /// `Ed25519KeyPair::from_seed_and_public_key()` instead.
-    ///
-    /// Since the public key is not given, the public key will be computed from
-    /// the private key. It is not possible to detect misuse or corruption of
-    /// the private key since the public key isn't given as input.
-    pub fn from_seed_unchecked(seed: untrusted::Input)
+
+    fn from_seed_unchecked(seed: untrusted::Input)
                                -> Result<Ed25519KeyPairComponents, error::Unspecified> {
         let seed = slice_as_array_ref!(seed.as_slice_less_safe(), SEED_LEN)?;
         Ok(Self::from_seed_(seed))
-    }
-
-    /// Returns a reference to the little-endian-encoded public key bytes.
-    pub fn public_key_bytes(&'a self) -> &'a [u8] {
-        &self.public_key
     }
 
     fn from_seed_(seed: &Seed) -> Ed25519KeyPairComponents {
--- a/src/ec/curve25519/ed25519.rs
+++ b/src/ec/curve25519/ed25519.rs
@@ -206,7 +206,7 @@ impl<'a> Ed25519KeyPair {
 /// for signing, see [Ed25519KeyPair](./Ed25519KeyPair.html) instead.
 pub struct Ed25519KeyPairComponents {
     /// The 32-byte private part of this pair, i.e. the seed.
-    pub private_key: [u8; 32],
+    pub seed: [u8; 32],
     /// The 32-byte public part of this key.
     pub public_key: [u8; 32],
 }
@@ -224,9 +224,10 @@ impl<'a> Ed25519KeyPairComponents {
     /// `Ed25519KeyPair::from_pkcs8_maybe_unchecked()` instead.
     pub fn from_pkcs8(input: untrusted::Input)
                       -> Result<Ed25519KeyPairComponents, error::Unspecified> {
-        let (seed, public_key) = unwrap_pkcs8(pkcs8::Version::V2Only, input)?;
-        Self::from_seed_and_public_key(seed, public_key.unwrap())
+        let (seed, _) = unwrap_pkcs8(pkcs8::Version::V2Only, input)?;
+        Self::from_seed_unchecked(seed)
     }
+
 
     /// Constructs an Ed25519 key pair from the private key seed `seed` and its
     /// public key `public_key`.
@@ -252,7 +253,6 @@ impl<'a> Ed25519KeyPairComponents {
 
         Ok(pair)
     }
-
     /// Constructs a Ed25519 key pair from the private key seed `seed`.
     ///
     /// It is recommended to use `Ed25519KeyPair::from_pkcs8()` instead. When
@@ -290,7 +290,7 @@ impl<'a> Ed25519KeyPairComponents {
         }
 
         Ed25519KeyPairComponents {
-            private_key: seed.clone(),
+            seed: seed.clone(),
             public_key: a.into_encoded_point(),
         }
     }
--- a/src/ec/curve25519/ed25519.rs
+++ b/src/ec/curve25519/ed25519.rs
@@ -202,6 +202,100 @@ impl<'a> Ed25519KeyPair {
     }
 }
 
+/// The components of an Ed25519 key pair. This type cannot be used
+/// for signing, see [Ed25519KeyPair](./Ed25519KeyPair.html) instead.
+pub struct Ed25519KeyPairComponents {
+    /// The 32-byte private part of this pair, i.e. the seed.
+    pub private_key: [u8; 32],
+    /// The 32-byte public part of this key.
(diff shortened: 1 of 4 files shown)


P-E-Meunier:
Success! I just implemented an ssh-agent client, might add a server later today.

P-E-Meunier:
Ok, I'm almost done with the agent server. It works fine for Ed25519 keys, and the only thing missing now is to generate dP and dQ from d, p and q using the bigint module. Is there an easy answer?

briansmith:
> Ok, I'm almost done with the agent server. It works fine for Ed25519 keys, and the only thing missing now is to generate dP and dQ from d, p and q using the bigint module. Is there an easy answer?

They're already in the input. Notice in the code these two lines:

```rust
let _dP = der::positive_integer(input)?;
let _dQ = der::positive_integer(input)?;
```

You just need to put dP and dQ into the RSAPrivateKeyComponents structure.

src/ec/curve25519/ed25519.rs
```diff
@@ -202,6 +202,100 @@ impl<'a> Ed25519KeyPair {
     }
 }
 
+/// The components of an Ed25519 key pair. This type cannot be used
+/// for signing, see [Ed25519KeyPair](./Ed25519KeyPair.html) instead.
+pub struct Ed25519KeyPairComponents {
+    /// The 32-byte private part of this pair, i.e. the seed.
+    pub private_key: [u8; 32],
```
briansmith: This should be called `seed`.

src/ec/curve25519/ed25519.rs
```diff
@@ -202,6 +202,100 @@ impl<'a> Ed25519KeyPair {
...
+    /// The input must be in PKCS#8 v2 format, and in particular it must contain
+    /// the public key in addition to the private key. `from_pkcs8()` will
+    /// verify that the public key and the private key are consistent with each
+    /// other.
+    ///
+    /// If you need to parse PKCS#8 v1 files (without the public key) then use
+    /// `Ed25519KeyPair::from_pkcs8_maybe_unchecked()` instead.
+    pub fn from_pkcs8(input: untrusted::Input)
+                      -> Result<Ed25519KeyPairComponents, error::Unspecified> {
+        let (seed, public_key) = unwrap_pkcs8(pkcs8::Version::V2Only, input)?;
```
briansmith: Once you have seed and public_key, you can `return Ed25519KeyPairComponents { seed, public_key }`. In particular, you don't need to do any further validation.

src/ec/curve25519/ed25519.rs
```diff
@@ -202,6 +202,100 @@ impl<'a> Ed25519KeyPair {
...
+    ///
+    /// It is recommended to use `Ed25519KeyPair::from_pkcs8()` instead.
+    ///
+    /// The private and public keys will be verified to be consistent with each
+    /// other. This helps avoid misuse of the key (e.g. accidentally swapping
+    /// the private key and public key, or using the wrong private key for the
+    /// public key). This also detects any corruption of the public or private
+    /// key.
+    pub fn from_seed_and_public_key(seed: untrusted::Input,
+                                    public_key: untrusted::Input)
```
briansmith: If you already have the seed and the public key, then you already have the components. Thus, this function isn't necessary in Ed25519KeyPairComponents.

src/ec/curve25519/ed25519.rs
```diff
@@ -202,6 +202,100 @@ impl<'a> Ed25519KeyPair {
...
+    /// Constructs a Ed25519 key pair from the private key seed `seed`.
+    ///
+    /// It is recommended to use `Ed25519KeyPair::from_pkcs8()` instead. When
+    /// that is not practical, it is recommended to use
+    /// `Ed25519KeyPair::from_seed_and_public_key()` instead.
+    ///
+    /// Since the public key is not given, the public key will be computed from
+    /// the private key. It is not possible to detect misuse or corruption of
+    /// the private key since the public key isn't given as input.
+    pub fn from_seed_unchecked(seed: untrusted::Input)
```
briansmith: Are you actually using this function in your key server? I suspect you're not, if not, let's remove it. Otherwise, I am curious why you need it, as I suspect you shouldn't need it.

src/ec/curve25519/ed25519.rs
```diff
@@ -202,6 +202,100 @@ impl<'a> Ed25519KeyPair {
...
+                               -> Result<Ed25519KeyPairComponents, error::Unspecified> {
+        let seed = slice_as_array_ref!(seed.as_slice_less_safe(), SEED_LEN)?;
+        Ok(Self::from_seed_(seed))
+    }
+
+    /// Returns a reference to the little-endian-encoded public key bytes.
+    pub fn public_key_bytes(&'a self) -> &'a [u8] {
```
briansmith: This function isn't needed because `self.public_key` is public.

src/rsa/signing.rs
```diff
@@ -386,6 +386,70 @@ impl RSAKeyPair {
...
+    /// The public exponent (big endian).
+    pub e: untrusted::Input<'a>,
+    /// The secret exponent (big endian).
+    pub d: untrusted::Input<'a>,
+    /// The inverse of q modulo p (big endian).
+    pub iqmp: untrusted::Input<'a>,
+    /// Prime factor of n (big endian).
+    pub p: untrusted::Input<'a>,
+    /// Prime factor of n (big endian).
+    pub q: untrusted::Input<'a>,
```
briansmith: This is missing dP and dQ. Also, make sure these are all in the same order as they're given in the RFC.

src/signature.rs
```diff
@@ -302,14 +302,15 @@ pub use ec::curve25519::ed25519::{
     ED25519,
 
     Ed25519KeyPair,
+    Ed25519KeyPairComponents,
```
briansmith: This should be exposed from `ring::signature::primitives` not from `ring::signature`.

src/signature.rs
```diff
@@ -302,14 +302,15 @@ pub use ec::curve25519::ed25519::{
...
     ED25519_PKCS8_V2_LEN,
     ED25519_PUBLIC_KEY_LEN,
 };
 
 pub use pkcs8::PKCS8Document;
 
 #[cfg(all(feature = "rsa_signing", feature = "use_heap"))]
-pub use rsa::signing::{RSAKeyPair, RSASigningState};
+pub use rsa::signing::{RSAKeyPair, RSAKeyPairComponents, RSASigningState};
```
briansmith: Ditto; this should be exposed from `ring::signature::primitives`, not `ring::signature`.

P-E-Meunier:
> They're already in the input. Notice in the code these two lines:

Sorry, I was imprecise. They are not always in the input in all cases. Sure, in PKCS#8, they are, but they are not stored by the SSH agent. Oh wait, there's a "comment" field. Nevermind.

Thanks for the rest of the review!

P-E-Meunier:
Ok, this works, but my server is only be able to exchange RSA keys with my client (not with OpenSSH). I really need a way to compute dP and dQ from the other parameters.

src/ec/curve25519/ed25519.rs
P-E-Meunier: Actually, I need it, because keys from SSH key files are not always in PKCS#8, and I still need to get their components.

P-E-Meunier:
By the way, I just fixed my problem by using the num-bigint crate. I now have a client and server for SSH-agent (the only missing feature is smartcards).

src/ec/curve25519/ed25519.rs
```diff
@@ -228,6 +228,13 @@ impl<'a> Ed25519KeyPairComponents {
         Self::from_seed_unchecked(seed)
     }
 
+    /// Constructs an `Ed25519KeyPair` from these components.
+    pub fn to_key_pair(&self) -> Result<Ed25519KeyPair, error::Unspecified> {
```
briansmith: I specifically want to avoid having a `to_key_pair` function here, so let's remove this commit. Users can implement this logic themselves if they want.

src/ec/curve25519/ed25519.rs
briansmith: Sorry, I don't understand. There's already a public `pub fn from_seed_and_public_key()` function in `Ed25519KeyPair` so you can just use it directly without using `Ed25519KeyPairComponents` at all, right?

src/ec/curve25519/ed25519.rs
briansmith: > I specifically want to avoid having a to_key_pair function here, so let's remove this commit. Users can implement this logic themselves if they want.

Never mind this, I misunderstood and I've changed my mind. Ignore the previous comment.

briansmith:
Sorry, I started to review this and then realized it's going to take too much time for me to do today. Ignore my earlier comments. I'll think through what I want to do here and return to the review another day.

P-E-Meunier:
Hi! Is there anything I can do to help? We'd like to release a new version of Pijul, but we have still too many users complaining about problems with their keys, which would be solved by this PR.

briansmith:
Thanks for submitting this. I understand it's not needed by the original submitter any longer so I'm closing this. I think we'll add this functionality one way or another soon.

PR #587, opened 2017-10-26
Give access to a validation key's digest algorithm

tomaka:


briansmith:
Thanks! Looks like the fact that this was missing is an oversight, since it's already present on `SigningKey`.

PR #591, opened 2017-11-06
Be more specific about 'MSVC 2015 or later' error

### LINKED ISSUE #590: "MSVC 2015 or later is required." should be ""MSVC 2015 Update 2 or later is required."
I have Visual Studio 2015 installed but I couldn't compile ring because it always said it needed MSVC 2015 or later.

I found this check in `limbs.inl`:
```c
#if _MSC_FULL_VER < 190023918
#error "MSVC 2015 or later is required."
#endif
```

I happen to have an older version (installed by my sysadmin):
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.00.23026 for x86
```

I manually removed that check to see how it goes and it finally compiled fine.

So here are my questions/requests:

- what risk do I have to use this older version of the compiler?
- can we make the message more explicit and say that we have version xxx while we are expecting version yyy. It took me a while to figure out that *MSVC 2015* is not enough. Or at least mention which update we need at minimum.

tafia:
Closes #590

### DIFF
--- a/crypto/limbs/limbs.inl
+++ b/crypto/limbs/limbs.inl
@@ -27,7 +27,14 @@
  * We test with MSVC 2015 update 2, so make sure we're using a version at least
  * as new as that. */
 #if _MSC_FULL_VER < 190023918
-#error "MSVC 2015 or later is required."
+  #if _MSC_VER == 1900
+      #if _MSC_FULL_VER == 190023026
+        #error "MSVC 2015 Update 2 or later is required, found MSVC 2015."
+      #elif _MSC_FULL_VER == 190023506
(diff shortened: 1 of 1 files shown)


tafia:
I believe the failing tests are not a consequence of this PR.

briansmith:
Hi, I decided not to include the extra logic to report the current version of MSVC found because that's not a pattern I'm eager to maintain for all compilers. However, I appreciate you submitting this PR. Thanks!

briansmith:
(See 8090623a2e24128a9249eb7ff054ff8b73ea5d62 for the commit that replaced this PR.)

Issue #331, opened 2016-11-05
X25519 (and/or ECDH in general) with static keys


danburkert:
Hi, while attempting to add ring as a backend to a [Noise](http://noiseprotocol.org/) implementation, I found that ring does not support DH key agreement with static keys.  Static DH is required for the protocol.  Is static DH support planned?  Are there any workarounds in the meantime (perhaps a way to unsafely clone the key?).

briansmith:
> Are there any workarounds in the meantime (perhaps a way to unsafely clone the key?).

We do everything we can do to ensure that isn't possible.

> Hi, while attempting to add ring as a backend to a Noise implementation, I found that ring does not support DH key agreement with static keys. Static DH is required for the protocol. Is static DH support planned? 

Yes, it is. In particular, I explicitly want to make it possible to implement Noise using *ring*.

Basically, we need to create a `ring::agreement::StaticKeyPair` to go alongside `ring::agreement::EphemeralPrivateKey`. In fact, `ring::signature::Ed25519KeyPair` isn't too far from what we need.

So, at this point, it's simply a matter of getting the code and tests written.

briansmith:
To clarify some things:

* The key generation function should be like `Ed25519KeyPair`s, not like `EphemeralPrivateKey`'s `generate_serializable()`, which returns the key pair object *and* a serialized representation of the private + public key. Otherwise, the caller won't be able to save the key pair bytes to disk.

* Also like `Ed25519KeyPair`, `StaticKeyPair` needs to save the public key within the `StaticKeyPair` object. This is different from `EphemeralPrivateKey`. `EphemeralPrivateKey`'s public key handling is optimized for the fact that the public key is only needed once and isn't always needed at the same time as the private key. This is why we have `EphemeralPrivateKey` and not `EphemeralKeyPair`.

* As stated above, `generate_serializable()` needs to give the caller a serialized sequence of bytes that encodes the key pair. That means there needs to be a function that de-serializes that sequence of bytes, again similar to `Ed25519KeyPair`. For now, it is OK to do this the same way as `Ed25519KeyPair` for X25519 keys. In the future I will almost definitely switch this to use a self-describing typed file format.

* Since you only need X25519 for Noise, it is OK to defer the work on the P-256 and P-384 variants. In fact, you can just name it `X25519StaticKeyPair` and avoid all the `agreement::Algorithm` stuff, for now. Eventually we'll generalize it to work for P-256 and P-384.

briansmith:
[...]

PR #593, opened 2017-11-28
An wrap to support using X25519 with static keys

kamyuentse:
For #331

### DIFF
--- a/src/agreement.rs
+++ b/src/agreement.rs
@@ -147,27 +147,27 @@ impl<'a> EphemeralPrivateKey {
 /// A static key pair for static X25519 scheme.
 ///
 /// Refer NIST.SP.800-56A Chapter 6 for more details.
-pub struct X22519StaticKeyPair {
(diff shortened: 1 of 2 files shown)


briansmith:
Thanks for the PR. I'm not that eager to support static ECDH key exchange but I do want to help people who are building higher-level constructs that need something like this.

Off the top of my head, I think "static" is a bad word to use here. I think "Reusable", to contrast with "Ephemeral" might be better.

Do you have any suggestions for how we can avoid the duplication of code between this new thing and the ephemeral key code? The amount of duplication seems quite unpleasant.

kamyuentse:
@briansmith Thanks for your review, I admit that I the "static" is not a good name, the "reusable" seems reasonable to me.

For the duplication of code, we can introduce something named "Inner"(maybe you have your idea about the naming), which consist of `ec::PrivateKey` and `Algorithm`, then we can avoid the duplication. Is this solution make sense?

Issue #598, opened 2017-12-22
file not found for module `montgomery`


sackery:
|
15 | pub mod montgomery;
   |         ^^^^^^^^^^
   |
   = help: name the file either arithmetic/montgomery.rs or arithmetic/montgomery/mod.rs inside the directory "/home/sackery/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.11.0/src/arithmetic"

andrewtj:
0.11 does include that file, have you per-chance done something that would have removed it? If not what platform, Rust version, etc are you using and can you provide steps to reproduce the problem?

thomashoneyman:
Same error here. Versions:

rustc 1.24.0-nightly
ring 0.11.0

I'm running on MacOS 10.13.2 

```rust
error[E0583]: file not found for module `montgomery`
  --> /.../.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.11.0/src/arithmetic/arithmetic.rs:15:9
   |
15 | pub mod montgomery;
   |         ^^^^^^^^^^
   |
   = help: name the file either arithmetic/montgomery.rs or arithmetic/montgomery/mod.rs inside the directory "/.../.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.11.0/src/arithmetic"

error: aborting due to previous error

error: Could not compile `ring`.
```

thomashoneyman:
Updated my comment above with the Cargo file @andrewtj

andrewtj:
Ta, on macOS 10.12.6 it works with nightly-2017-12-21 but not nightly-2017-12-22.

thomashoneyman:
I see. It does build correctly on the latest nightly in the Docker image supported at `rustlang/rust:nightly`.

thomashoneyman:
@andrewtj Are you aware of a way to set the specific nightly for a project?

sackery:
why don't follow rust standard module contract?!
use mod.rs not #path[...]

briansmith:
> why don't follow rust standard module contract?!
> use mod.rs not #path[...]

Because the old `mod.rs` naming scheme is terrible. https://github.com/rust-lang/rust/pull/46531 implements a new path scheme that we'll switch to when it reaches stable rust.

briansmith:
> Ta, on macOS 10.12.6 it works with nightly-2017-12-21 but not nightly-2017-12-22.

[...]

PR #599, opened 2017-12-22
Update directory structure

cramertj:
This PR moves around several files in order to
remove #![allow(legacy_directory_ownership)].
This will prevent breakage when the
legacy_directory_ownership warning becomes a hard
error.

cc https://github.com/briansmith/ring/issues/598


briansmith:
Instead of doing renames like src/aead/aead.rs → src/aead/mod.rs, please do renames like src/aead/aead.rs → src/aead.rs, using the new path rules.

cramertj:
@briansmith Oh, so you'd like to enable `#![feature(non_modrs_mods)]`?

briansmith:
> @briansmith Oh, so you'd like to enable #![feature(non_modrs_mods)]?

Sure, but the code needs to work in Stable Rust too. It sounds like the rustc developers need to come up with a migration plan.

cramertj:
@briansmith If you want this to work in stable rust, the only way to structure it is the way I have here.

Fuzen-py:
@cramertj You can always just move it to the way it was requested and then add a mod.rs to make what should be public public. Id say it would seem even messier though

jbg:
> It sounds like the rustc developers need to come up with a migration plan.

They did. If you disagree with it, this doesn't seem like the most productive way to indicate that, considering that ignoring the warnings that you weren't following the plan has broken `ring` for anyone who tracks rust nightly.

briansmith:
I went ahead and locked this thread to minimize wasted effort on everybody's part. There's a large number of pending changes to *ring* that I want to merge before I rename any files. I'll try to get all these changes done by the end of January.

briansmith:
The issue that this PR addresses was fixed in the 0.13.0-alpha release. Closing this.

PR #600, opened 2017-12-24
Use explicitly uninitialized constants in tests.

nick70:
A revision of https://github.com/briansmith/ring/pull/564 (i deleted my fork OOPS lol).

### DIFF
--- a/tests/aead_tests.rs
+++ b/tests/aead_tests.rs
@@ -34,6 +34,8 @@
 
 extern crate ring;
 
+mod constants;
+
 use ring::{aead, error, test};
 use std::vec::Vec;
 
@@ -148,7 +150,7 @@ fn test_aead(aead_alg: &'static aead::Algorithm, file_path: &str) {
             257, // 16 AES blocks or 8 ChaCha20 blocks, plus 1.
         ];
 
-        let mut more_comprehensive_in_prefix_lengths = [0; 4096];
+        let mut more_comprehensive_in_prefix_lengths = [constants::UNINITIALIZED_USIZE; 4096];
         let in_prefix_lengths;
         if cfg!(debug_assertions) {
             in_prefix_lengths = &MINIMAL_IN_PREFIX_LENS[..];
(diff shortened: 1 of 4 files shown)


tests/aead_tests.rs
```diff
@@ -34,6 +34,8 @@
 
 extern crate ring;
 
+mod constants;
```
briansmith: Instead of creating a new module, let's just put them in the `ring::test` module, so they will also be available for unit tests, in addition to integration tests.

tests/constants.rs
```diff
@@ -0,0 +1,5 @@
+#[allow(dead_code)]
+pub const UNINITIALIZED_BYTE: u8 = 0xDE;
```
briansmith: Probably you wanted `#[cfg(test)]` but if/when you put them in the `tests` module this will be a non-issue because they will be accessible outside of *ring* can thus never dead.

Let's use `U8` instead of `BYTE` in the name.

tests/constants.rs
```diff
@@ -0,0 +1,5 @@
...
+
+#[allow(dead_code)]
```
briansmith: Ditto.

briansmith: I think the value of `UNINITIALIZED_USIZE` shouldn't have any zero bytes in it.

However, I suggest that we just remove the one use of `UNINITIALIZED_USIZE` we have for now and remove `UNINITIALIZED_USIZE`.

tests/aead_tests.rs
```diff
@@ -148,7 +150,7 @@ fn test_aead(aead_alg: &'static aead::Algorithm, file_path: &str) {
             257, // 16 AES blocks or 8 ChaCha20 blocks, plus 1.
         ];
 
-        let mut more_comprehensive_in_prefix_lengths = [0; 4096];
+        let mut more_comprehensive_in_prefix_lengths = [constants::UNINITIALIZED_USIZE; 4096];
```
briansmith: I don't think we need to worry about this, because this isn't being passed to a function under test. So, I suggest just reverting it back to 0.

briansmith:
Closing this since progress on it has ceased. Thanks for your help here!

PR #601, opened 2017-12-24
Derive Clone for KeyPair, PrivateKey and Signature

shahn:
Here's a first little pull request for a small feature: Being able to Clone KeyPairs and Signatures.

### DIFF
--- a/src/rsa/bigint.rs
+++ b/src/rsa/bigint.rs
@@ -64,6 +64,7 @@ impl AsRef<BIGNUM> for Nonnegative {
 /// This set is sometimes called `Natural` or `Counting`, but texts, libraries,
 /// and standards disagree on whether to include zero in them, so we avoid
 /// those names.
+#[derive(Clone)]
 pub struct Positive(Nonnegative);
 
 impl Positive {
@@ -108,6 +109,7 @@ impl Positive {
 }
 
 /// Odd positive integers.
+#[derive(Clone)]
 pub struct OddPositive(Positive);
 
 impl OddPositive {
@@ -192,6 +194,7 @@ pub unsafe trait NotMuchSmallerModulus<L>: SmallerModulus<L> {}
 /// for efficient Montgomery multiplication modulo *m*. The value must be odd
 /// and larger than 2. The larger-than-1 requirement is imposed, at least, by
 /// the modular inversion code.
+#[derive(Clone)]
 pub struct Modulus<M> {
     value: OddPositive, // Also `value >= 3`.
 
@@ -281,6 +284,7 @@ impl Modulus<super::N> {
 // Defaulting `E` to `Unencoded` is a convenience for callers from outside this
 // submodule. However, for maximum clarity, we always explicitly use
 // `Unencoded` within the `bigint` submodule.
+#[derive(Clone)]
 pub struct Elem<M, E = Unencoded> {
     value: Nonnegative,
 
@@ -498,6 +502,7 @@ pub fn elem_sub<M, E>(mut a: Elem<M, E>, b: &Elem<M, E>, m: &Modulus<M>)
 
 
 // The value 1, Montgomery-encoded some number of times.
+#[derive(Clone)]
 pub struct One<M, E>(Elem<M, E>);
 
 #[cfg(feature = "rsa_signing")]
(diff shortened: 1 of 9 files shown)


shahn:
Is a Rust version bump to be expected soon? Otherwise I could impl Clone manually

briansmith:
> Is a Rust version bump to be expected soon? Otherwise I could impl Clone manually

Any change to *ring* can always increase the Rust version requirement to the latest stable version, currently Rust 1.22.1.

shahn:
Ah, then I suppose the PR might be mergable as is. The CI failures don't seem relevant

briansmith:
@shahn Please take a look at 7a6f3c8ddd430ec6dd542bc8762bb574f5ba3131 where I did what you asked for for `Signature` and created a tiny framework for testing these kinds of things.

In that PR, I didn't add `Clone()` for `KeyPair` because I am unsure whether we should do that. Maybe it is good enough to use `Arc<...KeyPair>` to share a `KeyPair` between threads? I guess I'm not sure we'll always be able to clone keypairs and private keys going forward, because we'll eventually have implementations of those types that are external to us and thus limited. However, I'd like to hear your use case to see if it's worth doing in the interim.

shahn:
Thanks! My use case is a data structure that owns a KeyPair. It doesn't require any kind of synchronization overhead currently as it is completely immutable, so wrapping the key in an Arc would be somewhat unfortunate.

Another big use case is that for testing I need to create a lot of keys. Creating one and then cloning it would save a ton of time for the test framework.

I'm not sure I understand the reasoning about Keypairs and private keys, couldn't there be a trait those need to implement which actually requires a Clone bound?

src/signature_impl.rs
```diff
@@ -18,6 +18,7 @@
 use ec;
 
 /// A public key signature returned from a signing operation.
+#[derive(Clone)]
```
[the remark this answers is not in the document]
shahn: Yes, and you already did in 7a6f3c8ddd430ec6dd5 :)

shahn:
Just commenting to note that I think the requested changes are irrelevant to this ticket now because they were already merged upstream in 7a6f3c8

src/rsa/bigint.rs
```diff
@@ -1052,6 +1058,7 @@ mod repr_c {
 
     // Keep in sync with `bignum_st` in openss/bn.h.
     #[repr(C)]
+    #[derive(Clone)]
```
briansmith: We cannot derive `Clone` for this type because `d` is a pointer. In order to clone `d` we need to make a copy of the data being pointed to by `d`.

briansmith: I suggest that you split this PR into two parts: One for RSA and one for non-RSA stuff. The RSA stuff is tricky and the non-RSA stuff is simple.

The `Nonnegative` type already implements a variant of `Clone::clone()` called `try_clone()`. We were expecting that there would be a `TryClone` trait added to the standard library, but it looks like that won't be the case; see https://github.com/rust-lang/rfcs/pull/2189#issuecomment-381439136. At least at the present time I don't want to change *ring* to panic when the C functions return errors, so I think we cannot implement `Clone` for RSA private keys. If you want an RSA private key that you can `Clone::clone()` in your app then you can get one by using `Arc<RSAKeyPair>` instead of `RSAKeyPair`.

briansmith:
I refactored the code so that there are no `BIGNUM` components within an `RSAKeyPair`. This made it possible to `derive(Clone)` for that type.

I'm closing this PR. If you need EdDSA keypairs to be `Clone`, please submit a PR for that. If you need ECDSA keypairs to be `Clone`, let's submit a separate PR for that too, which I'll probably wait to consider until ECDSA signing is actually implemented.

Thanks for the PR!

briansmith:
> I refactored the code so that there are no BIGNUM components within an RSAKeyPair. This made it possible to derive(Clone) for that type.

I undid the `derive(Clone)` for `RSAKeyPair` for the 0.13.0 release. I am open to the possibility of adding it back later. However, in general cloning `RSAKeyPair` is going to be super heavyweight and I don't understand why it wouldn't be better to use `Arc<RSAKeyPair>` instead.

PR #602, opened 2017-12-24
Support reusable private key for agreement

kamyuentse:
@briansmith I modified the PR according to your comment. Because this seems off-topic from the original PR, I create a new one.

Closed #593


chocol4te:
Any updates on this? At 1aim we really don't want to have to use OpenSSL and this is the last bit of functionality we need.  We're more than happy to work on any changes required to get this merge-ready.

PR #604, opened 2017-12-25
Add support for RSA keys where p < q

gh2o:
Such keys are generated by gnutls.


briansmith:
Thanks for the PR. I agree we should support keys with p < q. However, I don't want to use `transmute()`. Also I think we should try to find the simplest possible solution with a minimum amount of branches in the code path; simpler control flow would make (formal) verification easier.

How about this?: When constructing a key with p < q we simply swap p & q and dP and dQ and calculate a new qInv for the new p and q. For the new p and q, we could calculate q mod p with a conditional subtraction since we know that q < p < 2*q. We can do the inversion to get qInv using fermat's little theorem, using `elem_exp_consttime()` with `p - 2` as the exponent. WDYT?

gh2o:
Thanks for the suggestions! I've updated the PR according to them.

One question though: if we already have a code path where we generate our own qInv, would it be worthwhile to do that for all cases (for both p > q and p < q)?

src/rsa/signing.rs
```diff
@@ -340,22 +346,33 @@ impl RSAKeyPair {
...
                 let q_mod_p = q.try_clone()?.into_elem(&p.modulus)?;
-                let qInv_times_q_mod_p =
-                    bigint::elem_mul(&qInv, q_mod_p, &p.modulus)?;
-                if !qInv_times_q_mod_p.is_one() {
-                    return Err(error::Unspecified);
-                }
+                let qInv = if pq_swapped {
+                    // Calculate a new qInv. Steps 7.c-7.f can be skipped because
+                    // elem_inverse() already verifies qInv's invertibility.
+                    bigint::elem_inverse(
```
briansmith: IIRC, we cannot use `elem_inverse` here because `elem_inverse()` is not constant time. Instead we need to use Fermat's Little Theorem: calculate qInv = `q**(p - 2)`. (We probably should rename `elem_inverse()` to `elem_inverse_vartime()` or something like that.)

src/rsa/signing.rs
```diff
@@ -251,13 +251,20 @@ impl RSAKeyPair {
                 let n = n.into_modulus::<N>()?;
                 let oneRR_mod_n = bigint::One::newRR(&n)?;
 
+                // Swap p and q if p < q.
+                let (p, q, dP, dQ, pq_swapped) = match q.verify_less_than(&p) {
+                    Ok(_)  => (p, q, dP, dQ, false),
+                    Err(_) => (q, p, dQ, dP, true),
+                };
```
briansmith: Please also calculate the new `qInv` here instead of doing it below. It would be clearer. Or, if there's a reason not to calculate it here, add a comment about why the new `qInv` isn't calculated here.

src/rsa/signing.rs
briansmith: It seems reasonable to skip the check that q * qInv = 1 (mod p). However I think it's also reasonable to do the check anyway. In particular when we switch to Fermat's Little Theorem it's probably good to do the check since it wouldn't be (as) redundant.

briansmith:
> One question though: if we already have a code path where we generate our own qInv, would it be worthwhile to do that for all cases (for both p > q and p < q)?

Calculating the inverse, especially using Fermat's Little Theorem, is quite slow. That's why we use the trick of checking `qInv * q == 1 (mod p)` instead of computing `qInv` ourselves.

For this reason, for performance it would be better to have a solution more like your previous one but without `transmute()`. However, that kind of implementation is more difficult to write and much more difficult to review than this simpler one, so we can/should defer it until later.

It would be great if you could also file a bug with gnutls suggesting that they always generate keys with p > q, citing this issue and also noting that BoringSSL will probably do something similar.

briansmith:
In the RSA bigint code there are two exponentiation functions. The one that you want to use is `elem_exp_consttime()`. Unfortunately `elem_exp_consttime()` isn't *exactly* constant-time which is why I've been avoiding solving this issue for so long.

gh2o:
I made the requested changes. Thanks for the explanation.

> It would be great if you could also file a bug with gnutls suggesting that they always generate keys with p > q, citing this issue and also noting that BoringSSL will probably do something similar.

gnutls generates both p>q keys and p<q keys; they should both be valid under the linked NIST spec.

briansmith:
@gh2o OK, this looks right to me at a first glance. Do you have any 2048- and 4096- bit keys with p < q that you can add as test vectors? See the existing RSA signing tests for a model to use?

gh2o:
Added a few test vectors.

Capitrium:
@briansmith is this PR blocked on anything in particular, or is it just the test failures (which seem to be build server issues) that are holding this up?

gh2o:
@briansmith Thanks for the thumbs up! I've done the above steps, and all tests are passing.

thedeeno:
@briansmith This ready to merge? Looks like @gh2o made the changes you asked for.

briansmith:
Merge! Thanks!

briansmith:
I made some minor improvements to this. Please take a look: https://github.com/briansmith/ring/compare/e6d423c55cd711ca9779da25bdc2d14d95223a1b...6e68a0cf769c50fb4a3774b7c4897adf8ef9e165

gh2o:
@briansmith Looks good to me :+1:

thedeeno:
@briansmith can we do a release for this change? The following project depends on ring and currently isn't usable on kubernetes clusters using rsa keys. See this issue:

https://github.com/databricks/click/issues/33

I'd love to get them to update their dependency!

briansmith:
@thedeeno I am working on the new release, which I expect to cut next week. I have ~100 BoringSSL changes to merge between now and then.

briansmith:
It would be useful for people to check recent changes to rsa/ and crypto/fipsmodule/bn. I spent a significant amount of effort improving the (timing) side channel resistance of the RSA "constant-time" exponentiation. I did so particularly in support of this PR, because this PR uses the important secret `q` as the base of the exponentiation during the computation of `q**-1 (mod p)` as `q**(p - 2) (mod p)` using Fermat's Little Theorem. I'm not claiming it's perfect now but it should be clearly better--not only for this inversion, but for the RSA computation itself.

briansmith:
I see now that BoringSSL implemented this mostly the same way: https://boringssl.googlesource.com/boringssl/+/be837402a974d21071390e0d1a0ebc157901ec79%5E%21/#F0.

Issue #575, opened 2017-08-31
Allow cross-version linkage


SergioBenitez:
Bringing this over from #535.

Being able to link two versions of ring into one binary is incredibly important, and it will become even more important as ring gains adoption. In fact, being able to link two versions at once was the primary reason why I've used ring as opposed to openssl in my projects. I want to illustrate why in hopes that ability to link two versions of `ring` at once is prioritized.

The current state of affairs forces a lock-step upgrade of dependencies. Consider what Rocket had to do to upgrade from cookie 0.8 to cookie 0.9, which updated ring to 0.11. First, the cookie dependency was updated to 0.9. This caused a breakage since Rocket also depends on rustls and hyper-rustls which depended on ring 0.9. I maintain a fork of hyper-rustls, but I couldn't update it to ring 0.11 because it depends on rustls. So, I opened up an issue on rustls. rustls itself depended on webpki-roots, which depends on ring. So, webpki-roots had to be updated. Thankfully, it had no downstream uses of ring, so it updated just fine. Once webpki-roots was updated, rustls was updated. Once rustls was updated, hyper-rustls was updated. Finally, once hyper-rustls was updated, I could update the three ring-related dependencies in Rocket. To upgrade one dependency in Rocket, I had to force updates to 3 other crates. All because of ring.

Thankfully, @ctz owns rustls and webpki-roots, and I maintain cookie, so this was able to happen rather quickly. What if this hadn't been the case? How long would it have taken for the upgrade to happen? Your goal is for users to always be on the latest version of ring, but I think you'll see just the opposite as users are forced to the least common denominator.

Additionally, ring is most likely to be used as an internal library. It's unlikely that two libraries will interface through ring's types. This is where Cargo/Rust's ability to have two versions of the same library in one project shines! It's a shame that we've lost this benefit. And, what's worse, ring now directly controls when a library has to release a breaking change. This is a massive inversion of control; I really don't want a dependency controlling my version numbers. Again, I think you'll see many projects refusing to update ring as a result, leading to libraries on old versions of ring.

cc @alexcrichton

andrewtj:
Is this under consideration?

lbolla:
Is this ticket been worked on?

andrewtj:
@briansmith are you considering this? Would you entertain PRs that prefix the appropriate symbols?

briansmith:
> @briansmith are you considering this? Would you entertain PRs that prefix the appropriate symbols?

[...]

PR #605, opened 2017-12-25
Backport nightly fixes to V0.11

SergioBenitez:
This PR is made against the master branch because no v0.11 branch exists. I've locally created a branch off of the `0.11.0` tag which these commits should be applied to.

This backports the nightly fix changes made to master and released in `0.13.0-alpha` to `0.11.0`. The crate version is updated to `0.11.1` for easy releasing.

I want to keep the argument short: Rocket v0.3 uses ring 0.11 and is thus currently broken on the latest nightlies as a result (https://github.com/SergioBenitez/Rocket/issues/513). Because of https://github.com/briansmith/ring/issues/575, we cannot update `ring` to `0.13.0-alpha` without a breaking release, which Rocket is not ready for. Even then, none of Rocket's `ring` dependents have upgraded to `ring 0.13.0-alpha` (`cookie`, `rustls`, `hyper-sync-rustls`), so this wouldn't be _technically_ feasible either. This PR is the only way to get Rocket working on nightly.

I understand that you only want to support the latest version of _ring_. Were it not for https://github.com/briansmith/ring/issues/575, while still unfavorable to the community, it would be a tenable position; many crates could simply upgrade to the latest version risk-free. But coupling https://github.com/briansmith/ring/issues/575 with your present stance creates a truly difficult scenario for crates depending on `ring`. At present time, there are 69 crates depending on `ring`, none of which are on `0.13.0-alpha`, and thus the majority broken on nightly. Your position makes `ring` difficult to use, inverts control over release schedules, and is currently breaking other projects. If you don't want to support older versions of `ring`, at least accept PRs when the community wants to support them for you. Please don't simply silence opinions by locking issues and PRs such as https://github.com/briansmith/ring/pull/599, https://github.com/briansmith/ring/issues/598, and https://github.com/briansmith/ring/issues/575.

Please accept this PR and release `0.11.1`. I'll submit a PR for `0.12.2` as well.


briansmith:
Nope.

Issue #603, opened 2017-12-24
Travis CI Android emulator-based tests for Android are broken


briansmith:
It seems like the Android SDK changed in some non-backward-compatible way. If somebody has a fix, I'd take a PR for it. Otherwise I think I might try rewriting all the non-Windows CI stuff using Docker and then using the Docker-ified stuff, ETA in Feburary.

briansmith:
See https://github.com/rust-lang/rust/pull/45580 for some inspiration.

andrewtj:
I think a change in Rust might have inadvertently broken the test suite. [This glob in travis.sh](https://github.com/briansmith/ring/blob/a0d65dd207fae12255bd3a470bc9983391242c1a/mk/travis.sh#L101) matches two files so `adb push` attempts to place them both in the directory `/data/ring-test` instead of just copying the test executable to that path. [Patching that](https://github.com/andrewtj/ring/commit/00c81b9496f4320c50fa123d02a566a58d77b8e2) gets the run a little further to an [`Illegal instruction`](https://api.travis-ci.org/v3/job/331628497/log.txt) failure.

andrewtj:
And here's a [tombstone](https://gist.github.com/andrewtj/d91428c249cbaf08276f1078b5431eeb).

briansmith:
@andrewtj That's great news! It seems like there were several problems.

The problem *I* was experiencing previously was "adb: command not found" and "emulator: command not found" errors on Travis CI. Now it seems those errors aren't occurring! Maybe the problem was just that the Travis CI cache was in a corrupted state or something.

Because of those "command not found" errors, I wasn't able to notice the problem that your patch fixes, which apparently is due to a change in rustc. Thanks for providing an example fix.

I went back to tag 0.12.1 and applied that fix to the 0.12.1 tag and kicked off a Travis CI build. The illegal instruction error still occurs. However, back when I released 0.12.1, the build passed just fine; see https://travis-ci.org/briansmith/ring/builds/271520581. That indicates to me that there is probably a change in the emulator, or a change in rustc or some other part of the toolchain that makes it incompatible with the emulator.

andrewtj:
Okay I've run a few more builds and it looks like [1.21.0 works](https://travis-ci.org/andrewtj/ring/jobs/331685981) but [1.22.1 does not](https://travis-ci.org/andrewtj/ring/jobs/331685979). So I guess something must have changed in Rust's jemalloc fork in 1.22?

[...]

PR #608, opened 2017-12-26
Support reusable private key for agreement

kamyuentse:
Support reusable private key for agreement, introduce a structure `Inner`, which both used in `EphemeralPrivateKey` and `ReusablePrivateKey`.


davidrusu:
@kamyuentse mind rebasing against master so we can see some green builds (#603)?

kamyuentse:
@davidrusu rebased!

PR #610, opened 2017-12-28
Make per-nonce input size limit checks for AES-GCM consistent.

briansmith:
The checks in gcm.c are correct but the other checks are not correct.
Remove the redundant and not-tight-enough `aead_check_in_len()`
assertions. Fix the checks in the Rust code to match the checks in
gcm.c.

Thanks to Frithjof Schulze for reporting this.

### DIFF
--- a/src/aead/mod.rs
+++ b/src/aead/mod.rs
@@ -25,11 +25,6 @@
 //! [AEAD]: http://www-cse.ucsd.edu/~mihir/papers/oem.html
 //! [`crypto.cipher.AEAD`]: https://golang.org/pkg/crypto/cipher/#AEAD
 
-pub mod chacha20_poly1305_openssh;
-
-mod chacha20_poly1305;
-mod aes_gcm;
-
 use {constant_time, error, init, poly1305, polyfill};
 
 pub use self::chacha20_poly1305::CHACHA20_POLY1305;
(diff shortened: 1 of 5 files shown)


briansmith:
@schulze Thanks again for reporting this. Could you please review this PR?

src/aead/mod.rs
```diff
@@ -334,12 +342,14 @@ const TAG_LEN: usize = poly1305::TAG_LEN;
 const NONCE_LEN: usize = 96 / 8;
 
 
-/// |GFp_chacha_20| uses a 32-bit block counter, so we disallow individual
-/// operations that work on more than 256GB at a time, for all AEADs.
-fn check_per_nonce_max_bytes(in_out_len: usize)
+fn check_per_nonce_max_bytes(alg: &Algorithm, in_out_len: usize)
                              -> Result<(), error::Unspecified> {
-    if polyfill::u64_from_usize(in_out_len) >= (1u64 << 32) * 64 - 64 {
+    if polyfill::u64_from_usize(in_out_len) > alg.max_input_len {
```
briansmith: Note the change from `>=` to `>` here. I'm at a loss as to why I was using `>=` before.

briansmith:
Landed in a0d65dd207fae12255bd3a470bc9983391242c1a. Thanks for reporting this!

Issue #615, opened 2017-12-30
Add `ring::hmac::VerifyContext`


bennyyip:
There is `ring::hmac::SigningContext`, but its counterpart is missing. Shall we add it? I can send a PR.

briansmith:
@HeroicKatora implemented this in #671. Today I merged `hmac::SigningKey` and `hmac::VerificationKey` into a single `hmac::Key`. Previously I also renamed `hmac::Signature` to `hmac::Tag`. I propose:

1. Rename `hmac::SigningContext` to `hmac::Context`
2. Rename `SigningContext::sign() -> Tag` to `SigningContext::finish() -> Tag` (analogous to `digest::Context::finish()`).
3. Rename `hmac::sign() -> `hmac::hmac()` (analogous to `digest::digest()`).
4. `impl PartialEq<&[u8]> for Tag` and remove `hmac::verify()`.

Thoughts?

HeroicKatora:
The most immediate benefit of having `verify` take some existing byte tag is ensuring the programmer can not–by accident, negligence or ignorance–make the alleged signature value depend on the output of `hmac`, i.e. the actual signature value. 

This doesn't mean the current interface can't be enhanced, for example I see no harm in the extra comparison implementation (nevertheless it should be `impl PartialEq<[u8]> for Tag`, without an extra ref around the byte slice) and similarly `verify` could make use of this property similar to how `HashMap::get` works:

* `SigningContext::sign() -> Tag`
* `SigningContext::verify<T: PartialEq<Tag>>(alleged: &T) -> Result<(), error::Unspecified>`

An appropriate explanation in `check` could explain how to perform this manually by using `sign` and comparing the signatures but warning of the misuse. Related, `finish` seems more neutral instead of `sign` and part of the issue with misuse is the missing *simpler* alternative for verification. So the name `finish` could be fine to replace `sign`, being more general to cryptographically versed users without affecting use for beginners too much (although I have no studies on this and am simply guessing).

HeroicKatora:
Under the consideration that:

* Making signing and verification similar streamlines code for both parties.
* Verification of data should be possible without inspecting the final `Tag`.
* Partial updates (I-U-F) for the purpose of verification makes it easier to wrongly further process data that has not been verified.
* Functions with generic arguments have a monomorphisation overhead.
[...]

PR #616, opened 2017-12-31
Add VerifyingContext

bennyyip:
#615


briansmith:
Sorry for the long delay. Thanks for your contribution here. I'm going to close this because I want to work together to design the API for this in #615 before we look at the code. After we figure out what we want the API to look like (in the discussion of #615), I hope we can work together to adopt your solution in a modified form.

Issue #514, opened 2017-04-15
Remove the syscall(2) dependency.


briansmith:
syscall(2) is a varargs function, which we strive to avoid categorically. We should probably just issue the getrandom syscall with inline assembly, or, when we get to the end of removing C dependencies, straight assembly language.

briansmith:
Also, we're striving to remove dependencies on libc (both the actual libc and the libc crate). The use of `syscall()` also forces us to care about `errno`, which is mutable global state, which we're also trying to avoid depending on.

briansmith:
@jeandudey submitted a PR for this in #617. Thanks!

@jeandudey's PR uses hand-written (AFAICT) inline asm code for x86 and x86-64. It's missing 32-bit ARM and 64-bit ARM support so I cannot check it in as-is. Also, I'd like to avoid hand-written assembly code if at all possible.

I want to suggest an alternative approach:

The Android bionic project is an implementation of libc for Android; See https://android.googlesource.com/platform/bionic.git.  The libc/tools/gensyscalls.py program in that project reads libc/SYSCALLS.TXT and generates assembly language code that wraps the syscall in a C function for all platforms that Android supports. I would like to use that approach to generate a `GFp_getrandom()` function:

* Create an input file that contains a single line: `ssize_t getrandom(void*, size_t, unsigned) all`
* Extract libc/tools/gensyscalls.py from bionic and put it into third-party/bionic.
* Modify libc/tools/gensyscalls.py so that it generates functions with a "GFp_" prefix so that we don't collide with any `getrandom()` function defined by libc on any platform.
* Modify libc/tools/gensyscalls.py so the generated code does NOT do anything with `errno`.
* Run the modified libc/tools/gensyscalls.py and save the output assembly language files to third-party/bionic.
* Add all the files in third-party/bionic to version control so that we don't have to run libc/tools/gensyscalls.py during the build (so we don't require Python to build *ring*).
* Modify the assembly language files to build.rs so that they get assembled and linked into Linux builds (only).
* Modify the Rust code to use the new `GFp_getrandom()` function on all Linux targets instead of `syscall()`.

@jeandudey WDYT? It's basically the same idea as yours, but without any hand-written code.

jeandudey:
@briansmith Hi! thanks for the response, i'm going to take a look at it.

FauxFaux:
I had a look at doing this. I don't think it's a good idea.

[...]

PR #617, opened 2018-01-02
Remove dependency on syscall(2)

jeandudey:
I don't have tested it in x86-64 only in x86.


briansmith:
I'm going to close this. As I commented in #514, I really appreciate this PR but I think we need to implement some of the details differently. In particular, I really want to avoid needing us to manually write, review, and maintain the assembly language code. In #514 I suggested a way to generate the assembly code like Android does and use it. That, combined with other aspects of this PR, would be perfect.

If you have time to submit a PR that works along the lines I outlined in #514, I would be very happy to work with you on it.

Thanks again.

Issue #540, opened 2017-05-18
Implement `core::fmt::Debug` for all public Algorithm types


briansmith:
I suggest that the implementation should just output the same thing it would output if the Algorithm type were defined to be an empty struct type, like this:

```rust
struct Algorithm;
```

In particular, the fields of the algorithm fields are not interesting.

Ideally, we'd create a macro (e.g. in a new `ring::debug` submodule) to make it easy to add such a Debug implementation to new types, and use that macro for all these Algorithm types.

briansmith:
Oh, sorry, I messed up in the description above. Consider a type like `ring::aead::Algorithm`. I would expect the `Debug` implementation to output something like "AES_128_GCM", `AES_256_GCM", "CHACHA20_POLY1305" or "ring::aead::AES_128_GCM", `ring::aead::AES_256_GCM", "ring::aead::CHACHA20_POLY1305" or something.

briansmith:
I believe this is done.

PR #618, opened 2018-01-02
Issue #540 - Adds Debug implementation for public algorithm types

tcsc:
This changeset adds the ring::debug module, containing the
derive_debug_from_field!() macro, which generates an implementation of
fmt::Debug that defers the formatting to a given field.

This macro is then used to generate fmt::Debug implmentations for
aead::Algorithm and digest::Algorithm which print the algorithm id.

### DIFF
--- a/src/aead/mod.rs
+++ b/src/aead/mod.rs
@@ -305,8 +305,10 @@ impl Algorithm {
     pub fn nonce_len(&self) -> usize { NONCE_LEN }
 }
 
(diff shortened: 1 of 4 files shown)


briansmith:
Thanks! I landed this as ba9f3b2c27f823fd9a2e97dba68394ac65916d2a. The only change I made was to adjust the commit message to avoid referencing GitHub.

PR #620, opened 2018-01-05
truncate auth tag by passing a tag length.

flier:
As you known, [EVP_AEAD_CTX_init](https://man.openbsd.org/EVP_AEAD_CTX_init.3) allow to truncate the authentication tags by passing a tag length.

> **EVP_AEAD_CTX_init()** initializes the context ctx for the given AEAD algorithm aead. The impl argument must be NULL for the default implementation; other values are currently not supported. _Authentication tags may be truncated by passing a tag length. A tag length of zero indicates the default tag length should be used_.

Some protocol, such as [QUIC](https://www.chromium.org/quic), is using [a shorter authentication tag](https://chromium.googlesource.com/chromium/src/+/master/net/quic/core/crypto/aes_128_gcm_12_decrypter.h#25) (12 bytes), which may break `open_in_place` function with hard code `TAG_LEN`.


src/aead/mod.rs
```diff
@@ -42,6 +42,7 @@ pub use self::aes_gcm::{AES_128_GCM, AES_256_GCM};
 /// Go analog: [`crypto.cipher.AEAD`]
 pub struct OpeningKey {
     key: Key,
+    auth_tag_len: usize,
```
briansmith: Instead of adding this field to the key, it shuold be added to the algorithm.

src/aead/mod.rs
```diff
@@ -58,8 +59,33 @@ impl OpeningKey {
...
+    /// A tag length of zero indicates the default tag length should be used.
+    ///
+    /// C analogs: `EVP_AEAD_CTX_init_with_direction` with direction
+    ///            `evp_aead_open`, `EVP_AEAD_CTX_init`.
+    ///
+    /// Go analog:
+    ///   [`crypto.aes.NewCipher`](https://golang.org/pkg/crypto/aes/#NewCipher)
+    /// + [`crypto.cipher.NewGCM`](https://golang.org/pkg/crypto/cipher/#NewGCM)
+    #[inline]
+    pub fn with_auth_tag_len(algorithm: &'static Algorithm, key_bytes: &[u8], auth_tag_len: usize)
```
briansmith: Instead of doing this as new constructors, create new AES_128_GCM_TRUNCATED_TAG_96 and CHACHA_POLY1305_TRUNCATED_TAG_96 algorithms.

flier:
> In addition to the other changes requested here, we need to copy the test vectors out of BoringSSL and Google's QUIC code for 96-bit tags for AES-GCM and ChaCha20-Poly1305.

For Google's QUIC test vectors, it could passed with [test cases](https://github.com/flier/rust-quic/blob/develop/src/crypto/aead.rs#L155).

briansmith:
ugh...something seems wrong with GitHub now. My review comments are not getting attached to the right places it seems. I'll re-review this in the morning.

Anyway, this is looking much better. Please look at how the tests for the pre-existing AES-GCM and ChaCha20-Poly1305 code are written. You'll need to include similar test vectors for the new algorithms. You should find the test vectors in BoringSSL that use 96-bit tags and adapt them for *ring*'s test suite like I did for the 128-bit test vectors. Separate the test vectors for the new algorithms into their own .txt files: One new test file for AES-GCM with 96-bit tags, one new test file for ChaCha20-Poly1305 with 96-bit tags.

Also we need to add additional documentation stating that these 96-bit algorithms are used by Google's variant of QUIC. We should also recommend that people use the default 128-bit variants instead.

src/aead/mod.rs
```diff
@@ -338,3 +342,118 @@ fn check_per_nonce_max_bytes(in_out_len: usize)
...
+#[cfg(test)]
+mod tests {
+    use std::iter;
+    use std::vec::Vec;
+
+    use test;
+    use super::*;
+
+    #[test]
+    fn test_aes_128_gcm_truncated_tag_96() {
```
briansmith: In aead_test.rs, there are tests `aead_aes_gcm_128()`, `aead_aes_gcm_256()`, and `aead_chacha20_poly1305()`. Please move your tests there and make them work the exact same way as those existing tests. In particular, you should use the `test_aead()` function. Also move chacha_poly_truncated_tag_96.txt and aes_gcm_truncated_tag_96.txt to that same directory.

Note in particular that `test_aead()` uses the same inputs to test both encryption and decryption so you don't need separate [encrypt] and [decrypt] sections in your test files; just make them all one big section.

The tests are in tests/aead_test.rs because they're testing the public API; we generally put tests of the public API in tests under tests/ instead of with these kind of same-module-file unit tests.

src/aead/aes_gcm_truncated_tag_96.txt
```diff
@@ -0,0 +1,230 @@
+[encrypt]
```
briansmith: At the top of this test input file and at the top of the other test input file, include a note that says that these test vectors come from the Chromium test suite, and include the path to the test within the Chromium repo.

tests/aead_aes_128_gcm_truncated_tag_96_tests.txt
```diff
@@ -0,0 +1,235 @@
+# The AES GCM test vectors come from the file gcmDecrypt128.rsp
+# downloaded from http://csrc.nist.gov/groups/STM/cavp/index.html
+
+# test_group_0
+
+KEY = 11754cd72aec309bf52f7687212e8957
```
briansmith: Hi, why did you switch from the Chromium test vectors to these NIST test vectors? I was expecting you would adjust the Chromium test vectors to be in the same form as this. I think we should stick with the Chromium test vectors for this. Is there something missing from the Chromium test vectors that make it hard/impossible to adapt them to this form?

tests/aead_tests.rs
```diff
@@ -64,10 +75,11 @@ fn test_aead(aead_alg: &'static aead::Algorithm, file_path: &str) {
         let plaintext = test_case.consume_bytes("IN");
         let ad = test_case.consume_bytes("AD");
         let mut ct = test_case.consume_bytes("CT");
-        let tag = test_case.consume_bytes("TAG");
+        let mut tag = test_case.consume_bytes("TAG");
         let error = test_case.consume_optional_string("FAILS");
 
         let tag_len = aead_alg.tag_len();
+        tag.truncate(tag_len);
```
briansmith: Why truncate `tag` here instead of just changing the input file data to have the right length?

One of the reasons I want to use the Chromium test vectors is precisely to make sure we're chopping of the tag in the right place. That is, they give us confidence that chopping off the end of the 128-bit tag to get a 96-bit tag is the right thing to do.

tests/aead_aes_128_gcm_truncated_tag_96_tests.txt
flier: sure, I use the test vectors from Chromium, but they comment it [come from NIST](https://chromium.googlesource.com/chromium/src/+/master/net/quic/core/crypto/aes_128_gcm_12_decrypter_test.cc#19).

tests/aead_tests.rs
flier: In fact, I copied [the behaviors from the Chromium](https://chromium.googlesource.com/chromium/src/+/master/net/quic/core/crypto/aes_128_gcm_12_decrypter_test.cc#261), I guess they want to keep a full tag.

flier:
@briansmith need any more updates?

briansmith:
@flier Sorry for the lack of follow-through on my part for this. Is this still relevant? My understanding is that everybody is quickly implementing the new version of QUIC that doesn't use this truncated auth tag: "All ciphersuites currently defined for TLS 1.3 - and therefore QUIC - have a 16-byte authentication tag and produce an output 16 bytes larger than their input."

PR #623, opened 2018-01-15
Remove syscall(2) dependency.

jeandudey:


briansmith:
@jeandudey You (accidentally?) included lots (100,000 lines worth) of Android header files in this. We need to find a way to prune it down to the minimum amount.

For example, I think you need the syscall numbers in the .S files; we could just whittle down the Android header files to the point where they are like the `#define`s you deleted from *ring* in this PR.

Maybe there are other header files that contain macros that are needed by the asm files. If so, please call them out and we'll figure out a way to get rid of the need for them or reduce them to the minimum set.

build.rs
```diff
@@ -127,6 +127,11 @@ const RING_SRCS: &'static [(&'static [&'static str], &'static str)] = &[
     (&[AARCH64], "crypto/ec/asm/ecp_nistz256-armv8.pl"),
     (&[AARCH64], "crypto/poly1305/asm/poly1305-armv8.pl"),
     (&[AARCH64], SHA512_ARMV8),
+
+    (&[X86], "third-party/bionic/libc/arch-x86/syscalls/GFp_getrandom.S"),
+    (&[X86_64], "third-party/bionic/libc/arch-x86_64/syscalls/GFp_getrandom.S"),
+    (&[ARM], "third-party/bionic/libc/arch-arm/syscalls/GFp_getrandom.S"),
+    (&[AARCH64], "third-party/bionic/libc/arch-arm64/syscalls/GFp_getrandom.S"),
```
briansmith: Don't we need to add all the header files added in this commit to `RING_INCLUDES`?

briansmith:
I'm closing this because my original idea turned out to be not-so-good after all. See https://github.com/briansmith/ring/issues/514#issuecomment-404673694 and the comment above it. Thanks for submitting this!

PR #625, opened 2018-01-20
Change verify_slices_are_equal to use the subtle crate

Aaronepower:
[Source of `subtle::slices_equal`](https://github.com/dalek-cryptography/subtle/blob/master/src/lib.rs#L497-L512)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -278,6 +278,7 @@ name = "ring"
 
 [dependencies]
 libc = "0.2.34"
+subtle = "0.5"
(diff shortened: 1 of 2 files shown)


briansmith:
Thanks for the PR. I have some plans to improve the way the constant-time stuff is done in *ring*, but those don't involve using the `subtle` crate. Right now let's just leave it as-is.

briansmith:
See #626, where I wrote down my plan for this and similar functions.

PR #633, opened 2018-03-01
Check __ANDROID_API__ instead of defining it automatically

tomaka:
This pull requests removes the code that automatically defines the `__ANDROID_API__` constant, and replaces it by a code that checks whether the constant has been defined by something else.
In other words, we now assume that the user manually defines `__ANDROID_API__`, for example through the `CFLAGS` environment variable.

The main reason is practical. If you generate a standalone NDK toolchain in order to compile your project, the wrapper automatically passes the `__ANDROID_API__` constant to the actual compiler. Therefore if you try to compile *ring* for the standalone toolchain, you will most likely get a compilation error because the constant passed by the standalone conflicts with the one defined by ring.

I don't know how widely used *ring* for Android is, but I'm open to feedback about this change. I don't really have a very strong opinion ; my goal is mainly to make Android compilation as smooth as possible.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -58,6 +58,9 @@ Note in particular that if you are cross-compiling an x86 build on a 64-bit
 version of Linux, then you need to have the proper gcc-multilibs and
 g++-multilibs packages or equivalent installed.
 
+On Android, you will need to define the `__ANDROID_API__` constant to at least
+`21` on 64-bit and `18` on 32-bit.
+
 
 
 Additional Features that are Useful for Development
--- a/build.rs
+++ b/build.rs
@@ -587,17 +587,6 @@ fn cc(file: &Path, ext: &str, target: &Target, warnings_are_errors: bool,
         // http://www.openwall.com/lists/musl/2015/06/17/1
         let _ = c.flag("-U_FORTIFY_SOURCE");
     }
-    if target.os() == "android" {
-        // Define __ANDROID_API__ to the Android API level we want.
-        // Needed for Android NDK Unified Headers, see:
-        // https://android.googlesource.com/platform/ndk/+/master/docs/UnifiedHeaders.md#Supporting-Unified-Headers-in-Your-Build-System
-        if target.arch() == "aarch64" {
-            // Minimum API level where AArch64 is available is 21.
-            let _ = c.define("__ANDROID_API__", Some("21"));
-        } else {
-            let _ = c.define("__ANDROID_API__", Some("18"));
-        }
-    }
 
     let mut c = c.get_compiler().to_command();
     let _ = c.arg("-c")


tomaka:
The current CI failure is not caused by this PR, however this PR may require changes in the CI configuration.

build.rs
```diff
@@ -590,14 +592,15 @@ fn cc(file: &Path, ext: &str, target: &Target, warnings_are_errors: bool,
...
-            let _ = c.define("__ANDROID_API__", Some("18"));
+        // We need __ANDROID_API__ to be defined to the Android API we want.
+        // To check this, we compile a dummy file that errors if it's not what we want.
+        let required_api_level = if target.arch() == "aarch64" { 21 } else { 18 };
+        let mut tempfile = tempfile::NamedTempFile::new().unwrap();
+        write!(tempfile, "#if __ANDROID_API__ >= {} \n#else\n#error Bad\n#endif\nint main() {{}}",
+               required_api_level).unwrap();
+        if let Err(_) = cc::Build::new().warnings(false).file(tempfile.path()).try_expand() {
+            panic!("You need to define __ANDROID_API__ to at least {} for ring to compile",
+                   required_api_level);
```
briansmith: Rather than generate a temporary file to test this, autoconf style, it seems better to just put the minimum version check `#if` logic in crypto/internal.h. Is there a reason you didn't do it that way?

tomaka: I did it this way for the nicer-looking user-friendlier error message, and also to consider defining the constant ourselves automatically if it is not provided by the user.
But I can do it in `crypto/internal.h` if you want.

Cargo.toml
```diff
@@ -290,6 +290,7 @@ lazy_static = "1.0"
 # control about what should be parallised in which way
 cc = "1.0.9"
 rayon = "1.0.0"
+tempfile = "2.2.0"
```
briansmith: If we're going to keep this, then this should be a conditional dependency only when the target OS is Android, if that is possible.

build.rs
briansmith: I propose that we keep the code removal in your patch, but drop the code addition. Then, in the build documentation at https://github.com/briansmith/ring/blob/master/BUILDING.md, let's add a section on building for Android that indicates exactly what one must do to CFLAGS w.r.t. `__ANDROID_API__` when building for Android.

In particular, IIUC, the Android SDK already does a check that `__ANDROID_API__` is defined, so our check is somewhat superflous.

BUILDING.md
```diff
@@ -58,6 +58,9 @@ Note in particular that if you are cross-compiling an x86 build on a 64-bit
 version of Linux, then you need to have the proper gcc-multilibs and
 g++-multilibs packages or equivalent installed.
 
+On Android, you will need to define the `__ANDROID_API__` constant to at least
+`21` on 64-bit and `18` on 32-bit.
```
briansmith: How does this sound?:

If you generate a standalone NDK toolchain in order to compile your project, the wrapper automatically passes the __ANDROID_API__ constant to the actual compiler. Otherwise, the `__ANDROID_API__` must be defined with a value of at least 21 on 64-bit targets or 18 on 32-bit targets (e.g. `export CFLAGS=-D__ANDROID_API__=21`).

briansmith:
Thanks! Merged!

briansmith:
I pushed df1ee1191583552e89a6939c8738393044937bae to clarify the wording a bit like I suggested above.

briansmith:
Reverted in b98f61f9718df0fdf291541c1ca4d0ae45baedf3 and b98f61f9718df0fdf291541c1ca4d0ae45baedf3.

The Android build breaks in Travis CI; see https://travis-ci.org/briansmith/ring/jobs/379021311 for an example. I think the issue is that our Travis CI configuration is using GCC instead of clang; see https://github.com/android-ndk/ndk/wiki/Changelog-r14-beta1#wont-fix where it says standalone toolchains using GCC need to set define this macro manually. Probably we should switch to clang on Travis CI for all Android targets.

tomaka:
> standalone toolchains using GCC need to set define this macro manually. 

This should be done through the environment variables of cc-rs in my opinion.

pietro:
Unfortunately I really haven't had time to deal with this. When I tried using clang the tests would fail. @tomaka can you build and test ring using clang for android?

briansmith:
https://travis-ci.org/briansmith/ring/builds/382297181 is branch that attempts to reland this.

briansmith:
Relanded as 08ec4f374caf1ef2bc079035642b89fbefdb5ed2 and 08ec4f374caf1ef2bc079035642b89fbefdb5ed2. Thanks!

tomaka:
Cool, thank you!

PR #638, opened 2018-03-19
Remove -fdata-sections and -ffunction-sections (let cc set these if needed)

mehcode:
Since https://github.com/alexcrichton/cc-rs/pull/294 this crate has stopped compiling on iOS targets with:

```
clang: error: -fdata-sections is not supported with -fembed-bitcode
clang: error: -ffunction-sections is not supported with -fembed-bitcode
```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -191,8 +191,6 @@ fn c_flags(target: &Target) -> &'static [&'static str] {
 fn cpp_flags(target: &Target) -> &'static [&'static str] {
     if target.env != MSVC {
         static NON_MSVC_FLAGS: &'static [&'static str] = &[
-            "-fdata-sections",
-            "-ffunction-sections",
             "-pedantic",
             "-pedantic-errors",
             "-Wall",


build.rs
```diff
@@ -191,8 +191,6 @@ fn c_flags(target: &Target) -> &'static [&'static str] {
 fn cpp_flags(target: &Target) -> &'static [&'static str] {
     if target.env != MSVC {
         static NON_MSVC_FLAGS: &'static [&'static str] = &[
-            "-fdata-sections",
-            "-ffunction-sections",
```
briansmith: The design of *ring* is heavily oriented to make `-fdata-sections` and `-ffunction-sections`, or equivalent dead code removal, work. I don't want to remove these flags by default.

If I understand correctly, the problem is specific to iOS targets. The best solution seems to be to skip these flags on iOS targets only, if iOS targets are the ones that are broken. WDYT?

Also, do iOS targets need additional flags? It seems like, from the cc-rs bug report, that additional flags would be useful.

mehcode: I guess I didn't explain in the summary but the point is that `cc` adds those flags unconditionally as long as you aren't on the iOS platform. So this should be harmless to remove from ring's build script.

See:

https://github.com/alexcrichton/cc-rs/blob/a2c478febababd8cd921fda2839e414953bd4218/src/lib.rs#L1068-L1071

briansmith:
Thanks for the explanation! I merged this.

PR #639, opened 2018-03-21
Revert the library version changes

tomaka:
The idea of this PR is that we're compatible with the `ring 0.12.1` that is on crates.io, so that we can add an override inside the parity repo.


tomaka:
Err, sorry, wrong repo.

PR #646, opened 2018-05-05
Add missing forward declaration for |LIMBS_equal_limb|.

frewsxcv:
|LIMBS_equal_limb| was added in:

* e70787436f03b84cf08601352460d51b23ba4366

But the header was not updated, resulting in compile failures.


briansmith:
I just did this myself. Thanks for the PR though!

Issue #648, opened 2018-05-08
Windows build fails with VS 2017 15.7.0


kjeremy:
I just updated my VS2017 to the latest stable 15.7.0.  Now ring 0.12.1 fails to compile.

The first errors I see when running `cargo build` complain about errors being treated as warnings.

> c:\users\jkolb\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.12.1\crypto\limbs\limbs.inl(128) : error C2220: warning treated as error - no 'object' file generated
c:\users\jkolb\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.12.1\crypto\limbs\limbs.inl(128) : warning C5045: Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified
c:\users\jkolb\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.12.1\crypto\limbs\limbs.inl(127) : note: index 'i' range checked by comparison on this line
c:\users\jkolb\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.12.1\crypto\limbs\limbs.inl(128) : note: feeds call on this line
c:\users\jkolb\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.12.1\crypto\curve25519\curve25519.c(3510) : error C2220: warning treated as error - no 'object' file generated
c:\users\jkolb\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.12.1\crypto\curve25519\curve25519.c(3510) : warning C5045: Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified
c:\users\jkolb\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.12.1\crypto\curve25519\curve25519.c(3509) : note: index 'i' range checked by comparison on this line
c:\users\jkolb\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.12.1\crypto\curve25519\curve25519.c(3510) : note: feeds call on this line
c:\users\jkolb\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.12.1\crypto\curve25519\curve25519.c(3495) : warning C5045: Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified
[...]

PR #650, opened 2018-05-09
ignore new C5045 VS2017 warning

zpgaal:
Try to fix the compile error for Visual Studio 2017 (15.7.*)
#648


briansmith:
Merged! Thanks! I verified locally that this works. According to the AppVeyor builds, this doesn't break previous compiler versions.

Issue #643, opened 2018-04-27
Add NIST CAVP AES-GCM tests


briansmith:
See https://boringssl.googlesource.com/boringssl/+/3975ecffc22b1d98065d11e94713d258ecddbe93. Ideally we'd do something very similar, but without depending on Go. However, depending on Go in the short term for running those tests isn't a showstopper; we can use BoringSSL's Go program for now and then somebody could replace it later with a Rust equivalent.

briansmith:
See also https://boringssl.googlesource.com/boringssl/+/1d134eee9a7f456cc100ac450abcc67243d8fbdd

brycx:
I would like to take a swing at this, but I'm pretty new so it might take some time. Is that okay?

brycx:
So I took the test vectors from [NIST's homepage](https://csrc.nist.gov/Projects/Cryptographic-Algorithm-Validation-Program/CAVP-TESTING-BLOCK-CIPHER-MODES#GCMVS), and with the below Python script(I will have to rewrite it in Rust later on), modified it to fit the conventions used in the current test vector files for GCM. Then I added a test to `aead_tests.rs` and now I keep running into errors.

What gets me right now is, the test vectors have the "FAIL" option set, with no specification on what type of failure. I'm converting `FAIL` to `FAIL = ""`, and I added a match arm in aead_tests.rs to account for this:
```
                // Accounting for the NIST CAVP empty error string
                Some(ref error) if error == "" => {
                    assert_eq!(Err(error::Unspecified), s_result);
                    assert_eq!(Err(error::Unspecified), o_result);
                },
```
The two errors that keep popping up are:
- `'nist_cavp_aes_128_gcm_decrypt' panicked at 'internal error: entered unreachable code: Unexpected error test case: ""'`
- `thread 'nist_cavp_aes_128_gcm_decrypt' panicked at 'assertion failed: (left == right)
 left: Ok(55),
 right: Err(Unspecified)', tests/aead_tests.rs:179:21`

I would love some pointers on how to proceed with this.

Conversion script:

```
replacements = {
    'Key' : 'KEY',
    'IV' : 'NONCE',
    'PT' : 'IN',
    'AAD' : 'AD',
    'Tag' : 'TAG',
    'FAIL' : 'FAILS = '
}

lines = []

with open('./gcmDecrypt128.rsp') as infile:
    for line in infile:
        for src, target in replacements.items():
            line = line.replace(src, target)
        if line.endswith("= \n"):
[...]

PR #651, opened 2018-05-10
Add NIST CAVP GCMVS tests

brycx:
Issue: #643 

So this is not the most elegant solution, but I hope you can use this anyhow. The way I've done this is by having a script `third_party/NIST/GCMVS/make_cavp_format.py` format the test vectors from [NIST](https://csrc.nist.gov/projects/cryptographic-algorithm-validation-program/cavp-testing-block-cipher-modes). 

Mainly what this does is, because *ring* doesn't accept other than 128-bit tag, it removes all the test cases that do not meet this requirement. 

It also, in `gcmDecryp*` files, changes a `FAIL` to either `FAILS = FAILS_TO_DECRYPT` if the nonce is 96-bits, or `FAILS = WRONG_NONCE_LENGTH` if not. 

It changes all the keywords, so that in the end, all that needs to be added to *ring*'s code is test functions and an error match arm in `aead_tests.rs`.


briansmith:
Thanks for the PR. I'm closing this because I think there are now more useful ways of testing the AES-GCM code than those test vectors and I don't want to spend time on this.

At the time I filed the initial issue things were different. Sorry to waste your time on this.

brycx:
No problem at all, though I'm curious: what more useful ways are you thinking of?

PR #652, opened 2018-05-16
Use clang on travis with Android

tomaka:
Follow-up of #633

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ matrix:
       os: osx
       osx_image: xcode9.2
 
-    - env: TARGET_X=armv7-linux-androideabi CC_X=arm-linux-androideabi-gcc FEATURES_X=--features=rsa_signing MODE_X=DEBUG KCOV=0
+    - env: TARGET_X=armv7-linux-androideabi CC_X=arm-linux-androideabi-clang FEATURES_X=--features=rsa_signing MODE_X=DEBUG KCOV=0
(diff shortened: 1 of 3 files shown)


tomaka:
The travis failure is due to the fact that the inline assembly uses thumb-2 instructions.

According to [this page](https://developer.android.com/ndk/guides/standalone_toolchain#clang_targets_with_arm), we are supposed to use the `thumb-` targets, which makes sense to me.

I'm not sure why it works for gcc, but I suppose that gcc doesn't follow the same system.
I'm also confused as to how Parity successfully compiles despite the fact that is uses clang and [a git version of ring](https://github.com/paritytech/ring/) (so no precompiled header).

briansmith:
> According to this page, we are supposed to use the thumb- targets, which makes sense to me.

Let's change this PR to use the thumb targets so the CI build step passes.

briansmith:
> Let's change this PR to use the thumb targets so the CI build step passes.

Never mind,I don't think that's the right solution. I think I have a fix. Working on it.

briansmith:
> I'm also confused as to how Parity successfully compiles despite the fact that is uses clang and [...]

I bet Parity is using a newer version of the NDK and/or clang or whatever and I bet that is the solution here. Any tips on how to get *ring*'s CI to work more like Parity's?

tomaka:
Tests seem to pass on travis :tada:

PR #655, opened 2018-05-21
Add PBKDF2-HMAC-SHA256 test vectors from RFC 7914

brycx:
[RFC link](https://tools.ietf.org/html/rfc7914#section-11).


briansmith:
Thanks! Merged.

PR #662, opened 2018-06-12
Implement ECDSA nonce hardening to protect against bad PRNG.

briansmith:


ctz:
I'll write out what I see to check understanding.  With:

- *H*: hash function associated with signature, with image size |h| and MD block size |H| in bits
- *q*: group size
- *d*: private scalar, |*q*|-length encoding in the usual way
- *H(m)*: hash of message being signed
- $<sup>n</sup>: n random bits

Precompute:

-  *d<sub>nrk</sub>* = *H*($<sup>|*q*|</sup> || *d*)

Per-signature:

- *k* = *H*(*d<sub>nrk</sub>* || $<sup>|*H*| - |*h*|</sup> || *H(m)*)

If that's accurate it looks good to me.

Issue #661, opened 2018-06-09
Adding deterministic SecureRandom for testing


realcr:
One of the latest update to ring (v0.13.0-alpha5) broke some code we use for a mock deterministic random generator. We use it for testing. This is what it looks like:

```rust
use std::cell::RefCell;

use rand::{self, Rng, StdRng};
use ring::{error::Unspecified, rand::SecureRandom};

pub struct DummyRandom<R> {
    rng: RefCell<R>,
}

impl DummyRandom<StdRng> {
    pub fn new(seed: &[usize]) -> Self {
        let rng_seed: &[_] = seed;
        let rng: StdRng = rand::SeedableRng::from_seed(rng_seed);

        // Note: We use UnsafeCell here, because of the signature of the function fill.
        // It takes &self, it means that we can not change internal state with safe rust.
        DummyRandom {
            rng: RefCell::new(rng),
        }
    }
}

impl<R: Rng> SecureRandom for DummyRandom<R> {
    fn fill(&self, dest: &mut [u8]) -> Result<(), Unspecified> {
        self.rng.borrow_mut().fill_bytes(dest);
        Ok(())
    }
}
```

The compilation error I get is:

```bash
error[E0277]: the trait bound `crypto::test_utils::DummyRandom<R>: ring::private::Sealed` is not satisfied
  --> src/crypto/test_utils.rs:23:14
   |
23 | impl<R: Rng> SecureRandom for DummyRandom<R> {
   |              ^^^^^^^^^^^^ the trait `ring::private::Sealed` is not implemented for `crypto::test_utils::DummyRandom<R>`
```

I assume that a new ring::private::Sealed trait was introduced for safety reasons, to make sure exactly that I don't do something like the code above. Having the ability to generate deterministic random is very important for my tests, because I would like the tests to be reproducible if something goes wrong. I don't want to have a test that sometimes succeeds and sometimes fails.

Here is some part from a test case where I used DummyRandom:

```rust
    use super::super::test_utils::DummyRandom;

    #[test]
    fn test_rand_values_store() {
        let rng = DummyRandom::new(&[1, 2, 3, 4, 5]);

        // Generate some unrelated rand value:
        let rand_value0 = RandValue::new(&rng);

        let mut rand_values_store = RandValuesStore::new(&rng, 50, 5);
        let rand_value = rand_values_store.last_rand_value();
        // ...
```

Now the options I am left with are to use one of: 
[...]

PR #674, opened 2018-07-10
Add SeededFixedRandom

eoger:
Somewhat related to #661, this adds a fake rng initialized with a seed that always holds the same results. I used the LFSR algorithm since it's pretty concise.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -466,6 +466,31 @@ pub mod rand {
     use core;
     use {error, polyfill, rand};
     use private;
+    use std::cell::Cell;
+    use std::mem;
+
+    /// An implementation of `SecureRandom` that always holds the same results
+    /// provided the same seed.
+    #[derive(Debug)]
+    pub struct SeededFixedRandom {
+        pub start: Cell<u16>,
+    }
+
+    impl rand::SecureRandom for SeededFixedRandom {
+        fn fill(&self, dest: &mut [u8]) -> Result<(), error::Unspecified> {
+            let start = self.start.get();
+            for i in 0..dest.len() {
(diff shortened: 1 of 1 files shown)


briansmith:
I understand that you wrote this because you are writing tests. Did you see the test support code in https://briansmith.org/rustdoc/ring/test/rand/index.html? I would expect that the existing stuff, especially `FixedSliceSequenceRandom`, would be sufficient for almost all tests, since they are what I use for the NIST-supplied test vectors and similar.

I'd rather avoid adding any implementation of `SecureRandom` that isn't obviously only useful for tests like `FixedSliceSequenceRandom` since the purpose of `SecureRandom` is to ensure the output is not predictable even by the application using it.

For SGX and for other reasons we'll likely add a secure DRBG that takes a seed and that can be augmented with more seed material (reseeded). I don't know which one we would pick--probably HMAC-DRBG.

eoger:
I tried using `FixedSliceSequenceRandom` with `EphemeralPrivateKey::generate()` but it's a bit harder since the way we get these random slices are not trivial: https://github.com/briansmith/ring/blob/9814ac47dff08424dd75e31881e7ba257880a096/src/ec/suite_b/private_key.rs#L62-L68

briansmith:
> I tried using FixedSliceSequenceRandom with EphemeralPrivateKey::generate() but it's a bit harder since the way we get these random slices is not trivial:

In the case of P-256 ECDH, you need one 32-byte value per key generation as long as the value isn't "too big". In practice it is actually kind of difficult to make a "too big" value so almost any value will work. In particular, any 32-byte value where the high-order bit isn't set will work.

eoger:
I've made it work with your suggestions, thank you! Closing this for now.

PR #676, opened 2018-07-12
Export ring::signature::ECDSASigning

tarcieri:
Exports `ec::suite_b::ecdsa::signing::Algorithm` as `ring::signature::ECDSASigning` to match a similar export for `ECDSAVerification`.

### DIFF
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -288,6 +288,7 @@ use std;
 
 pub use ec::suite_b::ecdsa::{
     signing::{
+        Algorithm as ECDSASigning,
         Key as ECDSAKeyPair,
         ECDSA_P256_SHA256_ASN1_SIGNING, ECDSA_P256_SHA256_FIXED_SIGNING,
         ECDSA_P384_SHA384_ASN1_SIGNING, ECDSA_P384_SHA384_FIXED_SIGNING,


briansmith:
I think we exported the ECDSA verification algorithm type due to a limitation in how Rust's type/visibility system works or perhaps due to a misunderstanding of it. We have `signature::SigningAlgorithm` which I expected to work for every use case. Can you use `signature::SigningAlgorithm` instead?

tarcieri:
In this particular case I was trying to store it in a struct. I actually ended up avoiding that

briansmith:
> In this particular case I was trying to store it in a struct. I actually ended up avoiding that

You can store a reference to the algorithm in a struct. It should be easy to do because `SigningAlgorithm: 'static`.

tarcieri:
Ran into this again. I'd like to pass the signing algorithm to use as a parameter to a method which in turn calls `ECDSAKeyPair::from_pkcs8()`.

briansmith:
> I'm not sure how I'd do that with SigningAlgorithm? The from_pkcs8() method concretely wants a &'static ring::ec::suite_b::ecdsa::signing::Algorithm.

Use `ring::signature::key_pair_from_pkcs8()`.

tarcieri:
Aha, thanks

PR #678, opened 2018-07-13
Add NotRandom

juchiast:
Hi, I implemented `NotRandom`, as you suggested in #661 (I use `Fn` instead of `FnMut`).

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -532,9 +532,18 @@ pub mod rand {
         }
     }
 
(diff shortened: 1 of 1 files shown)


realcr:
@briansmith : Hi, are there plans to merge @juchiast 's pull request as a solution for issue https://github.com/briansmith/ring/issues/661 ?
Thank you for your help!

briansmith:
Hi, I'm going to close this for now. There are too many PRs open to address this issue and first we should figure out what we even want to do. If this turns out to be the approach we decide on then let's reopen this after we decide that.

PR #680, opened 2018-07-21
Added `RING_DISABLE_PREGENERATED` enviroment flag

akru:
Motivation: https://github.com/paritytech/parity-ethereum/issues/8612

Currently pregenerated asms used in all cases except sources as git
repo. This commit add special flag `RING_DISABLE_PREGENERATED` as
enviroment variable. It helps to disable using of pregenerated files
in some cases like a cargo-vendor builds.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -378,8 +378,9 @@ fn build_c_code(target: &Target, pregenerated: PathBuf, out_dir: &Path) {
     }).unwrap();
 
     let is_git = std::fs::metadata(".git").is_ok();
+    let disable_pregenerated = std::env::var("RING_DISABLE_PREGENERATED").is_ok();
 
-    let use_pregenerated = !is_git;
+    let use_pregenerated = !is_git && !disable_pregenerated;
     let warnings_are_errors = is_git;
 
     let asm_dir = if use_pregenerated { &pregenerated } else { out_dir };


briansmith:
Isn't it better to fix the cargo-asm builds so they can use the pregenerated asm? After all, isn't cargo-vendor supposed to work exactly like using the crate from crates.io, except without crates.io being involved? Why don't cargo-asm builds have the pregenerated asm files available?

briansmith:
What are the minimal steps to reproduce this problem with cargo-vendor? Is it only a problem when using Nix or does it apply to all uses of cargo-vendor?

akru:
As you can see [here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/rust/fetchcargo.nix#L26) nix build system just run `cargo vendor` command and copy `vendor` folder for future usage. You can reproduce it in `parity-ethereum` forlder, for example.

Unfortunately I have no ideas how to detect cargo-vendor build environment from `build.rs` code.

briansmith:
> As you can see here nix build system just run cargo vendor command and copy vendor folder for future usage. You can reproduce it in parity-ethereum forlder, for example.

Doesn't `cargo vendor` copy the `pregenerated` directory? In other words, isn't this a problem in `cargo vendor`, and not in *ring*?

In the (hopefully near) future it will be more important to be able to use the pregenerated files so I don't think a workaround that skips the pregeneration makes sense except as a last resort. We should figure out why the pregeneration doesn't work and fix it.

briansmith:
OK, it seems this isn't necessary, as explained in https://github.com/alexcrichton/cargo-vendor/issues/89#issuecomment-408564725. Thanks for submitting this, though!

Issue #562, opened 2017-07-20
Add support for mips


saprette:
Hello,

I'm trying to compile Rocket for an Omega2, but it's failing as the ring dependency will not cross compile. Would you have some pointer? Many thanks!

I've been able to cross compile and run an http server on the Omega2 with hello world examples using tiny_http and iron latest but I have an issue cross compiling Rocket due to **'failed to run custom build command for `ring v0.11.0`'** (it does work if I don't cross compile on my laptop). 
I've installed rust & cargo using rustup and I updated them before trying to build Rocket today.

```
sam@sam:~/omega/Rocket/examples/hello_world$ cargo build --target=mipsel-unknown-linux-musl 
   Compiling toml v0.4.2
   Compiling url v1.5.1
   Compiling ring v0.11.0
error: failed to run custom build command for `ring v0.11.0`
process didn't exit successfully: `/home/sam/omega/Rocket/target/debug/build/ring-a7af99b63f4ef999/build-script-build` (exit code: 101)
--- stderr
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', /checkout/src/libcore/option.rs:335:20
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:380
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:396
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:611
   5: std::panicking::begin_panic_new
             at /checkout/src/libstd/panicking.rs:553
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:521
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:497
   8: core::panicking::panic_fmt
             at /checkout/src/libcore/panicking.rs:92
   9: core::panicking::panic
             at /checkout/src/libcore/panicking.rs:51
  10: <core::option::Option<T>>::unwrap
[...]

PR #684, opened 2018-08-01
Add support for MIPS architectures in Travis

mpapierski:
This adds mips/mips64/mipsel/mips64el targets to the build system.

Unfortunately travis `trusty` image doesn't ship with `gcc-mips*`
packages. On these platforms `cargo` is replaced with `cross`[1] to run
those builds on a docker container.

For now all MIPS tests are allowed to fail to make way for future PRs
that will fill the gaps.

This is a first step to bring support to MIPS platforms (#562).

Note that the with this commit matrix really grew up to about 168 builds. This is fine since we have 'fast_finish` (no waiting for `allow_failures` builds).

I'm still waiting for a full build to finish on my fork (https://travis-ci.org/mpapierski/ring). Note that at first I've disabled few tests just to make way for `mips` builds, and I plan to re-run cancelled tests that are known to be green just to make sure everything works as expected.

Expected failure on all mips architectures:

```
--- stderr
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:335:21
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

[1]: https://github.com/japaric/cross

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,54 @@ matrix:
   fast_finish: true
   allow_failures:
     - rust: nightly
+    - env: TARGET_X=mips64-unknown-linux-gnuabi64 CC_X=mips64-linux-gnuabi64-gcc FEATURES_X=--features=rsa_signing MODE_X=DEBUG KCOV=0
+    - env: TARGET_X=mips64-unknown-linux-gnuabi64 CC_X=mips64-linux-gnuabi64-gcc FEATURES_X= MODE_X=DEBUG KCOV=0
+    - env: TARGET_X=mips64-unknown-linux-gnuabi64 CC_X=mips64-linux-gnuabi64-gcc FEATURES_X=--features=rsa_signing MODE_X=RELWITHDEBINFO KCOV=0
+    - env: TARGET_X=mips64-unknown-linux-gnuabi64 CC_X=mips64-linux-gnuabi64-gcc FEATURES_X= MODE_X=RELWITHDEBINFO KCOV=0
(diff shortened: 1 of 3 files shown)


briansmith:
OK, looking at this, it seems like MIPS is a lot more expensive to support than I expected.

Is there really no alternative to using docker and the cross toolkit? I think that on other targets we use `sudo: true` and then install packages. I would prefer to keep MIPS more like the other platforms if at all possible.

mpapierski:
@briansmith Instead of running builds through cross it could be probably achieved in a smiliar way by specifying linker to a cross compiling one, and the runner would be qemu-$ARCH. That would work. Anyway, I'd still recommend cross as it saves a lot of headaches. We have lots of successes with it especially when dealing with MIPS and big endian. We even run x86 tests through cross.

briansmith:
> OK, looking at this, it seems like MIPS is a lot more expensive to support than I expected.
> 
> Is there really no alternative to using docker and the cross toolkit? I think that on other targets we use `sudo: true` and then install packages. I would prefer to keep MIPS more like the other platforms if at all possible.

I have decided, at least for now, to use `cross` for all targets except i686 and x86_64. I've extended your PR to support this on this branch: https://travis-ci.org/briansmith/ring/builds/468249568 and I'll merge it once CI passes.

pietro:
@mpapierski @briansmith Travis now has a Xenial environment that so we can use it for cross compiling mips. I'll add it to nightly only on a branch in my fork. This way it will only add 8 more builds instead of 24.

briansmith:
I had to move away from Travis CI to GitHub Actions. Further, I learned a lot about how to do this cross-compilation stuff so that we don't need to use `cross`. So, this PR is obsolete and I'm going to close it. It isn't your fault; I just didn't get to it in a timely manner. Thanks for contributing this. I will update issue #562 with my current plans.

PR #687, opened 2018-08-13
Very small fix so the PrivateKey and EphemeralPrivateKey structs can be copied.

SmashedCanary:
Nothing exceptional.

### DIFF
--- a/src/agreement.rs
+++ b/src/agreement.rs
@@ -98,6 +98,7 @@ derive_debug_from_field!(Algorithm, i);
 /// An ephemeral private key for use (only) with `agree_ephemeral`. The
 /// signature of `agree_ephemeral` ensures that an `EphemeralPrivateKey` can be
 /// used for at most one key agreement.
+#[derive(Clone, Copy)]
 pub struct EphemeralPrivateKey {
     private_key: ec::PrivateKey,
     alg: &'static Algorithm,
--- a/src/ec/mod.rs
+++ b/src/ec/mod.rs
@@ -62,6 +62,7 @@ pub struct KeyPair {
     pub public_key: [u8; PUBLIC_KEY_MAX_LEN],
 }
 
+#[derive(Clone, Copy)]
 pub struct PrivateKey {
     bytes: [u8; SCALAR_MAX_BYTES],
 }


briansmith:
It is intentional that these types don't implement `Copy` but I'm not sure whether it is good or bad. What's the rationale for wanting them to be `Copy`?

SmashedCanary:
I had an application where I was passing them around a lot, sometimes it was smarter to copy than to reference. Why would it be intentional to not implement Copy for them?

briansmith:
> Why would it be intentional to not implement Copy for them?

In the future there may be alternative implementations of these types, e.g. if/when we add HSM support, where the copying would be expensive and where bytewise copying might not work at all, so `Clone` might make sense but `Copy` wouldn't.

src/agreement.rs
```diff
@@ -98,6 +98,7 @@ derive_debug_from_field!(Algorithm, i);
 /// An ephemeral private key for use (only) with `agree_ephemeral`. The
 /// signature of `agree_ephemeral` ensures that an `EphemeralPrivateKey` can be
 /// used for at most one key agreement.
+#[derive(Clone, Copy)]
```
briansmith: This part is definitely incompatible with the design for `EphemeralPrivateKey`.  `EphemeralPrivateKey` is intentionally move-only so that we can use the type system to ensure that the key is only used once.

src/ec/mod.rs
```diff
@@ -62,6 +62,7 @@ pub struct KeyPair {
     pub public_key: [u8; PUBLIC_KEY_MAX_LEN],
 }
 
+#[derive(Clone, Copy)]
```
briansmith: I think we don't want to rely on `Copy` semantics always being supported. At least, we'd need a very strong argument to make such a guarantee. In the future I'd like to add transparent HSM support and similar things, and `Copy` wouldn't work for that.

briansmith:
I'm going to close this based on the discussion above. I think it's actually pretty reasonable to `impl Clone for PrivateKey` but I probably won't do that until I've explored the PKCS#11 side of things.

PR #690, opened 2018-08-30
Add more variants to `der::Tag`

akuanti:


briansmith:
I probably won't get around to reviewing this unless/until I write/review a PR for webpki that uses the additional tags.

PR #694, opened 2018-10-03
crypto: replace aes_block_f and aes_ctr_f

kevinis:
This commit replaces aes_block_f with block128_f in various locations
including GCM128_CONTEXT. This decouples AES from GCM and leaves a
generic GCM interface for other block ciphers to implement GCM.

The same applies for aes_ctr_f, which is replaced with ctr128_f.

### DIFF
--- a/crypto/fipsmodule/cipher/e_aes.c
+++ b/crypto/fipsmodule/cipher/e_aes.c
@@ -158,12 +158,12 @@ int GFp_AES_set_encrypt_key(const uint8_t *user_key, unsigned bits,
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


briansmith:
> This decouples AES from GCM and leaves a
> generic GCM interface for other block ciphers to implement GCM.

What other block cipher would we want to use with GCM in *ring*? In the past, there was such an abstraction already but *ring* undid it to avoid undefined behavior in C code.  I believe BoringSSL also recently did the same thing.

kevinis:
Agree. Casting from a void * is bad. Closing this PR.

PR #695, opened 2018-10-05
Expose unauthenticated stream cipher API

ghedo:
This adds a new "unauthenticated_stream" module, to expose a plain stream
cipher API, which mostly follows the aead module, but instead of providing
"open" and "seal" operations, it just provides a "xor_keystream" operation
and uses 128 bits nonces instead of 96 bits.

The 128 bits nonce are necessary to match the usage of the stream ciphers
for the QUIC protocol packet number protection scheme, where 16 octets are
sampled from the packet payload and used as explicit nonce for AES-CTR
and ChaCha20.

This also implements support for the ChaCha20 cipher, exposed through the
new API described above, by reusing the existing ChaCha20 C interface. The
first 32 bits of the 128 bits nonce are extracted and used as an explicit
initial counter.

The ChaCha20 test vectors were copied from the AEAD ChaCha20-Poly1305 ones,
with the nonces modified to add an explicit little-endian counter of "1", and
with AD and TAG parameters removed.


src/unauthenticated_stream/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+use {chacha, error, init, polyfill};
+
+pub use self::chacha20::CHACHA20;
+
+/// A key for an unauthenticated stream cipher.
+///
+/// C analog: `EVP_CIPHER_CTX`
+///
+/// Go analog: [`crypto.cipher.Stream`]
+pub struct StreamingKey {
```
briansmith: Even though encryption and decryption are the same math for these ciphers, in *ring* we always have distinct public types for the encryption and decryption keys, and I'd like to continue that unless there is a reason not to. If there is a reason not to, what is it?

src/unauthenticated_stream/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    /// The key's unauthenticated stream cipher algorithm.
+    ///
+    /// C analog: `EVP_CIPHER_CTX.cipher`
+    #[inline(always)]
+    pub fn algorithm(&self) -> &'static Algorithm { self.key.algorithm() }
+}
+
+/// Applies the keystream to the input data in place.
+pub fn xor_keystream_in_place<'a>(key: &StreamingKey, nonce: &[u8],
```
briansmith: I thought in the discussion we agreed that we didn't need an in-place API. I'd rather skip the in-place API and provide an out-of-place API with separate in and out parameters unless the people implementing QUIC think in-place is better. (If in-place is better, why?)

Assuming we have separate encryption key and decryption key types, then we'd also have separate encrypt and decrypt functions.

I don't think we need `keystream` in the name. I find that to be confusing.

ghedo: > I thought in the discussion we agreed that we didn't need an in-place API. I'd rather skip the in-place API and provide an out-of-place API with separate in and out parameters unless the people implementing QUIC think in-place is better. (If in-place is better, why?)

Ah, sorry, I forgot to mention this in the description. I went back to the in-place implementation because it makes things a little bit better for QUIC implementations. The problem is that the QUIC packet header, including the _unencrypted_ packet number, is used as AD for the decryption of the main packet payload. So an implementation needs to first decrypt the packet number, write the packet number plaintext back into the packet header buffer, and then decrypt the payload by splitting the buffer between header (used as AD) and payload (used as ciphertext). Similarly for encryption, the payload is encrypted first, and then the packet number is also encrypted.

Having in-place operations makes this somewhat better because the implementation doesn't need to do the "write packet number plaintext back" operation since it's done automatically. It's not a huge improvement though, so if you'd rather have the operations not be -in-place anyway that can work as well.

djc: Yeah, if the API is only out-of-place we'd probably then have to copy it in place after the fact again.

ghedo:
@briansmith can you have another look please?

src/unauthenticated_stream/chacha20.rs
```diff
@@ -0,0 +1,65 @@
...
+// purpose with or without fee is hereby granted, provided that the above
+// copyright notice and this permission notice appear in all copies.
+//
+// THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHORS DISCLAIM ALL WARRANTIES
+// WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY
+// SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+// OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
```
briansmith: I think `unauthenticated_encryption` is a better module name than `unauthenticated_stream`.

It seems like a lot (most?) of this code is duplicated from the existing ChaCha20 AEAD code. I would really like to avoid this duplication by reorganizing the code so that the ChaCha20-Poly1305 AEAD code is implemented in terms of this new unauthenticated API as much as possible. Basically, I would expect this new API to replace the private API in chacha20.rs, i.e. starting with `git mv src/chacha20.rs src/unauthenticated_encryption/`. See below for specific examples.

If this is hard to do, let's discuss the difficulties and see if we can resolve them together without duplicating the code.

src/unauthenticated_stream/chacha20.rs
```diff
@@ -0,0 +1,65 @@
...
+                                        unauthenticated_stream::Algorithm {
+    key_len: chacha::KEY_LEN_IN_BYTES,
+    init: chacha20_init,
+    xor_keystream: chacha20_xor_keystream,
+    id: unauthenticated_stream::AlgorithmID::CHACHA20,
+    max_input_len: max_input_len!(CHACHA20_BLOCK_LEN, CHACHA20_OVERHEAD_BLOCKS_PER_NONCE),
+};
+
+const CHACHA20_BLOCK_LEN: u64 = 64;
+const CHACHA20_OVERHEAD_BLOCKS_PER_NONCE: u64 = 1;
```
briansmith: These need to be defined in a single place, not duplicated between the ChaCha20-Poly1305 AEAD code and the unauthenticated ChaCha20 code.

src/unauthenticated_stream/chacha20.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// Copies |key| into |ctx_buf|.
+pub fn chacha20_init(ctx_buf: &mut [u8], key: &[u8])
+                              -> Result<(), error::Unspecified> {
+    ctx_buf[..key.len()].copy_from_slice(key);
+    Ok(())
+}
```
briansmith: This is exactly `chacha20_poly1305_init` and shouldn't be duplicated.

src/unauthenticated_stream/chacha20.rs
```diff
@@ -0,0 +1,65 @@
...
+
+fn chacha20_xor_keystream(ctx: &[u64; unauthenticated_stream::KEY_CTX_BUF_ELEMS],
+                          nonce: &[u8; unauthenticated_stream::NONCE_LEN],
+                          in_out: &mut [u8]) -> Result<(), error::Unspecified> {
+    let chacha20_key = ctx_as_key(ctx)?;
+    let mut counter = make_counter(nonce);
+    chacha::chacha20_xor_in_place(&chacha20_key, &counter, in_out);
+    counter[0] = 0;
+    Ok(())
```
briansmith: This is exactly `chacha20_poly1305_seal()` except `chacha20_poly1305_seal()` also does a Poly1305 operation. `chacha20_poly1305_seal()` should be re-implemented in terms of `chacha20_xor_keystream`.

I still don't like the `_keystream` in the name. Let's just remove or replace that. Let's add `_in_place` to the name so that the reader of the calling code can more quickly understand that the input buffer is also the output buffer.

src/unauthenticated_stream/chacha20.rs
```diff
@@ -0,0 +1,65 @@
...
+        chacha::KEY_LEN_IN_BYTES / 4)
+}
+
+#[inline]
+fn make_counter(nonce: &[u8; unauthenticated_stream::NONCE_LEN]) -> chacha::Counter {
+    [u32_from_le_u8(slice_as_array_ref!(&nonce[0..4], 4).unwrap()),
+     u32_from_le_u8(slice_as_array_ref!(&nonce[4..8], 4).unwrap()),
+     u32_from_le_u8(slice_as_array_ref!(&nonce[8..12], 4).unwrap()),
+     u32_from_le_u8(slice_as_array_ref!(&nonce[12..16], 4).unwrap())]
+}
```
briansmith: I hope this is the same as the existing `make_counter`. Let's just have one `make_counter` instead of two.

src/unauthenticated_stream/chacha20.rs
```diff
@@ -0,0 +1,65 @@
...
+                          nonce: &[u8; unauthenticated_stream::NONCE_LEN],
+                          in_out: &mut [u8]) -> Result<(), error::Unspecified> {
+    let chacha20_key = ctx_as_key(ctx)?;
+    let mut counter = make_counter(nonce);
+    chacha::chacha20_xor_in_place(&chacha20_key, &counter, in_out);
+    counter[0] = 0;
+    Ok(())
+}
+
+fn ctx_as_key(ctx: &[u64; unauthenticated_stream::KEY_CTX_BUF_ELEMS])
```
briansmith: I assume this is the same as `ctx_as_key` in the AEAD code. Let's get rid of the duplication.

src/unauthenticated_stream/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+    #[inline(always)]
+    pub fn algorithm(&self) -> &'static Algorithm { self.key.algorithm() }
+}
+
+/// Decrypts the input data in place.
+pub fn decrypt_in_place<'a>(key: &DecryptingKey, nonce: &[u8],
+                            in_out: &'a mut [u8])
+                            -> Result<&'a mut [u8], error::Unspecified> {
+    let nonce = slice_as_array_ref!(nonce, NONCE_LEN)?;
+    check_per_nonce_max_bytes(key.key.algorithm, in_out.len())?;
```
briansmith: Shouldn't `xor_keystream` do the above two steps, since they are the same for encryption and decryption?

src/unauthenticated_stream/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+    key_len: usize,
+    id: AlgorithmID,
+
+    /// Use `max_input_len!()` to initialize this.
+    // TODO: Make this `usize`.
+    max_input_len: u64,
+}
+
+/// TODO: Make this a `const fn` when those become stable.
+macro_rules! max_input_len {
```
briansmith: I think we can/should avoid duplicating this.

src/unauthenticated_stream/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+impl PartialEq for Algorithm {
+    fn eq(&self, other: &Self) -> bool { self.id == other.id }
+}
+
+impl Eq for Algorithm {}
+
+// All the stream ciphers we support use 128-bit nonces.
+const NONCE_LEN: usize = 128 / 8;
```
briansmith: Let's avoid duplicating this definition too.

src/unauthenticated_stream/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+
+fn check_per_nonce_max_bytes(alg: &Algorithm, in_out_len: usize)
```
briansmith: Ditto.

tests/unauthenticated_stream.rs
```diff
@@ -0,0 +1,264 @@
...
+// `open_in_place`. It should be constructing a valid ciphertext using the
+// unsupported nonce size using a different implementation that supports
+// non-standard nonce sizes. So, when `open_in_place` returns
+// `Err(error::Unspecified)`, we don't know if it is because it rejected
+// the non-standard nonce size or because it tried to process the input
+// with the wrong nonce. But at least we're verifying that `open_in_place`
+// won't crash or access out-of-bounds memory (when run under valgrind or
+// similar). The AES-128-GCM tests have some WRONG_NONCE_LENGTH test cases
+// that tests this more correctly.
+fn test_stream_nonce_sizes(stream_alg: &'static unauthenticated_stream::Algorithm)
```
briansmith: I guess this test is exactly the same as the test for the AEAD code, except that unauthenticated. Let's find a way to factor out the common logic so that it can be shared for the AEAD and unauthenticated tests.

tests/unauthenticated_stream.rs
```diff
@@ -0,0 +1,264 @@
...
+            Some(error) => {
+                unreachable!("Unexpected error test case: {}", error);
+            },
+        };
+
+        Ok(())
+    });
+}
+
+fn test_stream_key_sizes(stream_alg: &'static unauthenticated_stream::Algorithm) {
```
briansmith: Let's find a way to abstract this test so that the common logic between AEAD and unauthenticated cases can be shared.

djc:
The high-level API for `EncryptingKey::encrypt_in_place()` and `DecryptingKey::decrypt_in_place()` seems good to me, though I don't yet have the implementation experience that will make me more confident in my judgement on this. I'm going to try and provisionally implement this using another crate to gain some deeper understanding.

`unauthenticated_encryption` is on the long side. Are there other kinds of unauthenticated that we have to differentiate from? A top-level name like `unauthenticated` seems like it would fit names like `aead` and `agreement` fine, too. (It won't matter that much, since it probably will only be in a single `use` statement in practice.)

src/unauthenticated_stream/chacha20.rs
ghedo: ACK.

src/unauthenticated_stream/chacha20.rs
ghedo: ACK.

src/unauthenticated_stream/chacha20.rs
ghedo: ACK.

src/unauthenticated_stream/chacha20.rs
ghedo: It's not quite the same. The existing one takes an explicit counter argument and 12-bytes nonce, while this takes 16 bytes nonce and extracts the first 4 bytes as counter. It could be possible to make a generic one though.

src/unauthenticated_stream/chacha20.rs
ghedo: Actually, there is another difference between this and `chacha20_poly1305_seal`: since the aead module uses 12 bytes nonces with an hardocded initial counter, while this uses the counter as passed by the application as part of a 16 bytes nonce (this specifically is needed for QUIC).

Though `chacha20_poly1305_seal` can still be implemented by using `chacha20_xor_keystream` by just making sure it passed the 16 bytes nonce as returned by `make_counter`.

src/unauthenticated_stream/chacha20.rs
ghedo: ACK.

src/unauthenticated_stream/chacha20.rs
ghedo: One problem of reusing `unauthenticated_encryption::chacha20` as part of `aead::chacha20_poly1305` is that, if I understand things correctly, it would require making `unauthenticated_encryption::chacha20` public. I think it would be better to keep `chacha20` and move the additional common bits and pieces to that instead, so we can keep them private, thoughts?

src/unauthenticated_stream/mod.rs
ghedo: This is different from `aead` (16 bytes here vs. 12 bytes in aead).

briansmith:
> `unauthenticated_encryption` is on the long side.

I don't disagree, but of all the proposed names so far, that's the name I think is best. I want to optimize the naming here for maximum clarity. I expect it to be rare for anybody to use this submodule anyway, and if the length of name discourages poor use, then I see that as an advantage.

src/unauthenticated_stream/chacha20.rs
briansmith: > Also, would need another private module to make common functions and macros shared by `unauthenticated_encryption` and `aead` (that is, unrelated to chacha itself) internally public. Though maybe I'm missing something about how Rust public/private modules work.

I think `pub(crate)` would probably work. It's hard for me to say without trying it.

The important thing is to minimize duplicated logic.

src/unauthenticated_stream/chacha20.rs
ghedo: Right, though the AES-CTR case uses a 16 byte nonce and doesn't mention an explicit counter, so in order to make a generic API we'd either have to make the application pass the explicit counter to encrypt_in_place/decrypt_in_place (and let the application figure out what that means for AES-CTR), or pass a 16 bytes nonce in both cases (and let the application figure out what 16 bytes nonce means for ChaCha20, although the solution is in the text you linked to) which is what I did here.

The latter is somewhat easier from the application POV since ring already has support for turning a 4 byte slice into a u32, while the application would need to implement it on top of using ring's API.

But I can look into doing the explicit counter option and see how much more difficult it is to use vs. how much better it makes the internal ring code.

briansmith: > Right, though the AES-CTR case uses a 16 byte nonce and doesn't mention an explicit counter,

I see. Thanks for explaining that.

For now, I suggest we expose the AES-CTR code as taking a 96-bit nonce and a 32-bit counter, like how QUIC specifies for ChaCah20. Internally *ring*'s AES-CTR implementation only supports incrementing the last 32 bits of the counter so any other style of API (e.g. one that takes a 128-bit initial counter) would be misleading as far as the capabilities go.

ghedo:
FTR here's a very rough implementation of QUIC's packet number decryption using the ring API I implemented here (this is based on my other ring branch that also has AES-CTR support, so I could verify ChaCha20, AES128-CTR and AES256-CTR all work and interop with other QUIC implementations): https://github.com/ghedo/quiche/blob/master/src/packet.rs#L257

src/unauthenticated_stream/chacha20.rs
ghedo: So, trying to figure out how that would work in an application, the problem I see is that that would make having a common API for both ChaCha20 and AES-CTR pointless, because the application would need to be aware of the specific algorithm being used to know how to generate different counters based on the different algorithms (i.e. for ChaCha it would use the first 4 bytes, while for AES-CTR it's actually the last 4 AFAICT).

So, if we go with the explicit counter, I'd say we might as well just expose the raw "ChaCha20()" and "AES-CTR()" operations directly as used in the QUIC spec without having to go to all the trouble of making a generic API which wouldn't really make the application's life easier.

The application would need to generate different key objects depending on the algorithm (so would need to expose key generation APIs as well for both ChaCha and AES), which I think is fine since this isn't intended for general use anyway, and it would discourage use unless really needed.

briansmith: > So, trying to figure out how that would work in an application, the problem I see is that that would make having a common API for both ChaCha20 and AES-CTR pointless,

I agree that it would be less than ideal, but I "pointless" is an overstatement.

> because the application would need to be aware of the specific algorithm being used to know how to generate different counters based on the different algorithms (i.e. for ChaCha it would use the first 4 bytes, while for AES-CTR it's actually the last 4 AFAICT).

With your original proposed API, the application would need to know these details too. It happens that for a QUIC implementation, the details just happen to work for what QUIC needs, but there would be no way for somebody other than us to know that given the API signature and the API documentation given in this PR to this point.

For QUIC, the details of the counter format matter less because QUIC only needs to encrypt one block (IIUC) so the counter never gets incremented during the operation. However, for non-QUIC applications that would use this API, the details of how the counter is constructed (big-endian for AES-GCM, little-endian for ChaCha20, IIUC) and how it is incremented (mod 2**32) and the amount of data that can be encrypted in a single operation (2**32 * block size, approximately) matter a lot.

Now, that said, I also eventually want to extend this API to work for non-CTR modes, in particular CBC mode. For those modes a 96-bit nonce and a 32-bit counter as separate arguments don't make sense. So I think it would be OK to keep the API in the same shape you presented it, especially if the difference in the counter/nonce split between ChaCha20 and AES-GCM is explained by the little-endian vs. big-endian nature of each; Do you agree that the difference comes down to endianness, or is it something different?

briansmith: > So, trying to figure out how that would work in an application, the problem I see is that that would make having a common API for both ChaCha20 and AES-CTR pointless,

I agree that it would be less than ideal, but I "pointless" is an overstatement.

> because the application would need to be aware of the specific algorithm being used to know how to generate different counters based on the different algorithms (i.e. for ChaCha it would use the first 4 bytes, while for AES-CTR it's actually the last 4 AFAICT).

With your original proposed API, the application would need to know these details too. It happens that for a QUIC implementation, the details just happen to work for what QUIC needs, but there would be no way for somebody other than us to know that given the API signature and the API documentation given in this PR to this point.

For QUIC, the details of the counter format matter less because QUIC only needs to encrypt one block (IIUC) so the counter never gets incremented during the operation. However, for non-QUIC applications that would use this API, the details of how the counter is constructed (big-endian for AES-GCM, little-endian for ChaCha20, IIUC) and how it is incremented (mod 2**32) and the amount of data that can be encrypted in a single operation (2**32 * block size, approximately) matter a lot.

Now, that said, I also eventually want to extend this API to work for non-CTR modes, in particular CBC mode. For those modes a 96-bit nonce and a 32-bit counter as separate arguments don't make sense. So I think it would be OK to keep the API in the same shape you presented it, especially if the difference in the counter/nonce split between ChaCha20 and AES-GCM is explained by the little-endian vs. big-endian nature of each; Do you agree that the difference comes down to endianness, or is it something different?

briansmith: > So, trying to figure out how that would work in an application, the problem I see is that that would make having a common API for both ChaCha20 and AES-CTR pointless,

I agree that it would be less than ideal, but I "pointless" is an overstatement.

> because the application would need to be aware of the specific algorithm being used to know how to generate different counters based on the different algorithms (i.e. for ChaCha it would use the first 4 bytes, while for AES-CTR it's actually the last 4 AFAICT).

With your original proposed API, the application would need to know these details too. It happens that for a QUIC implementation, the details just happen to work for what QUIC needs, but there would be no way for somebody other than us to know that given the API signature and the API documentation given in this PR to this point.

For QUIC, the details of the counter format matter less because QUIC only needs to encrypt one block (IIUC) so the counter never gets incremented during the operation. However, for non-QUIC applications that would use this API, the details of how the counter is constructed (big-endian for AES-GCM, little-endian for ChaCha20, IIUC) and how it is incremented (mod 2**32) and the amount of data that can be encrypted in a single operation (2**32 * block size, approximately) matter a lot.

Now, that said, I also eventually want to extend this API to work for non-CTR modes, in particular CBC mode. For those modes a 96-bit nonce and a 32-bit counter as separate arguments don't make sense. So I think it would be OK to keep the API in the same shape you presented it, especially if the difference in the counter/nonce split between ChaCha20 and AES-GCM is explained by the little-endian vs. big-endian nature of each; Do you agree that the difference comes down to endianness, or is it something different?

> The application would need to generate different key objects depending on the algorithm (so would need to expose key generation APIs as well for both ChaCha and AES), which I think is fine since this isn't intended for general use anyway, and it would discourage use unless really needed.

I don't think we need key generation APIs for QUIC since QUIC uses derived keys, right?

briansmith: > So, trying to figure out how that would work in an application, the problem I see is that that would make having a common API for both ChaCha20 and AES-CTR pointless,

I agree that it would be less than ideal, but I "pointless" is an overstatement.

> because the application would need to be aware of the specific algorithm being used to know how to generate different counters based on the different algorithms (i.e. for ChaCha it would use the first 4 bytes, while for AES-CTR it's actually the last 4 AFAICT).

With your original proposed API, the application would need to know these details too. It happens that for a QUIC implementation, the details just happen to work for what QUIC needs, but there would be no way for somebody other than us to know that given the API signature and the API documentation given in this PR to this point.

For QUIC, the details of the counter format matter less because QUIC only needs to encrypt one block (IIUC) so the counter never gets incremented during the operation. However, for non-QUIC applications that would use this API, the details of how the counter is constructed (big-endian for AES-GCM, little-endian for ChaCha20, IIUC) and how it is incremented (mod 2**32) and the amount of data that can be encrypted in a single operation (2**32 * block size, approximately) matter a lot.

Now, that said, I also eventually want to extend this API to work for non-CTR modes, in particular CBC mode. For those modes a 96-bit nonce and a 32-bit counter as separate arguments don't make sense. So I think it would be OK to keep the API in the same shape you presented it, especially if the difference in the counter/nonce split between ChaCha20 and AES-GCM is explained by the little-endian vs. big-endian nature of each; Do you agree that the difference comes down to endianness, or is it something different?

> The application would need to generate different key objects depending on the algorithm (so would need to expose key generation APIs as well for both ChaCha and AES), which I think is fine since this isn't intended for general use anyway, and it would discourage use unless really needed.

I don't think we need key generation APIs for QUIC since QUIC uses derived keys, right?

src/unauthenticated_stream/chacha20.rs
ghedo: Done.

src/unauthenticated_stream/chacha20.rs
ghedo: Done.

src/unauthenticated_stream/chacha20.rs
ghedo: Done.

src/unauthenticated_stream/chacha20.rs
ghedo: Done.

src/unauthenticated_stream/mod.rs
ghedo: Done.

src/unauthenticated_stream/mod.rs
ghedo: Done.

tests/unauthenticated_stream.rs
ghedo: Done (using a macro, not sure if there's a better way...).

tests/unauthenticated_stream.rs
ghedo: I tried to go the macro way like for the other test, but the problem here is that since the aead API also needs to pass tag and ad around, the code ends up being quite different from the unathenticated encryption case, so there's not much code reuse. But maybe there's a way to do this that I missed.

src/unauthenticated_stream/mod.rs
ghedo: Done (I did the "do_encryption_in_place" thing, so AES won't need changes).

src/unauthenticated_stream/chacha20.rs
ghedo: > Do you agree that the difference comes down to endianness, or is it something different?

Yesh, makes sense.

> I don't think we need key generation APIs for QUIC since QUIC uses derived keys, right?

The AES key still needs the whole initialization step AFAICT (that's what I meant with "key generation").

ghedo:
@briansmith I think I removed all the duplication as suggested (except for `test_encryption_nonce_sizes()` see https://github.com/briansmith/ring/pull/695#discussion_r227987699), can you please have another look?

ghedo:
@briansmith I now also rebased on top of master and fixed the conflicts (as well as updated to do "use crate::{...}").

djc:
> I have a question about how you're implementing QUIC: My understanding is that your implementation decrypts the first block using this unauthenticated API, and then opens the sealed record using the AEAD API. Now, do you always immediately open the sealed record after decrypting the first block? Or are there times where you decrypt the first block and then keep the ciphertext around encrypted for a while?

Current QUIC drafts use the stream ciphers to encrypt a variable-length encoded packet number in packet headers. Packet payloads are encrypted with AEAD algorithms, but they use a different key.

Packet Protection: https://tools.ietf.org/html/draft-ietf-quic-tls-16#section-5
Packet Number Protection: https://tools.ietf.org/html/draft-ietf-quic-tls-16#section-5.4

ghedo:
> The reason I ask is that I just realized that this QUIC is similar to the chacha20-poly1305@openssh.com construct that we implemented (and that you slightly modified in this PR). In particular I'm wondering if we should "just" make new AEADS that do the decypt+open/seal+encrypt atomically. WDYT?

I don't think it's that simple for QUIC. In addition to what @djc said (the packet number is variable length), the result of the packet number decryption (using unauthenticated stream cipher) is also used as part of the AEAD open operation in two ways: 1) the "raw" decrypted packet number in wire format is used as part of the AD (the AD is the full QUIC packet header after packet number decryption) and 2) the decoded packet number is mixed in the AEAD nonce.

So in order to do both packet number decryption/encryption and AEAD, ring would need to know an awful lot of information about how QUIC packet headers and packet numbers are encoded (https://tools.ietf.org/html/draft-ietf-quic-transport-16#section-17.1), which is not particularly simple https://tools.ietf.org/html/draft-ietf-quic-transport-16#appendix-A

src/unauthenticated_encryption/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+///
+/// C analog: `EVP_CIPHER_CTX`
+struct Key {
+    ctx_buf: [u64; KEY_CTX_BUF_ELEMS],
+    algorithm: &'static Algorithm,
+}
+
+/// TODO: once unauthenticated_encryption supports AES, we should move the
+/// definition of KEY_CTX_BUF_ELEMS here.
+const KEY_CTX_BUF_ELEMS: usize = aead::KEY_CTX_BUF_ELEMS;
```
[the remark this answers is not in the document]
ghedo: IIRC the problem I run into was that this is used in the argument of `ctx_as_key()` (to have an explicitly length) which is now shared between aead and unathenticated_encryption, so as far as I understand it, `KEY_CTX_BUF_ELEMS` needs to be the same for both modules (and hence the largest possible value that accomodates both), or we need two separate `ctx_as_key()`.

ghedo:
@briansmith incidentally the QUIC spec recently changed the way packet numbers are encrypted (it also now covers more of the header, so it's called "Header Protection" now), which will be pusblished in draft-17.
https://quicwg.org/base-drafts/draft-ietf-quic-tls.html#rfc.section.5.4

The way it works now, is that the encryption functions are only used to generate a "mask" which is then XOR'd by the application itself (rather than the crypto API) against the bytes that need to be protected. This is necessary because now _some_ of the bits of the first byte are encrypted, but not all of them (so you can't just use normal encryption and decryption functions on the actual header).

In the ChaCha20 case, this works by using ChaCha20() as before, but ontly to encrypt 5 "0" bytes (so, always lower than 16), and in the AES case, AES-ECB() (so no CTR anymore) is used to encrypt the 16 bytes that were previously used as nonce for AES-CTR.

So, in both cases we can force both the input and output to be 16 bytes (e.g. in the ChaCha case we would just encrypt more "0" bytes, and then discard the result in the application), and we don't need a decrypt operation anymore, so there might be a better name than "encrypt" for the function.

Since draft-17 hasn't been published yet I haven't had a chance to update my code to do that and test it against other implementations, so I'd say this PR is either blocked, or should be closed and then I will create a new one with the new approach.

Though this can be used for design discussions, e.g.:

1. What should the new module be called? "unauthenticated_encryption" still applies, but maybe something more specific to QUIC since ChaCha20+AES-ECB is a pretty weird pair.

2. What should the "encrypt" function be called? I'm fine with encrypt, but maybe something like "make_quic_mask()" or whatever.

3. Do we want to expose a generic API for both cases, or exposing the underlying ChaCha20() and AES-ECB() on their own would be enough?

briansmith:
> In the ChaCha20 case, this works by using ChaCha20() as before, but ontly to encrypt 5 "0" bytes (so, always lower than 16), and in the AES case, AES-ECB() (so no CTR anymore) is used to encrypt the 16 bytes that were previously used as nonce for AES-CTR.

I believe the AES case is the same as one block of AES-CTR with the sample as the IV, encrypting an input of zeros; i.e. the same as the ChaCha20 case.

I imagine then we can have an API something like this:

```rust
    // In src/aead/quic.rs
    trait HeaderProtectionKey {
        fn new(&[u8]) -> Result<Self, error::Unspecified>;
        fn new_mask(&self, sample: &[u8; 16]) -> [u8;  16];
    }
    struct AesHeaderProtectionKey { ... }
    impl HeaderProtectionKey for AesHeaderProtectionKey { ... }
    struct ChaCha20HeaderProtectionKey { ... }
    impl HeaderProtectionKey for ChaCha20HeaderProtectionKey { ... }
}
```

We should be able to implement this `HeaderProtectionKey` using the internal ChaCha20 and internal (upcoming) AES APIs very efficiently.

briansmith:
We now have a new internal API in *ring* that should make it much, much easier to implement this functionality. See:

* `aes::encrypt_block`: https://github.com/briansmith/ring/blob/c8b76df6e53063a6757f03a1bbf78af070b2c070/src/aead/aes.rs#L91
* `aes::encrypt_iv_xor_block`: https://github.com/briansmith/ring/blob/c8b76df6e53063a6757f03a1bbf78af070b2c070/src/aead/aes.rs#L129

I also refactored the chacha20 code to make it trivial to create an out-of-place `encrypt_iv_xor_block` operation; see https://github.com/briansmith/ring/blob/c8b76df6e53063a6757f03a1bbf78af070b2c070/src/aead/chacha.rs#L68.

ghedo:
Thanks @briansmith. Looking at your proposal again, I don't understand why we'd want to use traits for this? Shouldn't we use the `KeyInner` trick for this as well?

ghedo:
Replaced by https://github.com/briansmith/ring/pull/749.

Issue #113, opened 2016-02-18
AES_set_encrypt_key and variants return error codes, but they are ignored


briansmith:
Look at https://github.com/briansmith/ring/blob/80400f5fdfb3765acd490801c66ddf30d3dc6378/crypto/cipher/e_aes.c#L152 and other instances where those functions are called. The return value is ignored. It seems like the functions will never fail if given valid inputs, so they should have their return types changed to `void`. 

However, some variants aren't so clear as to whether they can fail on valid inputs or not. For example, `asm_AES_set_encrypt_key`/`_armv4_AES_set_encrypt_key` have code to jump to `.Labrt`. Is this just the "verify inputs are not null" check? If so, that code can be removed.

Note that in BoringSSL, but not OpenSSL, it is assumed that input parameters are not NULL unless the function's documentation specifically states they may be NULL, and NULL checks on parameters are not done.

ecstatic-morse:
I'd like to look into this.

The error conditions in the assembly are as follows.
- x86_64:
  - [null pointer checks](https://github.com/briansmith/ring/blob/9814ac47dff08424dd75e31881e7ba257880a096/crypto/fipsmodule/aes/asm/aes-x86_64.pl#L753)
  - [Ensure `bits` is 128 or 256](https://github.com/briansmith/ring/blob/9814ac47dff08424dd75e31881e7ba257880a096/crypto/fipsmodule/aes/asm/aes-x86_64.pl#L775)

PR #696, opened 2018-10-06
Use `#[repr(transparent)]` newtype to hold FFI error codes

### LINKED ISSUE #451: Mark C function return values as `#[must_use]`
One approach is to use (if/when it is implemented) the proposed `#[must_use]` annotation on functions: https://github.com/rust-lang/rfcs/pull/886.

Another approach is to change the return value to a `#[repr(transparent)]` wrapper that is marked `#[must_use]`, if/when this proposal is implemented: https://github.com/rust-lang/rfcs/pull/1758.

This is a good-first-bug since this is something anybody can do, once the facilities are available to actually do it, but note that it requires additional work on rustc.

ecstatic-morse:
Resolves #451.

This PR changes the definition of some foreign functions to return a `bssl::ErrorCode` instead of `c::int`. 

Converting to a `Result` is now done through a conversion function rather than `bssl::map_result`. It also defines marker traits for the two widely used return code conventions in this library:
  * 1 for success (used in most BoringSSL functions)
  * 0 for success

### DIFF
--- a/src/aead/aes_gcm.rs
+++ b/src/aead/aes_gcm.rs
@@ -44,7 +44,7 @@ pub static AES_256_GCM: aead::Algorithm = aead::Algorithm {
 
 fn aes_gcm_init(ctx_buf: &mut [u8], key: &[u8])
                 -> Result<(), error::Unspecified> {
-    bssl::map_result(unsafe {
+    Result::from(unsafe {
         GFp_aes_gcm_init(ctx_buf.as_mut_ptr(), ctx_buf.len(), key.as_ptr(),
                          key.len())
     })
@@ -55,7 +55,7 @@ fn aes_gcm_seal(ctx: &[u64; aead::KEY_CTX_BUF_ELEMS],
                 tag: &mut [u8; aead::TAG_LEN])
                 -> Result<(), error::Unspecified> {
     let ctx = polyfill::slice::u64_as_u8(ctx);
-    bssl::map_result(unsafe {
+    Result::from(unsafe {
         GFp_aes_gcm_seal(ctx.as_ptr(), in_out.as_mut_ptr(), in_out.len(), tag,
                          nonce, ad.as_ptr(), ad.len())
     })
@@ -66,7 +66,7 @@ fn aes_gcm_open(ctx: &[u64; aead::KEY_CTX_BUF_ELEMS],
                 in_out: &mut [u8], tag_out: &mut [u8; aead::TAG_LEN])
                 -> Result<(), error::Unspecified> {
     let ctx = polyfill::slice::u64_as_u8(ctx);
-    bssl::map_result(unsafe {
+    Result::from(unsafe {
         GFp_aes_gcm_open(ctx.as_ptr(), in_out.as_mut_ptr(),
                          in_out.len() - in_prefix_len, tag_out, nonce,
                          in_out[in_prefix_len..].as_ptr(), ad.as_ptr(),
@@ -99,19 +99,19 @@ const GCM128_SERIALIZED_LEN: usize = 16 * 16;
 
 extern {
     fn GFp_aes_gcm_init(ctx_buf: *mut u8, ctx_buf_len: c::size_t,
-                        key: *const u8, key_len: c::size_t) -> c::int;
+                        key: *const u8, key_len: c::size_t) -> bssl::Result;
 
     fn GFp_aes_gcm_seal(ctx_buf: *const u8, in_out: *mut u8,
                         in_out_len: c::size_t,
                         tag_out: &mut [u8; aead::TAG_LEN],
                         nonce: &[u8; aead::NONCE_LEN], ad: *const u8,
-                        ad_len: c::size_t) -> c::int;
+                        ad_len: c::size_t) -> bssl::Result;
 
     fn GFp_aes_gcm_open(ctx_buf: *const u8, out: *mut u8,
                         in_out_len: c::size_t,
                         tag_out: &mut [u8; aead::TAG_LEN],
                         nonce: &[u8; aead::NONCE_LEN], in_: *const u8,
-                        ad: *const u8, ad_len: c::size_t) -> c::int;
+                        ad: *const u8, ad_len: c::size_t) -> bssl::Result;
 }
 
 
@@ -172,6 +172,8 @@ mod tests {
     }
 
     extern "C" {
+        // XXX: This function does not adhere to the return value conventions.
+        #[must_use]
         fn GFp_AES_set_encrypt_key(key: *const u8, bits: usize,
                                    aes_key: *mut AES_KEY) -> c::int;
         fn GFp_AES_encrypt(in_: *const u8, out: *mut u8, key: *const AES_KEY);
--- a/src/bssl.rs
+++ b/src/bssl.rs
@@ -14,10 +14,22 @@
 
 use {c, error};
 
-pub fn map_result(bssl_result: c::int) -> Result<(), error::Unspecified> {
-    match bssl_result {
-        1 => Ok(()),
-        _ => Err(error::Unspecified),
+/// A `c::int` returned from a foreign function containing **1** if the function was successful
+/// or **0** if an error occurred. This is the convention used by C code in `ring`.
+#[derive(Clone, Copy, Debug)]
+#[must_use]
+#[repr(transparent)]
+pub struct Result(c::int);
+
+impl From<Result> for ::std::result::Result<(), error::Unspecified> {
+    fn from(ret: Result) -> Self {
+        match ret.0 {
+            1 => Ok(()),
+            c => {
+                debug_assert_eq!(c, 0, "`bssl::Result` value must be 0 or 1");
+                Err(error::Unspecified)
+            },
+        }
     }
 }
 
@@ -34,6 +46,7 @@ macro_rules! bssl_test {
         fn $fn_name() {
             use $crate::{c, init};
             extern {
+                #[must_use]
                 fn $bssl_test_main_fn_name() -> c::int;
             }
 
@@ -44,6 +57,27 @@ macro_rules! bssl_test {
                 $bssl_test_main_fn_name()
             };
             assert_eq!(result, 0);
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    mod result {
+        use {bssl, c};
+        use core::mem;
+
+        #[test]
+        fn size_and_alignment() {
+            type Underlying = c::int;
+            assert_eq!(mem::size_of::<bssl::Result>(), mem::size_of::<Underlying>());
+            assert_eq!(mem::align_of::<bssl::Result>(), mem::align_of::<Underlying>());
+        }
+
+        #[test]
+        fn semantics() {
+            assert!(Result::from(bssl::Result(0)).is_err());
+            assert!(Result::from(bssl::Result(1)).is_ok());
         }
     }
 }
--- a/STYLE.md
+++ b/STYLE.md
@@ -91,29 +91,33 @@ already a safer function for doing the conversion in
 [ring::polyfill](src/polyfill.rs). If not, add one to `ring::polyfill`.
 
 The C code generally uses the C `int` type as a return value, where 1 indicates
-success and 0 indicates failure. Sometimes the C code has functions that return
-pointers, and a NULL pointer indicates failure. The module
-[ring::bssl](src/bssl.rs) contains some utilities for mapping these return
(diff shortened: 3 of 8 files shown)


src/aead/aes_gcm.rs
```diff
@@ -44,34 +44,34 @@ pub static AES_256_GCM: aead::Algorithm = aead::Algorithm {
...
                 nonce: &[u8; aead::NONCE_LEN], ad: &[u8], in_out: &mut [u8],
                 tag: &mut [u8; aead::TAG_LEN])
                 -> Result<(), error::Unspecified> {
     let ctx = polyfill::slice::u64_as_u8(ctx);
-    bssl::map_result(unsafe {
+    unsafe {
         GFp_aes_gcm_seal(ctx.as_ptr(), in_out.as_mut_ptr(), in_out.len(), tag,
                          nonce, ad.as_ptr(), ad.len())
-    })
+    }.into()
```
briansmith: I don't find this as clear as the existing code but I guess people would learn it fast enough.

Would `Result::from` be a workable improvement?

src/bssl.rs
```diff
@@ -12,12 +12,62 @@
...
-        _ => Err(error::Unspecified),
+/// Many foreign functions return a `c::int` indicating whether an error occurred. This module
+/// wraps that error code in a `#[repr(transparent)]` newtype which can be converted into a
+/// `Result`.
+mod ffi_error {
+    use core::marker::PhantomData;
+    use {c, error};
+
+    /// The meaning of the error code returned by a foreign function.
+    pub trait Semantics {
```
briansmith: I don't see the advantage of factoring this out into a trait. One of the design principles of *ring* is to avoid unnecessary abstractions, and at least the way the code is currently, this abstraction isn't necessary.

src/bssl.rs
```diff
@@ -12,12 +12,62 @@
...
+    #[derive(Debug, Clone, Copy)]
+    pub enum BoringSsl {}
+
+    impl Semantics for BoringSsl {
+        fn is_success(code: c::int) -> bool {
+            match code {
+                1 => true,
+                c => {
+                    // Be pedantic in debug mode to catch some cases where function signatures are
+                    // wrongly defined.
```
briansmith: What are examples of such wrongly-defined functions?

src/bssl.rs
```diff
@@ -12,12 +12,62 @@
...
+        }
+    }
+
+    /// An error code returned by a foreign function.
+    ///
+    /// The template parameter indicates the semantics of the returned value.
+    #[must_use]
+    #[repr(transparent)]
+    #[derive(Debug, Clone, Copy)]
+    pub struct ErrorCode<S = BoringSsl> {
```
briansmith: I don't think `ErrorCode` is a very good name for this boolean flag. I think `Result` would make more sense since BoringSSL's boolean results are basically `Result<(), ()>`s.

src/bssl.rs
ecstatic-morse: By wrongly defined I meant that an author adding a new `extern` declaration might mistakenly use the wrong return type (e.g. `bssl::ErrorCode` for a function which returns a negative value on error).

src/bssl.rs
ecstatic-morse: Agreed. I avoided it initially since `modname::Result` is usually a type alias for `std::result::Result` with the error type defined.

src/bssl.rs
ecstatic-morse: Just two separate structs here then?

I had imagined that this would make it easy to share code later as features were added, but that's kind of the definition of overabstraction :)

src/bssl.rs
briansmith: In that case I would say "any cases where <whatever name we are using> might be misused."

src/bssl.rs
```diff
@@ -12,12 +12,62 @@
...
+
+    /// The semantics used by most BoringSSL functions:
+    ///
+    /// * `1` indicates success.
+    /// * `0` indicates failure.
+    #[derive(Debug, Clone, Copy)]
+    pub enum BoringSsl {}
+
+    impl Semantics for BoringSsl {
+        fn is_success(code: c::int) -> bool {
```
briansmith: I would say, rather than defining an `is_success() -> bool` function, just inline this logic directly into the `From` implementation that does the conversion to `Result<(), error::Unspecified>`.

src/bssl.rs
ecstatic-morse: Also, how do you feel about using `Posix` to mean returns `0` on failure. It was the best, shortest thing I could think of, but `bssl::PosixResult` reads a bit strange.

src/aead/aes_gcm.rs
```diff
@@ -172,8 +172,10 @@ mod tests {
     }
 
     extern "C" {
+        // AES_set_encrypt_key has a non-standard error code,.
+        // https://github.com/google/boringssl/blob/28babde159253bfa9003a445242605806fff5f1f/include/openssl/aes.h#L83
         fn GFp_AES_set_encrypt_key(key: *const u8, bits: usize,
-                                   aes_key: *mut AES_KEY) -> c::int;
+                                   aes_key: *mut AES_KEY) -> bssl::PosixStyleErrorCode;
```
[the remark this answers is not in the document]
ecstatic-morse: There are system APIs currently used by ring which return 0 on success, but to my knowledge this is the only currently used function from BoringSSL which has that convention.

ecstatic-morse:
There's also [a WIN32 function](https://github.com/briansmith/ring/blob/9814ac47dff08424dd75e31881e7ba257880a096/src/rand.rs#L153) which returns one on success, but uses a `c::win32::BOOLEAN` instead of a `c::int`.

Based on your feedback as well as this new information, I'd like to make the following sets of changes:
[ ] Define a new module called `c::result` to make clear that the newly added types should only be used for ffi.
[ ] Define transparent newtypes `c::result::BoringSsl`, `c::result::Posix` and `c::result::Win32Bool` to handle various commonly used return types and conventions.
[ ] Publicly alias either `c::Result` or `bssl::Result` to `c::result::BoringSsl`.
[ ] Add size/alignment tests for the newtypes via the `define_test_metrics` macro.

I could also eliminate the POSIX-style and Win32Bool wrappers, change `AES_set_encrypt_key` to return a BoringSSL style result ([or nothing at all](https://github.com/briansmith/ring/issues/113)), and add a `#[must_use]` declaration to the various system APIs which return an error code. Would you prefer this approach?

I was just perusing the list of good-first-bugss and saw this one. Sorry if this took more time for you to review than it would to have implemented it. Thanks for all the work you've done in the Rust community!

briansmith:
For that one single Win32 function, why not just change the FFI declaration of the function to use function-level `#[must_use]`?:

```rust
        #[link_name = "SystemFunction036"]
        #[must_use]
        fn RtlGenRandom(random_buffer: *mut u8,
                        random_buffer_length: c::win32::ULONG)
                        -> c::win32::BOOLEAN;
```

I suggest we skip the Posix case for now as I think it's trickier and also I hope we can eliminate the use of "POSIX" semantics from non-Posix functions like that AES API (in a separate PR).

For the GFp/BoringSSL API, maybe we should "just" use `#[must_use]` instead too? What are the pros/cons from your perspective?

> [ ]  Publicly alias either c::Result or bssl::Result to c::result::BoringSsl.

I suggest we just skip this, regardless.

>  Add size/alignment tests for the newtypes via the define_test_metrics macro.

Sounds good to me, if we need the newtypes.

ecstatic-morse:
Both approaches statically guarantee that error codes are being checked, so the issue is more about the conversion to `Result` currently done by `bssl::map_result`.

With the free function, the consumer must know the return code semantics at each call site, whereas with the newtype, the return code semantics must only be known at the declaration, and conversion can be done thoughtlessly via `Result::from`. Having to know the return code semantics at the call site isn't a large burden, especially if we normalize `GFp_AES_set_encrypt_key` to be consistent¸ but there will still be functions such as `GFp_memcmp` which return a `c::int` which doesn't indicate an error.

The presence or absence of `#[must_use]` shouldn't be used to indicate whether a function returns an error. Since `GFp_memcmp` could be marked `#[must_use]` as well so the compiler could warn when the result of a comparison is unused just as it does with `a == b`.

I understand the proliferation of return type wrappers being undesirable, especially given the efforts to oxidize `ring`, but I think it'd be good to provide one for the common case.

briansmith:
OK, I think in this PR we should go with your approach for the `GFp` API, the approach I suggested above for the single Win32 function, and punt Posix to later. That should allow us to make progress here.

ecstatic-morse:
Got it! I'll make the changes later today. Do you want the path to the type to be `bssl::Result`, `c::Result` or some spelling of c::result::BoringSsl`?

appveyor.yml
```diff
@@ -11,7 +11,7 @@ platform:
 environment:
   matrix:
     - TOOLCHAIN_VERSION: 14.0
-      RUST: 1.26.0
+      RUST: 1.28.0
```
briansmith: This should be done as a separate PR and we should be doing Rust 1.29.1.

src/aead/aes_gcm.rs
```diff
@@ -172,6 +172,9 @@ mod tests {
     }
 
     extern "C" {
+        // AES_set_encrypt_key returns 0 on success, unlike most BoringSSL functions.
+        // https://github.com/google/boringssl/blob/28babde159253bfa9003a445242605806fff5f1f/include/openssl/aes.h#L83
```
briansmith: Please remove the link to the BoringSSL code. If the comment being linked to is missing in *ring* then we should add the comment to *ring*.

src/bssl.rs
```diff
@@ -14,10 +14,25 @@
 
 use {c, error};
 
-pub fn map_result(bssl_result: c::int) -> Result<(), error::Unspecified> {
-    match bssl_result {
-        1 => Ok(()),
-        _ => Err(error::Unspecified),
+/// A `c::int` returned from a foreign function containing **1** if the function was successful
+/// or **0** if an error occurred. This is the convention used by BoringSSL.
```
briansmith: Let's remove the "This is the convention used by BoringSSL." I don't want to conflate *ring* with BoringSSL.

src/bssl.rs
```diff
@@ -14,10 +14,25 @@
...
+#[must_use]
+#[repr(transparent)]
+#[derive(Debug, Clone, Copy)]
```
briansmith: Let's put `#[derive(...)]` first unless there's some reason not to. I usually sort these alphabetically when I remember to.

src/bssl.rs
```diff
@@ -14,10 +14,25 @@
...
+pub struct Result(c::int);
+
+impl From<Result> for ::std::result::Result<(), error::Unspecified> {
+    fn from(ret: Result) -> Self {
+        match ret.0 {
+            1 => Ok(()),
+            c => {
+                // Be pedantic in debug mode to catch any cases where functions are mistakenly
+                // declared to return a BoringSSL-style error code when they really have
+                // different semantics.
```
briansmith: The comment is not necessary; let's just remove it.

src/bssl.rs
```diff
@@ -14,10 +14,25 @@
...
+                debug_assert_eq!(c, 0, "BoringSSL functions should return 0 or 1");
```
briansmith: I suggest "bssl::Result value must be be 0 or 1" instead.

src/bssl.rs
```diff
@@ -47,3 +63,24 @@ macro_rules! bssl_test {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    mod result {
+        use {bssl, c};
+        use std::mem;
```
briansmith: use `core::mem` instead of `std::mem`. In general we use `core` instead of `std` whenever possible.

appveyor.yml
ecstatic-morse: Is there a reason this isn't just "stable" instead of an explicit version number?

src/aead/aes_gcm.rs
ecstatic-morse: I added a comment to ring. I'm going to fix [this issue](https://github.com/briansmith/ring/issues/113) next, making it irrelevant.

briansmith:
@ecstatic-morse This looks great. Please squash this PR into a single commit that ends with the "I agree ..." statement. The commit message should be wrapped to ~72 characters.

I suggest this as the commit message (wrapped to ~72 characters):

```
Replace the `bssl::map_result` function with a new `bssl::Result` type.

I agree to ....
```

briansmith:
Thanks!

PR #697, opened 2018-10-07
chore: implement copy trait for Context

Brooooooklyn:


briansmith:
The `Context` types intentionally do not implement `Copy` because forking a context into two is relatively unusual operation and I want to make it implicit. Further, I would say that any type that contains a `Context` shouldn't be `Copy` either, for the same reason.

Instead of implementing `Copy` we should document why `Copy` isn't implemented.

I am, as always, open to being convinced otherwise.

briansmith:
I'm closing this as WONTFIX. Please comment if you think this is a mistake.

Issue #541, opened 2017-05-18
Implement `core::fmt::Debug` for all key types.


briansmith:
I suggest that the implementation should just output the same thing it would output if the type were defined to the algorithm field as its only field, and `#[derive(Debug)]` were used:

```rust
struct WhateverKey {
   alg: &'static WhateverAlgorithm;
}
```

In particular, we don't want the `Debug` implementation to include any part of the *value* of the key.

Ideally, we'd create a macro (e.g. in a new ring::debug submodule) to make it easy to add such a Debug implementation to new types, and use that macro for all these Algorithm types.

briansmith:
Considering this done "enough."

PR #698, opened 2018-10-08
impl core::fmt::Debug for hmac::SigningKey and hmac::SigningContext

andrewtj:
I don't have any particular preference for the output so I based it on #541.

### DIFF
--- a/src/debug.rs
+++ b/src/debug.rs
@@ -14,11 +14,11 @@
 
 // Generates an implementation of the Debug trait for a type that defers to the
 // Debug implementation for a given field.
-macro_rules! derive_debug_from_field {
-    ($typename:ty, $fieldname:ident) => {
+macro_rules! derive_debug_via_self {
+    ($typename:ty, self.$($tt:tt)+) => {
         impl ::core::fmt::Debug for $typename {
             fn fmt(&self, f: &mut ::core::fmt::Formatter) -> Result<(), ::core::fmt::Error> {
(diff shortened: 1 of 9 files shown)


briansmith:
Does `derive_debug_from_field!` work for this use? If so, let's use it. Otherwise, why not?

andrewtj:
Another option would be to modify `derive_debug_from_field!` to take `$($tt:tt)*` instead of an ident - would that be better?

briansmith:
> Another option would be to modify `derive_debug_from_field!` to take `$($tt:tt)*` instead of an ident - would that be better?

That does sound better to me in the abstract, assume we rename the `from_field` part.

andrewtj:
> That does sound better to me in the abstract, assume we rename the `from_field` part.

Okay. I've pushed 7fd1fa7 which renames the macro to `derive_debug_via_self!` and also makes the macro take `self.something` as the second param. If either change isn't too taste I'm happy to make further changes.

src/debug.rs
```diff
@@ -14,11 +14,11 @@
 
 // Generates an implementation of the Debug trait for a type that defers to the
 // Debug implementation for a given field.
-macro_rules! derive_debug_from_field {
-    ($typename:ty, $fieldname:ident) => {
+macro_rules! derive_debug_via_self {
+    ($typename:ty, self.$($tt:tt)+) => {
```
briansmith: Interesting, I didn't realize we could enforce that the expression starts with `self` like this.

PR #701, opened 2018-10-11
Use "stable" instead of an explicit version for rust on appveyor

ecstatic-morse:
*ring* only supports the latest version of each rust release channel.

I might be missing the rationale for the current behavior. Many CI templates also [use the channel name](https://github.com/japaric/trust/blob/bbeaa9b8282f5faf2a2e276d41e2960082edbd2b/appveyor.yml#L8).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,7 +11,7 @@ platform:
 environment:
   matrix:
     - TOOLCHAIN_VERSION: 14.0
-      RUST: 1.26.0
+      RUST: stable
     - TOOLCHAIN_VERSION: 14.0
       RUST: beta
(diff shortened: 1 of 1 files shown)


briansmith:
The "stable" build fails with this PR.

Those other projects are probably using rustup or something to download the Rust toolchain in Appveyor. *ring*'s Appveyor logic predates rustup. So if we want "stable" to work then we have more work to do. (I am fine using rustup in Appveyor.)

ecstatic-morse:
Will do.

ecstatic-morse:
I'm closing this so I don't run CI unnecessarily.

Issue #660, opened 2018-06-08
Implement KAT-based self-tests


briansmith:
With the large amount of platform-specific code, especially platform-specific assembly code, it's hard to test every implementation on every target. Further, in some deployment scenerios it is difficult to run the test suite on the target; the test suite might be run on a simulator/emulator or similar-but-not-the-same hardware.

In those instances, having NIST-/FIPS- style power-on tests for each algorithm would be helpful. We already have some one-time-init logic that is strategically triggered for each algorithm. We can change that to run an algorithm-specific KAT test after the one-time-init has completed.

PR #702, opened 2018-10-11
Make caller enforce invariants when setting AES encryption key

### LINKED ISSUE #113: AES_set_encrypt_key and variants return error codes, but they are ignored

ecstatic-morse:
Resolves #113.

`GFp_AES_set_encrypt_key` can fail if and only if the following invariants are violated.

* All pointer arguments must be non-null.
* The key size must be either 128 or 256 bits (192 bit keys [aren't supported](https://github.com/briansmith/ring/commit/b3e91be71edde28f5d2884d3c3c34260b6a79378)).

The first invariant should not be checked at runtime (except via assertion) as it is the responsibility of the caller.

The second could be expressed statically by accepting an enum with the valid key lengths, but enums don't provide any type safety in C. Instead, we document the acceptable key lengths and make them the responsibility of the caller.

As a result of these changes, `GFp_AES_set_encrypt_key` (as well as `GFp_aes_c_set_encrypt_key`) no longer needs to return an error code. The assembly functions still return an error code and do input validation.

### DIFF
--- a/crypto/fipsmodule/aes/aes.c
+++ b/crypto/fipsmodule/aes/aes.c
@@ -285,26 +285,16 @@ static const uint32_t rcon[] = {
     // for 128-bit blocks, Rijndael never uses more than 10 rcon values
 };
 
-int GFp_aes_c_set_encrypt_key(const uint8_t *key, unsigned bits,
-                              AES_KEY *aeskey) {
+// |bits| must be 128 or 256. 192-bit keys are not supported.
+void GFp_aes_c_set_encrypt_key(const uint8_t *key, unsigned bits,
+                               AES_KEY *aeskey) {
+  assert(key != NULL);
+  assert(aeskey != NULL);
+  assert(bits == 128 || bits == 256);
+
   uint32_t *rk;
   int i = 0;
   uint32_t temp;
-
-  if (!key || !aeskey) {
-    return -1;
-  }
-
-  switch (bits) {
-    case 128:
-      aeskey->rounds = 10;
-      break;
-    case 256:
-      aeskey->rounds = 14;
-      break;
-    default:
-      return -2;
-  }
 
   rk = aeskey->rd_key;
 
@@ -313,6 +303,8 @@ int GFp_aes_c_set_encrypt_key(const uint8_t *key, unsigned bits,
   rk[2] = from_be_u32_ptr(key + 8);
   rk[3] = from_be_u32_ptr(key + 12);
   if (bits == 128) {
+    aeskey->rounds = 10;
+
     while (1) {
       temp = rk[3];
       rk[4] = rk[0] ^ (Te2[(temp >> 16) & 0xff] & 0xff000000) ^
@@ -323,41 +315,41 @@ int GFp_aes_c_set_encrypt_key(const uint8_t *key, unsigned bits,
       rk[6] = rk[2] ^ rk[5];
       rk[7] = rk[3] ^ rk[6];
       if (++i == 10) {
-        return 0;
+        return;
       }
       rk += 4;
     }
   }
+
+  assert(bits == 256);
   rk[4] = from_be_u32_ptr(key + 16);
   rk[5] = from_be_u32_ptr(key + 20);
   rk[6] = from_be_u32_ptr(key + 24);
   rk[7] = from_be_u32_ptr(key + 28);
-  if (bits == 256) {
-    while (1) {
-      temp = rk[7];
-      rk[8] = rk[0] ^ (Te2[(temp >> 16) & 0xff] & 0xff000000) ^
-              (Te3[(temp >> 8) & 0xff] & 0x00ff0000) ^
-              (Te0[(temp) & 0xff] & 0x0000ff00) ^
-              (Te1[(temp >> 24)] & 0x000000ff) ^ rcon[i];
-      rk[9] = rk[1] ^ rk[8];
-      rk[10] = rk[2] ^ rk[9];
-      rk[11] = rk[3] ^ rk[10];
-      if (++i == 7) {
-        return 0;
-      }
-      temp = rk[11];
-      rk[12] = rk[4] ^ (Te2[(temp >> 24)] & 0xff000000) ^
-               (Te3[(temp >> 16) & 0xff] & 0x00ff0000) ^
-               (Te0[(temp >> 8) & 0xff] & 0x0000ff00) ^
-               (Te1[(temp) & 0xff] & 0x000000ff);
-      rk[13] = rk[5] ^ rk[12];
-      rk[14] = rk[6] ^ rk[13];
-      rk[15] = rk[7] ^ rk[14];
-
-      rk += 8;
+  aeskey->rounds = 14;
+  while (1) {
+    temp = rk[7];
+    rk[8] = rk[0] ^ (Te2[(temp >> 16) & 0xff] & 0xff000000) ^
+            (Te3[(temp >> 8) & 0xff] & 0x00ff0000) ^
+            (Te0[(temp) & 0xff] & 0x0000ff00) ^
+            (Te1[(temp >> 24)] & 0x000000ff) ^ rcon[i];
+    rk[9] = rk[1] ^ rk[8];
+    rk[10] = rk[2] ^ rk[9];
+    rk[11] = rk[3] ^ rk[10];
+    if (++i == 7) {
+      return;
     }
+    temp = rk[11];
+    rk[12] = rk[4] ^ (Te2[(temp >> 24)] & 0xff000000) ^
+             (Te3[(temp >> 16) & 0xff] & 0x00ff0000) ^
+             (Te0[(temp >> 8) & 0xff] & 0x0000ff00) ^
+             (Te1[(temp) & 0xff] & 0x000000ff);
+    rk[13] = rk[5] ^ rk[12];
+    rk[14] = rk[6] ^ rk[13];
+    rk[15] = rk[7] ^ rk[14];
+
+    rk += 8;
   }
-  return 0;
 }
 
 void GFp_aes_c_encrypt(const uint8_t *in, uint8_t *out, const AES_KEY *key) {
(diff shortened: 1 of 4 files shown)


crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -237,9 +244,12 @@ int GFp_aes_gcm_init(uint8_t *ctx_buf, size_t ctx_buf_len, const uint8_t *key,
     return 0;
   }
 
-  // XXX: Ignores return value. TODO: These functions should return |void|
-  // anyway.
-  GFp_AES_set_encrypt_key(key, (unsigned)key_len * 8, &ks);
+  unsigned bits = key_len * 8;
+  if (bits != 128 && bits != 256) {
```
briansmith: Instead, let's compare `key_len != 16 && key_len != 32` so we don't have to worry about any overflow issues with the multiplication.

crypto/fipsmodule/aes/aes.c
```diff
@@ -323,16 +315,18 @@ int GFp_aes_c_set_encrypt_key(const uint8_t *key, unsigned bits,
       rk[6] = rk[2] ^ rk[5];
       rk[7] = rk[3] ^ rk[6];
       if (++i == 10) {
-        return 0;
+        return;
       }
       rk += 4;
     }
   }
+
```
briansmith: set `aeskey->rounds = 14;` here.

crypto/fipsmodule/aes/aes.c
```diff
@@ -323,16 +315,18 @@ int GFp_aes_c_set_encrypt_key(const uint8_t *key, unsigned bits,
...
   rk[4] = from_be_u32_ptr(key + 16);
   rk[5] = from_be_u32_ptr(key + 20);
   rk[6] = from_be_u32_ptr(key + 24);
   rk[7] = from_be_u32_ptr(key + 28);
   if (bits == 256) {
```
briansmith: Remove the `if (bits == 256) {` here and do this unconditionally.

ecstatic-morse:
If/when this gets merged, the equivalent [error handling](https://github.com/briansmith/ring/blob/9814ac47dff08424dd75e31881e7ba257880a096/crypto/fipsmodule/aes/asm/aes-x86_64.pl#L753-L756) [code](https://github.com/briansmith/ring/blob/9814ac47dff08424dd75e31881e7ba257880a096/crypto/fipsmodule/aes/asm/aes-x86_64.pl#L775-L776) should be removed from the assembly implementations. 

Also, 192-bit key support was removed from the [vectorized assembly](https://github.com/briansmith/ring/commit/1103cf29dfbbf51f0dd8fb757084caa052863869) but not the assembly using [AES-NI](https://github.com/briansmith/ring/blob/9814ac47dff08424dd75e31881e7ba257880a096/crypto/fipsmodule/aes/asm/aesni-x86_64.pl#L1196).

I can open new issues for these if you like.

crypto/fipsmodule/aes/aes.c
```diff
@@ -323,41 +315,41 @@ int GFp_aes_c_set_encrypt_key(const uint8_t *key, unsigned bits,
...
       rk[7] = rk[3] ^ rk[6];
       if (++i == 10) {
-        return 0;
+        return;
       }
       rk += 4;
     }
   }
+
+  // bits == 256
```
briansmith: This should be `assert(bits == 256);`. That is, we prefer checked (at debug time at least) assertions to unchecked assertions.

src/aead/aes_gcm.rs
```diff
@@ -172,8 +174,8 @@ mod tests {
     }
 
     extern "C" {
-        fn GFp_AES_set_encrypt_key(key: *const u8, bits: usize,
```
briansmith: We should have done this change as a separate commit. But, OK.

briansmith:
@ecstatic-morse Are you still interested in this? It would be great if you could make the small changes I requested in the PR soon. I want to refactor the C code (or, if you want, you can do so, in another PR!) to remove the duplication between the 128-bit case and the 256-bit case, and to replace the weird loop constructions with straightforward `for` loops. I'd prefer to merge this PR before making those changes so that I don't create a merge conflict for you that you'd have to resolve manually.

briansmith:
@ecstatic-morse Are you still interested in this? It would be great if you could make the small changes I requested in the PR soon. I want to refactor the C code (or, if you want, you can do so, in another PR!) to remove the duplication between the 128-bit case and the 256-bit case, and to replace the weird loop constructions with straightforward `for` loops. I'd prefer to merge this PR before making those changes so that I don't create a merge conflict for you that you'd have to resolve manually.

briansmith:
@ecstatic-morse Are you still interested in this? It would be great if you could make the small changes I requested in the PR soon. I want to refactor the C code (or, if you want, you can do so, in another PR!) to remove the duplication between the 128-bit case and the 256-bit case, and to replace the weird loop constructions with straightforward `for` loops. I'd prefer to merge this PR before making those changes so that I don't create a merge conflict for you that you'd have to resolve manually.

briansmith:
@ecstatic-morse Are you still interested in this? It would be great if you could make the small changes I requested in the PR soon. I want to refactor the C code (or, if you want, you can do so, in another PR!) to remove the duplication between the 128-bit case and the 256-bit case, and to replace the weird loop constructions with straightforward `for` loops. I'd prefer to merge this PR before making those changes so that I don't create a merge conflict for you that you'd have to resolve manually.

briansmith:
@ecstatic-morse Are you still interested in this? It would be great if you could make the small changes I requested in the PR soon. I want to refactor the C code (or, if you want, you can do so, in another PR!) to remove the duplication between the 128-bit case and the 256-bit case, and to replace the weird loop constructions with straightforward `for` loops. I'd prefer to merge this PR before making those changes so that I don't create a merge conflict for you that you'd have to resolve manually.

briansmith:
@ecstatic-morse Are you still interested in this? It would be great if you could make the small changes I requested in the PR soon. I want to refactor the C code (or, if you want, you can do so, in another PR!) to remove the duplication between the 128-bit case and the 256-bit case, and to replace the weird loop constructions with straightforward `for` loops. I'd prefer to merge this PR before making those changes so that I don't create a merge conflict for you that you'd have to resolve manually.

briansmith:
@ecstatic-morse Are you still interested in this? It would be great if you could make the small changes I requested in the PR soon. I want to refactor the C code (or, if you want, you can do so, in another PR!) to remove the duplication between the 128-bit case and the 256-bit case, and to replace the weird loop constructions with straightforward `for` loops. I'd prefer to merge this PR before making those changes so that I don't create a merge conflict for you that you'd have to resolve manually.

briansmith:
@ecstatic-morse Are you still interested in this? It would be great if you could make the small changes I requested in the PR soon. I want to refactor the C code (or, if you want, you can do so, in another PR!) to remove the duplication between the 128-bit case and the 256-bit case, and to replace the weird loop constructions with straightforward `for` loops. I'd prefer to merge this PR before making those changes so that I don't create a merge conflict for you that you'd have to resolve manually.

ecstatic-morse:
Sorry, life got in the way last week. I made the changes and squashed.

My plan was to start working on #660 and #705 tonight. Refactoring AES round key generation would be a nice warm up.

On a more meta note, I'm trying to get accustomed to the programming style here before I tackle more involved issues. If you have any suggestions for next steps, I'd love to hear them.

ecstatic-morse:
I refactored the C AES key expansion routines [in this branch](https://github.com/ecstatic-morse/ring/blob/aes-key-expansion/crypto/fipsmodule/aes/aes.c). I'll submit a PR when this gets merged.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -132,15 +132,19 @@ int GFp_vpaes_set_encrypt_key(const uint8_t *userKey, unsigned bits,
...
   // Keep this in sync with |gcm128_init_gmult_ghash| and |aes_ctr|.
 
+  assert(user_key != NULL);
+  assert(key != NULL);
+  assert(bits == 128 || bits == 256);
+
 #if defined(HWAES)
   if (hwaes_capable()) {
-    return GFp_aes_hw_set_encrypt_key(user_key, bits, key);
+    GFp_aes_hw_set_encrypt_key(user_key, bits, key);
```
briansmith: Please prefix this with `(void)` to make it clear we're ignoring the return value intentionally.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -149,14 +153,15 @@ int GFp_AES_set_encrypt_key(const uint8_t *user_key, unsigned bits,
 #error "BSAES and VPAES are enabled at the same time, unexpectedly."
 #endif
   if (vpaes_capable()) {
-    return GFp_vpaes_set_encrypt_key(user_key, bits, key);
+    GFp_vpaes_set_encrypt_key(user_key, bits, key);
```
briansmith: Ditto: prefix with `(void)`.

briansmith:
Thanks for the quick turnaround @ecstatic-morse. I was just about to merge this but I noticed a couple of things: (1) We are missing a couple places where we should be using `(void)`, and (2) You didn't add "I agree to license..." to the end of your commit message. Please update the commit with those changes and I'll merge ASAP.

ecstatic-morse:
Whoops, sorry about that. I added the disclaimer to the commit message and made the requested changes.

Is the use `(void)` to explicitly ignore return values documented anywhere? I didn't see it in `STYLE.md`.

briansmith:
Thanks!

No, I guess it isn't mentioned in the style guide. One reason is that it is unusual for us to ignore return values in the first place. We're only doing so here temporarily until the types of those functions change to be `void`.

FWIW, I'd normally write `(void)x` instead of `(void) x` but NBD.

PR #708, opened 2018-10-24
Refactor C code for AES key expansion.

ecstatic-morse:
This begins the refactor discussed [here](https://github.com/briansmith/ring/pull/702#issuecomment-431735239). Hopefully this is close to what you were looking for.

The control flow of the original code was confusing. This moves the S-box computations into a separate function and simplifies the control flow. The while loop for 256-bit key expansion remains the same, as it facilitates loop unrolling.

### DIFF
--- a/crypto/fipsmodule/aes/aes.c
+++ b/crypto/fipsmodule/aes/aes.c
@@ -285,70 +285,90 @@ static const uint32_t rcon[] = {
     // for 128-bit blocks, Rijndael never uses more than 10 rcon values
 };
 
-// |bits| must be 128 or 256. 192-bit keys are not supported.
-void GFp_aes_c_set_encrypt_key(const uint8_t *key, unsigned bits,
-                               AES_KEY *aeskey) {
-  assert(key != NULL);
-  assert(aeskey != NULL);
-  assert(bits == 128 || bits == 256);
+// Applies the S-box to each byte in a 32-bit word. This is SubWord(x) in the
+// spec.
+//
+// [b0, b1, b2, b3] => [S(b0), S(b1), S(b2), S(b3)]
+static uint32_t rijndael_sbox(uint32_t x) {
+  return (Te2[(x >> 24)] & 0xff000000)
+    ^ (Te3[(x >> 16) & 0xff] & 0x00ff0000)
+    ^ (Te0[(x >> 8) & 0xff] & 0x0000ff00)
+    ^ (Te1[(x) & 0xff] & 0x000000ff);
+}
 
-  uint32_t *rk;
-  int i = 0;
-  uint32_t temp;
+// Applies the S-box to each byte in a 32-bit word, then rotates the output one
+// byte to the left. This is RotWord(SubWord(x)) in the spec.
+//
+// [b0, b1, b2, b3] => [S(b1), S(b2), S(b3), S(b0)]
+static uint32_t rijndael_sbox_then_rotate(uint32_t x) {
+  return (Te2[(x >> 16) & 0xff] & 0xff000000)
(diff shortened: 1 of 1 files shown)


crypto/fipsmodule/aes/aes.c
```diff
@@ -285,65 +285,55 @@ static const uint32_t rcon[] = {
     // for 128-bit blocks, Rijndael never uses more than 10 rcon values
 };
 
-// |bits| must be 128 or 256. 192-bit keys are not supported.
-void GFp_aes_c_set_encrypt_key(const uint8_t *key, unsigned bits,
-                               AES_KEY *aeskey) {
-  assert(key != NULL);
-  assert(aeskey != NULL);
-  assert(bits == 128 || bits == 256);
+// Applies the S-box to each byte in a 32-bit word. This is SubWord(x) in the
```
ecstatic-morse: Are these comments too wordy? They say the same thing 3 different ways.

crypto/fipsmodule/aes/aes.c
```diff
@@ -285,65 +285,55 @@ static const uint32_t rcon[] = {
...
+static uint32_t rijndael_sbox_then_rotate(uint32_t x) {
+  return (Te2[(x >> 16) & 0xff] & 0xff000000)
+    ^ (Te3[(x >> 8) & 0xff] & 0x00ff0000)
+    ^ (Te0[(x) & 0xff] & 0x0000ff00)
+    ^ (Te1[(x >> 24)] & 0x000000ff);
+}
+
+// Computes 10 additional 128-bit round keys from a 128-bit initial key
+// (11 round keys in total).
+static void expand_key_128(uint32_t rk[11 * 4]) {
```
ecstatic-morse: Is it good style to include array length in parameters?

crypto/fipsmodule/aes/aes.c
```diff
@@ -285,65 +285,55 @@ static const uint32_t rcon[] = {
...
+static void expand_key_256(uint32_t rk[15 * 4]) {
+  int i = 0;
 
-  assert(bits == 256);
-  rk[4] = from_be_u32_ptr(key + 16);
-  rk[5] = from_be_u32_ptr(key + 20);
-  rk[6] = from_be_u32_ptr(key + 24);
-  rk[7] = from_be_u32_ptr(key + 28);
-  aeskey->rounds = 14;
   while (1) {
```
ecstatic-morse: This while loop remains the same so that there is only one branch in its inner loop. An alternate way to go about this is the following:

```c
int i = 0;
goto LOOP_START;

for (; i < 8; ++i, rk += 8) {
    rk[12] = rk[4] ^ rijndael_sbox(rk[11]);
    rk[13] = rk[5] ^ rk[12];
    rk[14] = rk[6] ^ rk[13];
    rk[15] = rk[7] ^ rk[14];

LOOP_START:
    rk[8] = rk[0] ^ rijndael_sbox_then_rotate(rk[7]) ^ rcon[i];
    rk[9] = rk[1] ^ rk[8];
   rk[10] = rk[2] ^ rk[9];
   rk[11] = rk[3] ^ rk[10];
}
```

briansmith:
Thanks!

TBH, this was a bigger change then I was expecting to make. I really just wanted the "while" loops to be rewritten as "for" looks and have the duplicate code combined together. I don't know if I really want to make a bigger change within aes.c because aes.c has the bad license.

Moreover, since I had the idea to refactor this code, I found the paper that motivated issue #709. I think we should just close this issue for now with the expectation that #709 will get resolved in such a way that aes.c will be obliterated. WDYT?

ecstatic-morse:
Scary stuff.

Sounds like a good plan to me!

Issue #699, opened 2018-10-08
Construct RSAKeyPair from components


andrewtj:
DNSSEC has an informal `key: value` text format for storing private keys which for RSA material looks like the following (from [RFC 5702](https://tools.ietf.org/html/rfc5702#section-6.1)):

```text
Private-key-format: v1.2
Algorithm:       8 (RSASHA256)
Modulus:         wVwaxrHF2CK64aYKRUibLiH30KpPuPBjel7E8ZydQW1HYWHfoGmidzC2RnhwCC293hCzw+TFR2nqn8OVSY5t2Q==
PublicExponent:  AQAB
PrivateExponent: UR44xX6zB3eaeyvTRzmskHADrPCmPWnr8dxsNwiDGHzrMKLN+i/HAam+97HxIKVWNDH2ba9Mf1SA8xu9dcHZAQ==
Prime1:          4c8IvFu1AVXGWeFLLFh5vs7fbdzdC6U82fduE6KkSWk=
Prime2:          2zZpBE8ZXVnL74QjG4zINlDfH+EOEtjJJ3RtaYDugvE=
Exponent1:       G2xAPFfK0KGxGANDVNxd1K1c9wOmmJ51mGbzKFFNMFk=
Exponent2:       GYxP1Pa7CAwtHm8SAGX594qZVofOMhgd6YFCNyeVpKE=
Coefficient:     icQdNRjlZGPmuJm2TIadubcO8X7V4y07aVhX464tx8Q=
```

I have been re-encoding these to use them with *ring* but I'd rather construct them more directly if possible. Ideally I'd like to fill in a struct provided by *ring* with the components and then pass that to a constructor on `RSAKeyPair`. Does this sound like something that'd fit in?

briansmith:
Sure, although I've rejected doing this in the past, I am more open to doing it now.

I suggest we have a public `RSAPrivateKeyBytes` type, or similar, that looks something like this:

```rust
pub struct RSAPrivateKeyInput<'a> {
    pub n: untrusted::Input<'a>,
    pub e: untrusted::Input<'a>,
    ...
}
```

The user would fill in *all* the fields of the `RSAPrivateKeyInput` using regular record construction and then we'd implement `RSAKeyPair::from(RSAPrivateKeyInput)`. Then we'd rewrite the existing ASN.1 parsing code to be a wrapper around this new API. Then the existing tests would need to be adapted to test both the existing API and the new API. Then the new API could be exposed publicly.

[...]

PR #714, opened 2018-10-30
Implement RSAKeyPair from components

andrewtj:
This continues from #699.

There's a pretty clean break between the DER parsing logic and validation of the values so I don't think there's a need for any new tests.

I'm not sure the camel case fields of `KeyPairInput` are good style but I went with that since `KeyPair` has camel case fields. (Aside: the "Rust Guidelines" link in `STYLE.MD` no longer works.)


src/rsa/signing.rs
```diff
@@ -176,181 +201,194 @@ impl KeyPair {
                 let dP = der::positive_integer(input)?;
                 let dQ = der::positive_integer(input)?;
                 let qInv = der::positive_integer(input)?;
```
briansmith: Above, we shouldn't use `der::positive_integer` for the parsing, because the checking for a positive number should be done in `from_key_input()`.

src/rsa/signing.rs
```diff
@@ -176,181 +201,194 @@ impl KeyPair {
...
-                // Second, stop if `p > 2**(nBits/2) - 1`.
-                let half_n_bits = public_key.n_bits.half_rounded_up();
-                if p_bits != half_n_bits {
-                    return Err(error::Unspecified);
-                }
+    /// Constructs an RSA private key from components.
+    ///
+    /// Components are in big-endian, considered unsigned, and must not contain
+    /// leading zeros. Input aside this is just like `from_pkcs8()`. See the
+    /// documentation for `from_pkcs8()` for more details.
```
briansmith: Let's not repeat this part of the documentation about the form of each value in `KeyPairInput` here.

src/rsa/signing.rs
```diff
@@ -21,6 +21,31 @@ use super::{bigint, bigint::Prime, N, verification};
...
 
+/// Used to construct an RSA key pair from components. Feature: `rsa_signing`.
+///
+/// Components are in big-endian, considered unsigned, and must not contain
+/// leading zeros. To construct a `KeyPair` provide this struct to
+/// `RSAKeyPair::from_key_input()`.
+#[derive(Clone, Debug)]
+pub struct KeyPairInput<'a> {
+    /// Public modulus.
+    pub n: untrusted::Input<'a>,
```
briansmith: It seems like it would be better to instead use a type `der::PositiveInteger` that wraps `untrusted::Input`. Then `der::positive_integer` can return a `PositiveInteger` and each of these fields can be a `der::PositiveInteger`. Then we can document the endianness and lack of leading zeros in the documentation for `der::PositiveInteger` instead of above.

andrewtj: A new type would make sense there. Changing the return type of `der::positive_integer()` will have a knock-on effect. Before I go ahead, I just want to confirm that I should change all the associated `untrusted::Input`s to `der::PositiveInteger`s in things like [`ecdsa/verification.rs`](https://github.com/briansmith/ring/blob/b5153f0e388e9b35b7157b3efa93b8091d1876c7/src/ec/suite_b/ecdsa/verification.rs#L27-L30)? (I don't mind chasing the types, I just want to make sure I'm headed in the right direction in case this ends up being a bit invasive.)

briansmith: Yes. In fact, it would be great to have the change of the return type of `der::positive_integer` in a separate PR, that this PR would then be rebased on top of.

briansmith:
@andrewtj Any update on this? I want to run rustfmt on *ring* soon and it would disrupt your PR if you're still working on this.

andrewtj:
Go for it. I took a swing at adding the wrappers last week and ran out of time. I'll come back to it later this week. rustfmt won't be a hindrance.

andrewtj:
I'm not going to get back to this after all. I'll leave #699 open and leave a comment pointing to the `der::PositiveInteger` discussion.

PR #720, opened 2018-11-13
Fix `--no-default-features` build.

briansmith:
Disable the generic signing API when `use_heap` isn't enabled.


briansmith:
@jonas-schievink Please review this. This should fix the `--no-default-features` build.

PR #727, opened 2018-11-17
impl Clone for keys

Ralith:
I'm not sure if the omission was deliberate, but this makes the structs in question more convenient to work with in some cases. Not sure I got everything.


briansmith:
I don't know if `Clone` for keys is appropriate. I think `TryClone` would make more sense.

> the structs in question more convenient to work with in some cases

In what cases is it useful to clone instead of move a key, in your experience? Especially I am curious about the motivation for wanting to clone symmetric keys.

briansmith:
> Under what circumstances might a clone fail?

One example: In the future I hope to implement some integrations with native OS keystores and in those situations a key might fail to clone.

> While mostly I've been able to refactor pre-existing Clone requirements away when I encounter this issue, it's unclear why that should be necessary, and doing so sometimes becomes awkward. For example, I have a unit test that simulates a server restarting with persistent SigningKeys, and Clone would allow me to reconstruct the server using the same configuration struct, whereas currently I'm forced to store an intermediate buffer of random bytes and create a new SigningKey for each server instance.

Lots of aspects of the API could be changed to make writing tests easier, at the cost of making it easier to introduce bugs in non-test code. In general I don't add features that are bad for production code just to improve testing (though I won't rule it out completely).

> I suppose could move everything behind an Arc. It seems a little strange for that to be the recommended practice for what I expect to usually be small, trivially copyable types, though.

They are all just trivially-copyable bytes now, but I don't want to guarantee that will always be the case.

PR #731, opened 2018-12-05
[WIP] Rsa keygen

### LINKED ISSUE #219: Add RSA key generation
This should build on #208. We should consider just fixing the public exponent at 65537, since that is safe and fast (for verifiers). We should also consider limiting the public modulus size to 2048-8192 bits.

Instead of making this so that it returns an `RSAKeyPair` and adding a `private_key_bytes()` method for serializing it, I suggest that we just make a function that returns the bytes of an already-serialized RSA private key.

There is code implementing the math in the function `RSA_generate` in `crypto/rsa/rsa_impl.c`. I think NIST may have test vectors that we can use for verifying the implementation.

est31:
Opened the PR so that I can discuss with @briansmith about the code. Nothing works yet. DON'T MERGE!

List of tasks:

- [ ] Implement primality test (Rust port of C's `BN_primality_test` function)
- [ ] Implement FIPS 186-4 section B 3.3 algorithm to obtain `(p,q)` prime pair
- [ ] Use `(p,q)` to create a `ring::signature::RSAKeyPair` instance
- [ ] PKCS8 formatting of the created key

Fixes #219


est31:
@briansmith I've implemented some of the first steps of the primality testing algorithm. How does it look so far?

As for the further steps, I don't know how to do the subtraction in Rust. There is subtraction support in the bignum module but it implements arithmetic modulo some number. This code however doesn't do that, it just uses vanilla bignums. There are functions `limb_sub`, etc in `crypto/limbs/limbs.inl`. Can I expose them to Rust? How should I do that?

crypto/fipsmodule/bn/div_extra.c
```diff
@@ -59,7 +57,64 @@ static uint16_t shift_and_add_mod_u16(uint16_t r, uint32_t a, uint16_t d,
   return t;
 }
 
-uint16_t bn_mod_u16_consttime(const BIGNUM *bn, uint16_t d) {
+
+// Function taken from boringssl's bn.c file.
```
briansmith: Any code that comes from a file X in BoringSSL must be in a file named X in *ring*. Also, the license must be the same. The best way to add functions from bn.c would be to add one commit that is the latest unmodified version from BoringSSL, and then another commit that deletes everything except the license and the code you need.

Note that the licenses are different. That is actually why `div.c` and `div_extra.c` are separate files; the `_extra` files are ISC-licensed, the other files are "OpenSSL licensed," i.e. the bad license everybody is trying to get away from.

src/lib.rs
```diff
@@ -150,3 +150,6 @@ mod private {
 mod tests {
     bssl_test!(test_constant_time, bssl_constant_time_test_main);
 }
+
+// Only here for development
+pub use rsa::key_generation::probable_primality_test;
```
briansmith: It seems like you maybe forgot to `git add` this?

briansmith:
> * [ ]  Implement primality test (Rust port of C's `BN_primality_test` function)

We cannot "port" code from BoringSSL unless it is ISC-licensed like div-extra. In particular, `BN_primality_test` has the wrong license and we can't use it even as a model. Instead, look at the algorithm from Wikipedia or a textbook.

Unfortunately, because of the licensing issues, I need to close this PR. Please sort out the license-related stuff (email me if it is confusing: brian@briansmith.org) and submit a new PR. I appreciate your effort here but I need to make sure the licenses are respected and that the *ring* Rust code is under the ISC license.

est31:
Hmmm will make a new attempt tomorrow then

est31:
@briansmith okay I've removed the parts that I've taken from the C function `BN_primality_test` and re-did them myself. I've opened a new PR.

PR #733, opened 2018-12-06
[WIP]: Rsa keygen

### LINKED ISSUE #219: Add RSA key generation

est31:
Opened the PR so that I can discuss with @briansmith about the code. Nothing works yet. DON'T MERGE!

List of tasks:

- [x] Implement division by small primes as suggested [here](https://github.com/briansmith/ring/issues/219#issuecomment-443644184)
- [x] Implement miller-rabin primality test based on FIPS 186-4 C.3.1
- [x] Tests for the miller-rabin primality test
- [x] Implement FIPS 186-4 section B 3.3 algorithm to obtain `(p,q)` prime pair
- [x] Use `(p,q)` to create the needed `p,q,dP,dQ,qInv,r_i,d_i,t_i`
- [x] DER formatting of the created key
- [ ] Implement primitive for multiplication
- [x] Implement primitive for inverse modulo m

Important reading:

* FIPS 186-4: https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.186-4.pdf
* https://eprint.iacr.org/2018/749
* https://tools.ietf.org/html/draft-mavrogiannopoulos-pkcs8-validated-parameters-00
* Formatting of the keys: https://tools.ietf.org/html/rfc5958 and https://tools.ietf.org/html/rfc3447#section-3

Fixes #219


src/rsa/key_generation.rs
```diff
@@ -0,0 +1,148 @@
...
+        let mut z_elem = bigint::elem_exp_consttime(b_elem, &m_exponent, &w_modulus)?;
+        // Step 4.4
+        if z_elem.is_one() || limb::limbs_equal_limbs_consttime(&z_elem.limbs, &w_m1) == LimbMask::True {
+            continue;
+        }
+        // Step 4.5
+        for _ in 1 .. a {
+            // Step 4.5.1
+            // TODO figure out how to make this compile
+            //z_elem = bigint::elem_squared(z_elem, &w_modulus.as_partial());
```
est31: @briansmith do you know how this can be made compile? Right now I'm getting sth like:

```error[E0308]: mismatched types                                                                                         
  --> src/rsa/key_generation.rs:87:22                                                                                  
   |                                                                                                                   
87 |             z_elem = bigint::elem_squared(z_elem, &w_modulus.as_partial());                                       
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `arithmetic::montgomery::Unencoded`, found enum `arithmetic::montgomery::RInverse`
   |                                                                                                                   
   = note: expected type `rsa::bigint::Elem<_, arithmetic::montgomery::Unencoded>`                                     
              found type `rsa::bigint::Elem<_, arithmetic::montgomery::RInverse>` 

```

crypto/limbs/limbs.c
```diff
@@ -99,6 +99,11 @@ Limb LIMBS_less_than_limb(const Limb a[], Limb b, size_t num_limbs) {
   return constant_time_select_w(lo, hi, lo);
 }
 
+void LIMBS_sub_limb(Limb r[], const Limb a[], const Limb b,
+               size_t num_limbs) {
+  limbs_sub_limb(r, a, b, num_limbs);
```
briansmith: Replace this with

```c
void LIMBS_odd_sub_one(Limb r[], size_t num_limbs) {
    assert(num_limbs > 0);
    assert(!LIMBS_are_even(r, num_limbs));
    r[0] &= ~1;
}
```

src/limb.rs
```diff
@@ -69,6 +69,42 @@ pub fn limbs_less_than_limbs_consttime(a: &[Limb], b: &[Limb]) -> LimbMask {
...
+pub fn limbs_count_low_zero_bits(v: &[Limb]) -> u16 {
+    extern "C" {
+        fn BN_count_low_zero_bits(bn: *const Limb, num_limbs: c::size_t) -> c::int;
+    }
+    let r = unsafe { BN_count_low_zero_bits(v.as_ptr(), v.len()) };
+    r as u16
+}
+
+#[cfg(feature = "rsa_keygen")]
+pub fn limbs_copy(r: &mut [Limb], a: &[Limb]) {
```
briansmith: You can do the copying in pure Rust code. (In *ring*, we can copy secret values from one place to another in Rust, we just don't do any arithmetic in Rust get.)

src/limb.rs
```diff
@@ -69,6 +69,42 @@ pub fn limbs_less_than_limbs_consttime(a: &[Limb], b: &[Limb]) -> LimbMask {
...
+    extern "C" {
+        fn LIMBS_copy(r: *mut Limb, a: *const Limb, num_limbs: c::size_t);
+    }
+    unsafe { LIMBS_copy(r.as_mut_ptr(), a.as_ptr(), a.len()) }
+}
+
+
+#[cfg(feature = "rsa_keygen")]
+pub fn limbs_rshift(r: &mut [Limb], n: u16) {
```
briansmith: This should have `secret_shift` in the name.

src/rsa/bigint.rs
```diff
@@ -67,7 +67,7 @@ struct Width<M> {
 }
 
 /// All `BoxedLimbs<M>` are stored in the same number of limbs.
-struct BoxedLimbs<M> {
+pub struct BoxedLimbs<M> {
```
briansmith: Let's make the primality test code a submodule of `bigint` so that we don't need to increase the visibility of things in `bigint`.

src/rsa/key_generation.rs
```diff
@@ -0,0 +1,147 @@
+// Copyright 2018 Brian Smith.
```
briansmith: If you copied some of my code, please keep this copyright line. Please add your own copyright line.

src/rsa/key_generation.rs
```diff
@@ -0,0 +1,147 @@
...
+    error, rand,
+    rsa::bigint,
+};
+
+/// Performs the Miller-Rabin primality test
+///
+/// Implementation of the algorithm as it
+/// is described in FIPS 186-4 C.3.1
+fn miller_rabin_test(
+    w: &[u64],
```
briansmith: You have written this whole thing in terms of `Limb` but it would be a lot clearer to write it in terms of `bigint::Nonnegative`, I think.

src/rsa/key_generation.rs
```diff
@@ -0,0 +1,147 @@
...
+    iterations: u32,
+    rng: &rand::SecureRandom,
+) -> Result<bool, error::Unspecified> {
+    // Step 1.
+    let mut w_m1 = vec![0; w.len()];
+    limb::limbs_sub_limb(&mut w_m1, w, 1);
+    let a = limb::limbs_count_low_zero_bits(&w_m1);
```
briansmith:
```rust
// for w: &Nonnegative
let w_minus_1 = w.odd_minus_1()?;
let a = w.trailing_zeros();
```

src/rsa/key_generation.rs
```diff
@@ -0,0 +1,147 @@
...
+    let mut m = vec![0; w_m1.len()];
+    limb::limbs_copy(&mut m, &w_m1);
+    limb::limbs_rshift(&mut m, a);
+
+    // Step 3 not needed.
+
+    // preparation
+    let boxed_w = bigint::BoxedLimbs::<()>::minimal_width_from_unpadded(&w);
+    let w_modulus = bigint::Modulus::from_boxed_limbs(boxed_w)?.0;
+    let boxed_m = bigint::BoxedLimbs::<()>::minimal_width_from_unpadded(&m);
```
briansmith: see `Modulus::from_nonnegative_with_bit_length()`.

src/rsa/key_generation.rs
```diff
@@ -0,0 +1,147 @@
...
+    };
+    // Step 4.
+    let mut b: Vec<limb::Limb> = vec![0; w.len()];
+    for _ in 0 .. iterations {
+        loop {
+            // Step 4.1
+            unsafe {
+                let b_mut_ptr = (&mut b[..]).as_mut_ptr() as * mut u8;
+                let b_mut_slice = core::slice::from_raw_parts_mut(b_mut_ptr, b.len() * 8);
+                rng.fill(b_mut_slice)?;
```
briansmith: This should probably be factored out into a `impl Elem { fn  random(rng: &SecureRandom) -> Result<Self, error::Unspecified> }` method.

src/rsa/key_generation.rs
briansmith: For now: multiply `z_elem` by `w_modulus.oneRR` before the loop.

Later: Refactor `elem_exp_consttime` so that it returns an `Elem<R>` instead of `Elem<Unencoded>` by moving the `into_unencoded()` to the caller.

src/rsa/key_generation.rs
```diff
@@ -0,0 +1,147 @@
...
+        if z_elem.is_one() || limb::limbs_equal_limbs_consttime(&z_elem.limbs, &w_m1) == LimbMask::True {
+            continue;
+        }
+        // Step 4.5
+        for _ in 1 .. a {
+            // Step 4.5.1
+            // TODO figure out how to make this compile
+            //z_elem = bigint::elem_squared(z_elem, &w_modulus.as_partial());
+            // Step 4.5.2
+            if limb::limbs_equal_limbs_consttime(&z_elem.limbs, &w_m1) == LimbMask::True {
```
briansmith: You will need to convert `w_m1` to an `Elem<W>` and then use the `Elem` comparison functions.

src/rsa/key_generation.rs
```diff
@@ -0,0 +1,147 @@
...
+                continue;
+            }
+            // Step 4.5.1
+            if z_elem.is_one() {
```
briansmith: You will need to create a value `oneR: Elem<W, R>` with value `1 * w_modulus.oneRR` and then compare using the element comparison functions.

src/rsa/key_generation.rs
```diff
@@ -0,0 +1,147 @@
...
+    limb::{self, LimbMask},
+    error, rand,
+    rsa::bigint,
+};
+
+/// Performs the Miller-Rabin primality test
+///
+/// Implementation of the algorithm as it
+/// is described in FIPS 186-4 C.3.1
+fn miller_rabin_test(
```
briansmith: I think this function's type should be something like like `fn miller_rabin_test<M: Prime>(w: Nonnegative, iterations: u32, rng: &SecureRandom) -> Result<Modulus<M>, error::Unspecified>`.

briansmith:
OK, I can see now that this is going to be a lot of work from both of us. Because we do any more work, let's clarify the IPR part:

1. Who is the copyright holder for new the code that contributing? Please add the Copyright lines that says who the copyright holder is, to the tops of the files with new code before we continue.
2. Do you have the permission of the copyright holder to contribute this code?
3. Please add the contributor statement at https://github.com/briansmith/ring#contributing to the end of every commit message (you can skip the ones that only add unmodified BoringSSL code).

briansmith:
In order for me to effectively review this code, we need to refactor it into at least two parts: low-level routines and high-level routines. The M-R test itself should  be a high-level routine that takes a `Nonnegative`, does some operations on `Nonnegative`s, `Elem`s, and `Modulus`s, and returns a `Modulus`. There probably should not be any mention of `Limb` or `u64` in the M-R code at all.

Anything that deals with `Limb` or `&[Limb]` or ``BoxedLimbs` needs to be encapsulated in a new extension of the `bigint` API. For example, subtracting 1 from an odd number `m` to produce an even `m - 1` needs to be a method of `Nonnegative`. Counting the trailing zeros of a `Nonnegative` needs to be a method of `Nonnegative`. When possible, use the standard Rust names for these functions, e.g. `trailing_zeros()`.

src/rsa/key_generation.rs
est31: What should the returned modulus contain? Will we need that value somewhere? To me it seems that only bool yes/no output is needed.

briansmith: > What should the returned modulus contain? Will we need that value somewhere? To me it seems that only bool yes/no output is needed.

At least, for example, we need to calculate `q**-1 (mod p)` to construct the RSA key. IIUC, this function already constructs the `Modulus` for the prime so we might as well return it. OTOH, if the function doesn't end up constructing the `Modulus` already then we can just return a `()`.

src/rsa/key_generation.rs
est31: When I put this before the loop,
```Rust
        let z_elem_u = bigint::elem_exp_consttime(b_elem, &m_exponent, &w_modulus)? ;
        let mut z_elem = bigint::elem_mul(z_elem_u, w_modulus.oneRR, w_modulus);
```
and I do
```Rust
            z_elem = bigint::elem_squared(z_elem, &w_modulus.as_partial());
```
it gives me:
```
error[E0308]: mismatched types                                                                                         
  --> src/rsa/bigint/key_generation.rs:82:22                                                                           
   |                                                                                                                   
82 |             z_elem = bigint::elem_squared(z_elem, &w_modulus.as_partial());                                       
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `arithmetic::montgomery::Unencoded`, found enum `arithmetic::montgomery::RInverse`
   |                                                                                                                   
   = note: expected type `rsa::bigint::Elem<_, arithmetic::montgomery::Unencoded>`                                     
              found type `rsa::bigint::Elem<_, arithmetic::montgomery::RInverse>`
```

if instead I do
```Rust
z_elem = bigint::elem_mul(&z_elem, z_elem.clone(), &w_modulus);
```

I get a type mismatch as well.

```
error[E0308]: mismatched types                                                                                         
  --> src/rsa/bigint/key_generation.rs:83:22                                                                           
   |                                                                                                                   
83 |             z_elem = bigint::elem_mul(&z_elem, z_elem.clone(), &w_modulus);                                       
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `arithmetic::montgomery::Unencoded`, found enum `arithmetic::montgomery::RInverse`
   |                                                                                                                   
   = note: expected type `rsa::bigint::Elem<_, arithmetic::montgomery::Unencoded>`                                     
              found type `rsa::bigint::Elem<_, arithmetic::montgomery::RInverse>` 
```

est31:
Rebased. ATM blocked on getting advice: https://github.com/briansmith/ring/issues/219#issuecomment-445479144

est31:
@briansmith do you know how to do the PKCS8 formatting part?

briansmith:
> @briansmith do you know how to do the PKCS8 formatting part?

ECDSA does it like this: https://github.com/briansmith/ring/blob/aa562a17dae92022497dedc19c18f0174585c4f1/src/ec/suite_b/ecdsa/signing.rs#L89-L100.

The trouble is that the current PKCS#8 serialization stuff is a little too simple for RSA because it assumes the private key and public key are small, fixed-length fields. That's not the case for RSA. Basically, you'll need to implement the inverse of `der::read_tag_and_get_value` and then use it a bunch of times to encode all the fields. I recommend first getting it working to serialize a `RSAPrivateKey`, then do the PKCS#8 part in another step.

est31:
@briansmith since you made commit 67425b7463edb3b90f45c4307188876424a71295, should I also remove the RSA key generation feature?

briansmith:
> @briansmith since you made commit [67425b7](https://github.com/briansmith/ring/commit/67425b7463edb3b90f45c4307188876424a71295), should I also remove the RSA key generation feature?

Yes, that's exactly why I did it.

est31:
Cool! Will do.

briansmith:
@est31 Please check out the new `ring::io::der_writer` module. In particular, look at how it is used by the RSA `KeyPair` to serialize the public key. The private key should be pretty much exactly the same, except you'll probably need to add support for serializing the integer value 0 for the version field.

Are you waiting on me for anything?

est31:
Hmmm, having a function return io::Positive from a Nonnegative already doesn't seem to work for lifetime reasons (the limbs can't be used as basis, thus the function has to do an allocation), but the basic question still remains.

est31:
Nevermind, I'll just use `bn_div_consttime` and `bn_mul_consttime` from boringssl.

briansmith:
> @briansmith so I was wondering whether I could use the `write_positive_integer` function for key pair serialisation. In general, it seems possible. I'd do a function that converts a rsa::bigint::Nonnegative to a io::Positive value.

Yes, and the building block is already available in `limb::big_endian_from_limbs`.

> But there is one problem: I need to be able to write zero for the version field

For the version field, use this (add it in a separate commit):
```rust
pub fn write_version_0_tlv(output: &mut Accumulator) {
    write_tlv(output, Tag::Integer, &|output| {});
}
```

> What should I do? Rename the `Positive` type?

I want to keep `Positive` as meaning "positive number" because it has a very useful property: it is non-empty. Some parts of the existing code rely on that.

briansmith:
> Hmmm, having a function return io::Positive from a Nonnegative already doesn't seem to work for lifetime reasons (the limbs can't be used as basis, thus the function has to do an allocation),

There are various ways of solving that, some of which avoid allocations and some of which have allocations. For now, I don't care which way we do things; i.e. any allocations you do are OK in the initial version.

briansmith:
> * A operation to multiply two Nonnegative (limbs) numbers, without applying any modulo or without any montgomery representation or anything.

Please explain why you need this operation. This is the operation I was hoping we'd be able to avoid adding to *ring*. Please share a link to the part of the code where you would use this operation.

> * Then an operation to compute `a mod b`, for two limbs numbers as well.

Also, why do we need this? Please share a link to the part of the code where you would use this operation and we'll see what we can do (instead).

briansmith:
> Please explain why you need this operation. This is the operation I was hoping we'd be able to avoid adding to _ring_. Please share a link to the part of the code where you would use this operation.

Of course, I understand we need to calculate N = P * Q. Is there any other reason we'd need it? If so, I can probably come up with a special case thing for the N = P * Q calculation.

est31:
I've ported the modulo code to the ring internal format and am also working on the multiplication code.... the multiplication code is a LOT of work. If it can be avoided somehow, it would be great.

est31:
Oh and I'll also need modulo for [this](https://github.com/briansmith/ring/issues/219#issuecomment-445479144) computation.

est31:
Hmmm I just realized that `(p - 1) * (q - 1) = p * q - p - q - 2 = n - p - q - 2` so I guess that we'll only have to compute n through multiplication.

est31:
Except for two primitives (multiplication and inverse mod m), the bulk of the code is done. Waiting for input from @briansmith on how to do n = p * q. Some checks are still missing though in the algo to obtain p,q and I need to test the whole thing, as well as remove the commented out C code.

est31:
Rebased. Multiplication still left to do, waiting on input by @briansmith .

est31:
@briansmith any news? you said:

> I can probably come up with a special case thing for the N = P * Q calculation.

I stopped my work on porting the multiplication functions due to that, waiting for your input. Should I continue?

est31:
Thanks for the info. You have a concrete idea how to implement it. That's great to hear! I guess I'll wait then.

briansmith:
For future reference: https://eprint.iacr.org/2018/367.pdf

briansmith:
@est31 Are you planning to finish this? If so, please email me at brian@briansmith.org to discuss the IPR stuff. Otherwise, let's just close this.

est31:
@briansmith no my intention is not to close this PR. I've put already a ton of time into it and closing would mean that the time was wasted which I don't want it to be. I've also sent you an E-Mail.

briansmith:
Regarding the divisibility check, check out section "3.2 Fast Divisibility Check with a Single Multiplication" in https://arxiv.org/pdf/1902.01961.pdf. Note in particular that reference [5] may or may not a better starting point than the new method that paper presents.

est31:
@briansmith do you think I need to implement that before merging? Otherwise I'd leave it as a post-merge project.

est31:
@briansmith is the IPR stuff cleared now? I haven't gotten any E-Mail replies.

est31:
Still no reaction...

So now I stand before the choice of bumping this yet again, waiting more time, or simply closing for being ghosted by the maintainer. I choose to close it, because I'm doubtful it'll ever be merged: At the start of this PR, it got tons of reactions by the maintainer and judging from that the general view was that this feature is wanted and the PR welcome. The atmosphere was quite positive until [this](https://github.com/briansmith/ring/pull/733#issuecomment-462060506) comment which suggested to close the PR without any reason. It took me quite by surprise because I was waiting for @briansmith to contribute their part like they said above, and now I'm expected to finish it while the IPR situation is still not clear. There has neither been a yes nor a no in all the months since that communication and this kind of treatment is simply not okay. Also, given that I have invested so much time into my PR, the suggestion to "just close this" is quite insulting.

Of course, ultimately it's their project and up to them to accept or reject stuff.

I've considered closing this multiple times already but every time I stopped in the last moment, hoping that maybe if I wait a little there'd be a reply or something. After all I'm throwing away a ton of work and invested time here. What a shame :/.

In retrospect, I probably shouldn't have started this project without previously doing smaller PRs to change readmes or whatever to probe how receptive the maintainer is. I used to do this but in this case, I let my guard down. My fault. Ultimately, I don't need RSA keygen as much as I used to, as I made it for rcgen which is using multiple other encryption algorithms now. But it would still have been cool to have contributed such a major thing to a crypto library.

That your contribution gets rejected is always a risk of open source but it still hurts every time, especially as I have spent a good deal of time with it. Time to move on.

PS: For all the people who want to add RSA key generation support, first a warning. I don't know why the maintainer has ghosted me, maybe it's because of IPR, maybe it's because the maintainer doesn't like RSA and hopes to be able to wait until RSA is deprecated, never having to worry about key generation at all. Who knows what the reason is. So before you waste time, please make sure you are ready to take the risk of your change being rejected like my PR was. I also suggest not looking at my PR or building onto it unless there are confirmations by the maintainer that this is okay (I'd personally be okay with it). Sorry that I couldn't help you.

PR #738, opened 2018-12-12
Add support for x86_64-fortanix-unknown-sgx target

akash-fortanix:
The x86_64-fortanix-unknown-sgx recently gained Tier 3 support upstream.

### DIFF
--- a/mk/update-travis-yml.py
+++ b/mk/update-travis-yml.py
@@ -47,6 +47,7 @@ compilers = {
     "arm-unknown-linux-gnueabihf" : [ "arm-linux-gnueabihf-gcc" ],
     "i686-unknown-linux-gnu" : linux_compilers,
     "x86_64-unknown-linux-gnu" : linux_compilers,
+    "x86_64-fortanix-unknown-sgx" : linux_compilers,
     "x86_64-apple-darwin" : osx_compilers,
 }
 
@@ -76,17 +77,20 @@ targets = {
         "aarch64-unknown-linux-gnu",
         "i686-unknown-linux-gnu",
         "arm-unknown-linux-gnueabihf",
+        "x86_64-fortanix-unknown-sgx",
     ],
 }
 
 def format_entries():
-    return "\n".join([format_entry(os, target, compiler, rust, mode, features)
-                      for rust in rusts
-                      for os in oss
-                      for target in targets[os]
-                      for compiler in compilers[target]
-                      for mode in modes
-                      for features in feature_sets])
+    return "\n".join([entry for entry in
+                      (format_entry(os, target, compiler, rust, mode, features)
+                       for rust in rusts
+                       for os in oss
+                       for target in targets[os]
+                       for compiler in compilers[target]
+                       for mode in modes
+                       for features in feature_sets)
+                      if entry is not None])
 
 # We use alternative names (the "_X" suffix) so that, in mk/travis.sh, we can
 # ensure that we set the specific variables we want and that no relevant
@@ -111,6 +115,12 @@ entry_sources_template = """
             %(sources)s"""
 
 def format_entry(os, target, compiler, rust, mode, features):
+    if target == "x86_64-fortanix-unknown-sgx" and rust != "nightly":
+        return
+    # Tracked in https://github.com/fortanix/rust-sgx/issues/64
+    if target == "x86_64-fortanix-unknown-sgx" and compiler == "clang":
+        return
+
     target_words = target.split("-")
     arch = target_words[0]
     vendor = target_words[1]
@@ -127,13 +137,11 @@ def format_entry(os, target, compiler, rust, mode, features):
             mode == "DEBUG")
 
     if sys == "darwin":
-        abi = sys
         sys = "macos"
     elif sys == "androideabi":
-        abi = sys
         sys = "linux"
-    else:
-        abi = target_words[3]
+    elif target == "x86_64-fortanix-unknown-sgx":
+        sys = "linux"
 
     def prefix_all(prefix, xs):
         return [prefix + x for x in xs]
@@ -263,7 +271,7 @@ def main():
         file.seek(0)
         file.write(new_contents)
         file.truncate()
-        print new_contents
+        print new_contents,
 
 if __name__ == '__main__':
     main()
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -60,6 +60,8 @@ pub trait SecureRandom: sealed::Sealed {
 /// On Windows, `fill` is implemented using the platform's API for secure
 /// random number generation.
 ///
+/// On SGX, `fill()` is implemented using the `RDRAND` instruction.
+///
 /// Otherwise, `fill()` is implemented by reading from `/dev/urandom`. (This is
 /// something that should be improved for any platform that adds something
 /// better.)
@@ -100,6 +102,7 @@ impl sealed::Sealed for SystemRandom {}
         target_os = "macos",
         target_os = "ios",
         target_os = "fuchsia",
+        target_env = "sgx",
         windows
     ))
 ))]
@@ -119,6 +122,9 @@ use self::darwin::fill as fill_impl;
 
 #[cfg(any(target_os = "fuchsia"))]
 use self::fuchsia::fill as fill_impl;
+
+#[cfg(target_env = "sgx")]
+use self::rdrandom::fill as fill_impl;
 
 use crate::sealed;
 
@@ -309,6 +315,36 @@ mod fuchsia {
     #[link(name = "zircon")]
     extern "C" {
         fn zx_cprng_draw(buffer: *mut u8, length: usize);
+    }
+}
+
+#[cfg(all(target_env = "sgx", target_feature = "rdrand"))]
+mod rdrandom {
+    use crate::{bssl, error};
+
+    fn rdrand_loop() -> Result<[u8; 8], error::Unspecified> {
+        extern "C" {
+            fn CRYPTO_rdrand(out: &mut [u8; 8]) -> bssl::Result;
+        }
+
+        for _ in 0..10 {
+            let mut buf = [0u8; 8];
+            match Result::from(unsafe { CRYPTO_rdrand(&mut buf) }) {
+                Ok(()) => return Ok(buf),
+                Err(_) => continue
+            }
+        }
+        Err(error::Unspecified)
+    }
+
+    pub fn fill(dest: &mut [u8]) -> Result<(), error::Unspecified> {
+        for dst in dest.chunks_mut(8) {
+            let src = rdrand_loop()?;
+            let dst_len = dst.len();
+            dst.copy_from_slice(&src[..dst_len])
+        }
+
+        Ok(())
     }
 }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -416,6 +416,38 @@ matrix:
           sources:
             - ubuntu-toolchain-r-test
 
+    - env: TARGET_X=x86_64-fortanix-unknown-sgx  FEATURES_X= MODE_X=DEBUG KCOV=0
+      rust: nightly
+      os: linux
+      dist: trusty
+
+    - env: TARGET_X=x86_64-fortanix-unknown-sgx  FEATURES_X= MODE_X=RELWITHDEBINFO KCOV=0
+      rust: nightly
+      os: linux
+      dist: trusty
+
+    - env: TARGET_X=x86_64-fortanix-unknown-sgx CC_X=gcc-7 FEATURES_X= MODE_X=DEBUG KCOV=0
+      rust: nightly
+      os: linux
+      dist: trusty
+      addons:
+        apt:
+          packages:
+            - gcc-7
+          sources:
+            - ubuntu-toolchain-r-test
+
+    - env: TARGET_X=x86_64-fortanix-unknown-sgx CC_X=gcc-7 FEATURES_X= MODE_X=RELWITHDEBINFO KCOV=0
+      rust: nightly
+      os: linux
+      dist: trusty
+      addons:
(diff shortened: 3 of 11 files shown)


briansmith:
@akash-fortanix Thanks.

I guess it is hopeless to expect to be able to test this in CI. Howe er, I do have a SGX-enabled laptop myself (IIUC). Please supply a script that allows me to run the *ring* test suite in SGX that people can run manually.

jethrogb:
@briansmith Unfortunately, the libtest crate is not yet supported on `x86_64-fortanix-unknown-sgx`. We are working on making it happen, but it'll take a few weeks (unwinding is one of the biggest dependencies).

briansmith:
OK, thanks. I think that we can handle this like Android, where for some reason we had trouble getting the tests to run. Let's change the CI to add one new target that will do everything except run the tests, and then leave the lines that run the tests commented out.

jethrogb:
Right now, this is a tier 3 target, meaning no binaries fo `std` are distributed. It will need to be built from source. Are you ok with installing xargo in CI to do the build?

briansmith:
> Right now, this is a tier 3 target, meaning no binaries for `std` are distributed. It will need to be built from source. Are you ok with installing xargo in CI to do the build?

We need *some* script that can be run by somebody who wants to test this code. If it requires Xargo then please include the installation of Xargo and its build dependencies in the script. I suggest just adding this https://github.com/rust-embedded/cross since cross already uses xargo for some plaforms.

briansmith:
In the documentation for the `ring::rand` module, please describe how this is done (using RDRAND).

*ring* normally uses CPUID to detect CPU features. How is that supposed to work on this target? In particular, how does *ring* know whether or not it is OK to use AES-NI, AVX2, etc.

briansmith:
Can we assume that at least SSSE3 is available? If so then the difference between a trustworthy CPUID and an untrustworthy one would be primarily performance + the possibility of crashing on an illegal instruction. If we can't even rely on SSSE3 always being available then that is more problematic.

In particular, in the AES-GCM code, there is logic to fall back from AES-NI to other implementations if AES-NI isn't detected. the "VPAES" implementation is (probably, mostly) free of side channels but it requires SSSE3. The "fallback" implementation isn't free of side-channels at all. So, it seems like on SGX the best course of action would be to do whatever SGX gives us for an untrustworthy CPUID, but then hard-code the "SSSE3 enabled" bit so that the VPAES implementation would always be used, and also completely disable via a `#[cfg(target_arch)]` the "fallback" implementation of AES.

Similarly, in gcm.rs, we probably want to assume in SGX that the hardware polynomial multiplication instructions are always available, again via `#[cfg(target_arch)]` .

We also need to document these assumptions somewhere and note the consequence, namely the possibility of getting an illegal instruction exception.

On Twitter, it was suggested that we could detect certain CPU instruction extensions via `sigsetjmp` and `siglongjmp` if they work in SGX. Do they?

jethrogb:
I'm totally fine with requiring and documenting the existence of various features and relying on those in a hard-coded manner.

jethrogb:
It's not clear to me what *ring*s story is w.r.t. `std` support, because of this:

```rust
// src/lib.rs
#![no_std]
...
extern crate std;
```

It looks like you briefly tried to be `no_std` in ae7567553f31283c92b6a1cf7f1f483d5c6bf6fd, but this was later broken in a281f58027cf8aabc2e2fe349dc20bb456ee37dc. Since then a fair amount of `std`-dependent code was added. I'm going to assume `std` is fair game.

In that case, I'm going to suggest `GFp_cpuid_setup` in `cpu-intel.c` is replaced by Rust code that fills in `GFp_ia32cap_P` based on the [`is_x86_feature_detected`](file:///tmp/stdsimd/target/doc/stdsimd/macro.is_x86_feature_detected.html) macro. This way, *ring* will automatically benefit from whatever CPUID-detection mechanism might or might not be added in the future to this or other SGX targets.

Somewhat relatedly, I think this is also how #357 should be implemented once `is_aarch64_feature_detected`/`is_arm_feature_detected` is stable.

akash-fortanix:
AppVeyor build failures are not caused by this PR.

jethrogb:
> There is no extern crate std in ring master.

The current `cfg_attr` almost never gets activated. You need to be Redox or UNIX but not Linux or Mac/IOS. Building for FreeBSD/Redox targets with `--no-default-features` is very broken right now.

```
error[E0433]: failed to resolve: use of undeclared type or module `std`
  --> src/ec/curve25519/ops.rs:22:5
   |
22 | use std::marker::PhantomData;
   |     ^^^ use of undeclared type or module `std`

error[E0432]: unresolved import `std`
  --> src/cpu.rs:19:13
   |
19 |         use std;
   |             ^^^ no `std` external crate

error[E0658]: imports can only refer to extern crate names passed with `--extern` on stable channel (see issue #53130)
   --> src/rand.rs:208:21
    |
191 |     use std;
    |         --- not an extern crate passed with `--extern`
...
208 |                 use std::io::Read;
    |                     ^^^
    |
    = help: add #![feature(uniform_paths)] to the crate attributes to enable
note: this import refers to the unresolved item imported here
   --> src/rand.rs:191:9
    |
191 |     use std;
    |         ^^^

error[E0432]: unresolved import `std`
   --> src/rand.rs:191:9
    |
191 |     use std;
    |         ^^^ no `std` external crate
```

> I would accept a PR to do that.

Ok. Do you want to land that with this or separately?

briansmith:
> src/ec/curve25519/ops.rs:22:5 | 22 | use std::marker::PhantomData;

Fixed in master, along with a couple of other `use std` instances.

> src/cpu.rs:19:13 | 19 | use std;

This still remains; we depend on `std::sync::Once`, which isn't available in libcore. Probably we could work around this.

> src/rand.rs:208:21 | 191 | use std; | --- not an extern crate passed with `--extern`

That's in the `dev/urandom` code. Currently, every target needs to submit a PR to *ring* to tell it how to accesss the PRNG, basically.

briansmith:
> Ok. Do you want to land that with this or separately?

It should be done in a separate PR. I think it is a non-trivial amount of work because the person making the PR needs to communicate clearly how to interpret what the OpenSSL/BoringSSL (assembly) language code is doing, so the reviewer can efficiently verify the correctness of the changes. That means, basically, documenting every access of `GFp_ia32cap_P`.

briansmith:
> `is_x86_feature_detected`

According to https://github.com/rust-lang-nursery/stdsimd/issues/464#issuecomment-392559140, we can't use `is_x86_feature_detected`; instead we need to use whatever lower-level interface that libcore provides.

jethrogb:
Just FYI, the error output in https://github.com/briansmith/ring/pull/738#issuecomment-450096921 was just an excerpt. There were many more errors, I didn't want to put all the lines in my comment. You can trivially discover this by running `cargo build --target x86_64-unknown-freebsd --no-default-features` or `cargo build --target x86_64-unknown-redox --no-default-features`.

> According to rust-lang-nursery/stdsimd#464, we can't use is_x86_feature_detected; instead we need to use whatever lower-level interface that libcore provides.

Well, yes, in my discussion I think I was pretty clear that std support was required for this. I thought this was ok because ring doesn't actually work without std. There is no "lower level interface" in libcore, feature detection requires `std` on several platforms.

briansmith:
I fixed most, but not all, of the `no_std` related build breakage, in 7ad3bb79f9d1646e0d45a80fd4f024241f89a402, abe7b2d7d5dbf6d61c405ce6f1cad6119ec5c321, and b48696520db472448ee491a9f2c4929650c50c82.

The issue of how to make *ring* working in `no_std` configurations is now #744.

A proposal for changing how we do feature detection should be in a separate issue.

I'll email @jethrogb next week about how make progress on this.

jethrogb:
NB I've documented how to build & run for the target at https://github.com/fortanix/rust-sgx/issues/49. As you can see, it's kind of unwieldy right now, but I expect things to get better quickly.

akash-fortanix:
@briansmith Any update on how to take this ahead?

jethrogb:
Updated to include tests. This should start passing in Travis after the new nightly gets released about 15 hours from now (around 0:00 UTC).

W.r.t. to the constant_time_test changes, alternatively, more of the driving logic could be moved into Rust, depending on your preference.

I should also note that all KCOV runners were removed in 1670080572ec5f1a09dd87b007588637bb747587. This PR does nothing to change that situation.

@briansmith We'd like to see this merged soon, please let us know if there's anything else blocking merging this.

elichai:
@briansmith about the CI, what I do is I have set up drone.io in my own server that supports SGX (a couple of cloud providers already support that but not cheaply)

jethrogb:
LTO fix should land at the same time as the Ubuntu 14.04 fix

crypto/constant_time_test.c
```diff
@@ -57,14 +58,14 @@ static int test_binary_op_w(crypto_word (*op)(crypto_word a, crypto_word b),
                             int is_true) {
   crypto_word c = op(a, b);
   if (is_true && c != CONSTTIME_TRUE_W) {
-    fprintf(stderr,
+    printerr(
```
briansmith: These tests will pretty much never fail so I think we should just remove all the print statements completely, instead of jumping through hoops.

mk/update-travis-yml.py
```diff
@@ -111,6 +115,12 @@ def format_entries():
             %(sources)s"""
 
 def format_entry(os, target, compiler, rust, mode, features):
+    if target == "x86_64-fortanix-unknown-sgx" and rust != "nightly":
```
briansmith: When will this target be in stable Rust?

src/rand.rs
```diff
@@ -279,6 +284,19 @@ mod darwin {
     }
 }
 
+#[cfg(target_env = "sgx")]
+mod rdrandom {
+    use crate::error;
+    use rdrand::RdRand;
+    use rand_core::RngCore;
```
briansmith: I don't want to use `rand_core` or `rdrand` in this project. Instead, let's just use the BoringSSL rdrand code via FFI.

Please take the Cargo.toml and rand.rs changes out of this PR and let's do them in another PR.

jethrogb: I'm fine with not using the external crates.

> Please take the Cargo.toml and rand.rs changes out of this PR and let's do them in another PR.

Not sure what you mean here. This is the entire PR?

mk/update-travis-yml.py
jethrogb: It will land in beta if https://github.com/rust-lang/rust/pull/57659 is accepted for backporting, which will put it in stable at 1.33.0. However, that version won't include the fixes for LTO and building on Ubuntu 14.04.

jethrogb:
NB. The commit message of 1c48185a5cb4216aef510f775832cab08a7c201f has an audit of all the CPU features currently used by *ring*. Some of them are pretty strange.

jethrogb:
Filed a PR for MOVBE: https://github.com/rust-lang/rust/pull/57999

jethrogb:
Well, looks like LTO is now failing on *every* platform. But on the plus side, it should work for SGX on the latest nightly.

jethrogb:
Fixed merge conflicts (i.e. duplicate work) with ab0726d, 03e6ef8, 32cf372, 1194b80, bd96baa.

jethrogb:
@briansmith this PR has been open for 2 months. What is blocking merging this?

src/rand.rs
```diff
@@ -307,6 +313,47 @@ mod fuchsia {
...
+    }
+
+    pub fn fill(dest: &mut [u8]) -> Result<(), error::Unspecified> {
+        unsafe {
+            let (prefix, rest) = dest.split_at_mut(dest.len() - (dest.len() % 8));
+            Result::from(CRYPTO_rdrand_multiple8_buf(prefix.as_mut_ptr(), prefix.len()))?;
+
+            if !rest.is_empty() {
+                let mut buf = [0; 8];
+                Result::from(CRYPTO_rdrand(&mut buf))?;
```
briansmith: I think it would be better to do everything with`CRYPTO_rdrand` and avoid `CRYPTO_rdrand_multiple8_buf`. Then we can implement the `retry 10 times` advice from Intel in this Rust code.

src/rand.rs
```diff
@@ -307,6 +313,47 @@ mod fuchsia {
...
 }
 
+#[cfg(all(target_env = "sgx", target_feature = "rdrand"))]
+mod rdrandom {
+    use crate::error;
+    use crate::c;
+
+    #[must_use]
+    #[repr(transparent)]
+    pub struct OneMeansSuccess(c::int);
```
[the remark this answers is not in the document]
jethrogb: Yes, thanks

src/rand.rs
jethrogb: Done

jethrogb:
Ping...

jethrogb:
@briansmith do you have any other comments on this PR?

kostko:
I've been successfully using this branch for the past few weeks, it would be great if it could be merged barring any other concerns?

briansmith:
This target would be one of the most expensive ones to maintain because of the specific toolchain and hardware requirements. it would also be one of the least used; especially I don't expect many non-commercial users to use it. That doesn't seem like a good cost:benefit ratio to me.

The best thing to do is find somebody who is cares about the x86_64-fortanix-unknown-sgx target enough to sponsor at least a year of maintenance of *ring* on it. I have asked around and I've not found anybody particularly interested in it so far. Please email me privately at brian@briansmith.org if you are interested.

briansmith:
I think this is substantially easier now. See https://github.com/briansmith/ring/issues/744#issuecomment-1756037111 and the review I did for UEFI support in https://github.com/briansmith/ring/pull/1406.

briansmith:
Closing due to inactivity.

PR #739, opened 2018-12-12
Implement static key agreement.

### LINKED ISSUE #331: X25519 (and/or ECDH in general) with static keys

briansmith:
Fixes #331.

@danburkert @realcr @kamyuentse @jprider63 @chocol4te @TheBestJohn @kamyuentse Please review.

### DIFF
--- a/src/agreement.rs
+++ b/src/agreement.rs
@@ -33,21 +33,18 @@
 //! // Make `my_public_key` a byte slice containing my public key. In a real
 //! // application, this would be sent to the peer in an encoded protocol
 //! // message.
-//! let mut my_public_key = [0u8; agreement::PUBLIC_KEY_MAX_LEN];
-//! let my_public_key = &mut my_public_key[..my_private_key.public_key_len()];
-//! my_private_key.compute_public_key(my_public_key)?;
+//! let my_public_key = my_private_key.compute_public_key()?;
+//! let my_public_key = my_public_key.as_ref();
 //!
 //! // In a real application, the peer public key would be parsed out of a
 //! // protocol message. Here we just generate one.
 //! let mut peer_public_key_buf = [0u8; agreement::PUBLIC_KEY_MAX_LEN];
-//! let peer_public_key;
-//! {
+//! let peer_public_key = {
 //!     let peer_private_key =
 //!         agreement::PrivateKey::<agreement::Ephemeral>::generate(&agreement::X25519, &rng)?;
-//!     peer_public_key = &mut peer_public_key_buf[..peer_private_key.public_key_len()];
-//!     peer_private_key.compute_public_key(peer_public_key)?;
-//! }
-//! let peer_public_key = untrusted::Input::from(peer_public_key);
+//!     peer_private_key.compute_public_key()?
+//! };
+//! let peer_public_key = untrusted::Input::from(peer_public_key.as_ref());
 //!
 //! // In a real application, the protocol specifies how to determine what
 //! // algorithm was used to generate the peer's private key. Here, we know it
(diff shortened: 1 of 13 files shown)


alex:
Did you actually intend to ask me to review 3 times, or were there perhaps some other Alex's you were looking for?

briansmith:
> Did you actually intend to ask me to review 3 times, or were there perhaps some other Alex's you were looking for?

Sorry, I think you are the top of the autocomplete for some reason. I was having trouble mentioning somebody else (I think they deleted their GitHub account) and didn't intend to mention you at all.

briansmith:
> Edit: Put some time on this. I've had to change my Ed25519KeyPair to a `ring::agreement::PrivateKey<ring::agreement::Static`... not sure if I can then save this out to a pkcs8 file. also having issues with the borrow checker of course. Can't put much more time on this today. Will update

You are trying to do that because you want to serialize and deserialize the static `PrivateKey`, right? I didn't implement that but I agree it makes sense.

realcr:
Thank you for your work!  I read all the commits and the new public interface. From what I understand I should use `PrivateKey<Static>`. Looks good to me. 
I didn't yet try to use this in my codebase. I agree with @TheBestJohn , it might be useful to be able to serialize/deserialize the static private keys.

TheBestJohn:
> You are trying to do that because you want to serialize and deserialize the static `PrivateKey`, right?

Yeah the double ratchet algorithm I'm trying to reproduce uses the previous "sending key" to derive the next I believe. Either way this is for ongoing communication and storing the keys for future messages may be necessary. I haven't figured this out all the way and may be overlooking obvious issues here though. Does a signing keypair and an agreement keypair differ in any real way?

realcr:
> Does a signing keypair and an agreement keypair differ in any way fundamentally?

I'm think that you can not reuse an agreement keypair for signing. Mathematically those keys are equivalent,  but I'm pretty sure they are stored differently. From what I remember Ring does not allow to use one key for those two tasks (Maybe with the new static PrivateKey interface you can?)
Brian probably knows this better.

TheBestJohn:
Makes sense. It likely is a misunderstanding on my part. I am not a crypto guy nor do I think I can even approach making something worthwhile. I'm just trying stuff out for now. It makes sense that you wouldn't want to mix the two.

>  I don't think that the raw private key value is used in the double ratchet except in the ways that are already exposed by _ring_ (as of this PR).

Yeah it need to do the agreement and have the key as an argument. The only way I can really think to do that would be to grab it's root and do all the setup with that. I very likely misunderstood something.

couchand:
> usually you do not want to use the same keypair for both key agreement and signing, and ring enforces this by having separate types for agreement keys and signing keys. However, I know some people have tried to define protocols that use the same key for both. If you are implementing one of those then file a new issue describing (with a link) the protocol you are trying to implement.

While Double Ratchet per se doesn't use signatures, the X3DH key exchange algorithm does, and it is indeed described as using the same key for signing and agreement.  See, for example, [sections 3.2 and 3.3](https://signal.org/docs/specifications/x3dh/#publishing-keys).

The details of the keys are more fully specified in [another document](https://signal.org/docs/specifications/xeddsa/#security-considerations), where this observation is made:

> Key reuse: It is safe to use the same key pair to produce XEdDSA and VXEdDSA signatures.
> 
> In theory, under some circumstances it is safe to use a key pair to produce signatures and also use the same key pair within certain Diffie-Hellman based protocols [15]. In practice this is a complicated topic requiring careful analysis, and is outside the scope of the current document.

As far as I can tell (I'm certainly no expert), their XEdDSA signatures are just regular Ed25519 signatures.  [My Signal implementation](https://github.com/couchand/signal-rs/) (which uses quite the hodgepodge of crypto libraries) currently depends on the Dalek crates for this specific reason: the conversion is easy with them.

Signatures are relatively rare compared to key agreement and none of the keys except the identity key need to be used for signatures, so it would be preferred to store the X25519 keys and just convert to the signing format as needed.

I haven't had a chance to try it out yet, but it looks like this PR has what I'd need to update my example to ring, assuming the above discussion is resolved.

briansmith:
I'm going to change this to get rid of the `Lifetime` arguments in favor of just having separate `EphemeralPrivateKey` and `ReusablePrivateKey` types. The `Lifetime` stuff just was too confusing to people.

briansmith:
Closing this old thing. I'll open a new one when ready.

PR #746, opened 2019-01-02
Remove last SHA-1 signature algorithm (RSA SHA-1).

briansmith:


briansmith:
> I'm not sure if this is a consideration but .org is only signed with RSASHA1-NSEC3-SHA1 at the moment.

Thanks! That's exactly why I posted this PR before the removal.

Do you have a link to any information about plans for .org to change, and/or whether any other TLDs are in the same boat? /cc @bluejekyll

briansmith:
OK, thanks for the prompt feedback. I'm going to close this without merging it, and then file an issue about deprecating SHA-1 stuff in *ring*.

PR #749, opened 2019-01-03
Implement QUIC header protection API

ghedo:
This adds a new API to the AEAD module for generating QUIC Header Protection
masks, as described in draft-ietf-quic-tls-17.

Only AES support is currently implemented, but ChaCha20 can be added later.

---

This implements the API discussed in https://github.com/briansmith/ring/pull/695. After some initial testing this seems to work fine with other QUIC draft-17 implementations, though I haven't finished implementing draft-17 in my own implementation so I haven't gotten to full interop yet.

Also notice the "FIXME" in the `new_mask()` method: is there a way to turn `Block` into `[u8; 16]` that doesn't involve copying?

Finally, I've also started looking at ChaCha20 support (see https://github.com/ghedo/ring/tree/quic_chacha) though it seems the `chacha20_xor_inner()` API was removed so I had to add an `encrypt_block()` method since `encrypt_iv_xor_blocks_in_place()` seems to only work with two blocks. I also had to add a constructor for `Iv` to use a `Block` as input. Is this ok? (havent tested this at all and don't really know if it actually works in practice, so I left it out for now).


src/aead/quic.rs
```diff
@@ -0,0 +1,134 @@
...
+        Ok(HeaderProtectionKey {
+            inner: (algorithm.init)(key_bytes)?,
+            algorithm,
+        })
+    }
+
+    /// Generate a new QUIC Header Protection mask.
+    ///
+    /// `sample` must be exactly `SAMPLE_LEN` bytes long.
+    pub fn new_mask(&self, sample: &[u8]) -> Result<[u8; SAMPLE_LEN], error::Unspecified> {
```
briansmith: In draft 17, IIUC, we need only the first 5 bytes of the output. Is that correct? If so, let's make the return type `Result<[u8; 5], error::Unspecified>`.

In particular, draft 17 says "The output of this algorithm is a 5 byte mask which is applied to the protected header fields using exclusive OR."

src/aead/quic.rs
```diff
@@ -0,0 +1,134 @@
...
+        let sample = <&[u8; SAMPLE_LEN]>::try_from_(sample).unwrap();
+        let sample = block::Block::from(sample);
+        let block = (self.algorithm.new_mask)(&self.inner, sample);
+
+        // FIXME: turn Block into [u8; 16] without copying?
+        let mut out: [u8; SAMPLE_LEN] = unsafe { std::mem::uninitialized() };
```
briansmith: We don't use `mem::uninitialized()` like this in *ring*. Instead we rely on (hope for) dead store elimination to remove the zeroing, and/or an efficient vector store of zeros. If there is a performance issue with doing so then please lmk how severe it is.

src/aead/quic.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    /// Generate a new QUIC Header Protection mask.
+    ///
+    /// `sample` must be exactly `SAMPLE_LEN` bytes long.
+    pub fn new_mask(&self, sample: &[u8]) -> Result<[u8; SAMPLE_LEN], error::Unspecified> {
+        let sample = <&[u8; SAMPLE_LEN]>::try_from_(sample).unwrap();
+        let sample = block::Block::from(sample);
+        let block = (self.algorithm.new_mask)(&self.inner, sample);
+
+        // FIXME: turn Block into [u8; 16] without copying?
```
briansmith: If it actually helps performance, we could return a `Block` (and make `Block` public) to allow users to avoid the copy. However, keep in mind that a 16-byte copy should be just one or a few vector instructions on most targets.

src/aead/quic.rs
```diff
@@ -0,0 +1,134 @@
...
+        let mut out: [u8; SAMPLE_LEN] = unsafe { std::mem::uninitialized() };
+        out.copy_from_slice(block.as_ref());
+
+        Ok(out)
+    }
+
+    /// The key's QUIC Header Protection algorithm.
+    #[inline(always)]
+    pub fn algorithm(&self) -> &'static Algorithm { self.algorithm }
```
briansmith: Is this useful for your application? If you don't actually use it, I suggest we try to go without it until a need crops up for it.

src/aead/quic.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+const SAMPLE_LEN: usize = block::BLOCK_LEN;
```
briansmith: IMO `super::TAG_LEN` would be more descriptive than `block::BLOCK_LEN`.

src/aead/quic.rs
```diff
@@ -0,0 +1,134 @@
...
+            inner: (algorithm.init)(key_bytes)?,
+            algorithm,
+        })
+    }
+
+    /// Generate a new QUIC Header Protection mask.
+    ///
+    /// `sample` must be exactly `SAMPLE_LEN` bytes long.
+    pub fn new_mask(&self, sample: &[u8]) -> Result<[u8; SAMPLE_LEN], error::Unspecified> {
+        let sample = <&[u8; SAMPLE_LEN]>::try_from_(sample).unwrap();
```
briansmith: This shouldn't be `.unwrap()`. it should be `?`. Please add at least three test cases that trigger this: zero-length sample, 15 byte sample, and 17 byte sample.

briansmith:
> Finally, I've also started looking at ChaCha20 support (see https://github.com/ghedo/ring/tree/quic_chacha) though it seems the `chacha20_xor_inner()` API was removed so I had to add an `encrypt_block()` method since `encrypt_iv_xor_blocks_in_place()` seems to only work with two blocks.

`chacha20_xor_inner` is now `chacha::Key::encrypt`.

I suggest adding a method of type `self, &[u8; 16]` -> `[u8; 5]` directly to `chacha::Key` that does exactly what you need, and avoid the `Block` interface completely, changing the `new_mask` internal API you created to do the same.

> I also had to add a constructor for `Iv` to use a `Block` as input. Is this ok?

Yes, please call it `assume_unique_for_key`.

>  (havent tested this at all and don't really know if it actually works in practice, so I left it out for now).

Besides the tests that check the input length handling, please add a simple known answer test for each key type (AES 128, AES 256, and ChaCha20).

briansmith:
Again, I noticed that this seems similar to the length obfuscation in the chacha20-poly1305@openssh.com cipher suite. One question about the protocol: SSH authenticates the masked bytes. Does the current QUIC proposal authenticate the masked bytes or the unmasked bytes? In other words, is the mask itself ever authenticated?

src/aead/quic.rs
ghedo: I'm fine leaving the copy, I just didn't know whether there was something already available to do the conversion. Anyway, I removed the comment.

src/aead/quic.rs
ghedo: Not really, removed.

src/aead/quic.rs
ghedo: Ok, done.

src/aead/quic.rs
ghedo: Ooops, fixed. And added tests.

ghedo:
> I suggest adding a method of type self, &[u8; 16] -> [u8; 5] directly to chacha::Key that does exactly what you need, and avoid the Block interface completely

I ended up adding a new_mask method to aes::Key as well, which also handles the Block -> [u8; 5] conversion.

Though I left the Block as input parameter in the internal new_mask API for now, since it seems that both aes and chacha can use one (aes because of the encrypt API, and chacha because Iv is made of a Block), so passing the Block directly saves having to create one in both cases.

> please add a simple known answer test for each key type (AES 128, AES 256, and ChaCha20).

I added a test case for each, generated using an independent implementation (the one in ngtcp2 https://github.com/ngtcp2/ngtcp2/tree/draft-17). Though I'll see if I can get some more "official" ones.

briansmith:
@quininer Do you have any concerns with this?

briansmith:
Thanks! Looks great. I appreciate you taking the time to learn how *ring* works internally, especially the testing and `KeyInner` type stuff.

briansmith:
When I merged this, I had a bunch of pending merges from BoringSSL locally. I didn't realize it at the time, but it was problematic to merge this before my local changes were merged. Consequently, I force-pushed my changes, undoing the merge here. Then I rebased this change on top of the new master, unchanged, so this became 351ed162516ef2061f42955453b82003bbb63124.

PR #750, opened 2019-01-04
Use `lazy_static!` for CPU feature detection to work with `#[no_std]`.

### LINKED ISSUE #745: Replace the use of `std::sync::Once` with `no_std`-compatible code
See #744. One of the two pain ports in porting *ring* to a `no_std` platform is our use of `std::sync::Once`, which (obviously) requires libstd. There is only one use of `std::sync::Once`. Our single use of `std::sync::Once` is very simple: We look up the CPU capabilities (e.g. using CPUID) and store the results.

I think it is safe to assume that looking up the CPU capabilities is idempotent, so we could probably even let threads race with each other to cache the CPU capabilities, letting them stomp on each other's caching (in a controlled way).

/cc @jethrogb

briansmith:
Fixes #745.

@jethrogb Thoughts?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -290,13 +290,11 @@ include = [
 name = "ring"
 
 [dependencies]
+lazy_static = "1.2"
 untrusted = "0.6.2"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 libc = { version = "0.2.45" }
-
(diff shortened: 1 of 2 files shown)


briansmith:
Never mind. I think it would be better to just use `spin.rs` or similar directly instead.

PR #759, opened 2019-01-15
Replace libstd with spin crate in `cpu::cache_detected_features`.

briansmith:
Eliminate one of the two remaining problems with `#![no_std]` support
and reduce platform variance.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -300,6 +300,9 @@ name = "ring"
 [dependencies]
 untrusted = "0.6.2"
 
+[target.'cfg(not(target_os = "ios"))'.dependencies]
+spin = { version = "0.5.0" }
+
 [target.'cfg(any(target_os = "android", target_os = "linux"))'.dependencies]
 libc = { version = "0.2.45" }
 
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -28,8 +28,8 @@ pub(crate) fn features() -> Features {
     // assumed to be present; see `arm::Feature`.
     #[cfg(not(target_os = "ios"))]
     {
-        static INIT: std::sync::Once = std::sync::ONCE_INIT;
-        INIT.call_once(|| {
+        static INIT: spin::Once<()> = spin::Once::new();
+        let () = INIT.call_once(|| {
             #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
             {
                 extern "C" {


elichai:
It seems that the failure of the test here is because of a network problem, I would just try to rerun it.
https://travis-ci.org/briansmith/ring/jobs/479666488

elichai:
@briansmith If you could use version `0.4.10` instead it would be great, because `0.5.0`  doesn't have `feature(min_const_fn)` and it's still an unstable feature in `nightly-2018-10-01`, so it doesn't compile at that version.
Then this would work with my PR to support sgx: https://github.com/briansmith/ring/pull/764

Cargo.toml
```diff
@@ -297,6 +297,9 @@ name = "ring"
 [dependencies]
 untrusted = "0.6.2"
 
+[target.'cfg(not(target_os = "ios"))'.dependencies]
+spin = { version = "0.5.0", default-features = false }
```
elichai:
```suggestion
spin = "0.4.10"
```
This would support @baidu/rust-sgx-sdk and there are no features in spin so that disabling default-features isn't needed

briansmith: With this:
```toml
[target.'cfg(not(target_os = "ios"))'.dependencies]
spin = { version = "0.4.10", features = ["once"] }
```

I get:
```
error[E0554]: #![feature] may not be used on the stable release channel
#![cfg_attr(feature = "const_fn", feature(const_fn))]
```

So we have to use spin 0.5 in order for it to work on the stable channel.

IMO it would be much better to have the Baidu SGX SDK update the version of Rust it uses to a newer version. Is that not going to happen any time soon?

elichai: That's too bad.
I hope it will be soon
https://github.com/baidu/rust-sgx-sdk/issues/48

elichai: So this isn't relevant anymore, Sorry for the trouble.
I still don't think the `default-features = false`, I see no features in the toml: https://github.com/mvdnes/spin-rs/blob/master/Cargo.toml

briansmith:
So, we can go ahead and do this, but if we do, this would be the first and only non-`rust-lang` crate that *ring* depends on. OTOH `lazy_static` does depend on `spin` itself it any crate has activated the `spin_no_std` feature, and we already depend on `lazy_static` (though we don't active `spin_no_std` ourselves).

Cargo.toml
briansmith: > I still don't think the `default-features = false`, I see no features in the toml: https://github.com/mvdnes/spin-rs/blob/master/Cargo.toml

Thanks! I updated the PR to remove `default-features = false`.

PR #763, opened 2019-01-16
Use Fn bound for kdf on agree_ephemeral.

couchand:
The purpose of the kdf parameter would seem to be to restrict the use
of the key material.  The most restrictive function bound is Fn, which
requires that the kdf be read-only with respect to its environment.

### DIFF
--- a/src/agreement.rs
+++ b/src/agreement.rs
@@ -163,7 +163,7 @@ pub fn agree_ephemeral<F, R, E>(
     peer_public_key: untrusted::Input, error_value: E, kdf: F,
 ) -> Result<R, E>
 where
-    F: FnOnce(&[u8]) -> Result<R, E>,
+    F: Fn(&[u8]) -> Result<R, E>,
 {
     // NSA Guide Prerequisite 1.
     //


couchand:
Yes, `FnOnce` restricts the closure to being called once, but I'm referring to the restrictions on the closures that it is valid for a user to provide.  A specific example is probably helpful:

```
fn bound_fnonce<R, F: FnOnce(&usize) -> R>(n: usize, f: F) -> R {
    f(&n)
}

fn bound_fn<R, F: Fn(&usize) -> R>(n: usize, f: F) -> R {
    f(&n)
}

fn main() {
    // In normal usage, it doesn't seem to matter.
    let six_of_one = bound_fnonce(5, |x| x + 1);
    let half_a_dozen = bound_fn(5, |x| x + 1);

    // But, we can add misuse resistance!
    let mut exfiltrated: Option<usize> = None;
    let this_compiles = bound_fnonce(5, |x| {
        exfiltrated = Some(*x);
        x + 1
    });
    /*
    let this_does_not = bound_fn(5, |x| {
        exfiltrated = Some(*x);
        x + 1
    });
    */
}
```

If you'd try to compile the last snippet, the compiler would catch it:

```
error[E0594]: cannot assign to `exfiltrated`, as it is a captured variable in a `Fn` closure
  --> src/main.rs:21:9
   |
21 |         exfiltrated = Some(*x);
   |         ^^^^^^^^^^^^^^^^^^^^^^ cannot assign
   |
help: consider changing this to accept closures that implement `FnMut`
  --> src/main.rs:20:37
   |
20 |       let this_does_not = bound_fn(5, |x| {
   |  _____________________________________^
21 | |         exfiltrated = Some(*x);
22 | |         x + 1
23 | |     });
   | |_____^
```

briansmith:
The misuse resistance of the proposal is quite limited. Consider this modification of your program:

```rust
fn bound_fn<R, F: Fn(&usize) -> R>(n: usize, f: F) -> R {
    f(&n)
}

fn main() {
    let exfiltrated = bound_fn(5, |x| {
        Some(*x)
    });
    println!("{:?}", exfiltrated);
}
```

If we want to prevent exfiltration like this then we'd need to provide a (sealed) collection of KDFs and force the user to use one of those KDFs.

briansmith:
OK, I'm going to close this. We can't prevent even mild-mannered exfiltration without drastically changing things; switching from `FnOnce` to `Fn` only prevents one way of doing it, which isn't even the most obvious way. Further, switching from `FnOnce` to `Fn` would break multiple users who need the `move` semantics. 

Thanks for the PR though! Let me know if I'm misunderstanding anything and i'll reopen this.

couchand:
No problem if you don't think it's worth it. I'll agree the benefit is marginal. What I like is just that it eliminates the side channel: if you're sneaking out the key the signature would give it away, rather than needing to inspect the code. But the impact on moves is a good point!

briansmith:
@couchand Thanks again for the PR. I do think it would be interesting to propose an RFC about being able to bound closure parameters as `FnOnce + !FnMut` or similar, or some other kind of "pure" closure arguments.

Issue #744, opened 2018-12-27
Document how to port *ring* to a `no_std`-only platform


briansmith:
1. Submit a PR to implement `fill_impl` for the platform in `ring::rand`.
2. (Currently) Submit a PR to get `ring::cpu` working without `std::sync::Once`.

briansmith:
The other, harder, issue, is how to get the test suite to run on such platforms, because the tests use libstd a lot currently. Some issues are probably, e.g. we could use `ArrayVec` or similar instead of `std::vec::Vec` in most cases, though we'd have to worry about stack overflows.

I suggest we prototype this first by solving it specifically for SGX, which seems to be a `no_std` platform that people are especially interested in. See https://github.com/baidu/rust-sgx-sdk#v104-release in particular; it would be nice to help Baidu avoid the need for forking *ring*.

elichai:
@briansmith Are you working on this?
I would be glad to try and do this for sgx.

jonas-schievink:
This solution is unlikely to work in general for embedded platforms. AFAIK, for SGX there is only one obvious implementation of a random number generator (namely the `rdrand` instruction) that can be used for the entire target.

This is not the case for eg. embedded ARM targets, which don't have a single random number generator. Instead, vendors often provide a peripheral for random number generation, which could implement `SecureRandom`.

For those targets I suggest `#[cfg]`ing the `SystemRandom` struct away, and expecting users (or microcontroller support crates) to implement the `SecureRandom` trait for specific peripherals. Would that be an acceptable solution?

briansmith:
> EDIT: Just noticed that `SecureRandom` is sealed, why is that? Can that be removed or would that break any assumptions?

That way we can guarantee it is actually secure by ensuring all the implementations are in *ring*.

> For those targets I suggest #[cfg]ing the SystemRandom struct away, and expecting users (or microcontroller support crates) to implement the SecureRandom trait for specific peripherals. Would that be an acceptable solution?

In such situations the peripherals are almost definitely implementing a raw, unmixed, entropy source. We should take the output of that and postprocess it with a DRBG.

In other words, for `#[cfg(target_os="none")]` we can provide an alternative API for plugging in an entropy source, which feeds a DRBG. Probably we should do that for SGX too, in particular.

jonas-schievink:
Thanks, that makes sense!

> In such situations the peripherals are almost definitely implementing a raw, unmixed, entropy source. We should take the output of that and postprocess it with a DRBG.

[...]

PR #764, opened 2019-01-17
 Added sgx feature for rand with `no-std`

elichai:
Hi, I decided to implement sgx support to ring according to https://github.com/briansmith/ring/issues/744

This should work after this is merged: https://github.com/briansmith/ring/pull/759
(you can run `cargo build --no-default-features --features sgx` )
I added `min_const_fn` because rust-sgx right now supports only up to `nightly-2018-10-01`, and at that version this is still an unstable feature.

@briansmith I would love a feedback :)


elichai:
I tired to understand the `travis.yml` file so I could add a test for that feature but it has too many stuff in it and I'm not sure how to treat it. but there should be a test for `x86_64-unknown-linux-gnu` with `cargo build --no-default-features --features=sgx`

src/rand.rs
```diff
@@ -248,6 +256,20 @@ mod sysrand_or_urandom {
...
+#[cfg(all(
+    target_os = "linux",
+    feature = "sgx",
+    not(any(feature = "dev_urandom_fallback", feature = "use_heap"))))]
+mod sgxrand {
+    use sgx_trts::trts;
+    use crate::error;
+
+    pub fn fill(dest: &mut [u8]) -> Result<(), error::Unspecified> {
+        trts::rsgx_read_rand(dest).map_err(|_| error::Unspecified)
```
briansmith: Is this really a good idea? I looked at the source code for `sgx_raed_rand` in the Intel SGX SDK. It does feature detection using CPUID to try to detect whether RDRAND is available, and falls back to `rand()` if feature detection says RDRAND isn't available. But, feature detection like that isn't secure in SGX because the `CPUID` has to be executed outside the enclave.

In #738, @akash-fortanix and @jethrogb suggested to just hard-code the use of `RDRAND`, i.e. require `RDRAND` be available within SGX, since in practice it seems like all SGX-capable CPUs include RDRAND, despite Intel's warnings. That seems more secure to me. What do you think?

Importantly, I'd rather avoid having two different implementations for this for the two different SGX targets.

elichai: It sounds like you're saying that Intel's secure randomness isn't really secure.
Assembly is out of my yard.
@dingelish What do you think?

dingelish: > It sounds like you're saying that Intel's secure randomness isn't really secure.
> Assembly is out of my yard.
> @dingelish What do you think?

I've not dug into the randomness of `RDRAND` much. What I know is that people choose to trust it or not.

Technically the randomness of `RDRAND` [seems good enough](https://eprint.iacr.org/2014/504.pdf).

src/lib.rs
```diff
@@ -67,14 +67,18 @@
...
     ),
     no_std
 )]
+
 #![cfg_attr(feature = "internal_benches", allow(unstable_features))]
 #![cfg_attr(feature = "internal_benches", feature(test))]
 
+#![cfg_attr(feature = "sgx", allow(unstable_features))] // This is because rust-sgx cuurently use nightly-2018-10-01-x86_64-unknown-linux-gnu
+#![cfg_attr(feature = "sgx", feature(min_const_fn))] // In that version `min_const_fn` is still unstable
+
```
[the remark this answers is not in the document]
elichai: Now that dinglesh confirmed that the rustc will be updated soon I can remove this feature.
just tested with `2019-01-16` and it works great.
https://github.com/baidu/rust-sgx-sdk/issues/48#issuecomment-455283528

elichai:
@briansmith Update me when you decide how you want to do this.
I removed the unstable features.
Do you want to point me to what to do? Or do you prefer to just do it yourself and close this PR? (or maybe a combination of the PR with your own work)

src/rand.rs
briansmith: @dingelish Whether or not RDRAND is secure, both of these approaches ultimately use RDRAND, right? That is, the `sgx_read_rand` function uses RDRAND under the hood, right? The main questoin is whether it is better to use the `sgx_read_rand` wrapper or whether it's better to just directly use `RDRAND`, right?

I have to say, I don't understand the Intel SDK source code at all. My concern was already raised as a bug at https://github.com/intel/linux-sgx/issues/111. It is confusing to have two different implementations of the function (if I understand correctly): the one you pointed me to and the one at https://github.com/intel/linux-sgx/blob/master/common/src/sgx_read_rand.cpp. This makes me lean in favor of just using RDRAND directly for all SGX targets. WDYT?

dingelish: Yes everything will ultimately use RDRAND. I believe in foreseeable future this is the only randomness source in SGX.

You can forget about `sgx_read_rand`, a function exported by Intel SGX SDK, because ring does not need to depends on Intel SGX SDK. Simply wrap `RDRAND` or just use intrinsic functions and that'll be 100% fine for SGX enclave.

Yes. There are two implementations of `sgx_read_rand`. One is for the untrusted side, and another is for the trusted side (SGX Enclave). The code snipped he referred to is used in the untrusted side.

jethrogb:
Platform support should be added with `cfg(target_*)`, not with cargo features. This is what #738 does. By the way, contrary to the PR title, #738 which should support all SGX platforms (not just x86_64-fortanix-unknown-sgx).

elichai:
@briansmith Hi, Any updates on this? Thanks!

elichai:
Another problem is the `cpuid` instruction in `crypto/cpu-intel.c` isn't possible inside of the enclave.

in `linux-sgx` they did an ocall for this for outside of the enclave, but I think that will be too much of a hassle to implement inside of ring.
(here's what's being called when you use Intel's sdk `sgx_cpuid` function: https://github.com/intel/linux-sgx/blob/master/sdk/tlibc/gen/se_cpuid.c#L45 which then call: https://github.com/intel/linux-sgx/blob/master/psw/urts/se_ocalls.cpp#L42 through: https://github.com/intel/linux-sgx/blob/master/common/inc/sgx_tstdc.edl#L34 and `__cpuidex` calls the `cpuid` instruction)

src/rand.rs
briansmith: @dingelish Thanks for the explanation. I agree that we should just more directly use RDRAND and avoid `sgx_read_rand`.

briansmith:
Hi, based on the discussion in this issue and elsewhere, I created Issue #775 to track what needs to be done to support SGX, which seems to be a synthesis of what Baidu has done and what Fortanix has done. I think we should discuss how to get the SGX port working in that issue, instead of in the individual Baidu and Fortanix PRs.

My understanding is that Baidu already updated their SGX SDK so that the new feature flag being proposed here isn't necessary, right?

I'm going to close this specific PR because I think there are some details here that should be changed, but I do appreciate the PR and I think everything that's proposed here will happen, perhaps just slightly differently.

PR #769, opened 2019-01-23
Remove misplaced code block boundary in STYLE.md

ian-h-chamberlain:
The "Submodules and file naming" section ended with three backticks (` ``` `), which caused the next section to be formatted as code instead of using proper Markdown formatting.


briansmith:
Thanks!

PR #771, opened 2019-01-23
Added NotRandom: A dummy random generator for testing.

realcr:
This is an up to date pull request attempting to solve https://github.com/briansmith/ring/issues/661

This pull request adds `NotRandom`, a structure that implements `SecureRandom` and `Sealed`, allowing to write deterministic tests for random related cryptographic code.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -527,9 +527,24 @@ pub mod rand {
         }
     }
 
+    /// An implementation of `SecureRandom` that fills the output slice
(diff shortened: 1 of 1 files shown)


briansmith:
Hi, I'm going to close this for now. There are too many PRs open to address this issue and first we should figure out what we even want to do. If this turns out to be the approach we decide on then let's reopen this after we decide that.

PR #772, opened 2019-01-23
Change Sealed to be public and unsafe trait

realcr:
This is a different approach in attempting to solve https://github.com/briansmith/ring/issues/661. (See a more basic attempt for solution at PR https://github.com/briansmith/ring/pull/771). 

Although only small changes are made here, this is a big design change which should be considered carefully.

## Abstract

This PR proposes to mark the `Sealed` trait unsafe and make it public. This change allows users of the library to do things that might not be intended by ring's developers, but at the same time forces the user to put an `unsafe` clause in their code.

## Motivation

Ring is strict by design, using Rust's safety guarantees to make it difficult to misuse cryptographic primitives. This is a good thing, because it makes mistakes difficult to do. 
However, it is sometimes possible that users of the Ring library might want to do something outside of those restrictions.

By making the `Sealed` trait unsafe and public, we allow users of Ring a safety trade-off.
This idea is in line with many existing Rust interfaces. Most Rust interfaces offer safety by default, but still leave an unsafe hatch for users that know what they are doing. 

One such example could be seen in issue https://github.com/briansmith/ring/issues/661. Currently if one has any above moderately sophisticated code that uses random from Ring, it becomes very difficult to test. The current solution offered by Ring is adding internal dummy random structs that implement `Sealed`. From `src/test.rs`:
```rust
impl sealed::Sealed for FixedByteRandom {}
impl<'a> sealed::Sealed for FixedSliceRandom<'a> {}
impl<'a> sealed::Sealed for FixedSliceSequenceRandom<'a> {}
```
Note that PR https://github.com/briansmith/ring/pull/771 suggests to add `NotRandom` to this list.

If we introduce a public unsafe Sealed trait, we can remove all the mock random structures, and allow the user to make a mock user struct herself, the way she likes to, for example:

```rust
use ring::sealed::Sealed;
unsafe impl Sealed for MyDummyRandom {}
```

### DIFF
--- a/src/ec/curve25519/ed25519/verification.rs
+++ b/src/ec/curve25519/ed25519/verification.rs
@@ -67,7 +67,7 @@ impl signature::VerificationAlgorithm for EdDSAParameters {
     }
 }
 
-impl sealed::Sealed for EdDSAParameters {}
+unsafe impl sealed::Sealed for EdDSAParameters {}
 
 extern "C" {
     fn GFp_x25519_ge_double_scalarmult_vartime(
--- a/src/ec/suite_b/ecdsa/signing.rs
+++ b/src/ec/suite_b/ecdsa/signing.rs
@@ -57,7 +57,7 @@ impl PartialEq for Algorithm {
 
 impl Eq for Algorithm {}
 
-impl sealed::Sealed for Algorithm {}
+unsafe impl sealed::Sealed for Algorithm {}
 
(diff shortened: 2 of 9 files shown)


briansmith:
I definitely don't want to do this. I understand it is desirable to solve the random issue but I don't want to weaken *every* guarantee we have to do that.

PR #773, opened 2019-01-26
Redo most AEAD and Curve25519 CPU feature detection in Rust.

briansmith:


src/aead/gcm.rs
```diff
@@ -116,3 +230,35 @@ struct GCM128_CONTEXT {
     H_unused: Block,
     key: GCM128_KEY,
 }
+
+enum Implementation {
+    CLMUL,
```
briansmith: When we do the SGX port, The `CLMUL` variant should be the only one that is available, since that is the only side-channel-free implementation of GCM.

src/aead/gcm.rs
```diff
@@ -116,3 +230,35 @@ struct GCM128_CONTEXT {
...
+    Fallback,
+}
+
+#[inline]
+fn detect_implementation(cpu: cpu::Features) -> Implementation {
+    if (cpu::intel::FXSR.available(cpu) && cpu::intel::PCLMULQDQ.available(cpu))
+        || cpu::arm::PMULL.available(cpu)
+    {
+        return Implementation::CLMUL;
+    }
```
briansmith: In the future SGX port, if we don't detect `Implementation::CLMUL` then we need to fail here.

src/aead/aes.rs
```diff
@@ -240,31 +248,21 @@ pub enum Implementation {
...
         return Implementation::HWAES;
     }
 
     #[cfg(any(target_arch = "x86_64", target_arch = "x86"))]
     {
-        extern "C" {
-            fn GFp_vpaes_capable() -> c::int;
-        }
-        if unsafe { GFp_vpaes_capable() } != 0 {
+        if cpu::intel::SSSE3.available(cpu_features) {
```
briansmith: In the future SGX port, if we choose neither `Implementation::HWAES` nor `Implementation::VPAES` then we need to fail, because those are the only side-channel-protected implementations.

src/cpu.rs
```diff
@@ -12,15 +12,237 @@
...
+    pub(crate) const AES: Feature = Feature {
+        word: 1,
+        mask: 1 << 25,
+    };
+
+    #[cfg(target_arch = "x86_64")]
+    pub(crate) const AVX: Feature = Feature {
+        word: 1,
+        mask: 1 << 28,
+    };
```
briansmith: For SGX, assuming we're going to use `RDRAND`, we'll need to augment this with a new 

```rust
    pub(crate) const RDRAND: Feature = Feature {
        word: 1,
        mask: 1 << 30,
    };
```

elichai:
I haven't went over all the code, but it seems that you still use `GFp_cpuid_setup()` which calls to  `OPENSSL_cpuid()` which runs the `cpuid` instruction. which will give an illegal instruction in SGX.

briansmith:
Thanks for looking at this!

briansmith:
> I haven't went over all the code, but it seems that you still use `GFp_cpuid_setup()` which calls to `OPENSSL_cpuid()` which runs the `cpuid` instruction. which will give an illegal instruction in SGX.

This is the same before and after this change, right?

Since people were already using *ring* in SGX, how was it working w.r.t. this?

PR #776, opened 2019-01-29
Add support for fuchsia system feature detection on aarch64

erickt:
This adds support for detecting arm features on a fuchsia device,
which uses the `zx_system_get_features` syscall to extract out
this information, which is described here:

https://fuchsia.googlesource.com/zircon/+/master/docs/syscalls/system_get_features.md

The feature constants can be found here:

https://fuchsia.googlesource.com/zircon/+/master/system/public/zircon/features.h

The type `zx_status_t` and ZX_OK are defined here:

https://fuchsia.googlesource.com/zircon/+/master/system/public/zircon/types.h#39
https://fuchsia.googlesource.com/zircon/+/master/system/public/zircon/errors.h#14

### DIFF
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -47,6 +47,11 @@ pub(crate) fn features() -> Features {
             {
                 arm::linux_setup();
             }
+
+            #[cfg(target_os = "fuchsia")]
+            {
+                arm::fuchsia_setup();
+            }
         });
     }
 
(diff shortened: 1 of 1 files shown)


src/cpu.rs
```diff
@@ -110,6 +115,44 @@ pub(crate) mod arm {
...
+        const ZX_ARM64_FEATURE_ISA_PMULL: u32 = 1 << 4;
+        const ZX_ARM64_FEATURE_ISA_SHA2: u32 = 1 << 6;
+
+        let mut caps = 0;
+        let rc = unsafe { zx_system_get_features(ZX_FEATURE_KIND_CPU, &mut caps) };
+
+        // OpenSSL and BoringSSL don't enable any other features if NEON isn't
+        // available.
+        if rc == ZX_OK && (caps & ZX_ARM64_FEATURE_ISA_ASIMD == ZX_ARM64_FEATURE_ISA_ASIMD) {
+            let mut features = NEON.mask;
```
joshlf: I assume we're confident that NEON is enabled in this codepath?

joshlf:
Other than the one minor comment, LGTM!

src/cpu.rs
erickt: I believe so. This is the same logic in https://fuchsia.googlesource.com/zircon/+/master/third_party/ulib/uboringssl/crypto/cpu-aarch64-fuchsia.c, and up above at line 85.

erickt:
It turns out there actually are some issues building on aarch64. I suspect I just need to get the build.rs handling our setup. I'm working to resolve it.

erickt:
Since arm feature detection landed in master, I decided to just squash this PR with #765.

PR #783, opened 2019-02-01
Run aarch64 android tests on travis.

### LINKED ISSUE #486: Add aarch64-linux-android targets to Travis CI
At the time we added 32-bit ARM Android to Travis CI, Aarch64 builds of Rust's libstd weren't available. However, now they are available, so we can add AAarch64 targets now.

pietro:
Closes #486.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,19 +31,11 @@ matrix:
       rust: stable
       os: linux
       dist: trusty
-      addons:
-        apt:
-          packages:
-            - expect
 
     - env: TARGET_X=armv7-linux-androideabi CC_X=arm-linux-androideabi-clang FEATURES_X= MODE_X=RELWITHDEBINFO KCOV=0
       rust: stable
       os: linux
       dist: trusty
-      addons:
-        apt:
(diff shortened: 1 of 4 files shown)


pietro:
@briansmith I suspect this will work if you clear the build caches in travis. I'm sorry I haven't had time to look into improving the android sdk/ndk install to auto clear the cache.

briansmith:
@pietro Thanks for your work on this.

Even after I cleared the build caches and whatnot, I couldn't get it to work.

I think we should try to get the emulator-based tests for AAarch64 Android running on top of the current (just changed) Android CI setup, which would require a very different PR. If we really have to resort to manually downloading the NDK and/or emulator to get this working then we'll do so, but I'd like that to be done only as a last resort.

A silver lining: I now finally understand more or less how the Android toolchain works and so I should be able to review future Android-targeted changes much faster.

With that in mind, I'm going to close this unless/until we can figure out a new PR that works reliably.

Thanks again for your amazing help with this and other CI issues! Really your help has been amazing.

Issue #775, opened 2019-01-29
Port to SGX


briansmith:
Various people want *ring* to work on SGX. Here's what needs to be done to have a decent port, AFAICT:

- [ ] Replace the use of CPUID in cpu-intel.c with something SGX-compatible for SGX targets. Baidu patched *ring* to use `sgx_cpuid`.
- [ ] Change `detect_implementation()` in aes.rs so that SGX targets will only use the constant-time AES-NI or constant-time VPAES implementations, and never any of the other leaky implementations.
- [ ] Change `detect_implementation()` in gcm.rs so that SGX will only use the constant-time `CLMUL` implementation or the ghash-ssse3-x86_64.pl implementation (which hasn't been imported from BoringSSL yet), and never any of the other leaky implementations.
- [ ] For SGX targets only, change rand.rs for SGX targets to use RDRAND via the BoringSSL [rdrand-x86_64.pl](https://boringssl.googlesource.com/boringssl/+log/master/crypto/fipsmodule/rand/asm/rdrand-x86_64.pl). I filed https://bugs.chromium.org/p/boringssl/issues/detail?id=262 to ask the BoringSSL developers about the lack of retry logic in that code. Document clearly in the `ring:rand` documentation that on SGX only, `RDRAND` is used.
- [ ] Remove use of `fprintf` from the constant_time_test.c.
- [ ] Merge #759: Use `spin::Once` instead of `std::sync::Once` in cpu.rs. (Needed by Baidu only, not Fortanix).
- [ ] Update Travis CI to build SGX test target. Take this from #738.
- [ ] Create a new document for SGX-specific issues:
   - [ ]  Document SGX-specific risks with how feature detection is done. In particular, document that because the CPU feature detection is done outside of the enclave, it can be tampered with, and because we rely on the results of this feature detection, it is possible for untrusted code outside of the enclave to trick us into thinking that a CPU instruction is available when it is not, and thus they can force us to halt the enclave with an illegal instruction fault.
    - [ ] Warn in this document, in addition to the `ring::rand` API documentation, that `ring::rand` is implemented solely in terms of `RDRAND` for SGX targets.
    - [ ] Document that SGX tests aren't run in CI. File a new issue in the issue tracker and link to the issue in this section of this new SGX-specific  documentation.
    - [ ] Document that the SGX port is tested to build correctly only in Nightly Rust. File a new issue in the issue tracker about adding stable and beta channel support, and link to the issue in this section of this new SGX-specific documentation.

[...]

PR #787, opened 2019-02-06
Add support for x86_64-unknown-uefi

josephlr:
Right now, ring contains a lot of long `cfg` clauses to determine which RNG implementation to use or if the crate should be `no_std`. This change simplifies them, mainly by using the fact that `std` in only ever used in tests or if `use_heap` is enabled. It also avoids using the confusing `windows` and `unix` attributes, and now makes all target decisions based on `target_os`.

This also fixes a bug arising from the fact that `#[cfg(windows)] != #[cfg(target_os = "windows")]`. [Some custom targets](https://github.com/rust-osdev/uefi-rs/blob/master/uefi-test-runner/x86_64-uefi.json) base themselves off of the windows family of LLVM, but these targets do _not_ have any of the Windows OS services.

The remaining issues blocking use of `x86_64-unknown-uefi` are:
- [ ] support for `libc` types: (https://github.com/rust-lang/libc/pull/1244)
- [ ] RDRAND support: part of https://github.com/briansmith/ring/pull/738
- [ ] Implement `fill_impl` to check for and then use RDRAND.
- [ ] Add UEFI testing to CI: see https://github.com/rust-osdev/uefi-rs

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,19 +59,8 @@
     unused_results,
     warnings
 )]
-#![cfg_attr(
-    any(
-        target_os = "redox",
-        all(
-            not(test),
-            not(feature = "use_heap"),
-            unix,
-            not(any(target_os = "macos", target_os = "ios")),
-            any(not(target_os = "linux"), feature = "dev_urandom_fallback")
-        )
-    ),
-    no_std
-)]
+// std is only used in tests or when gated with #[cfg(use_heap)]
+#![cfg_attr(not(any(test, feature = "use_heap")), no_std)]
 #![cfg_attr(feature = "internal_benches", allow(unstable_features))]
 #![cfg_attr(feature = "internal_benches", feature(test))]
 
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -100,14 +100,14 @@ impl sealed::Sealed for SystemRandom {}
         target_os = "macos",
         target_os = "ios",
         target_os = "fuchsia",
-        windows
+        target_os = "windows"
     ))
 ))]
 use self::urandom::fill as fill_impl;
 
 #[cfg(any(
     all(target_os = "linux", not(feature = "dev_urandom_fallback")),
-    windows
+    target_os = "windows",
 ))]
 use self::sysrand::fill as fill_impl;
 
@@ -158,7 +158,7 @@ mod sysrand_chunk {
     }
 }
 
-#[cfg(windows)]
+#[cfg(target_os = "windows")]
 mod sysrand_chunk {
     use crate::{error, polyfill};
     use core;
@@ -183,7 +183,7 @@ mod sysrand_chunk {
     }
 }
 
-#[cfg(any(target_os = "linux", windows))]
+#[cfg(any(target_os = "linux", target_os = "windows"))]
 mod sysrand {
     use super::sysrand_chunk::chunk;
     use crate::error;
@@ -198,13 +198,13 @@ mod sysrand {
     }
 }
 
-// Keep the `cfg` conditions in sync with the conditions in lib.rs.
 #[cfg(all(
     feature = "use_heap",
-    any(target_os = "redox", unix),
-    not(any(target_os = "macos", target_os = "ios")),
     not(all(target_os = "linux", not(feature = "dev_urandom_fallback"))),
-    not(any(target_os = "fuchsia")),
+    not(target_os = "macos"),
+    not(target_os = "ios"),
+    not(target_os = "fuchsia"),
+    not(target_os = "windows"),
 ))]
 mod urandom {
     use crate::error;
@@ -215,7 +215,7 @@ mod urandom {
 
         #[cfg(target_os = "redox")]
         static RANDOM_PATH: &str = "rand:";
-        #[cfg(unix)]
+        #[cfg(not(target_os = "redox"))]
         static RANDOM_PATH: &str = "/dev/urandom";
 
         lazy_static! {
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -59,19 +59,8 @@
...
-            not(feature = "use_heap"),
-            unix,
-            not(any(target_os = "macos", target_os = "ios")),
-            any(not(target_os = "linux"), feature = "dev_urandom_fallback")
-        )
-    ),
-    no_std
-)]
+// std is only used in tests or when gated with #[cfg(use_heap)]
+#![cfg_attr(not(any(test, feature = "use_heap")), no_std)]
```
briansmith: The comment here isn't necessary since it's just duplicating the code, so let's remove it.

src/rand.rs
```diff
@@ -100,14 +100,14 @@ impl sealed::Sealed for SystemRandom {}
         target_os = "macos",
         target_os = "ios",
         target_os = "fuchsia",
-        windows
+        target_os = "windows"
```
briansmith: How should `fill_impl` be implemented for UEFI targets? Please provide an implementation.

src/rand.rs
```diff
@@ -198,13 +198,13 @@ mod sysrand {
...
 #[cfg(all(
     feature = "use_heap",
-    any(target_os = "redox", unix),
-    not(any(target_os = "macos", target_os = "ios")),
     not(all(target_os = "linux", not(feature = "dev_urandom_fallback"))),
-    not(any(target_os = "fuchsia")),
+    not(target_os = "macos"),
+    not(target_os = "ios"),
+    not(target_os = "fuchsia"),
+    not(target_os = "windows"),
```
briansmith: I am planning to switch this to a whitelist approach because I want to avoid using `/dev/urandom` whenever practical. So I think that while this may work right now, it will soon (in a matter of days) stop working.

src/rand.rs
```diff
@@ -215,7 +215,7 @@ mod urandom {
 
         #[cfg(target_os = "redox")]
         static RANDOM_PATH: &str = "rand:";
-        #[cfg(unix)]
+        #[cfg(not(target_os = "redox"))]
```
briansmith: What was wrong with `cfg(unix)` here? It seems better to me.

briansmith:
Generally, I am happy to help people port *ring* to new environments, but to minimize the maintenance burden I ask that ports be reasonably complete and easy for me to test.

In the documentation for `ring::rand`, please add a note about how `ring::rand` is implemented for UEFI targets, similar to the other ones.

According to https://github.com/rust-osdev/uefi-rs/blob/master/.travis.yml, it seems like it is relatively straightforward to add UEFI as a target to Travis CI. Please do so.

briansmith:
Also, thanks for the PR!

josephlr:
@briansmith I was thinking to have this PR only simplify the `cfg` attributes. I was goint to add support for UEFI in a seperate PR. Do you want me to just do both in this PR?

Also, UEFI (like SGX) has to use hardware RNG, so actually completing the port will have to wait on some of the material from #775

josephlr:
It looks like #738 adds `RDRAND` support should I factor that change out and add it here, or wait for that PR to land?

briansmith:
> Do you want me to just do both in this PR?

I would like to have everything together in one PR, with a separate commit for each separate thing being changed. I will probably merge them all at once.

Please add the contribution statement at https://github.com/briansmith/ring#contributing at the end of each commit's commit message. Is Google the copyright holder for these changes?

josephlr:
> I would like to have everything together in one PR, with a separate commit for each separate thing being changed. I will probably merge them all at once.

Sounds good, I'll have this commit, rdrand support, and UEFI support

> Please add the contribution statement at https://github.com/briansmith/ring#contributing at the end of each commit's commit message. Is Google the copyright holder for these changes?

Will do, and yes Google is still the copyright holder. I can go though a process to transfer copyright to you (if you prefer), but the projects we contribute to are normally "Copyright The <project-name> Authors", avoiding any need for approval or rights transfer. This is _mostly_ what OpenSSL/[Chromium](https://chromium.googlesource.com/chromium/src/+/master/LICENSE)/Linux/etc... do.

I can change the `Copyright 2015-2016 Brian Smith.` statements to be `Copyright The Ring Authors`. Would that work with you?

briansmith:
Regarding `RDRAND`, I am sad about the idea of relying on that, although right now I don't have a better suggestion. It would be great if you could help review the PR that adds `RDRAND`. It is fine by me if you want to rebase this on top of the SGX PR.

briansmith:
Assigning the copyright would be best. Otherwise, in other files, we have `Portions Copyright 2019 Google, Inc.`

briansmith:
I'm currently planning for *ring* 0.17 to switch away from `spin` back to `std::sync::Once` for x86_64 targets in cpu.rs, which may have a negative impact here. If/when we do that, we'll still be open to accepting a PR for getting the `-uefi` target working, if something is needed.

josephlr:
This will be fine. I think bare-metal x86 targets shouldn't need spinlocks. All we have to do is check CPUID, and that operation is idempotent. So we should just be able to use atomic load/store, where we might run cpuid twice, but that seems fine.

For reference, `getrandom` has to check CPUID on bare-metal x86 targets. [This](https://github.com/rust-random/getrandom/blob/8365277450e6c88d603698154dfd14e31c13d566/src/rdrand.rs#L56-L60) is our spinlock-free implementation of the check.

josephlr:
Closing this as the issues here are best solved generically for all bare-metal x86 targets.

PR #795, opened 2019-02-22
Implement RSAKeyPair from components take #2

d-e-s-o:
This pull request picks up where https://github.com/briansmith/ring/pull/714 left off and addresses https://github.com/briansmith/ring/issues/699. It adds an additional constructor for `rsa::KeyPair` that accepts a key's constituent parts.


d-e-s-o:
So what is the deal with `Positive`? Is that supposed to be user facing or not, because it does not seem to have any sensible constructors?

briansmith:
> So what is the deal with `Positive`? Is that supposed to be user facing or not, because it does not seem to have any sensible constructors?

Right now, it is user-facing but not user-constructable. As part of this PR we should add the public constructor that one would need. Also, in `rsa_test.rs`, we should have at least one test case that uses this new API, so that we don't break the API in backward-compatible ways accidentally, e.g. by removing the public constructors needed.

d-e-s-o:
Okay, thanks for the feedback.

d-e-s-o:
So the problem I seemingly still have (and this is not directly related to the change at hand, but it could affect users of the newly introduce `RsaKeyPair` constructor), is that I don't think (but haven't verified yet) that I have the `dP`, `dQ`, and `qInv` components of the key. 

My understanding is that conceptually, I can just calculate them ([this site](https://www.di-mgt.com.au/crt_rsa.html) has details), but I'd honestly want avoid having that logic in my own code. Rather, it seems that `ring` should be more flexible and do that for me, by means of a different constructor, for example. It seems likely that other users have the same problem. What are your thoughts?

briansmith:
> So the problem I seemingly still have (and this is not directly related to the change at hand, but it could affect users of the newly introduce `RsaKeyPair` constructor), is that I don't think (but haven't verified yet) that I have the `dP`, `dQ`, and `qInv` components of the key.

I suggest we tackle that separately. If we were to add support for incomplete inputs we could add a new constructor for doing so.

> Rather, it seems that `ring` should be more flexible and do that for me, by means of a different constructor, for example. It seems likely that other users have the same problem. What are your thoughts?

Up to this point we've avoided implementing any modular arithmetic where the modulus is even.

What kind of application are you making that is frequently encountering RSA private keys without the CRT parameters?

d-e-s-o:
> > So the problem I seemingly still have (and this is not directly related to the change at hand, but it could affect users of the newly introduce `RsaKeyPair` constructor), is that I don't think (but haven't verified yet) that I have the `dP`, `dQ`, and `qInv` components of the key.
> 
> I suggest we tackle that separately. If we were to add support for incomplete inputs we could add a new constructor for doing so.

I'll open an issue once this request has gotten merged, if I think it's necessary.

> What kind of application are you making that is frequently encountering RSA private keys without the CRT parameters?

It's an [SSH agent](https://github.com/d-e-s-o/ssh-gpg-agent). I haven't studied the format closely, but just use a crate to load them and the resulting objects are lacking those parameters.

d-e-s-o:
Is there anything blocking this pull request from being merged?

src/io.rs
```diff
@@ -24,10 +24,18 @@ mod writer;
 pub(crate) mod der_writer;
 
 /// A serialized positive integer.
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
```
briansmith: We intentionally don't derive `Debug` for this type because it contains/references potentially-secret data. We should document this policy in a comment.

src/io.rs
```diff
@@ -24,10 +24,18 @@ mod writer;
...
 pub struct Positive<'a>(untrusted::Input<'a>);
 
 impl<'a> Positive<'a> {
+    /// Constructs a serialized positive integer from serialized
+    /// untrusted input.
+    pub fn from_untrusted(
```
briansmith: I suggest we just call this `new` or `parse`.

src/io/der.rs
```diff
@@ -121,8 +121,15 @@ where
...
 
-fn nonnegative_integer<'a>(
+fn read_nonnegative_integer<'a>(
     input: &mut untrusted::Reader<'a>, min_value: u8,
+) -> Result<untrusted::Input<'a>, error::Unspecified> {
+    let value = expect_tag_and_get_value(input, Tag::Integer)?;
+    nonnegative_integer(value, min_value)
+}
+
+pub fn nonnegative_integer<'a>(
```
briansmith: This wouldn't belong in the `der` submodule any more since it doesn't have anything to do with DER anymore.

One possibility: 1. Move all this logic to the new `Positive` constructor you added. 2. Rewrite `der::nonnegative_integer` to handle two cases separately: the zero-value case and the other case.

src/io.rs
```diff
@@ -24,10 +24,18 @@ mod writer;
 pub(crate) mod der_writer;
 
 /// A serialized positive integer.
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub struct Positive<'a>(untrusted::Input<'a>);
 
 impl<'a> Positive<'a> {
+    /// Constructs a serialized positive integer from serialized
+    /// untrusted input.
```
briansmith: This specification is not precise enough.

In the test cases you provided, the inputs have a leading zero if it has the highest bit set. This is a DER-specific encoding rule that is used to disambiguate negative and positive values; That rule isn't appropriate for this constructor. Instead, like I mentioned, leading zeros should be disallowed, and all values should be assumed to be positive.

Further, the input must be in big-endian order, and empty inputs are not allowed. Further, leading zeros must be rejected. This means, in particular, that zero is encoded as a single zero byte.

All of these aspects must be documented here, and each of these aspects of the constructor must be tested in integration tests (a test under tests/, e.g. in a file `positive_tests.rs`). In particular, "", 00, 0000, 000000, 0001, 00ff, ff, 00ff00, 01, and some long inputs should be tested (this list isn't intended to be exhaustive).

src/rsa/signing.rs
```diff
@@ -26,6 +26,30 @@ use crate::{
 };
 use untrusted;
 
+/// Used to construct an RSA key pair from components.
+///
+/// Components are in big-endian, considered unsigned, and must not contain
```
briansmith: This documentation about how a `Positive` must be encoded should be in the `Positive` constructor, not here.

briansmith:
Thanks for this PR. I appreciate the time you put into it. I am now taking on this task and I have a WIP implementation based on the experience of trying to use it to build something that uses it. Like when *ring* used `untrusted::Input` instead of `&[u8]` in its public API, it was also a mistake for me to insist that `Positive` be part of this API and the related APIs. I have a new implementation of this that I will post shortly as a PR, so I'm going to close this PR.

PR #319, opened 2016-10-23
[WIP] Issue #316 - Replace `/dev/urandom` usage on OpenBSD

tatsuya6502:
**Work In Progress. Please do not merge this PR yet.**

I started to work on issue [#136 "Replace `/dev/urandom` usage on *BSD"](https://github.com/briansmith/ring/issues/316). I have replaced `/dev/urandom` usage in OpenBSD and FreeBSD with `arc4rand_buf()` function in their C stdlib.

I am testing this on OpenBSD 6.0 (amd64) and FreeBSD 10.3-RELEASE (amd64).

**TODOs:**
- [ ] Update docs.
- [ ] Remove debug print in crypt/rand/sysrand.c.
- [x] Ensure FreeBSD's `arc4rand_buf()` is as safe as OpenBSD's. -> (Nov 1, 2016). It is ***not*** as safe as OpenBSD's. I will revert changes for FreeBSD. For more details, see [this comment](https://github.com/briansmith/ring/issues/316#issuecomment-257479426).

PR #797, opened 2019-02-24
On OpenBSD, use arc4random_buf to implement SystemRandom::fill()

### LINKED ISSUE #316: Replace `/dev/urandom` usage on OpenBSD
See https://github.com/briansmith/ring/pull/313#issuecomment-253959079, where @tatsuya6502 wrote:

> > If you are interested in better BSD support, one obvious next step would be to replace the implementation of ring::rand with one that uses arc4random or whatever is best on BSD, instead of using reads from /dev/urandom.
> 
> OK. Let me work on that. I did a bit of research and arc4random seems the best on BSD by the following reasons:
> 
> It is widely available on BSD systems.
> It produces high quality data.
> It is non blocking.
> http://man.openbsd.org/OpenBSD-6.0/arc4random.3
> 
> This family of functions provides higher quality data than those described in rand(3), random(3), and rand48(3).
> 
> Use of these functions is encouraged for almost all random number consumption because the other interfaces are deficient in either quality, portability, standardization, or availability. These functions can be called in almost all coding environments, including pthreads(3) and chroot(2).

That sounds good to me. The main questions I have are:
- You say it is non-blocking. I think this is because *BSD blocks userspace programs from running until the kernel entropy pool has been initialized, right? Or, does it potentially block in the situation where the kernel entropy pool isn't initialized? Or some other behavior?
- In your description of the functions you use "almost" twice. In what situations would using that function be inappropriate?

oherrala:
OpenBSD provides [`arc4random_buf()`](https://man.openbsd.org/arc4random_buf) to provide secure pseudo-random numbers. This PR mimics how fuchsia's `SystemRandom::fill()` is implemented, but just calls `arc4random_buf()` instead.

There have been work done by @tatsuya6502 in #319, but it hasn't seen progress. This PR should implement what has been there (without C) for OpenBSD, but not for FreeBSD. However, FreeBSD can be trivially added.

I tested this with OpenBSD-current (2019-02-22 snapshot) using Rust 1.32.0.

Fixes #316

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -308,7 +308,7 @@ spin = { version = "0.5.0" }
 
 [target.'cfg(any(target_os = "android", target_os = "linux"))'.dependencies]
 
-[target.'cfg(any(target_os = "redox", all(unix, not(any(target_os = "macos", target_os = "ios")))))'.dependencies]
+[target.'cfg(any(target_os = "redox", all(unix, not(any(target_os = "macos", target_os = "ios", target_os = "openbsd")))))'.dependencies]
 lazy_static = "1.2"
 
 [target.'cfg(target_os = "windows")'.dependencies]
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -55,6 +55,8 @@ pub trait SecureRandom: sealed::Sealed {
 ///
 /// On macOS and iOS, `fill()` is implemented using `SecRandomCopyBytes`.
 ///
+/// On OpenBSD, `fill()` is implemented using `arc4random_buf`.
+///
 /// On Redox, `fill()` is implemented by reading from `rand:`.
 ///
 /// On Windows, `fill` is implemented using the platform's API for secure
@@ -100,6 +102,7 @@ impl sealed::Sealed for SystemRandom {}
         target_os = "macos",
         target_os = "ios",
         target_os = "fuchsia",
+        target_os = "openbsd",
         windows
     ))
 ))]
@@ -119,6 +122,9 @@ use self::darwin::fill as fill_impl;
 
 #[cfg(any(target_os = "fuchsia"))]
 use self::fuchsia::fill as fill_impl;
+
+#[cfg(any(target_os = "openbsd"))]
+use self::openbsd::fill as fill_impl;
 
 use crate::sealed;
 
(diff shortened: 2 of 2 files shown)


briansmith:
@oherrala Thanks for this. What's the easiest way to test this from a Linux or Windows shell? Does `cross` work yet?

oherrala:
@briansmith I have not tested cross. I'll see if it could be used.

Personally I have OpenBSD VM running in VirtualBox to develop and test.

briansmith:
I'd rather not use `arc4random_buf()` because it's a ChaCha20-based PRNG on top of `getentropy()`. I would rather just use `getentropy()` directly. If we want to have a userspace PRNG then we should make one works over all of the OS-specific APIs that also supports personalization strings and/or "additional data."

briansmith:
Thanks for the PR.

The `ring::rand` code has been simplified to make a PR like this easier to do. Also I see that [libc::getentropy()](https://rust-lang.github.io/libc/x86_64-unknown-openbsd/libc/fn.getentropy.html) already exists so we should be able to do this without any direct use of `unsafe` in *ring* at all.

If it is important that we use a userspace CSPRNG like `arc4random_buf` implements then we should implement that userspace CSPRNG in Rust in *ring*, and then have that CSPRNG seed itself using `getentropy()`.

Also, what we really need is some kind of testing. At least we should make `cargo test --no-run` work in CI for OpenBSD and FreeBSD.

I'm going to close this PR because I think it would have to be totally rewritten based on what we know nowq. But I do appreciate the effort here!

oherrala:
> Also, what we really need is some kind of testing. At least we should make cargo test --no-run work in CI for OpenBSD and FreeBSD.

I agree about testing and CI. CirrusCI supports FreeBSD natively so something could be tested there. OpenBSD and Rust still feels lacking, but I hope things improve in future.

> I'm going to close this PR because I think it would have to be totally rewritten based on what we know nowq. But I do appreciate the effort here!

Thanks and I agree. I'll see what has happened to `ring::rand` and see if there's anything I could contribute in future. I have also thought about CSPRNG implemented in Rust. Dunno how fun project it might be to port OpenBSD's `arc4random` into Rust..

PR #799, opened 2019-02-26
Add diagnostic when compiling for a non-supported platform

tomaka:
Tried to compile for WASM, had to read the source code to figure out. This may save someone a few minutes in the future.


briansmith:
Thanks for this. Unfortunately, this gets in the way of people trying to add support for those unsupported platforms. I already did a bunch of work to support WASM but I haven't found anybody to pay for it yet so it's just sitting around in an incomplete state on my laptop. When we find the funding to finish the work then this won't be an issue. So, I'm just going to close this. Thanks again, though.

PR #801, opened 2019-03-06
Add more variants of ECDSA / SHA2 signing

npmccallum:


briansmith:
I intentionally didn't implement those variants. What's the use case for adding them?

npmccallum:
[AMD SEV](https://developer.amd.com/sev/)

The choices are now burned into CPUs, so we're stuck with them.

npmccallum:
(I don't work for AMD.)

npmccallum:
@briansmith Can this be merged?

Darkspirit:
Naive question from a naive outsider: To me it looks like AMD SEV would support usual curve/hash combinations. So what wouldn't work without merging this PR, and what could be done to stop others doing it wrong?

npmccallum:
@Darkspirit I have at least one certificate from the hardware with this combination. "Fixing" this problem means (1) convincing AMD to fix the problem and (2) make new CPUs and (3) buying a new CPU. This is an enormous amount of energy to overcome and will probably cost millions, maybe billions, of dollars.

npmccallum:
FYI, I'm already trying to convince AMD to fix this in the future. But there are millions of CPUs out there.

Darkspirit:
For reference: https://github.com/AMDESE/AMDSEV/issues

npmccallum:
@briansmith I'm blocked on my release until this is merged.

briansmith:
>  I have at least one certificate from the hardware with this combination. 

Which combination? P-384 + SHA-256 or P-256 + SHA-384? I'd rather just add the one that we have evidence, if we're going to add any of them.

Do you need to verify signatures with these combinations or do you really need to generate signatures with these combinations?

At a minimum, the `digest_scalar` tests need to be extended to include the combination(s) you want added, and also the ECDSA signature verification and/or signing tests need to be similarly extended. I think the existing tests (for the existing combinations) already cite the sources of the test vectors; make sure all the relevant test vectors from those sources are merged in.

josephlr:
> Which combination? P-384 + SHA-256 or P-256 + SHA-384? I'd rather just add the one that we have evidence, if we're going to add any of them.

We're also interested in supporting AMD SEV,  and almost certainly @npmccallum is talking about `P384_SHA256` support. From the [AMD SEV spec](https://developer.amd.com/wp-content/resources/55766.PDF) Chapter 2, Table 1, all of the asymmetric signing keys are either RSA-2048 or ECDSA-p384. From Appendix C, Table 90, the supported hash algorithms for ECDSA-p384 are SHA256 and SHA384.  Given that SHA256 is better supported (both [in hardware](https://en.wikipedia.org/wiki/Intel_SHA_extensions) and software), it's not surprising that, when given the option, `P384_SHA256` ends up being more common than `P384_SHA384`.

> Do you need to verify signatures with these combinations or do you really need to generate signatures with these combinations?

Reading Chapter 2 of the spec, there are three ECDSA-p384 signing keys: the CEK, PEK, and OCA. The CEK is burned into the chip, and the PEK is generated by the firmware, so *ring* seems seems to support CEK/PEK operations just fine. Although I don't really know the difference between the `*_FIXED` and `*_ASN1` algorithms, and I don't understand why we would have one but not the other. 

 However, the OCA is generated by the owner of the platform, so that would actually be done by *ring* (hypothetically). @npmccallum is there a reason you would want to generate your certificates/signatures with SHA256? (assuming that you're the platform owner in this context).

I think this PR should be revised to add in only `P384_SHA384` and also add tests.

briansmith:
> I think this PR should be revised to add in only `P384_SHA384` and also add tests.

We already have `ECDSA_P384_SHA384_FIXED_SIGNING` and `ECDSA_P384_SHA384_ASN1_SIGNING` for signing and `ECDSA_P384_SHA256_ASN1` for verification.

Based on a quick reading of https://developer.amd.com/wp-content/resources/55766.PDF, the signature format is fixed, not ASN.1. It seems like you're saying that the signatures that are generated from the hardware tend to use P-384 + SHA-256. Is that correct? If so, then that would be evidence that we need to add a `ECDSA_P384_SHA256_FIXED` for verification, which seems reasonable to do.

As far as signing goes: Can't you use P-384+SHA-384 for signing, regardless of what the hardware does? Or is there something that only verifies P-384+SHA-256 signatures that forces you to use SHA-256 instead? If so, please explain what that is.

josephlr:
> Based on a quick reading of https://developer.amd.com/wp-content/resources/55766.PDF, the signature format is fixed, not ASN.1. It seems like you're saying that the signatures that are generated from the hardware tend to use P-384 + SHA-256. Is that correct?

Yes, that it correct. The SEV firmware (which is not user controlled) specifically generates two P384 signatures:
  - `CEK` signs the `PEK`
  - `PEK` signs the `PDH`

They don't _have_ to use SHA256 but in practice seem to. You get them with `PDH_CERT_EXPORT` command, and can rotate and resign them with `PDH_GEN` and `PEK_GEN`. 

> If so, then that would be evidence that we need to add a `ECDSA_P384_SHA256_FIXED` for verification, which seems reasonable to do.

Agreed, this PR also adds `ECDSA_P256_SHA384_FIXED` which is _not_ needed.

> As far as signing goes: Can't you use P-384+SHA-384 for signing, regardless of what the hardware does? Or is there something that only verifies P-384+SHA-256 signatures that forces you to use SHA-256 instead? 

There's nothing that prevents the user from using `P-384+SHA-384`, so *ring* does _not_ need to add any of the signing algorithms added in this PR.

The user generates the OCA key and signs the PEK. The key must be P384, but the signature can use either SHA256 or SHA384. This signature is imported using `PEK_CERT_IMPORT` (details of exactly what the platform checks are found in section 5.8.1).

josephlr:
> "Fixing" this problem means (1) convincing AMD to fix the problem and (2) make new CPUs and (3) convincing users to buy a new CPU. This is an enormous amount of energy to overcome and will probably cost millions, maybe billions, of dollars and years of effort. Alternatively, we could just merge this patch.

@npmccallum just for reference, the code implementing AMD SEV is _not_ burned into hardware. It's ARM code running on the AMD Secure Processor. This code can be (and _should_ be) updated just like any other firmware. The current version of AMD firmware uses P384/SHA256 (IIRC), but that implementation could easily be changed in the future with a software update.

briansmith:
OK, it seems reasonable to add the new variant of the verification algorithm. It needs to be added to the test suite like the other algorithms.

PR #802, opened 2019-03-07
Remove the buildpath from debug info in files compiled by gcc and clang.

### LINKED ISSUE #715: Reproducible builds: Buildpath embedded in binaries
hello,

I was looking at ring from a [reproducible builds][0] point of view and I noticed that rust programs that depend on ring can not be built reproducibly with a dynamic build path.

The following screenshot is from [diffoscope][1] output on a binary that includes ring and was built twice in different directories. The binaries are almost identical, except the full path to the pre-generated assembly that is embedded into the binary. I'm not sure where this happens, so a pointer would be very appreciated.

image

[0]: https://reproducible-builds.org/
[1]: https://diffoscope.org/

pietro:
Fixes #715.


build.rs
```diff
@@ -524,6 +524,16 @@ fn obj_path(out_dir: &Path, src: &Path, obj_ext: &str) -> PathBuf {
     out_path
 }
 
+#[allow(box_pointers)]
```
briansmith: Let's remove the use of box pointers. We shouldn't need it for this.

build.rs
```diff
@@ -524,6 +524,16 @@ fn obj_path(out_dir: &Path, src: &Path, obj_ext: &str) -> PathBuf {
...
+fn remap_debug_prefix() -> &'static str {
+    use std::env;
+    let cwd_path = env::current_dir().expect("Couldn't get current directory");
+    let mut flag = String::from("-fdebug-prefix-map=");
+    flag.push_str(cwd_path.to_str().expect("Invalid path"));
+    flag.push_str("=.");
+    Box::leak(flag.into_boxed_str())
```
briansmith: I don't understand why `Box::leak` is required here. I suggest we just return a `String`.

build.rs
```diff
@@ -570,6 +580,7 @@ fn cc(
 
     if target.env() != "msvc" {
         let _ = c.define("_XOPEN_SOURCE", Some("700"));
+        let _ = c.flag(remap_debug_prefix());
```
briansmith: Rather than call `remap_debug_prefix()` for each compilation, let's call it once before any compilation, cache it, and pass it as an argument to `cc()` . In particular, from my experience writing makefiles it's best to avoid evaluating `env::current_dir()` more than once.

build.rs
briansmith: Please add a comment here explaining that this strips the build path for the purpose of reproducible builds.

pietro: I removed the box pointers and I added a comment but I'm not super sure about the wording.

build.rs
pietro: Done.

build.rs
pietro: I'm calling it once in `compile()`. But I just realized that's still called once per file. Sorry about that. I'll do another iteration on this PR.

briansmith:
@pietro Sorry, I didn't notice that you updated this PR. I'll be changing the build system soon and I'll look into this when I do so.

briansmith:
I changed the base branch to "main" from "master" so I can remove "master" today.

build.rs
```diff
@@ -538,11 +543,22 @@ fn obj_path(out_dir: &Path, src: &Path, obj_ext: &str) -> PathBuf {
...
+    flag.push_str("=.");
+    flag
+}
+
 fn cc(
     file: &Path,
     ext: &str,
     target: &Target,
     warnings_are_errors: bool,
+    extra_flags: Vec<String>,
```
briansmith, line 561: I understand you did this at my request so that we wouldn't duplicate the work in `remap_debug_prefix()` per source file. I realize that this complication is really caused by the structure of the build.rs code. I have refactored it in PR #1697 so that changes like the one in this PR don't have to try to solve the issues themselves.

build.rs
```diff
@@ -538,11 +543,22 @@ fn obj_path(out_dir: &Path, src: &Path, obj_ext: &str) -> PathBuf {
     out_path
 }
 
+// This option will strip the build path from C and assembly files compiled by gcc and clang.
+fn remap_debug_prefix() -> String {
+    use std::env;
+    let cwd_path = env::current_dir().expect("Couldn't get current directory");
+    let mut flag = String::from("-fdebug-prefix-map=");
+    flag.push_str(cwd_path.to_str().expect("Invalid path"));
```
briansmith, line 551: Instead we need to use `OsString` here; see @NobodyXu's work in PR #1580.

PR #805, opened 2019-03-20
Improve the usability of the AEAD API

briansmith:


elichai:
Looks good, right now the API is built in a way that requires you to clone the data in `Nonce` (i.e. this isn't possible: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3770f25c7a92539e48a50c01a34c2be9 )

briansmith:
I pushed an updated version of this to master. Closing this.

briansmith:
> Looks good, right now the API is built in a way that requires you to clone the data in `Nonce` (i.e. this isn't possible

The API requires the `Nonce` to be *moved*, not cloned. I don't think there is a significant cost to this. But, if this isn't what you want, please file a separate issue describing the problem in more detail. Thanks!

briansmith:
I updated this PR with the current WIP.

PR #807, opened 2019-03-30
Derive clone for hmac key

### LINKED ISSUE #806: Should hmac::SigningKey (also) implement Clone?
I see that `hmac::SigningKey`'s only field is of type `SigningContext` and it implements `Clone`. Is there a reason `hmac::SigningKey` does not implement `Clone`?

I can open a PR to add this if needed.

dvic:
This closes #806.

### DIFF
--- a/src/hmac.rs
+++ b/src/hmac.rs
@@ -131,6 +131,7 @@ impl AsRef<[u8]> for Tag {
 }
 
 /// A key to use for HMAC signing.
+#[derive(Clone)]
 pub struct Key {
     ctx_prototype: Context,
 }
--- a/tests/hmac_tests.rs
+++ b/tests/hmac_tests.rs
@@ -69,32 +69,36 @@ fn hmac_test_case_inner(
     output: &[u8],
     is_ok: bool,
 ) -> Result<(), error::Unspecified> {
-    let key = hmac::Key::new(digest_alg, key_value);
-
-    // One-shot API.
-    {
-        let signature = hmac::sign(&key, input);
-        assert_eq!(is_ok, signature.as_ref() == output);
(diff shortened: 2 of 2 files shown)


src/aead/aes.rs
```diff
@@ -19,6 +19,7 @@ use super::{
 use crate::{bits::BitLength, cpu, endian::*, error, polyfill};
 use libc::size_t;
 
+#[derive(Clone)]
```
briansmith: Please modify aead_test.rs to verify that the clone of a key produces the same results as the original key.

src/hmac.rs
```diff
@@ -167,6 +167,7 @@ pub type Signature = Tag;
 pub struct Tag(digest::Digest);
 
 /// A key to use for HMAC signing.
+#[derive(Clone)]
```
briansmith: Similarly, please modify hmac_test.rs to verify that the clone of a `Signing` (resp. `VerificationKey`) gives the same results as the original

dvic:
I have updated `test_aead` and `hmac_test_case_inner` and added the commit message suffix. Let me know if anything else needs to be done 👍

src/aead/aes.rs
dvic: Done.

src/hmac.rs
dvic: Done.

dvic:
🤦‍♂️ You are right, I could have guessed that (you have a comment about this in the code, see below). I'll remove the AEAD changes (HMAC was indeed what I initially requested) 👍. 

image

dvic:
I have updated this PR for the latest changes related to merging into one `hmac::Key` and to remove the AEAD changes, as you suggested.

briansmith:
Thank you!

PR #812, opened 2019-04-26
Provide a verify function in hmac context

HeroicKatora:
Second try at a consistent api for hmac verification in a multi-step
context. The verification steps never expose the correct tag to the user
to avoid misuse or leakage.

Closes: #615

Second iteration hmac context verification, open to more feedback.

### DIFF
--- a/src/hmac.rs
+++ b/src/hmac.rs
@@ -96,11 +96,11 @@
 //! // The receiver (somehow!) knows the key value, and uses it to verify the
 //! // integrity of the message.
 //! let v_key = hmac::Key::new(&digest::SHA384, key_value.as_ref());
-//! let mut msg = Vec::<u8>::new();
+//! let mut v_ctx = hmac::Context::with_key(&v_key);
 //! for part in &parts {
-//!     msg.extend(part.as_bytes());
+//!     v_ctx.update(part.as_bytes());
 //! }
-//! hmac::verify(&v_key, &msg.as_ref(), tag.as_ref())?;
+//! v_ctx.verify(tag.as_ref())?;
 //!
 //! # Ok::<(), ring::error::Unspecified>(())
 //! ```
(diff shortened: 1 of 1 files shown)


HeroicKatora:
I've removed the generic parameter idea from the design. It's completely orthogonal to the main issue addressed and not necessary.

HeroicKatora:
AppVeyor failed due to `spurious network error` on one build, other succeeded.

briansmith:
@HeroicKatora Thanks. I am interested in making this improvement soon. For your needs, do you really need the I-U-F interface or would an API that takes a `&[&[u8]]` as input or a `impl Iterator<Item=[u8]>` as input suffice? Obviously the I-U-F interface is more general but I think most use cases are better covered by one of the latter two kinds of API.

HeroicKatora:
I'm not sure, actually. What do you mean by `impl Iterator<Item=[u8]>`? Since the `Item` needs to be `Sized` it would likely be `impl Iterator<Item=&'_[u8]>`. Implementing [`Extend`](https://doc.rust-lang.org/std/iter/trait.Extend.html) could make I-U-F more usable. I'm not sure if similar is possible for a one-step solution, in particular [`FromIterator`](https://doc.rust-lang.org/std/iter/trait.FromIterator.html) does not take a seed value–i.e. the signing key.

briansmith:
I'm thinking of something like this:

```rust
    fn verify_parts<'a, T: AsRef<[u8]>>(data: impl IntoIterator<Item=T>, tag: &[u8]) {
        let ctx = ...;
        for data in data.into_iter() {
            ctx.update(part.as_ref());
        }
        ctx.finish.check(tag)
    }
```

Really what I'm asking is which use case you are trying to solve:
1. The data to HMAC isn't contiguous in memory, but it is all available at once.
2. The data to HMAC isn't all available at once.

This kind of `verify_parts` function would solve #1 but not #2. Your PR solves #2 and is general enough to be used to solve #1.

We used to have an interface like the one you are proposing, but I removed it based on feedback that it is too easy to misuse. So, if something like `verify_parts` solves the issues you are having, maybe it is worth shipping `verify_parts` first and avoiding the I-U-F interface?

HeroicKatora:
> We used to have an interface like the one you are proposing, but I removed it based on feedback that it is too easy to misuse.

Can you give a pointer to commit/PR with the reasoning? That sounds like something I could learn and apply to general interface design in the future.

Overall, this is a very good characterization. The solution of `1.` is general enough *iff* such an interface is also enough to create signed tags, I think.

briansmith:
> Can you give a pointer to commit/PR with the reasoning? That sounds like something I could learn and apply to general interface design in the future.

Unfortunately, the only note I have is:
https://github.com/briansmith/ring/commit/12596a32f9e8cb15dbe7f84d062a9a8c6207a7e8#diff-c76d1ccc0023b9b98f395f6c6da93b16.

The basic idea is that one should avoid having streaming APIs for AEADs and MACs so the application isn't as tempted to process the not-yet-verified input.

> The solution of 1. is general enough iff such an interface is also enough to create signed tags, I think.

I think you are implying that maybe I-U-F isn't needed for signing either. I think that's mostly true and maybe that should motivate us to have a `sign_parts` counterpart to `verify_parts`.

OTOH, in the case of signing, I know that HKDF benefits from having an I-U-F interface available. So I have doubts we can remove the I-U-F interface for signing even if we could avoid adding an I-U-F interface for verification.

HeroicKatora:
That all seems reasonable. So then back to the drawing board, i.e. #615 . The best place to add the functions taking `impl Iterator<Item=&[u8]>` would maybe be `Key` itself instead of at the module level, does that make sense?

Issue #412, opened 2017-01-13
Add AES-GCM-SIV Nonce-Misuse-Resistant AEAD


briansmith:
The draft spec is at https://tools.ietf.org/html/draft-irtf-cfrg-gcmsiv. One version of the paper describing one version of the construct (which might not be what's in the draft) is at https://eprint.iacr.org/2015/102.pdf. Note that the draft is almost definitely going to be updated in a significant way as foretold in https://www.ietf.org/mail-archive/web/cfrg/current/msg08853.html.

BoringSSL and Shay Gueron appear to be working on a highly-optimized implementation. Hopefully we can just take the BoringSSL asm code as-is, so we'd just have to write the upper layer in Rust. With this in mind, we should land changes prototyped in https://github.com/briansmith/ring/pull/369 first.

This would be the first purportedly-nonce-misuse-resistant AEAD in *ring*. We may also need to do additional work, e.g. defining a trait for nonce-misuse-resistant AEADs that is implemented by `AES_128_GCM_SIV` and `AES_256_GCM_SIV` but not by other AEADs. In order for that to work, we may also need to add an AEAD algorithm trait and hide the struct that's currently `ring::aead::Algorithm`.

We should do a literature review of papers analyzing the AES-GCM-SIV construct, including in particular whatever new paper Shay will probably write to describe the final version.

briansmith:
See also [RIV for Robust Authenticated Encryption](http://link.springer.com/chapter/10.1007/978-3-662-52993-5_2).

briansmith:
See also [Stronger Security Variants of GCM-SIV](https://eprint.iacr.org/2016/853.pdf).

briansmith:
See these threads:
* https://mailarchive.ietf.org/arch/msg/cfrg/k2mpWgod4mbdOxsvN6EtXHb0BAg
* https://www.ietf.org/mail-archive/web/cfrg/current/msg08894.html

Also, I'm curious if the idea from https://eprint.iacr.org/2016/564.pdf can be used in some way here, and/or if there are ways of strengthening the nonce construction.

briansmith:
See https://boringssl.googlesource.com/boringssl/+/41c10e2b5f37edce8b9f292f7f3bacb7e30e25c4

misnomer1:
Hey,
I am interested in taking up adding AES-GCM-SIV to ring. I have a question though.

[...]

PR #816, opened 2019-05-06
Adding AES_GCM_SIV ASM (x86_64) code to Ring

misnomer1:
This code depends on AES_GCM_SIV ASM implementation located at crypto/cipher_extra/asm/aes128gcmsiv-x86_64.pl. 

Note: I will send out non-asm version of aes_gcm_siv in another pull request


briansmith:
Please include the non-ASM implementation in this PR too.

I don't think it makes sense to use the same API for AES-GCM-SIV and regular AEADs, especially regarding the nonce. How does your application generate the nonce for AES-GCM-SIV now?

misnomer1:
We discussed the same here: https://github.com/briansmith/ring/issues/412
Let me know if this is not what you are talking about

The non-asm version is in process but it is hard to make it work, I will have to spend more time on that.

briansmith:
> The non-asm version is in process but it is hard to make it work, I will have to spend more time on that.

OK, no problem.

I'm going to close this for now. Please open up a new PR when you have it working for all platforms. Thanks!

Issue #532, opened 2017-05-06
PPC32 acceleration (for certain cryptographic primitives)


tarcieri:
Foreword: I can understand reluctance to add support for "exotic" platforms, and I'd personally class PPC32 as an "exotic" platform.. I agree supporting a wide range of legacy platforms has complexified and harmed the agility of e.g. OpenSSL in the past. I wholly admit I have purely selfish aims here, and perhaps if I didn't I would be against this sort of thing.

PPC32 is presently used in, as far as I can tell, a reasonable amount of deployed network gear (something confirmed to me by repeatedly asking people "why should we care about PPC32 anymore?"). My personal interest is in the Thales nShield platform.

I can't claim to have done a thorough evaluation of this, but from what I can tell ample Perl scripts exist in OpenSSL which generate PPC32 ASM implementations, but support for these architectures seems to have been dropped from BoringSSL.

Is it worth supporting optimized assembly implementations PPC32? Having wrangled some of OpenSSL's assembly into the general *ring* hash function framework and tested it on my particular PPC32 platform, I noted a 5X performance advantage over https://github.com/DaGenix/rust-crypto/blob/master/src/sha2.rs on the same platform. I have done no more testing than that.

[Rust-proper classes PPC32 as a "Tier 2" platform](https://forge.rust-lang.org/platform-support.html). This means test failures on these platforms are blocking, and Mozilla performs releases for PPC32. That said, actually exploring running Rust on a PPC32 test box (G4 Mac Mini) I quickly encountered [some catastrophic cargo failures](https://github.com/rust-lang/cargo/issues/3852). This was fixed by changing the Rust build system to target `powerpc` instead of `power4` (https://github.com/rust-lang/rust/pull/41080). Even after this was corrected, I'm still experiencing [problems building Rust code natively on PPC32](https://github.com/rust-lang/rust/issues/41253). Cross-compiling works fine.

The specific algorithm I'd like to have PPC32 acceleration for at the moment is SHA-512. AES acceleration might be nice as well. (Note: My particular platform provides hardware acceleration for both of these primitives, however it requires using a batch-oriented asynchronous message protocol and pipelining operations for efficiency. I'd like to avoid that in the short term)

I'd be interested in immediately contributing PPC32 acceleration for SHA-512, which I think I mostly have ready.

briansmith:
I don't have a convenient way of testing PowerPC code. If we accept contributions for supporting PowerPC then we need to add a note about the lack of testing to the README.

[...]

PR #819, opened 2019-05-08
wip: ppc64/ppc support

q66:
I decided to do my own take at ppc64 (and maybe 32, but comments suggest that might not be practical) support...

**Current status/TODO:**

- [x] complete testsuite pass on big and little endian ppc64
- [x] figure out why `ChaCha20_ctr32_vsx` segfaults [**edit:** it's because ring does not check length for zero and the vsx assembly assumes non-zero, fixed by checking separately in `GFp_ChaCha20_ctr32`]
- [x] add decision logic in `build.rs` to support big endian ppc64 (ELFv2 and ELFv1) and 32-bit ppc
- [x] clean up build hacks
- [x] import AES ppc assembly bits and integrate them
- [x] add CI bits [powerpc64le only, there is no powerpc64 toolchain in enivornment as ancient as ubuntu 14.04]
- [ ] minimize the assembly code (remove AES decryption and so on)
- [ ] rename remaining functions to avoid conflicts
- [ ] ppc32 support? `ecp_nistz256` missing (no portable impl, no ppc32 assembly)

The big endian byteswapping hacks in `chacha.rs` and `poly1305.rs` are most likely wrong. I don't know what the right fix is though.


q66:
Looks like the crash is when `len` for `GFp_ChaCha20_ctr32` is zero.

q66:
Build now supports all perlasm formats for ppc64 correctly. Still need to test though (and maybe fill in missing code).

q66:
The `ecp_nistz256` funcs don't have any 32-bit ppc assembly *or* generic portable implementations in OpenSSL, which might be a problem.

q66:
Okay, big endian builds but quite a few tests panic because of incorrect results.

tarcieri:
See the previous issue I opened about PPC32 support: https://github.com/briansmith/ring/issues/532

Most PPC32s cannot do constant time multiplication, and provide a "fast" multiply by 0 and 1. Together with their generally low clock speed, this can be leveraged as an exploitable timing sidechannel.

q66:
I guess the options are either to leave the 32-bit support out, or conditionally leave out the specific parts for ppc32 then. The latter seemed like it would take a fair amount of changes.

q66:
Okay, it's not just chacha20, in the main module these are failing:

```
failures:
    aead::chacha::tests::chacha20_tests
    aead::poly1305::tests::test_poly1305
```

No clue what to make of it, though.

q66:
The newly added commit "fixes" testsuite on ppc64be. I kinda doubt that this solution is correct or anything, but it's where it fails. A better (or well, actual) solution would be much appreciated.

crypto/fipsmodule/ec/ecp_nistz256_table.c
```diff
@@ -0,0 +1,9542 @@
+/*
```
briansmith: This table is already in *ring* and we won't need this file.

The PPC select functions are using the scatter/gather technique, which we don't need in *ring*. It will be better to just remove the "scatter-gather subroutines" section of the ecp_nistz256-ppc64.pl file and then remove ecp_nistz256_table.c too.

include/GFp/base.h
```diff
@@ -85,11 +85,17 @@
 #elif defined(__mips__) && defined(__LP64__)
 #define OPENSSL_64_BIT
 #define OPENSSL_MIPS64
+#elif defined(__powerpc64__)
```
briansmith: Please split the changes to this file out into the "base PR."

pregenerate_asm/Cargo.toml
```diff
@@ -10,3 +10,4 @@ path = "../build.rs"
 # Keep this in sync with `[build-dependencies]` in ../Cargo.toml.
 [dependencies]
 cc = "1.0.26"
+tempfile = "3.0"
```
briansmith: Please add a comment explaining why this dependency was added. If it is only needed for PPC then the dependency should be limited to PPC.

src/aead/chacha.rs
```diff
@@ -122,12 +122,30 @@ impl Key {
...
+                u32::from_le(key[6]), u32::from_le(key[7])
+            ];
+            let ctr: [u32; 4] = [
+                u32::from_le(ctr[0]), u32::from_le(ctr[1]),
+                u32::from_le(ctr[2]), u32::from_le(ctr[3])
+            ];
+            GFp_ChaCha20_ctr32(output, input, in_out_len, &key, &ctr);
+        } else {
+            GFp_ChaCha20_ctr32(output, input, in_out_len, key, ctr);
+        }
```
briansmith: Please split all changes to Rust code needed to support big-endian to a separate "big endian" PR. I expect this will be used by MIPS big endian too.

src/aead/poly1305.rs
```diff
@@ -58,7 +58,18 @@ impl Context {
...
+            let nonce: &[u32; 4] = unsafe { core::mem::transmute(key_and_nonce[1].as_ref()) };
+            let nonce: [u32; 4] = [
+                u32::from_le(nonce[0]), u32::from_le(nonce[1]),
+                u32::from_le(nonce[2]), u32::from_le(nonce[3])
+            ];
+            let nonce: &[u8; 16] = unsafe { core::mem::transmute(&nonce) };
+            Nonce(Block::from(nonce))
+        } else {
+            Nonce(key_and_nonce[1].clone())
+        };
```
briansmith: This should go in the "big endian" PR too.

src/ec/suite_b/ops.rs
```diff
@@ -64,6 +64,28 @@ macro_rules! limbs {
...
+            u32::to_be($limb_6),
+            u32::to_be($limb_7),
+            u32::to_be($limb_8),
+            u32::to_be($limb_9),
+            u32::to_be($limb_a),
+            u32::to_be($limb_b),
+        ]
+    };
+}
+
```
briansmith: Please move this to the "big endian" PR.

src/ec/suite_b/ops.rs
```diff
@@ -80,6 +102,22 @@ macro_rules! limbs {
...
+            ((($limb_1 | 0u64) << 32) | $limb_0).to_be(),
+            ((($limb_3 | 0u64) << 32) | $limb_2).to_be(),
+            ((($limb_5 | 0u64) << 32) | $limb_4).to_be(),
+            ((($limb_7 | 0u64) << 32) | $limb_6).to_be(),
+            ((($limb_9 | 0u64) << 32) | $limb_8).to_be(),
+            ((($limb_b | 0u64) << 32) | $limb_a).to_be(),
+        ]
+    };
+}
+
```
briansmith: Ditto.

q66:
Hmm, as far as I can see, the only thing that *does* have a portable fallback in the tree is AES encryption. As for the assembly code, I think it should be safe to use, it passes OpenSSL's testsuite, `ring`'s testsuite, and is used on `ppc64le` distros all over the place, without any issues.

What about splitting it into three, like this:

1) only the code required to get little endian ppc64 running
2) followed by another PR that gets big endian working
3) followed by another PR that adds VPAES and HWAES encryption in (split into multiple commits so that one adds the support and one minimizes the asm)
4) maybe followed later by something for 32-bit ppc...

Still not sure about 32-bit ppc, I have no way to test that (only modern POWER hardware here) and AFAIK it's a pain to get a ppc32 rust toolchain running in the first place...

q66:
I got a 32-bit ppc environment available now (I compiled my distribution for ppc32, then made up a chroot of that in a 64-bit big endian environment, it works fine, including Rust) so I can try handling that as well. Looks like BoringSSL has some code that will possibly enable ppc32 to work.

edelsohn:
You can request a VM or direct Jenkins CI account at OSUOSL for CI access.
[OSUOSL Power Development Hosting](https://osuosl.org/services/powerdev/).
You can list me (David Edelsohn) as the IBM Advocate.

johnjmar:
Building using ring v0.14.6 (/home/ubuntu/ring_q66) in a Power 9 (ppc64le)  system is yielding several 
`error: trait objects without an explicit `dyn` are deprecated`

q66:
i don't really care for working on this anymore, sorry

anyone who feels like picking up on it themselves can take my changes, but i won't be putting in any more active effort

PR #822, opened 2019-05-12
Support building with Bazel

akhilles:
This doesn't affect Cargo builds. Some build tools don't pass CARGO_PKG_NAME and DEBUG environment variables and I think it's reasonable to have some default behavior instead of panicking in the build script.

I realize non-Cargo build systems aren't supported but this is a relatively small change.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -251,14 +251,10 @@ const MSVC_OBJ_OPT: &str = "/Fo";
 const MSVC_OBJ_EXT: &str = "obj";
 
 fn main() {
-    if let Ok(package_name) = std::env::var("CARGO_PKG_NAME") {
-        if package_name == "ring" {
-            ring_build_rs_main();
-            return;
-        }
-    }
-
-    pregenerate_asm_main();
+    match std::env::var("CARGO_PKG_NAME") {
+        Ok(ref var) if var == "pregenerate_asm" => pregenerate_asm_main(),
+        _ => ring_build_rs_main(),
+    };
 }
 
 fn ring_build_rs_main() {
@@ -280,7 +276,11 @@ fn ring_build_rs_main() {
         ("o", "-o")
     };
 
-    let is_debug = env::var("DEBUG").unwrap() != "false";
+    let is_debug = match env::var("DEBUG") {
+        Ok(ref var) if var == "true" => true,
+        _ => false
+    };
+
     let target = Target {
         arch,
         os,


briansmith:
I'm generally happy to support Bezel. However, realistically we should have CI coverage if our goal is to support it. What can we do in CI to test that the Bazel build works?

josephlr:
> What can we do in CI to test that the Bazel build works?

We would need to have (or generate) a `BUILD` file for *ring*, and then just make sure `bazel build` succeeds. I don't think it would be crucial to do `bazel test`, unless we want to fully support `bazel`.

@akhilles I'm interested in adding blaze/bazel support to *ring*. Do you have a `BUILD` file that works for you on current master?

@briansmith this might be best addressed in a different issue, if you want to merge this and continue other discussion elsewhere.

akhilles:
@josephlr I do have a working `BUILD` file. I generated it using [cargo raze](https://github.com/google/cargo-raze) and then patched it to actually work (`raze` struggles with build scripts because there's a lot of environment variables that Cargo injects). The purpose of this pull request was to improve compatibility with `raze`.

However, this might not be an appropriate long-term solution. See https://github.com/google/cargo-raze/issues/97#issuecomment-488887288. It seems like Rust build scripts and Bazel just aren't that compatible.

I see the only proper way to support Bazel is to port all of _ring_'s build script functionality into a Bazel `cc_library` target and then add that target under `additional_deps` when using `raze`.

josephlr:
> I see the only proper way to support Bazel is to port all of ring's build script functionality into a Bazel cc_library target and then add that target under additional_deps when using raze.

So the main issue with supporting a `BUILD` file seems to be the complex way the assembly and C files are compiled (mostly due to OpenSSL itself being quite complex). If *ring* provided a `BUILD` file, would that just work with `cargo raze`? 

I took a stab at making this work myself, and I was even able to get *ring* linking and running, provided I didn't use any assembly-backed functionality. So it seems like if we could get some `cc_library` rules for the `.c` files and some other sort of genrule for the assembly file, it should be fairly straightforward to make *ring* fully build.

josephlr:
@akhilles and update on this, I got a [`BUILD` file working on `x86_64-unknown-linux-gnu`](https://gist.github.com/josephlr/ab94c1c3060be0b5c13c86b66e5a35dc) using *ring*'s `pregenerate_asm` tool. The last step would be to write a Bazel rule that does the generation, or just have *ring* commit the generated `*.S` files.

josephlr:
A further update on Bazel, I now have everything working!!

[`BUILD` file](https://gist.github.com/josephlr/520bd9dae579408e3466b8abd36173dd) and [`perlasm.bzl` file](https://gist.github.com/josephlr/ec1d64fd17d631da7dd9d7ef3fb245c4) now generate, build, and test everything for *ring*.

Note that this file is also a really good way to visualize *ring*'s build dependencies, even it you aren't using Bazel to build your code.

[Full output of `bazel test --test_output=all //ring:all`](https://gist.github.com/josephlr/39a723baa1f36346d1b904eaa8e98479)

akhilles:
I wish there's some way for `cargo raze` to detect that a `BUILD` file exists in the repository and to not override it. Currently, the generated `BUILD` file has to be manually patched right?

Also, `bazel query` is really cool since you can check exactly which source files are compiled and even the system/toolchain dependencies.

briansmith:
I'm planning to make changes to the build system that will make these changes obsolete. If you are interested in building *ring* with Bazel, please email me: brian@briansmith.org.

PR #826, opened 2019-05-20
Allow the verification of an ECDSA signature given just a digest

turbolent:
It seems like currently is only possible to verify an ECDSA signature given a message, the digest is produced internally.

Allow users to verify an ECDSA signature given an existing digest.

### DIFF
--- a/src/ec/suite_b/ecdsa/digest_scalar.rs
+++ b/src/ec/suite_b/ecdsa/digest_scalar.rs
@@ -51,8 +51,7 @@ pub fn digest_scalar(ops: &ScalarOps, msg: digest::Digest) -> Scalar {
     digest_scalar_(ops, msg.as_ref())
 }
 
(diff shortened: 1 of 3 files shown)


briansmith:
What is the problem that you are trying to solve by adding this API?

turbolent:
I'd like to verify signatures just given the hash/digest, not the plain unhashed message. This functionality is available in other libraries, e.g. see https://golang.org/pkg/crypto/ecdsa/#Verify

briansmith:
The *ring* API is designed to work consistently between EdDSA and ECDSA and RSA. Normal EdDSA doesn't allow verification or signing of messages with just the hash.

If you could be more specific about your application, in particular why the *ring* API can't work for it (vs. just preference), then I might be able to help you come up with some solution that works.

briansmith:
Closing this due to lack of interest.

PR #831, opened 2019-06-04
Generating a Nonce from random

elichai:
I think it's pretty self explanatory.
right now whomever uses the library needs to do this by themselves and they also might use a non bad randomness source, this should help that.

### DIFF
--- a/src/aead/nonce.rs
+++ b/src/aead/nonce.rs
@@ -24,7 +24,7 @@
 //!    is incremented, the current value is returned.
 
 use super::Block;
-use crate::{endian::*, error, polyfill::convert::*};
+use crate::{endian::*, error, polyfill::convert::*, rand};
 use core::marker::PhantomData;
 
 /// A nonce for a single AEAD opening or sealing operation.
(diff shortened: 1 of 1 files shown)


ctz:
I'm not sure this is a good thing to make easy? Fully random 96-bit nonces don't have a good security margin after a realistic number of ciphertexts with a given key.

elichai:
What are you suggesting? Using Unix timestamps?
And 2^(96/2) (birthday) isn't a small number of ciphertexts

elichai:
@briansmith I'm not sure why the tests fail, but it looks like something unrelated

src/aead/nonce.rs
```diff
@@ -52,6 +52,14 @@ impl Nonce {
     pub fn assume_unique_for_key(value: [u8; NONCE_LEN]) -> Self {
         Self(value)
     }
+
+    /// Generate a `Nonce` with a random value generated from `rng`.
+    #[inline]
+    pub fn generate(rng: &rand::SecureRandom) -> Result<Self, error::Unspecified> {
+        let mut nonce = [0u8; 12];
```
briansmith: Instead of 12, you could use `NONE_LEN`.

src/aead/nonce.rs
```diff
@@ -52,6 +52,14 @@ impl Nonce {
...
+        rng.fill(&mut nonce)?;
+        Ok(Self(nonce))
```
briansmith: Instead of this, you could use `Ok(Self::assume_unique_for_key(nonce))`.

briansmith:
Thanks for the PR @elichai.

The current API already allows one to use random nonces using the same code you propose for the new constructor function, modified as I suggested above. I intend to keep supporting such a possibility going forward.

On the other hand, I don't want to *encourage* the use of random nonces by providing an explicit API for them, like @ctz mentioned.

elichai:
@briansmith so how do you think it's best to construct the nonce?

PR #835, opened 2019-06-13
Fix bare_trait_objects warning on nightly

josephlr:
All this does is replace `&Trait` with `&dyn Trait`.


josephlr:
@briansmith it looks like [one of the failures](https://travis-ci.org/briansmith/ring/jobs/545136132) is a false-positive, but other than that, everything on Travis is back to green!!

briansmith:
Thanks for alerting me to the problem! I think this is doing the same thing that commit fdc558da0f109f8a9b51664e19c1e26a691a7ea8 does so I'm closing it in favor of that commit.

PR #836, opened 2019-06-13
Simplify core vs std usage

josephlr:
Right now, `src/lib.rs` defines *ring* to be a `no_std` crate conditionally based on the features and CPU architecture. This is sub-optimal, as it requires a lot of `cfg` attributes to be kept in sync. Both [`rand`](https://crates.io/crates/rand) and [`rdrand`](https://crates.io/crates/rdrand) have dealt with this kind of problem, and they use the following solution.
  - The create is unconditionally `#![no_std]`
  - If `std` is needed, it can be brought in with `extern crate std`
  - There are no bare `use std` or `use core` statements (as they are never necessary)

This is what I did in this PR. Some advantages to this PR:
  - It is now very easy to see what ring is depending on from `std`. This helps with:
    - https://github.com/briansmith/ring/issues/744
    - Eventually making ring's tests work with `no_std`
  - After `extern crate alloc` lands in the next stable Rust release, most of the `use std::` statements can be turned into `use alloc::`.
  - It will make it easier to eventually split the `use_heap` feature into two features
    - `std`: Enables getting randomness from files and `std::error::Error` implementations
    - `alloc`: Enables operations that need dynamic memory allocation (RSA signing, etc...)
  - Helps with the porting work in https://github.com/briansmith/ring/pull/787 and https://github.com/briansmith/ring/issues/775

### DIFF
--- a/src/aead/chacha.rs
+++ b/src/aead/chacha.rs
@@ -18,7 +18,6 @@ use super::{
     Block, BLOCK_LEN,
 };
 use crate::{endian::*, polyfill::convert::*};
-use core;
 use libc::size_t;
 
 #[repr(C)]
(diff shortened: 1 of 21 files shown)


briansmith:
Thanks!

PR #839, opened 2019-06-13
Use libc::SYS_getrandom

josephlr:
This has been in `libc` for a while, so we don't need to increment the
`libc` version in Cargo.toml. Note that we cannot use the `getrandom()`
function from `libc`, as it isn't available on Android.

Originally part of #744

### DIFF
--- a/src/c.rs
+++ b/src/c.rs
@@ -28,6 +28,3 @@ pub(crate) type uint = libc::c_uint;
     any(target_arch = "aarch64", target_arch = "arm")
 ))]
 pub(crate) type ulong = libc::c_ulong;
-
-#[cfg(any(target_os = "android", target_os = "linux"))]
-pub(crate) type long = libc::c_long;
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -115,26 +115,11 @@ use crate::sealed;
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
 mod sysrand_chunk {
-    use crate::{c, error};
+    use crate::error;
 
     #[inline]
     pub fn chunk(dest: &mut [u8]) -> Result<usize, error::Unspecified> {
-        // See `SYS_getrandom` in #include <sys/syscall.h>.
-
-        #[cfg(target_arch = "aarch64")]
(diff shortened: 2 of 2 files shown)


briansmith:
Can we just fix https://github.com/rust-lang/libc/issues/659#issuecomment-495610073 and then use `libc::getrandom()` instead? I'd like to remove the direct use of syscall by *ring* completely.

josephlr:
> Can we just fix rust-lang/libc#659 (comment) and then use libc::getrandom() instead? I'd like to remove the direct use of syscall by ring completely.

One particular issue is that `getrandom` is only supported on fairly new glibc/musl versions ([despite the syscall being around for awhile](https://lwn.net/Articles/711013/)), and isn't supported on Android. If you're OK with:
  - Requiring a recent glibc/musl
  - Potentially going back to using the raw syscall if *ring* ever supports Android

Then I think going with https://github.com/rust-lang/libc/pull/1399 is the best bet.

josephlr:
> Potentially going back to using the raw syscall if ring ever supports Android

@briansmith disregard the above, I forgot that *ring* currently supports Android. So it looks like using `libc::getrandom` is not an option, so we will have to stick with the raw syscall (i.e. this PR). 🙁

josephlr:
@briansmith this is ready to merge. https://github.com/rust-lang/libc/pull/1399 was merged, but we can't use it because the bindings are only available for Linux, not Android.

briansmith:
What's the specific goal that you are trying to achieve here? Is there something wrong with the constants that *ring* is currently using? Are you hoping to do another Linux port that isn't supported by *ring* yet? Or something else?

josephlr:
> What's the specific goal that you are trying to achieve here?

- Making *ring* easier to port (I was taking a shot at a writeup for #744). Also see: #562 and #819 
  - Obviously many more steps are needed, but it's little things like this that increase complexity and the cost of a port.
- Ease of code review (I had to go check that these numbers matched the one's in `libc`).

josephlr:
Closing for similar reasons to #856, this is trying to bring over chunks from the `getrandom` create which is the wrong approch.

Issue #151, opened 2016-03-12
Remove all pointer casts in C code


briansmith:
The current shortlist, which I got after grepping for “*)” and doing a bit of filtering, is below. All of them seem pretty easy to eliminate. However, it would be better to find a tool that finds such casts that is smarter than grep, that we can run as a pre-commit hook and that we can run during builds to verify we don't add any new casts, especially when we merge changes from other forks.

I will take a stab at eliminating these soon.

```
crypto/bn/exponentiation.c(347):  BN_ULONG *table = (BN_ULONG *) buf;
crypto/bn/exponentiation.c(364):  volatile BN_ULONG *table = (volatile BN_ULONG *)buf;
crypto/bn/exponentiation.c(444):  ((unsigned char *)(x_) +               \
crypto/bn/exponentiation.c(552):  tmp.d = (BN_ULONG *)(powerbuf + sizeof(m->d[0]) * top * numPowers);
crypto/bn/exponentiation.c(664):      const uint8_t *p_bytes = (const uint8_t *)p->d;
crypto/bn/exponentiation.c(686):        wvalue = *(const uint16_t *) (p_bytes + (first_bit >> 3));
crypto/bn/internal.h(196):    (BN_ULONG *)x, sizeof(x) / sizeof(BN_ULONG),        \
crypto/bn/montgomery.c(338):    nrp = (BN_ULONG *)(((uintptr_t)rp & ~m) | ((uintptr_t)ap & m));
crypto/bn/random.c(294):                                        (const uint8_t *)attempt,
crypto/bn/random.c(295):                                        sizeof(attempt), (const uint8_t *)done,
crypto/bn/rsaz_exp.c(218):  const uint8_t *p_str = (const uint8_t *)exponent;
crypto/bn/rsaz_exp.c(230):      wvalue = *((const unsigned short*)&p_str[index / 8]);
crypto/ec/p256-64.c(82):  out[0] = *((const u64 *)&in[0]);
crypto/ec/p256-64.c(83):  out[1] = *((const u64 *)&in[8]);
crypto/ec/p256-64.c(84):  out[2] = *((const u64 *)&in[16]);
crypto/ec/p256-64.c(85):  out[3] = *((const u64 *)&in[24]);
crypto/ec/p256-64.c(91):  *((u64 *)&out[0]) = in[0];
crypto/ec/p256-64.c(92):  *((u64 *)&out[8]) = in[1];
[...]

Issue #101, opened 2016-01-28
Remove integer arithmetic on values converted from pointers


briansmith:
We have done some work already to remove some uses of `uintptr_t`, such as  bccbeb28d737fadbed27a836957d6c821d5670bd. We should remove all of them. If, for some reason, we need to keep any, then we should encapsulate those uses in functions as described in http://trust-in-soft.com/when-in-doubt-express-intent-and-leave-the-rest-to-the-compiler/.

The C standard says that it must be possible to convert a pointer to a `uintptr_t` and then convert the `uintptr_t` back to a pointer, losslessly. However, the values of the resultant integers are implementation-defined. Also, `uintptr_t` is optional in C11 and it is not guaranteed that any integer type can hold the value of a pointer.

"'6.2.8 Alignment of objects' in C11 hints that it should work in a natural way" and the mapping is "'intended to be consistent with the addressing structure of the exec. env.'" as pointed out by @AlekseyCherepanov on Twitter. However, I don't think we can rely on these, especially when it comes to undefined behavior prevention and compiler bug avoidance.

Note that some pointer -> integer conversions are not using `uintptr_t` or `intptr_t`. It is common for the code to use `size_t` and I've also seen casts to `int`. Those need to be found and fixed.

briansmith:
Ideally, we would use C's `alignas` more, and require more inputs to be aligned. However, #36 is in the way; we don't have a way of aligning values in Rust other than doing the same `uintptr_t` arithmetic that is done in C, in Rust, where such things are even less well-specified.

AlekseyCherepanov:
Non-technical comment: you refer to @ch3root on Twitter, he is @ch3root on github too. My name is Aleksey, his name is Alexander. Thanks!

ch3root:
Yep, I'm @ch3root on github too. I filled the name in the profile now:-)

Quick search with coccinelle gives the following cases of casts of pointers to non-pointers:
https://github.com/briansmith/ring/blob/72db3b27b52dedc2fc0fb99e925d4a52b9584f43/crypto/bn/montgomery.c#L338
https://github.com/briansmith/ring/blob/72db3b27b52dedc2fc0fb99e925d4a52b9584f43/crypto/poly1305/poly1305_vec.c#L85
https://github.com/briansmith/ring/blob/72db3b27b52dedc2fc0fb99e925d4a52b9584f43/crypto/modes/gcm.c#L581
[...]

Issue #68, opened 2016-01-05
Audit and fix type aliasing and alignment problems in C code


briansmith:
See https://rt.openssl.org/Ticket/Display.html?id=4218 for an example.

briansmith:
`CRYPTO_ctr128_encrypt` and the other CTR-mode code was removed recently. But, there still may be more instances of this.

briansmith:
Closing in favor of #151. I believe that there's no way to have aliasing problems without casts in C, or at least in the subset of C that we use. Please correct me if I'm wrong!

PR #840, opened 2019-06-14
Bring back the `ring::c` internal C types module.

briansmith:


briansmith:
@josephlr @elichai @jonas-schievink PTAL.

A long time ago we had a `ring::c` module like this and we defined all the C types ourselves. We even had tests to verify that the size and alignment of each type matched between Rust and C. We got rid of all of that because somebody suggested it would be better to just delegate all of this to libc, thinking there's no reason libc wouldn't define these types when it makes sense to do so. Apparently that's wrong. So, this reverts to an internal API that is similar to the old `ring::c` API, but where the types are still implemented as aliases of `libc` types. This should reduce the amount of code that would need to be directly impacted by supporting non-libc sources of these types. In particular, this gives us the flexibility to, on a per-target basis, choose to use libc's types, another crate's types, or an internal definitions of the types (like *ring* originally did).

src/c.rs
```diff
@@ -0,0 +1,33 @@
...
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+//! C types.
+//!
+//! The libc crate provide the C types for most, but not all, targets that
+//! *ring* supports.
+
+use libc;
+
+pub(crate) type size_t = libc::size_t;
```
josephlr: We could just use `usize` instead throughout *ring* as suggested in https://github.com/briansmith/ring/pull/837#issuecomment-501877466. Note that many of the places *ring* uses `size_t` it should really be using `uintptr_t`; however, [BoringSSL also makes this "mistake" in thinking that `sizeof(size_t) == sizeof(void *)`](https://github.com/briansmith/ring/blob/8fa0cf5b0169b1ca99ef4b0787fbf5f02ce387f2/crypto/fipsmodule/modes/gcm.c#L57), so it might just be easier to assume `size_t == uintptr_t` in which case (as `libc::uintptr_t == usize`), we can just use `usize` throughout *ring*.

This could be addressed now, or in a future PR. We might also want to wait until https://github.com/rust-lang/libc/issues/1400 is resolved.

src/c.rs
```diff
@@ -0,0 +1,33 @@
...
+pub(crate) type int = libc::c_int;
```
josephlr: Rust has a `target-c-int-width` attribute that we could use to define `c::int` instead of relying on `libc`. However, this is best addressed in a future PR.

src/c.rs
briansmith: > Note that many of the places _ring_ uses `size_t` it should really be using `uintptr_t`

Could you provide an example?

Rust has made a mistake where it is only well defined on platforms where `uintptr_t` and `size_t` and `usize` are the same type; whenever `size_t` and `uintptr_t` are incompatible then Rust can't work given how `usize` is defined. I'm hoping a future edition of Rust will correct this. This is actually the reason I prefer to keep using `size_t` everywhere in Rust where `size_t` is used in the C code.

> [BoringSSL also makes this "mistake" in thinking that sizeof(size_t) == sizeof(void *)](https://github.com/briansmith/ring/blob/8fa0cf5b0169b1ca99ef4b0787fbf5f02ce387f2/crypto/fipsmodule/modes/gcm.c#L57)

I think you may be misunderstanding the example you pointed out; the test in that code fragment has to do with whether 64-bit arithmetic is native or not, nothing about pointers. https://boringssl.googlesource.com/boringssl/+/8f574c37dae935a29bff56e750101e5b354503de/crypto/fipsmodule/bn/exponentiation.c#895 would be an example where `size_t` is used instead of `uintptr_t` in BoringSSL. I believe we've eliminated all such confusion in *ring* already; see #68, #151, and in particular #101.

src/c.rs
briansmith: > Rust has a `target-c-int-width` attribute that we could use to define `c::int` instead of relying on `libc`. However, this is best addressed in a future PR.

That would be awesome to use if it works! Can you give an example? I tried `#[cfg(target_c_int_width = "32")] type int = i32;` but `target_c_int_width` doesn't seem to be exposed for `#[cfg(...)]` to use.

src/c.rs
```diff
@@ -0,0 +1,33 @@
...
+pub(crate) type uint = libc::c_uint;
+
+#[cfg(all(
```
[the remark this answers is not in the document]
briansmith: In general I don't add comments like this because they get out of date. One can use `Find Usages` or similar in their IDE to see what is using what.

Issue #558, opened 2017-07-20
Verify `/dev/urandom` has been seeded before using it.


DemiMarie:
libsodium polls on `/dev/random` to make sure the kernel’s PRNG is seeded.  This can uncover real problems in virtualized environments.

cedenday:
Ring only uses `/dev/urandom` as a last resort because it is slow.

Can't you do this yourself?

briansmith:
@DemiMarie I recommend people disable the default dev_urandom_fallback feature, which has the same effect with less code.

briansmith:
See https://boringssl.googlesource.com/boringssl/+/bc7daec4d833ed399571cf847221bb3065fbc9ab

briansmith:
I generalized the title of this issue so that it doesn't presume a particular implementation strategy. I agree that it is worth doing something here.

> @DemiMarie I recommend people disable the default dev_urandom_fallback feature, which has the same effect with less code.

I wish this worked better. In practice it is pretty hard to disable the dev_urandom_fallback feature because any crate that forgets to do so will implicitly enable it. :(

josephlr:
There are many ways we could go about doing this:

1. Read a single byte from `/dev/random`
  a. This is how [`getrandom`](https://github.com/rust-random/getrandom) (the OS-specific code for [`rand`](https://github.com/rust-random/rand)) does it: [link](https://github.com/rust-random/getrandom/blob/master/src/linux_android.rs#L54-L60)
  b. It's very simple to implement in safe Rust
  c. It has the issue that it consumes a byte from `/dev/random`, which could make other users of `/dev/random` very slightly slower.
  d. Opens two files, although only one is ever open at a time.
1. Use the `RNDGETENTCNT` ioctl on `/dev/urandom` to query the amount of kernel randomness in a loop
  a. This is what [BoringSSL](https://boringssl.googlesource.com/boringssl/) does: [link](https://github.com/google/boringssl/blob/cef9d3f38d72f13412c79157c25753e22cb05f7e/crypto/fipsmodule/rand/urandom.c#L216-L231)
  b. Only uses a single file.
  b. More complex and requires unsafe Rust.
  c. We would need to decide what we do inside the loop to not burn CPU. BoringSSL sleeps for 0.25 seconds. We could use [`thread::yield_now()`](https://doc.rust-lang.org/std/thread/fn.yield_now.html).
[...]

PR #846, opened 2019-06-14
Requesting Feedback: WIP changes to `rand.rs`

josephlr:
This PR is a collection of all the potential improvements to `rand.rs`. This won't be merged as-is, and will need to be split off into smaller PRs like #839. Not all of the ideas here will end up being merged. I'm just putting this up to get feedback from @briansmith @sconybeare @myfreeweb @oherrala @tatsuya6502 and anyone else who interested.

This was motivated by the work I've been doing for the [`getrandom`](https://github.com/rust-random/getrandom) crate.

Things this patch does:
  - Organizes code based on how we try to get randomness:
    - `sysrand`: get randomness via a syscall or system function
    - `file`: get randomness from a file
    - some implementations choose between `sysrand` and `file`
  - Unifies the Linux and Android implementations
  - Read 1 byte from `/dev/random` before reading from `/dev/urandom`
  - Removes `dev_urandom_fallback` feature, as it is no longer needed for security. If you want to force Linux to use `getrandom` just don't enable the `use_heap` feature.
  - Linux: use `libc::SYS_getrandom` (#839)
  - Linux: use `GRND_NONBLOCK` for `getrandom()` detection
  - Redox: Adds support back
  - NetBSD: Adds support (easy as it just does the same thing as Linux)
  - Windows: Simplifies implementation 
  - Minor formatting and naming fixes
  - Adds more implementations for `error::Unspecified` to make error handling easier.
  - Documentation update

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -321,8 +321,7 @@ cc = "1.0.26"
 
 [features]
 # These features are documented in the top-level module's documentation.
-default = ["use_heap", "dev_urandom_fallback"]
-dev_urandom_fallback = ["use_heap"]
+default = ["use_heap"]
 internal_benches = []
 slow_tests = []
 test_logging = []
--- a/src/c.rs
+++ b/src/c.rs
@@ -28,6 +28,3 @@ pub(crate) type uint = libc::c_uint;
     any(target_arch = "aarch64", target_arch = "arm")
 ))]
 pub(crate) type ulong = libc::c_ulong;
-
-#[cfg(any(target_os = "android", target_os = "linux"))]
-pub(crate) type long = libc::c_long;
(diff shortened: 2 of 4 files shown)


briansmith:
Sorry, I don't think it makes sense to mix so many changes at once, even for a conversation. Each of the changes you suggest deserve their own thread, and many of these topics already have their own thread. 
Please discuss in the appropriate issue. For example, checking whether `/dev/urandom` is initialized is issue #558, which I found through this search: https://github.com/briansmith/ring/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+rand.

(Also, because CI is so slow due to the number of configurations tested, and because I'm in a hurry to make some particular changes today, I've canceled the CI jobs for these PRs. Don't interpret this as an implicit rejection of your PRs. However, I think it would be better for us to agree on what to do through a discussion in an issue rather than through PRs if there is any potential for disagreement.)

PR #848, opened 2019-06-15
Add missing positive.rs to Cargo package includes

ramosbugs:
`cargo package` currently fails on master without this file.


josephlr:
I currently cant get `cargo package` to succeed on my system. Adding a test to the CI that makes sure this breakage won't happen again seems like a good idea.

briansmith:
Thanks for the PR. I do already do this and other checks in bulk when I package releases so there's no reason to make PRs for this issue. I appreciate the effort though!

PR #851, opened 2019-06-17
Use modern C initialization syntax

### LINKED ISSUE #844: Switch to modern C initialization syntax
At https://github.com/briansmith/ring/pull/842#discussion_r293963619 @josephlr suggested using the newer initialization syntax, e.g. `alignas(32) Elem x = {0};` to initialize the array to zero. This seems like a good idea. I believe we already tried and failed to do so before, but we should try again and at least document the reasoning this time.

josephlr:
Closes #844

I reviewed the C code for how we handle uninitialized variables. It
looks like the uses of limbs_zero were the only times we immdiately
zero a value right after declaring it in `crypto/`.

We could switch some of the initializations in `third_party/fiat`,
but those are in the "Proven by Coq" section, so I'll just leave them
be.

### DIFF
--- a/crypto/fipsmodule/ec/gfp_p256.c
+++ b/crypto/fipsmodule/ec/gfp_p256.c
@@ -77,9 +77,9 @@ void GFp_nistz256_select_w5(P256_POINT *out, const P256_POINT table[16],
   assert(index >= 0);
   size_t index_s = (size_t)index; /* XXX: constant time? */
 
-  alignas(32) Elem x; limbs_zero(x, P256_LIMBS);
-  alignas(32) Elem y; limbs_zero(y, P256_LIMBS);
-  alignas(32) Elem z; limbs_zero(z, P256_LIMBS);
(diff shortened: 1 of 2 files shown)


briansmith:
> We could switch some of the initializations in `third_party/fiat`,
> but those are in the "Proven by Coq" section, so I'll just leave them
> be.

Which ones are you referring to, in the third-part/fiat code?

josephlr:
> Which ones are you referring to, in the third-part/fiat code?

[Here](https://github.com/briansmith/ring/blob/ce56ab1be5f4982f300392e5915c766f0fa3e01e/third_party/fiat/curve25519.c#L1824) and potentially other uses of `fe_0`. I stopped looking when I saw all the stuff about proven secure though.

briansmith:
I'm going to keep all these cases consistent. I think it would be fine to change the curve25519 code because it's not auto-generated, just manually (IIUC) checked against a Coq implementation. Probably something worth doing in boringssl.

PR #853, opened 2019-06-18
Only fallback to `/dev/urandom` on ENOSYS from getrandom.

briansmith:
@josephlr I think this is one of the changes you wanted. PTAL.

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -113,10 +113,19 @@ use crate::sealed;
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
 mod sysrand_chunk {
-    use crate::{c, error};
+    use crate::c;
+
+    pub struct Error(c::int);
+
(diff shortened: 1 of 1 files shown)


josephlr:
So the only change I was considering adding to Linux `getrandom()` error checking was to check for support using the `GRND_NONBLOCK` flag (which required checking for EAGAIN return). However, after looking at the code again, I realized that's not necessary (as it will only block if we have the syscall, in which case we want to block anyway).

I think the existing implementation is fine, as [the only error codes we could get](http://man7.org/linux/man-pages/man2/getrandom.2.html) (given our invocation) are `EINTR` and `ENOSYS`.

josephlr:
Specifically, I think this PR should be closed.

briansmith:
Closing this in favor of the issue #1532.

PR #854, opened 2019-06-18
impl Clone for SystemRandom

### LINKED ISSUE #808: Make SystemRandom Clone and Copy
`SystemRandom` is [an empty struct](https://github.com/briansmith/ring/blob/569bb6ed6a6eceaea296a8ca9f6f77c3f2a66047/src/rand.rs#L81-L87) and some things would become simpler if it had `#[derive(Clone, Copy)]`.

josephlr:
Fixes #808

Per https://github.com/briansmith/ring/issues/808#issuecomment-502572386 I did not add CloneableSecureRandom as to avoid cluttering *ring*'s external API. It's also just equivalent to `SecureRandom + Clone` if anyone needs it.

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -76,6 +76,7 @@ pub trait SecureRandom: sealed::Sealed {
 /// `getrandom` and `read`.
 ///
 /// [`getrandom`]: http://man7.org/linux/man-pages/man2/getrandom.2.html
+#[derive(Clone)]
 pub struct SystemRandom;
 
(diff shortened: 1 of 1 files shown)


briansmith:
`SystemRandom` implements `Clone` and `Debug` as of 490f7eeb7fefd0c2874374d751334a02515f732c. Please check out the tests I added. The clone test isn't the best but it will prevent us from regressing the `Clone`-ability, at least.

Issue #852, opened 2019-06-18
Does Android 8.0 support the `getrandom` syscall?


briansmith:
In *ring* 0.14 and earlier, Android always used `/dev/urandom` and never used the `getrandom` syscall. The current code in the *ring* master branch *always* uses the `getrandom` syscall. Is this new behavior too aggressive?

In the absence of any reason to do otherwise, *ring*'s minimum supported version of Android is the same as Google Play's minimum version of Android. From https://developer.android.com/distribute/best-practices/develop/target-sdk: Google Play requires that [...] app updates target Android 9.0 starting November 1, 2019. Until then, Android 8.0 is the minimum supported version.

* Android 8.0 = API level 26 = Android O (Oreo)
* Android 9.0 = API level 28 = Android P (Pie)

I know that bionic libc only adds a `getrandom()` wrapper around the syscall starting in API level 28 according to https://android.googlesource.com/platform/bionic/+/android-9.0.0_r22/docs/status.md#libc.

However, it is not clear whether the syscall available for use through `syscall(SYS_getrandom)` in Android 8.0. If not, we need to enable the fallback to `/dev/urandom`. If so, then the current code is correct.

@josephlr Can you find an authoritative answer to this question?

josephlr:
TL;DR; We need to have a `/dev/urandom` fallback and will need it at least until the 3.16 kernel ends LTS support in mid 2020. 

An Android app has two values that matter when talking about versions. For more info [see this article](https://medium.com/androiddevelopers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd).
- `minSdkVersion` - The minimum version your app supports, for Rust this is `14` (Andorid 4.0, Ice Cream Sandwich) on 32-bit and `21` (Android 5.0, Lollipop) on 64-bit [as determined by their CI](https://github.com/rust-lang/rust/blob/master/src/ci/docker/dist-android/Dockerfile). 
- `targetSdkVersion` - The SDK's whose [behavior changes](https://developer.android.com/reference/android/os/Build.VERSION_CODES) your app has opted into. This allows Android to break backwards compatibility and then have apps opt into this breakage (not unlike Rust editions). This is the value that must increase by a certain date.

[...]

PR #855, opened 2019-06-18
Use `/dev/urandom` fallback on Andorid

josephlr:
Addresses #852 

Given [current Android device support](https://developer.android.com/about/dashboards) it is unreasonable to expect full support for the `getrandom(2)` syscall.

This PR also adds some minor style fixes for `errno` handling.


briansmith:
Thanks! I already have a change pending for this so I'm going to close this out.

PR #856, opened 2019-06-18
Use slice::chunks on Windows

josephlr:
We can do the same overflow check but in a much more readable/efficient way. This also helps toward overall simplification of `rand.rs` and the eventual elimination of `mod sysrand_chunk` ([see followup](https://github.com/josephlr/ring/commit/3d775b8c4102f598d8a74b62de67f05fdc2faa77)).

Also removes `polyfill::usize_from_u32` as we already check compatiblity in the `assert!()` (so we know the `as` cast is safe). We should probably cleanup `polyfill` anyway, as most of the proposed features have gone in a different direction or are unlikely to ever be stabilized.

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -154,30 +154,25 @@ mod sysrand_chunk {
 }
 
 #[cfg(windows)]
-mod sysrand_chunk {
-    use crate::{error, polyfill};
+mod sysrand {
+    use crate::error;
 
-    #[inline]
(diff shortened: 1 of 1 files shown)


briansmith:
I don't have time to consider this now. Thanks though!

josephlr:
> I don't have time to consider this now. Thanks though!

That fine! This is just one of many changes from `getrandom` that could be incorporated into _ring_, but those should be addressed wholesale after https://github.com/rust-lang/rust/pull/62082 is resolved.

PR #863, opened 2019-07-01
Add Rust SHA-2 implementation

briansmith:


briansmith:
This implementation is designed to be the base to enable some porting efforts. I expect in the future we'll implement some performance tweaks on top of it.

Issue #827, opened 2019-05-27
hmac::Key is too large


ctz:
At the minute a `hmac::Key` is fairly large (the order of 448 bytes on x64): it is two `digest::Context`s. Most of the parts of these `digest::Context`s are statically unused for the lifetime of a given `hmac::Key`: `pending` is unused once the object is constructed, `num_pending` is statically zero, `completed_data_blocks` is statically one -- obviously removing `pending` is the big win here (256 of those 448 bytes).

The cost of this would be closer coupling between the hmac and digest modules, but the benefit would be a `hmac::Key` closer to the optimal size of 128 bytes plus one pointer size.  Would you accept a PR for this?

briansmith:
I agree. I actually have prototyped this kind of change before. I'll see if I can dig up the code.

briansmith:
I consider this fixed by #868.

ctz:
Thanks!

PR #868, opened 2019-07-08
Reduce the size of `hmac::Key` and `hmac::Context`

briansmith:
Shrink the size of `hmac::Key` by 256 bytes. Shrink the size of `hmac::Context` by 128 bytes.

In theory it would be possible to eliminate two 'completed_data_blocks: u64' and one `pub algorithm: &'static Algorithm` from each `Key`, and one `completed_data_blocks: u64` and one `pub algorithm: &'static Algorithm` from each `Context` but I don't think those are large enough wins to bother right now.

### DIFF
--- a/src/digest.rs
+++ b/src/digest.rs
@@ -30,6 +30,89 @@ use core::num::Wrapping;
 mod sha1;
 mod sha2;
 
+#[derive(Clone)]
+pub(crate) struct BlockContext {
+    state: State,
+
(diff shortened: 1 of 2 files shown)


briansmith:
@ctz Please take a look at this. It attempts to fix #827.

briansmith:
I went ahead and merged this. Let me know if you have any opinions after the fact.

PR #869, opened 2019-07-09
Make `std` an optional dependency. Require `alloc` instead.

briansmith:
This raises the minimum version of Rust required to 1.36.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -36,7 +36,7 @@ use untrusted;
 /// enum Error {
 ///     CryptoError,
 ///
-/// #  #[cfg(feature = "use_heap")]
+/// #  #[cfg(feature = "alloc")]
 ///     IOError(std::io::Error),
 ///     // [...]
 /// }
@@ -88,7 +88,7 @@ impl core::fmt::Display for Unspecified {
     }
 }
 
-#[cfg(feature = "use_heap")]
+#[cfg(feature = "std")]
 impl std::error::Error for Unspecified {
     #[inline]
     fn cause(&self) -> Option<&dyn std::error::Error> {
@@ -168,12 +168,12 @@ impl KeyRejected {
         KeyRejected("PublicKeyIsMissing")
     }
 
-    #[cfg(feature = "use_heap")]
+    #[cfg(feature = "alloc")]
     pub(crate) fn too_small() -> Self {
         KeyRejected("TooSmall")
     }
 
-    #[cfg(feature = "use_heap")]
+    #[cfg(feature = "alloc")]
     pub(crate) fn too_large() -> Self {
         KeyRejected("TooLarge")
     }
@@ -186,7 +186,7 @@ impl KeyRejected {
         KeyRejected("WrongAlgorithm")
     }
 
-    #[cfg(feature = "use_heap")]
+    #[cfg(feature = "alloc")]
     pub(crate) fn private_modulus_len_not_multiple_of_512_bits() -> Self {
         KeyRejected("PrivateModulusLenNotMultipleOf512Bits")
     }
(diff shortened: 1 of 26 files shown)


josephlr:
This is great! If you're making backwards-incompatible changes to the Cargo features, it might be worth reconsidering the `dev_urandom_fallback` feature.

If I'm understanding the motivation correctly, the main reason to not have the `dev_urandom_fallback` feature is to avoid depending on `libstd`. However, if we now have separate `std` and `alloc` features, it might make sense to have the `std` feature control if we read from `/dev/urandom` on Linux.

Once #558 is fixed, there's not a good _security_ reason to disable the fallback. However, this might be better addressed in a followup PR.

src/error.rs
```diff
@@ -36,7 +36,7 @@ use untrusted;
 /// enum Error {
 ///     CryptoError,
 ///
-/// #  #[cfg(feature = "use_heap")]
+/// #  #[cfg(feature = "alloc")]
```
josephlr: `feature = "std"`

src/test.rs
```diff
@@ -116,13 +116,13 @@
 //! stack trace to the line in the test code that panicked: entry 9 in the
 //! stack trace pointing to line 652 of the file `example.rs`.
 
-#[cfg(feature = "use_heap")]
-use crate::bits;
+use crate::{bits, digest, error};
```
josephlr: Should the `bits` module be gated on the `alloc` feature here? Like it is in `lib.rs`?

src/lib.rs
```diff
@@ -62,13 +65,15 @@
...
+extern crate alloc;
+
+#[cfg(any(test, feature = "std"))]
 extern crate std;
 
 #[macro_use]
 mod debug;
 
-#[cfg(any(test, feature = "use_heap"))]
 #[macro_use]
```
josephlr: This still needs a `#[cfg(test)]` in order for `cargo build --no-default-features` to work.

josephlr: Actually, `#[cfg(feature = "alloc")]` is what you want here, so that the integration tests keep working.

src/lib.rs
```diff
@@ -62,13 +65,15 @@
 #![no_std]
 #![cfg_attr(feature = "internal_benches", allow(unstable_features), feature(test))]
 
-#[cfg(any(test, feature = "use_heap"))]
+#[cfg(any(test, feature = "alloc"))]
```
josephlr: `#[cfg(feature = "alloc")]` is all that's needed here, as `test` requires `alloc`.

In general, there a lot of `#[cfg(any(test, feature = "alloc"))]` in the code. These should be simplified to `#[cfg( feature = "alloc")]

src/lib.rs
```diff
@@ -62,13 +65,15 @@
...
+extern crate alloc;
+
+#[cfg(any(test, feature = "std"))]
```
josephlr: `#[cfg(feature = "std")]` is what's needed here, otherwise the current `cfg` make tests require `std`.

src/test.rs
```diff
@@ -116,13 +116,13 @@
...
 
-use crate::{digest, error};
+#[cfg(feature = "alloc")]
```
josephlr: If `#[cfg(feature = "alloc")]` is added to `lib.rs` then these annotations are not needed in this file (here and elsewhere).

briansmith:
Thanks. I updated the PR to address most of your concerns.

There is more work that needs to be done to properly support testing for `no_std` environments, e.g. adding the non-default configurations to CI. Those changes will have to wait until my pipeline is cleared.

briansmith:
Thanks for the review!

Issue #870, opened 2019-07-10
Support for weaker RSA keys used in DNSSEC


andrewtj:
While 2048-bit RSA is in use for many KSKs at the top of the hierarchy, weaker keys are still in use for ZSKs. Here's what is used in a selection of names that are of interest to me:

|Name    |ZSK                        |KSK                        |
|--------|---------------------------|---------------------------|
|       .|2048-bit RSASHA256         |2048-bit RSASHA256         |
|     au.|2048-bit RSASHA256         |2048-bit RSASHA256         |
| com.au.|2048-bit RSASHA256         |1024-bit RSASHA256         |
|  id.au.|2048-bit RSASHA256         |1024-bit RSASHA256         |
| net.au.|2048-bit RSASHA256         |1024-bit RSASHA256         |
| org.au.|2048-bit RSASHA256         |1024-bit RSASHA256         |
|    com.|2048-bit RSASHA256         |1024-bit RSASHA256         |
|    net.|2048-bit RSASHA256         |1024-bit RSASHA256         |
|    org.|2048-bit RSASHA1-NSEC3-SHA1|1024-bit RSASHA1-NSEC3-SHA1|

This is likely to continue to be the case for some time yet. Verisign operate a few of the names above and are only now in the midst of switching from [1024-bit to 1280-bit KSKs](https://indico.dns-oarc.net/event/29/contributions/660/).

As these keys are unavoidable, would you consider supporting them in some form for signature verification? Perhaps there could be dedicated `ring::signature::RsaParameters` behind a feature flag whose parameters are tied to some metric in the hierarchy, such as the weakest parameters needed to support com, net, and org.

I am currently using OpenSSL to fill the gap so lack of support for weak keys isn't that big a deal, but it would be nice to drop OpenSSL.

/cc @bluejekyll and @partim as I think they're in the same boat with their projects.

briansmith:
Sure, it's fine to add new variants that accept RSA 1024-8192 with SHA-1 and SHA-256 named with the `_FOR_LEGACY_USE_ONLY` suffix like RSA_PKCS1_2048_8192_SHA1_FOR_LEGACY_USE_ONLY now is. When doing so, please go through all the test vectors in *ring* and import the test vectors from the original source that were removed because the RSA key was too small.

My understanding is there is no need for non-SHA-1, non-SHA-256 variants, so let's not bother with those.

andrewtj:
[...]

PR #871, opened 2019-07-12
Support for 1024-bit RSA

andrewtj:
Follows from #870.


tests/rsa_tests.rs
```diff
@@ -151,8 +151,8 @@ fn test_signature_rsa_pkcs1_verify() {
 
             let digest_name = test_case.consume_string("Digest");
             let alg = match digest_name.as_ref() {
-                "SHA1" => &signature::RSA_PKCS1_2048_8192_SHA1_FOR_LEGACY_USE_ONLY,
```
briansmith: I didn't look at the whole PR. However, I did look at this test. The test needs to be modified so that it verifies the correct results for each of `RSA_PKCS1_*` separately. For example, we need to verify that RSA_PKCS1_2048_8192_SHA256 still rejects the too-small inputs while RSA_PKCS1_1024_8192_SHA256 accepts the superset that are at least 1024 bits wide.

To accelerate the process of getting this all done, please fix this problem for `RSA_PKCS1_3072_8192_SHA384` too, which already had this problem. Otherwise, I'll do that myself before merging the PR.

andrewtj:
Can do, but not for two or three hours.

vavrusa:
Rebased in https://github.com/vavrusa/ring/commits/master

irtefa:
Hi @briansmith, can we please move this forward?

briansmith:
Thanks for telling me about the issue with the panic. I am going to fix that. Then, I will work on merging this (or a modified version of it). Please rebase this onto master.

andrewtj:
@briansmith no worries, done.

vavrusa:
@briansmith I fixed the panic in https://github.com/vavrusa/ring/commit/4ff176b7b33e13815b552f3c0354d52d66ad74be but I'm not sure what is the best approach here. I suppose you can either create an oversize array to accommodate for padding like here, or change the `big_endian_from_limbs` to allow for writing truncated result when `out < width of the limbs`. What would you prefer?

jmagnuson:
Any interest in moving this forward? I've been using this branch + @vavrusa's panic fix for a while, and it works as advertised.

vavrusa:
I've been running it for a while now, the only annoyance is I have to patch ring with `[patch.crates-io]`. The fork has also support for 1024b RSASHA512 which is not planned to be supported here AFAIK. It is unfortunately required in real world DNS as plenty of zones still use this configuration.

vavrusa:
Is this mergeable with https://github.com/vavrusa/ring/commit/4ff176b7b33e13815b552f3c0354d52d66ad74be or can I do something else to improve this?

Darkspirit:
> The fork has also support for 1024b [RSASHA512](https://github.com/bluejekyll/trust-dns/blob/6c4e08e62384563bf4304f3e0af22477c00546ff/crates/proto/src/rr/dnssec/algorithm.rs#L104)

https://github.com/bluejekyll/trust-dns/issues/907 depends on this.

briansmith:
My plan is to do an immediate release that has #937 merged, then a release that has RSA 1024 support and OAEP support soon after.

src/rsa/convert_nist_rsa_test_vectors.py
```diff
@@ -197,17 +197,20 @@ def main(fn, test_type, padding_alg):
         e = int(case['e'], 16)
         d = int(case['d'], 16)
 
-        if n.bit_length() // 8 < 2048 // 8:
-            debug("Skipping due to modulus length (too small).", DEBUG)
-            continue
-
         if test_type == 'sign':
+            if n.bit_length() < 2048:
```
briansmith: I think it was intentional that we used `n.bit_length() // 8 < 2048 // 8` instead of `n.bit_length() < 2048` previously, right? we want to test the odd key lengths that are close to being correct. Or am I misunderstanding why this was changed?

src/rsa/convert_nist_rsa_test_vectors.py
```diff
@@ -197,17 +197,20 @@ def main(fn, test_type, padding_alg):
...
+                debug("Skipping due to modulus length (too small).", DEBUG)
+                continue
             if n.bit_length() > 4096:
                 debug("Skipping due to modulus length (too large).", DEBUG)
                 continue
 
             print_sign_test(case, n, e, d, padding_alg)
         else:
+            legacy = case['SHAAlg'] in ["SHA1", "SHA256"]
+            if (n.bit_length() < 2048 and not legacy) or n.bit_length() < 1024:
```
briansmith: Again, I think we want to use the rounding check `n.bit_length() // 8 < 2048 // 8` instead of checking `bit_length()` exactly, unless somebody can point out why not.

In particular, it would be interesting to know what happens when we have a, say, 1022 bit public key, right?

tests/rsa_tests.rs
```diff
@@ -163,26 +185,27 @@ fn test_signature_rsa_pkcs1_verify() {
...
+                        let n_len = der::positive_integer(input)?
+                            .big_endian_without_leading_zero()
+                            .len();
                         let _ = der::positive_integer(input)?;
-                        let _ = der::positive_integer(input)?;
-                        Ok(())
-                    }
-                ))
-                .is_ok());
+                        Ok(n_len * 8)
```
briansmith: Instead of multiplying out `n_len * 8`, we need to use the exact bit length as returned by `Modulus::from_be_bytes_with_bit_length`, don't we?

tests/rsa_tests.rs
```diff
@@ -163,26 +185,27 @@ fn test_signature_rsa_pkcs1_verify() {
...
 
             let msg = test_case.consume_bytes("Msg");
             let sig = test_case.consume_bytes("Sig");
             let is_valid = test_case.consume_string("Result") == "P";
-
-            let actual_result =
-                signature::UnparsedPublicKey::new(alg, &public_key).verify(&msg, &sig);
-            assert_eq!(actual_result.is_ok(), is_valid);
+            for &(alg, min_bits) in params {
+                let width_ok = key_bits >= min_bits;
```
briansmith: In particular, we should have test cases for public keys of lengths 1016, 1017, 1018, 1019, 1020, 1021, 1022, 1023, 1024, etc., `width_ok` would be different than what this test would indicate, right?

briansmith:
Very sorry about the unexpected long delay on this. Now I'm available to work on this again. Could you please rebase it once against and respond to the comments I just left about key sizes that are very close to 1024 bits but not exactly 1024 bits?

briansmith:
Let's defer improving the tests to cover odd bit lengths to another PR.

PR #874, opened 2019-07-16
Support FreeBSD, NetBSD, OpenBSD, and Solaris using `/dev/urandom`.

briansmith:


josephlr:
For OpenBSD,  this is fine. [ `/dev/random` and `/dev/urandom` are the same](https://www.freebsd.org/cgi/man.cgi?query=random&apropos=0&sektion=4&manpath=OpenBSD+6.5&arch=default&format=html), and aren't dumb.

For NetBSD, this is fine. NetBSD has nearly identical interfaces as Linux (including the https://github.com/briansmith/ring/issues/558).

For FreeBSD, we should use `/dev/random`. `/dev/urandom` is not guaranteed to exist, and `/dev/random` doesn't have the weird Linux problems, [see the docs](https://www.freebsd.org/cgi/man.cgi?query=random&sektion=4&apropos=0&manpath=FreeBSD+12.0-RELEASE).

For Solaris, it's really not clear which to use, [the docs are bad, mentioning "higher-quality" and "lower-quality" randomness without explaining what those terms mean](https://docs.oracle.com/cd/E36784_01/html/E36884/urandom-7d.html). The `getrandom` crate falls back to using `/dev/random` ([see this note](https://github.com/rust-random/getrandom/blob/b3e609f8b5d716851da8c1c423d86447d1760172/src/solaris_illumos.rs#L11-L14)). Regardless of which is used, the data needs to be read in chunks, as there are hard limits to how many bytes can be `read(2)` at a time:
 - `1024` for `/dev/random`
 - `128*1024` for `/dev/urandom`

Some of these platforms (OpenBSD, FreeBSD, Solaris) have APIs like Linux's `getrandom(2)` that allow you to get randomness without holding on to a file descriptor. But that requires a bunch of target specific code.

An alternative would be to just use the `getrandom` crate.

briansmith:
> For FreeBSD, we should use `/dev/random`. `/dev/urandom` isn't mentioned in the man pages, and `/dev/random` doesn't have the weird Linux problems, [see the docs](https://www.freebsd.org/cgi/man.cgi?query=random&sektion=4&apropos=0&manpath=FreeBSD+12.0-RELEASE). **EDIT:** fixed an incorrect reference to FreeBSD docs.

It's unfortunate that they never documented `/dev/urandom`. However, I verified `/dev/urandom` exists in every currently-supported (by freebsd.org) FreeBSD release.

josephlr:
> > For FreeBSD, we should use `/dev/random`. `/dev/urandom` isn't mentioned in the man pages, and `/dev/random` doesn't have the weird Linux problems, [see the docs](https://www.freebsd.org/cgi/man.cgi?query=random&sektion=4&apropos=0&manpath=FreeBSD+12.0-RELEASE). **EDIT:** fixed an incorrect reference to FreeBSD docs.
> 
> It's unfortunate that they never documented `/dev/urandom`. However, I verified `/dev/urandom` exists in every currently-supported (by freebsd.org) FreeBSD release.

So I found something weird, [FreeBSD 4.11 _does_ document `/dev/urandom`](https://www.freebsd.org/cgi/man.cgi?query=urandom&apropos=0&sektion=4&manpath=FreeBSD+4.11-RELEASE&arch=default&format=html), but they stopped documenting it in 5.0, so using `/dev/urandom` is probably fine. The Solaris support could just be removed.

briansmith:
> For Solaris, it's really not clear which to use

My goal here is to make Solaris work again as well as it did in 0.14.6 (slightly better, in fact). I filed #876 to track improving Solaris support.

> The Solaris support could just be removed.

It's now required by some commercial users of *ring*.

josephlr:
> It's now required by some commercial users of _ring_.

Sounds good, makes sense to me.

PR #883, opened 2019-08-01
Add RNG for Fortanix SGX

npmccallum:
Currently, ring doesn't provide a random number generator for SGX at all.
This is probably by accident, but is, in fact, the correct behavior.

Due to distrust of the OS, SGX enclave applications need a source of
randomness that is not forgable by the OS. This pretty much leaves only
the RDRAND instruction. Fortunately, LLVM provides an intrinsic for this
instruction.

Compiling to the Fortanix SGX target currently requires Rust nightly.
The implementation in this patch takes advantage of this to call the
LLVM intrinsic directly; a Rust feature that will never be stabilized.
At some point in the future, the Fortanix SGX target may be stabilized
and then we will need a way to invoke RDRAND wihout nightly Rust.

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -193,6 +193,9 @@ use self::darwin::fill as fill_impl;
 #[cfg(any(target_os = "fuchsia"))]
 use self::fuchsia::fill as fill_impl;
 
+#[cfg(any(target_env = "sgx"))]
+use self::sgx::fill as fill_impl;
+
 #[cfg(any(target_os = "android", target_os = "linux"))]
 mod sysrand_chunk {
     use crate::{c, error};
@@ -430,5 +433,36 @@ mod fuchsia {
     #[link(name = "zircon")]
     extern "C" {
         fn zx_cprng_draw(buffer: *mut u8, length: usize);
+    }
+}
+
+#[cfg(any(target_env = "sgx"))]
+mod sgx {
+    use core::arch::x86_64::_rdrand64_step;
+    use crate::error;
+
+    #[inline]
+    fn rdrand() -> Result<u64, error::Unspecified> {
+        // See https://software.intel.com/en-us/articles/intel-digital-random-number-generator-drng-software-implementation-guide
+        // Section 5.2.1
+        for _ in 0..10 {
+            let mut r = 0;
+            if unsafe { _rdrand64_step(&mut r) } == 1 {
+                return Ok(r);
+            }
+        }
+
+        Err(error::Unspecified)
+    }
+
+    pub fn fill(mut dest: &mut [u8]) -> Result<(), error::Unspecified> {
+        while dest.len() > 0 {
+            let rand = rdrand()?.to_ne_bytes();
+            let len = core::cmp::min(dest.len(), rand.len());
+            dest[..len].copy_from_slice(&rand[..len]);
+            dest = &mut dest[len..];
+        }
+
+        Ok(())
     }
 }


josephlr:
This PR could be greatly simplified by just using the implementation from `getrandom`, just without the CPUID detection (as you're on SGX) and without the AMD `RDRAND` bug detection.

See the implementation: https://github.com/rust-random/getrandom/blob/master/src/rdrand.rs

This implementation also fixes two issues present in this PR:
  - You do not need to use unstable features. Rust has a [stable `RDRAND` intrinsic](https://doc.rust-lang.org/core/arch/x86_64/fn._rdrand64_step.html). 
  - You should be calling RDRAND in a loop (per section 5.2.1 of the [Intel Documentation](https://software.intel.com/en-us/articles/intel-digital-random-number-generator-drng-software-implementation-guide)).

npmccallum:
@josephlr Given the desire for a small TCB inside an SGX enclave, I'm in favor of reducing the number of crates required. Given that Rust already has a stable `RDRAND` intrinsic, I'd rather use that directly and fix the loop issue. Does that work for you?

npmccallum:
@josephlr The latest push uses the stable `RDRAND` intrinsic in a tight loop.

josephlr:
> @josephlr Given the desire for a small TCB inside an SGX enclave, I'm in favor of reducing the number of crates required. Given that Rust already has a stable `RDRAND` intrinsic, I'd rather use that directly and fix the loop issue. Does that work for you?

Oh ya, I should have clarified what I meant. Just use the `getrandom` implementation as a guide for this PR. We don't want `ring` to use `getrandom` for this (that's a bigger decision for a later time).

src/rand.rs
```diff
@@ -432,3 +435,34 @@ mod fuchsia {
...
+    use core::arch::x86_64::_rdrand64_step;
+    use crate::error;
+
+    #[inline]
+    fn rdrand() -> Result<u64, error::Unspecified> {
+        // See https://software.intel.com/en-us/articles/intel-digital-random-number-generator-drng-software-implementation-guide
+        // Section 5.2.1
+        for _ in 0..10 {
+            let mut r = 0;
+            if unsafe { _rdrand64_step(&mut r) } == 1 {
```
josephlr: Explain why this is safe (i.e. that SGX always supports RDRAND).

npmccallum:
It looks like the only test failure was a spurious one.

Regarding the test coverage decrease, I'm not sure how to test this.

@lkatalin Do you know how to write tests that run inside SGX?

lkatalin:
@npmccallum I could certainly give it a try.

src/rand.rs
```diff
@@ -432,3 +435,34 @@ mod fuchsia {
...
+                return Ok(r);
+            }
+        }
+
+        Err(error::Unspecified)
+    }
+
+    pub fn fill(mut dest: &mut [u8]) -> Result<(), error::Unspecified> {
+        while dest.len() > 0 {
```
josephlr: This can me made more clear (and more efficient by eliding `memcpy` calls) by using [`chunks_exact_mut`](https://doc.rust-lang.org/core/slice/struct.ChunksExactMut.html), like the `getrandom` implementation does.  

See the comparison here: https://rust.godbolt.org/z/gwprxd

src/rand.rs
```diff
@@ -432,3 +435,34 @@ mod fuchsia {
         fn zx_cprng_draw(buffer: *mut u8, length: usize);
     }
 }
+
+#[cfg(any(target_env = "sgx"))]
+mod sgx {
+    use core::arch::x86_64::_rdrand64_step;
+    use crate::error;
+
+    #[inline]
```
josephlr: `#[inline]` does nothing on a private function, the attribute is only useful for `pub` functions/methods.

src/rand.rs
```diff
@@ -432,3 +435,34 @@ mod fuchsia {
...
+    fn rdrand() -> Result<u64, error::Unspecified> {
```
josephlr: If you end up having multiple call-sites for this function (i.e. if you make the `chunks_exact_mut` changes), it might be worth having this function return a `[u8; 8]` instead of a `u64`.

josephlr:
> Regarding the test coverage decrease, I'm not sure how to test this.

If you just want to check that the RDRAND implementation is good, you could test this on a normal x86_64 machine (without SGX) and just do the CPUID check.

briansmith:
> Just use the `getrandom` implementation as a guide for this PR.

Do not do this, because the license is not compatible.

I will probably implement this myself so I don't have to worry about the license implications.

Have somebody with spending authority email me: brian@briansmith.org if/when you want to see progress on this.

npmccallum:
@briansmith I did not use `getrandom` as a guide. I never even looked at it. I only looked at the Intel documentation. Please tell me that ring is not pay-to-merge.

josephlr:
> > Just use the `getrandom` implementation as a guide for this PR.
> 
> Do not do this, because the license is not compatible.

This is incorrect, `getrandom` is dual licensed under Apache 2.0 and MIT, which is compatible with Ring's ISC style license. If significant code is copied (which isn't necessary for this PR), then the copyright notice would need to be included in the file.

briansmith:
There's already a PR for this.

josephlr:
> There's already a PR for this.

https://github.com/briansmith/ring/pull/738 I'm assuming

briansmith:
> Please tell me that ring is not pay-to-merge.

I have a small number of days per month I work on stuff that isn't paid for. This has resulted in a backlog, unfortunately. Three are a variety of factors I use to prioritize the backlog. The #1 thing somebody can do to get items on the backlog prioritized higher is sponsor this project financially. That's not the only criteria, though.

PR #886, opened 2019-08-13
Clone for aead::UnboundKey, aes::Key, gcm::Key, chacha::Key.

likeanocean:
Hello!

There have been times where I've wanted to clone these but not been able to (especially upgrading legacy code to newer versions of ring).

If allowing clones sounds okay with you, I'll add the relevant tests!

### DIFF
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -401,6 +401,7 @@ impl core::fmt::Debug for UnboundKey {
 }
 
 #[allow(variant_size_differences)]
+#[derive(Clone)]
 enum KeyInner {
     AesGcm(aes_gcm::Key),
     ChaCha20Poly1305(chacha20_poly1305::Key),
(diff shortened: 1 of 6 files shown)


src/aead.rs
```diff
@@ -385,6 +385,7 @@ impl Aad<[u8; 0]> {
 }
 
 /// An AEAD key without a designated role or nonce sequence.
+#[derive(Clone)]
```
briansmith: We cannot allow cloning for `UnboundKey` because it would undo the protections that `OpeningKey` and `SealingKey` provide. Please remove the `#[derive(Clone)]` and instead add a comment to this effect.

We can allow cloning only for `LessSafeKey`. I think that adding  `#[derive(Clone)]` for `LessSafeKey` will solve your problem.

briansmith:
I'm going to close this because I think we only need `LessSafeKey` and the private types to be `Clone`, not `UnboundKey`. In general I'm happy to make *that* change though, if properly tested.

PR #889, opened 2019-08-20
Introduce EcdsaKeyPair::from_private_key_unchecked.

### LINKED ISSUE #882: EcdsaKeyPair from private key component only?
Would you consider adding a function similar to `Ed25519KeyPair::from_seed_unchecked`?

Background is that in DNSSEC, existing implementations (bind, ldns) store keys in files with a special format. For ECDSA, these files only contain the private key component. Technically, there is also an accompanying files with the DNSKEY record containing the public key, but for robustness, I would prefer ignoring that and generating the record from the key pair only.

partim:
This PR adds `EcdsaKeyPair::from_private_key_unchecked` which creates an ECDSA key pair from the private key only.

Note: The two private keys in the test case are random specimen from the NIST CAVP 186-4 ECDSA2VS Test Vectors. I think testing for short, long, and correctly sized keys should about cover it?

If merged, fixes #882.

### DIFF
--- a/src/ec/suite_b/ecdsa/signing.rs
+++ b/src/ec/suite_b/ecdsa/signing.rs
@@ -136,6 +136,29 @@ impl EcdsaKeyPair {
         Ok(Self::new(alg, key_pair))
     }
 
+    /// Constructs an ECDSA key pair direclty from the private key bytes.
+    ///
+    /// It is recommended to use `EcdsaKeyPair::from_pkcs8()` instead. When
+    /// that is not practical, it is recommended to use
+    /// `EcdsaKeyPair::from_private_key_and_public_key` instead.
+    ///
+    /// Since the public key is not given, the public key will be computed from
+    /// the private key. It is not possible to detect misuse or corruption of
+    /// the private key since the public key isn't given as input.
+    pub fn from_private_key_unchecked(
+        alg: &'static EcdsaSigningAlgorithm,
+        private_key: &[u8],
+    ) -> Result<Self, error::KeyRejected> {
+        let private_key = ec::Seed::from_bytes(
+            alg.curve,
+            untrusted::Input::from(private_key),
+            cpu::features()
+        ).map_err(|_| error::KeyRejected::invalid_encoding())?;
+        ec::KeyPair::derive(private_key)
+            .map(|key_pair| Self::new(alg, key_pair))
+            .map_err(|_| error::KeyRejected::invalid_encoding())
+    }
+
     fn new(alg: &'static EcdsaSigningAlgorithm, key_pair: ec::KeyPair) -> Self {
         let (seed, public_key) = key_pair.split();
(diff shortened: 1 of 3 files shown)


src/ec/suite_b/ecdsa/signing.rs
```diff
@@ -136,6 +136,29 @@ impl EcdsaKeyPair {
...
+    ) -> Result<Self, error::KeyRejected> {
+        let private_key = ec::Seed::from_bytes(
+            alg.curve,
+            untrusted::Input::from(private_key),
+            cpu::features()
+        ).map_err(|_| error::KeyRejected::invalid_encoding())?;
+        ec::KeyPair::derive(private_key)
+            .map(|key_pair| Self::new(alg, key_pair))
+            .map_err(|_| error::KeyRejected::invalid_encoding())
+    }
```
briansmith: In Ed25519, we have:

```rust
    pub fn from_seed_and_public_key(
        seed: &[u8],
        public_key: &[u8],
    ) -> Result<Self, error::KeyRejected> {
        let pair = Self::from_seed_unchecked(seed)?;

        // This implicitly verifies that `public_key` is the right length.
        // XXX: This rejects ~18 keys when they are partially reduced, though
        // those keys are virtually impossible to find.
        if public_key != pair.public_key.as_ref() {
```

That is, `from_seed_and_public_key` is implemented in terms of `from_seed_unchecked`. It seems to me like we should do something similar in ECDSA. What do you think?

partim: That seems reasonable. Let me update the PR.

partim:
From `from_private_key_and_public_key` I only added test cases for P-256. I think that’s enough, but we can easily add more to the txt file.

briansmith:
Sorry I lost track of this. I'm updating the base branch to be `main` so I can remove `master` today. The main issue I see now is with the tests: We should use the same test files (*.txt) to verify all the variants of this API, instead of having separate files just for this API. I.e. every place we test `from_private_key_unchecked` we should also be verifying the checked variant.

partim:
No worries, I got caught up in all sorts of other work as well.

I’m pretty sure there was a reason why I added two test files, but I can’t remember now. I’ll look into merging the two and provide an update soon.

partim:
Now it is my turn to apologize for the long silence.

I have pushed a commit to bring the branch in line with current main.

I have also looked into merging the two test files, but I’m not sure it makes sense. For one, the meaning of the `PublicKey` field in the two cases differs: In one case it contains input, in the other it contains a value the result needs to be checked against. In addition, there are test cases in `ecdsa_from_private_key_and_public_key_test` that test that the function fails for invalid public key input which don’t make sense in `ecdsa_from_private_key_unchecked_test`.

This could be fixed by adding additional fields to distinguish the semantics, but I am not sure if this improves clarity?

Issue #387, opened 2016-12-22
Ensure crypto instructions are used on AAarch32 & AAarch64 when available.


briansmith:
Grep for `BORINGSSL_CLANG_SUPPORTS_DOT_ARCH`. We need to ensure that that code is enabled in all AAarch32 and AAarch64 builds that use dynamic feature detection. Note that the Perl code seems to say that it is 64-bit only. I'm not sure if that's intended.

briansmith:
See https://llvm.org/bugs/show_bug.cgi?id=26016.

briansmith:
See also https://boringssl-review.googlesource.com/c/4411/1/crypto/aes/asm/aesv8-armx.pl#43, in particular the way that PerlAsm sometimes treats `.arch` as a no-op.

briansmith:
Depends on the refactoring in #408, at least.

pietro:
After #473 I think we can always define `BORINGSSL_CLANG_SUPPORTS_DOT_ARCH` on android.

myfreeweb:
I recently landed FreeBSD/aarch64 support in stdsimd (by using the `mrs` instruction in userspace) https://github.com/rust-lang-nursery/stdsimd/pull/611 — that method is also supported on Linux [since 4.11](https://github.com/torvalds/linux/commit/4aa8a472c33f4024381ade56bef008a9c566366a#diff-9764502b1d5418a9c00b3b235fd51bdc).

PR #892, opened 2019-08-25
Add AArch64 feature detection for FreeBSD

myfreeweb:
and fix build on any ARM platform that's not Linux/Fuchsia.

I hope I didn't screw up any platforms.. :D

The `all(target_os = "freebsd", target_arch = "aarch64")` inside of an `all(…, any(target_arch = "arm", target_arch = "aarch64"))` is to exclude FreeBSD/arm 32-bit.

cc #387

```
# before
test digest::sha256::_8192                                    ... bench:      47,207 ns/iter (+/- 228) = 173 MB/s
test aead::seal_in_place::aes_128_gcm::tls12_16               ... bench:  12,084,200 ns/iter (+/- 15,248,988)

# after
test digest::sha256::_8192                                    ... bench:       8,350 ns/iter (+/- 142) = 981 MB/s
test aead::seal_in_place::aes_128_gcm::tls12_16               ... bench:     722,151 ns/iter (+/- 917,484)
```

### DIFF
--- a/crypto/cpu-aarch64.c
+++ b/crypto/cpu-aarch64.c
@@ -1,0 +1,32 @@
+// Copyright 2019 Greg V
+//
+// Permission to use, copy, modify, and/or distribute this software for any
+// purpose with or without fee is hereby granted, provided that the above
+// copyright notice and this permission notice appear in all copies.
+//
+// THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHORS DISCLAIM ALL WARRANTIES
+// WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY
+// SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+// OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+
+// Run-time feature detection for aarch64 on any OS that emulates the mrs instruction.
+//
+// On FreeBSD >= 12.0, Linux >= 4.11 and other operating systems, it is possible to use
+// privileged system registers from userspace to check CPU feature support.
+//
+// For proper support of SoCs where different cores have different capabilities
+// the OS has to always report only the features supported by all cores, like FreeBSD does.
+//
+// Only FreeBSD uses this right now.
+
+#include <stdint.h>
+
+uint64_t GFp_aarch64_read_isar0(void) {
+  uint64_t val;
+  __asm __volatile("mrs %0, ID_AA64ISAR0_EL1" : "=&r" (val));
+  return val;
+}
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -170,10 +170,10 @@ pub(crate) mod arm {
 
     pub(crate) struct Feature {
         #[cfg_attr(
-            any(
-                target_os = "ios",
-                not(any(target_arch = "arm", target_arch = "aarch64"))
-            ),
+            not(all(
+                any(target_os = "android", target_os = "linux", target_os = "fuchsia"),
+                any(target_arch = "arm", target_arch = "aarch64")
+            )),
             allow(dead_code)
         )]
         mask: u32,
@@ -198,7 +198,10 @@ pub(crate) mod arm {
                 return self.mask == self.mask & unsafe { GFp_armcap_P };
             }
 
-            #[cfg(not(any(target_arch = "arm", target_arch = "aarch64")))]
+            #[cfg(not(all(
+                any(target_os = "android", target_os = "ios", target_os = "linux", target_os = "fuchsia"),
+                any(target_arch = "arm", target_arch = "aarch64")
+            )))]
             {
                 return false;
             }
@@ -206,7 +209,10 @@ pub(crate) mod arm {
     }
 
     // Keep in sync with `ARMV7_NEON`.
-    #[cfg(any(target_arch = "aarch64", target_arch = "arm"))]
+    #[cfg(all(
+        any(target_os = "android", target_os = "ios", target_os = "linux", target_os = "fuchsia"),
+        any(target_arch = "arm", target_arch = "aarch64")
+    ))]
     pub(crate) const NEON: Feature = Feature {
         mask: 1 << 0,
         ios: true,
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -60,6 +60,11 @@ pub(crate) fn features() -> Features {
             {
                 arm::fuchsia_setup();
             }
+
+            #[cfg(all(any(target_os = "freebsd"), target_arch = "aarch64"))]
+            {
+                arm::aarch64_mrs_setup();
+            }
         });
     }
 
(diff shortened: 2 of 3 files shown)


crypto/cpu-aarch64.c
```diff
@@ -0,0 +1,46 @@
...
+// On FreeBSD >= 12.0, Linux >= 4.11 and other operating systems, it is possible to use
+// privileged system registers from userspace to check CPU feature support.
+//
+// For proper support of SoCs where different cores have different capabilities
+// the OS has to always report only the features supported by all cores, like FreeBSD does.
+//
+// Only FreeBSD uses this right now.
+
+#ifdef __FreeBSD__
+#pragma GCC diagnostic ignored "-Wgnu-statement-expression"
```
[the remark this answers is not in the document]
myfreeweb: This is because of the FreeBSD `armreg.h` macros, we won't need this if I switch to directly using assembly

crypto/cpu-aarch64.c
```diff
@@ -0,0 +1,46 @@
...
+//
+// On FreeBSD >= 12.0, Linux >= 4.11 and other operating systems, it is possible to use
+// privileged system registers from userspace to check CPU feature support.
+//
+// For proper support of SoCs where different cores have different capabilities
+// the OS has to always report only the features supported by all cores, like FreeBSD does.
+//
+// Only FreeBSD uses this right now.
+
+#ifdef __FreeBSD__
```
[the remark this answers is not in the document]
myfreeweb: No. `elf_aux_info` (like `getauxval` but slightly different call signature - with a buffer length) was not present on aarch64 for a long time. It is there in -CURRENT now, but not in any release.

This mechanism is better, it seems to be what ARM intended, it's more similar to x86 cpuid. Linux supports it since 4.11 too.

crypto/cpu-aarch64.c
```diff
@@ -0,0 +1,46 @@
...
+#pragma GCC diagnostic ignored "-Wgnu-statement-expression"
+#include <GFp/cpu.h>
+#include <GFp/arm_arch.h>
+#include <machine/armreg.h>
```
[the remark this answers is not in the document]
myfreeweb: Yeah, that's a good idea.

myfreeweb:
> Are you trying to get FreeBSD + Raspberry Pi working? If so, which Raspberry Pi?

I'm on a full aarch64 desktop (Marvell MACCHIATObin + AMD GPU), typing on it right now :)

As for testing, EC2 a1 instances are super easy if you have an AWS account.

- go to https://lists.freebsd.org/pipermail/freebsd-snapshots/
- → latest month
- → latest "New FreeBSD snapshots available: head" (and/or "…: stable/12") post
- scroll to "FreeBSD/aarch64 EC2 AMIs are available in the following regions"
- copy the AMI id for your region and use it to start an instance

briansmith:
I believe I have an EspessoBin sitting around. Is AAarch64 ported to that? If so, I could try that. AWS is a little unrealistic because I'm not going to rent a (vritual) machine to test this over time.

myfreeweb:
> EspressoBin

Yeah, that works, you can boot the generic memstick image, but you need a DTB built from our tree ([here](https://unrelentingtech.s3.dualstack.eu-west-1.amazonaws.com/armada-3720-espressobin.dtb) I just built one) and [here's an example u-boot command](https://lists.freebsd.org/pipermail/freebsd-arm/2019-August/020302.html) to boot. (That example uses USB, commands are a bit different for SD, and `s/armada-3720-db.dtb/armada-3720-espressobin.dtb` of course)

Also, for RPi3 and Pine64(-LTS), their specific memstick images should just work.

https://download.freebsd.org/ftp/snapshots/arm64/aarch64/ISO-IMAGES/13.0/

briansmith:
I suggest we pick Raspberry Pi 3 or 4 or EspressoBin if that works for you. Then the step-by-step instructions would include basically how to build the SD card image for FreeBSD from Linux starting with the `curl` or `wget` of whatever, the `dd` or whatever to produce the image, how to configure the image to allow remote (probably SSH) access over ethernet, the `dd` or whatever to write it onto an SD card, connecting to the device over SSH, building *ring*'s tests, getting them onto the device, running the tests on the device, and getting the results back to the host device.

myfreeweb:
dang, you're really serious about this testing thing, why? Most projects, like Rust itself, do not test on every os/arch combination but do merge the support.

> produce the image

You don't produce anything, just write the existing one o_0

> building ring's tests, getting them onto the device

Why would you not build everything on the device?? 0_o

This is the most step-by-step I can get (for RPi 3):

- `wget -c "https://download.freebsd.org/ftp/snapshots/ISO-IMAGES/12.0/FreeBSD-12.0-STABLE-arm64-aarch64-RPI3-20190829-r351571.img.xz"`
- `unxz FreeBSD-12.0-STABLE-arm64-aarch64-RPI3-20190829-r351571.img.xz`
- `dd if=FreeBSD-12.0-STABLE-arm64-aarch64-RPI3-20190829-r351571.img of=/dev/… bs=1m`
- (there's a way to unxz to stdout and dd from stdin, but whatever)
- insert the card, boot, it should get an IP address over DHCP on the ethernet port, and sshd will be running already
- (i'm not going to tell you how to find the IP address… okay - watch the serial console, or watch your router)
- `ssh freebsd@what.ever.ip.address`, password: `freebsd`
- `su`, password: `root`
- `pkg install -y rust git` (and whatever else you want but you shouldn't need anything else i think)
- `exit` the root shell
- `git clone`, `cd`, `cargo test` as usual

briansmith:
> dang, you're really serious about this testing thing, why? Most projects, like Rust itself, do not test on every os/arch combination but do merge the support.

To cut it short, I think this is a better way than that.

> Why would you not build everything on the device?? 0_o

I have a 32 core ~4.0ghz x86-64 machine that runs an operating system I'm familiar with, which is much faster than the espressobin that I've booted like once :)

t6:
Can this be merged?  I do not quite understand what the necessary condition to get this accepted is.  It has been ~8 months since there was any activity here. PR still applies cleanly.

crypto/cpu-aarch64.c
```diff
@@ -0,0 +1,32 @@
...
+// For proper support of SoCs where different cores have different capabilities
+// the OS has to always report only the features supported by all cores, like FreeBSD does.
+//
+// Only FreeBSD uses this right now.
+
+#include <stdint.h>
+
+uint64_t GFp_aarch64_read_isar0(void) {
+  uint64_t val;
+  __asm __volatile("mrs %0, ID_AA64ISAR0_EL1" : "=&r" (val));
```
briansmith, line 30: Is there a `std::arch` intrinsic for this now?

src/cpu.rs
```diff
@@ -168,12 +173,39 @@ pub(crate) mod arm {
...
+            fn GFp_aarch64_read_isar0() -> u64;
+        }
+
+        #[inline]
+        fn bits_shift(x: u64, high: usize, low: usize) -> u64 {
+            (x >> low) & ((1 << (high - low + 1)) - 1)
+        }
+
+        let mut features = 0;
+        let aa64isar0 = unsafe { GFp_aarch64_read_isar0() };
```
briansmith, line 188: How do we know it is safe to execute the `mrs` instruction?

My understanding is that a conservative person should use `getauxval()` (or equivalent) to determine whether it is safe to execute `mrs`. Am I wrong?

briansmith (commented):
I'm leaning towards using #1007 instead of this one due to the concerns I raise in my other comments. WDYT?

src/cpu.rs
myfreeweb: It's guaranteed to be safe on FreeBSD 12.0 and newer.

`HWCAP` auxval (#1007) has been merged into 12.2, so I guess it's fine now too. (Used to only be available in -CURRENT.)

crypto/cpu-aarch64.c
myfreeweb: No, but `stdarch` uses this for its own detection:

https://github.com/rust-lang/stdarch/blob/a129c40f56882cae535b6861be88d848a6848bb8/crates/std_detect/src/detect/os/aarch64.rs

(but that's nightly only right now)

src/cpu.rs
briansmith: Thanks for the respond. I am really liking the idea of using `getauxval` because then it works just like Linux and we can avoid adding this tiny chunk of C code, so I think we should continue with #1007.

briansmith:
Thanks for your PR. I'm going to close this as a duplicate of #1007. To be frank, it's because it's a lot easier for me to maintain it if it shares the logic of the Linux implementation. I have nothing against FreeBSD but I don't have any FreeBSD systems.

PR #900, opened 2019-09-18
Support WASI

nhynes:


josephlr:
This should (at the very least) add stuff to the CI to make sure that *ring* builds and the tests pass when run on WASI.

nhynes:
> add stuff to the CI 

I totally agree. I was actually just looking into doing this for both Wasm and WASI using Travis. They're not Travis envs, so they don't jive with the current, very cleanly written test runner script.

nhynes:
NB: the newly added tests pass. CI appears to be failing because [one of them](https://travis-ci.org/briansmith/ring/jobs/586418481) is flaky.

josephlr:
@nhynes I'm not sure why the arm tests are acting up, but we also need to actually _run_ the tests on WASI. It looks like your current CI changes just build the code but don't execute it.

nhynes:
Good point. Turns out that wasm32 won't work without somewhat involved runtime support until the GFp functions are moved to not-assembly.

briansmith:
@nhynes I reopened this because I also want to make forward progress on WASI support.

It is true that some functionality won't be available yet in WebAssembly, but that should actually change soon.

In the meantime, there is a lot of functionality that is available already: basically `ring::rand` and all the stuff that is based on only digest functions. People are using the stuff that's working in wasm32 applications today.

How are you yourself running (not just building) the tests in WASI? And which specific WASI environment are you targetting?

(It looks like we don't have the wasm32-unknown-unknown tests running in CI in this repo now. I'll see what I can

nhynes:
This is great news, thanks!

I'm compiling to a standard wasi-unstable environment to be used in a custom (blockchain) WASI runtime. I was trying to use ECDSA which is why lld couldn't prune the offending symbols.
 
Building the tests shows what's needed in not-asm:

First, you'll need the [wasi-sdk](https://github.com/CraneStation/wasi-sdk) to build the _ring_ C extensions. Then compile using cargo:

```
$ RANLIB=/opt/wasi-sdk/bin/llvm-ranlib CC=/opt/wasi-sdk/bin/clang c t --no-run --target wasm32-wasi
```

Finally, running [wasm-nm](https://crates.io/crates/wasm-nm) shows the required functions:

```
$ wasm-nm -i target/wasm32-wasi/debug/ring-*.wasm
i GFp_p256_scalar_mul_mont
i GFp_p256_scalar_sqr_rep_mont
i GFp_nistz256_add
i GFp_nistz256_mul_mont
i GFp_nistz256_sqr_mont
i GFp_nistz256_point_add
i GFp_nistz256_neg
i GFp_nistz256_point_add_affine
i GFp_nistz256_point_double
i GFp_poly1305_emit
i GFp_poly1305_blocks
i GFp_poly1305_init_asm
i GFp_ChaCha20_ctr32
i GFp_bn_mul_mont
i GFp_gcm_init_clmul
i GFp_gcm_ghash_clmul
i GFp_gcm_gmult_clmul
i GFp_aes_hw_set_encrypt_key
i GFp_aes_nohw_set_encrypt_key
i GFp_aes_nohw_encrypt
i GFp_aes_hw_encrypt
i GFp_aes_hw_ctr32_encrypt_blocks
i GFp_nistz256_select_w5
i GFp_nistz256_select_w7
```

josephlr:
For running `wasm32` tests, the [`getrandom` crate](https://github.com/rust-random/getrandom/) currently has CI for both `wasm32-wasi` and `wasm32-unknown-unknown`.

We use [`wasmtime`](https://github.com/CraneStation/wasmtime) for WASI and the [`wasm-bindgen` tools](https://github.com/rustwasm/wasm-bindgen) to run `wasm32-unknown-unknown` tests. See our [`.cargo/config`](https://github.com/rust-random/getrandom/blob/master/.cargo/config) and [CI script](https://github.com/rust-random/getrandom/blob/master/.travis.yml#L68-76).

This lets us test the more conventional `wasm32-unknown-unknown`/`wasm-bindgen` targets on Node, Chrome, and Firefox, as well as testing on the non-Web, non-JS WASI.

Obviously this is easier for us as we have no C code, but if you can figure out how to build *ring*, you should be able to use something like our setup.

dunnock:
If nobody is working on this PR I would like to continue working on this topic applying CI changes that @josephlr mentioned. Perhaps since there are only little code changes and most are CI changes which need to be rewritten I think it makes sense to start from scratch.

dunnock:
@nhynes @briansmith I still want to make it to WASI even with custom functions: at least it will allow to run TlsStream in a customized WASI runtime. Direct approach I am following so far is to compile test runner exporting C functions into WASI environment, so that tests compiled to wasm32-wasi target might call those functions via a wrapper over original `export "C"`.

I started building WASI runtime, though there are few complications:
 - code duplication around exported functions and data, see one of function implementations https://github.com/dunnock/ring/blob/wasi-support/util/wasi/src/poly1305.rs due to types are private to module. What would help is making those types public, unless there is a reason they should stay private
 - code duplication would lead to incompatibility changes
 - wasiringruntime is separate crate to not pollute ring codebase, though since it links to same C libraries it heavily reuses build script and C artifacts, e.g. https://github.com/dunnock/ring/blob/wasi-support/util/wasi/build.rs

I want to start discussion and get your opinion before spending more time on this, is it feasible approach? Is there any better way to work around those complexities? 

One of possible approaches would be to convert all crypto code to pure rust, though probably it won't provide required crypto warranties and is it too big effort for benefit?

Another possible options might be for wasi runtime would just export custom top level cryptography functions so that TlsStream would use those functions to make it work - what would be candidates?

Or probably WASI runtime should just export TlsStream Sink/Send...

build.rs
```diff
@@ -489,6 +490,16 @@ fn build_library(
                 .and_then(|f| f.to_str())
                 .expect("No filename"),
         );
+
+        if &target.os == "wasi" {
+            let ranlib_cmd = std::env::var("RANLIB");
```
briansmith, line 495: Please write a comment explaining why this chunk of code is here.

Wouldn't we need to output a `cargo:rerun-if-env-changed=RANLIB` here?

include/GFp/base.h
```diff
@@ -87,6 +87,9 @@
 #elif defined(__mips__) && defined(__LP64__)
 #define OPENSSL_64_BIT
 #define OPENSSL_MIPS64
+#elif defined(__wasm32__)
+#define OPENSSL_32_BIT
+#define OPENSSL_X86_64
```
briansmith, line 92: It definitely isn't right to define `OPENSSL_X86_64` for WASI.

We already have:
```
#elif defined(__wasm__)
#define OPENSSL_32_BIT
```

Is that not good enough? I think we can just revert this chunk.

mk/travis.sh
```diff
@@ -102,6 +102,11 @@ else
   target_dir=target/$TARGET_X/debug
 fi
 
+if [[ "$TARGET_X" =~ "wasm32-" ]]; then
+  cargo build -vv -j2 ${mode-} ${FEATURES_X-} --target=$TARGET_X
```
briansmith, line 106: I will leave a comment in the issue about the CI/CD.

src/rand.rs
```diff
@@ -270,6 +270,17 @@ mod sysrand_chunk {
     }
 }
 
+#[cfg(all(target_arch = "wasm32", target_os = "wasi",))]
+mod sysrand_chunk {
+    use crate::error;
+
+    #[inline]
+    pub fn chunk(dest: &mut [u8]) -> Result<usize, error::Unspecified> {
+        wasi::wasi_unstable::random_get(dest).map_err(|_| error::Unspecified)?;
```
briansmith, line 279: It seems like, rather than adding that heavyweight dependency, we should just call `__wasi_random_get` directly and avoid that crate.

briansmith (changes requested):

briansmith:
My plan is to, soon, have C or Rust implementations of all features of *ring*, so that all *ring* features will work on WebAssembly, MIPS, etc.

I understand that right now some people are working around the functions that *ring* is currently missing by substituting their own. This will break as we add the C/Rust versions of those functions to *ring*. Also, there are other changes in the pipeline that would break such an approach. And, such an approach has never been guaranteed to work; we frequently change the "ABI" of the internal functions of *ring*, so it's pretty unreasonable to plug in implementations from the outside.

PR #901, opened 2019-09-29
Add support for Redox

AdminXVII:
Add redox to the urandom for src/rand.rs. Use the `rand:` URI on Redox rather than `/dev/urandom`.

cc @jackpot51

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -310,7 +310,7 @@ spin = { version = "0.5.2", default-features = false }
 [target.'cfg(any(target_os = "android", target_os = "linux"))'.dependencies]
 libc = { version = "0.2.48", default-features = false }
 
-[target.'cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux", target_os = "netbsd", target_os = "openbsd", target_os = "solaris"))'.dependencies]
+[target.'cfg(any(target_os = "android", target_os = "redox", target_os = "freebsd", target_os = "linux", target_os = "netbsd", target_os = "openbsd", target_os = "solaris"))'.dependencies]
 lazy_static = { version = "1.3", default-features = false, optional = true }
 
 [target.'cfg(all(target_arch = "wasm32", target_vendor = "unknown", target_os = "unknown", target_env = ""))'.dependencies]
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -183,7 +183,8 @@ use self::sysrand_or_urandom::fill as fill_impl;
     target_os = "freebsd",
     target_os = "netbsd",
     target_os = "openbsd",
-    target_os = "solaris"
+    target_os = "solaris",
+    target_os = "redox"
 ))]
 use self::urandom::fill as fill_impl;
 
@@ -357,7 +358,8 @@ mod sysrand_or_urandom {
     target_os = "freebsd",
     target_os = "netbsd",
     target_os = "openbsd",
-    target_os = "solaris"
+    target_os = "solaris",
+    target_os = "redox"
 ))]
 mod urandom {
     use crate::error;
@@ -368,9 +370,14 @@ mod urandom {
 
         use lazy_static::lazy_static;
 
+        #[cfg(not(target_os = "redox"))]
+        const RAND_LOCATION: &str = "/dev/urandom";
+        #[cfg(target_os = "redox")]
+        const RAND_LOCATION: &str = "rand:";
+
         lazy_static! {
             static ref FILE: Result<std::fs::File, std::io::Error> =
-                std::fs::File::open("/dev/urandom");
+                std::fs::File::open(RAND_LOCATION);
         }
 
         match *FILE {


jackpot51:
@briansmith Any news on this?

briansmith:
@jackpot51 Does redox provide an API that's more like `getrandom()` in Linux? PRNG APIs that require file handles are really dangerous and a last resort option. It seems like you are a leader of the Redox project; if so, I would be happy to discuss this with you on a call or via email.

jackpot51:
At the moment, no. Why are file handles considered dangerous?

briansmith:
> At the moment, no. Why are file handles considered dangerous?

POSIX requires something along the lines of "The file descriptor returned by a successful call [to open] will be the lowest-numbered file descriptor not currently open for the process." I'm guessing Redox has a similar constraint.

The main problem I'm aware of is when somebody accidentally closes the file descriptor that the crypto library is assuming to be valid for `rand:`, and then opens a different file with that file descriptor. Then the crypto library will read from that other file instead of `/dev/urandom`. For example, let's say the crypto library opens `rand:` and gets file descriptor 3. Then somebody loops through all open file descriptors and closes them (I have seen this happen in real programs), and then opens three files. Now *ring* will still read from file descriptor 3 but that will be the wrong file.

This issue comes up in almost every crypto library. See https://bugs.python.org/issue21207 for example. Thus I recommend every operating system provide a syscall like `getrandom()` instead of relying on a file interface.

jackpot51:
Redox is a microkernel. By design most things go through file handles.

briansmith:
> Redox is a microkernel. By design most things go through file handles.

I understand the appeal of the design that's been chosen so far. However, I do want to emphasize this is a significant downside. Zircon is a microkernel-based OS too, and provides a `getrandom()`-like API called `cprng_draw`.

Going back to *ring*, I am hoping to soon deprecate and/or remove the fallback to the file-based approach on other operating systems because I see this as too much of a hazard based on the real-world failures I have seen. Obviously, programs/libraries who close the file descriptors they shouldn't close are severely broken. However, it's really unfortunate when a library like *ring*, that is otherwise able to protect itself very well, has no defense against this. So I hope a new syscall interface for Redox is considered.

That said, I'm not going to use this PR as a means of forcing the issue. When I loop back around to adding support for new targets I will merge this as-is is there is no better alternatives.

briansmith:
A bigger issue that is likely to slow this down: I don't want to merge support new targets for which we do not have CI/CD coverage. So first I will need to learn how to set up a CI/CD environment that runs on a Linux host. IDK when I will get around to that. It would be very nice if somebody could add Redox to `cargo cross` or otherwise help with the CI/CD.

briansmith:
On Twitter, @dcarosone suggested that some operating systems might provide a "do not close" `fctl()`. I do think it might make sense to be ask `open()` to reject all attempts to close or reuse the file descriptor, similarly to how we can request the file descriptor is closed on `exec()`. Even better, a way to reject all operations on the file descriptor other than `read()`.

jackpot51:
I'll think about kernel randomness, and a syscall for getrandom. There are other good uses for kernel randomness, and there is probably a reasonable way to have a kernel random source get fed entropy from userspace drivers.

As for CI, it is pretty simple with the [redoxer](https://gitlab.redox-os.org/redox-os/redoxer/) tool. The readme describes its use. To test this PR, for example, use the following:

```
git clone https://github.com/AdminXVII/ring -b fix-redox ring_redoxer
cd ring_redoxer/
cargo install redoxer
redoxer toolchain
redoxer test
```

Redoxer currently uses a nightly rustc that may not compile ring with warnings forbidden. This is due to use of inline assembly in Redox projects for hardware access, such as drivers and the kernel. Hopefully this is fixed soon with the new asm! macro.

jackpot51:
After allowing warnings in build.rs and src/lib.rs, I was able to run the tests with Redoxer successfully

briansmith:
> I'll think about kernel randomness, and a syscall for getrandom. There are other good uses for kernel randomness, and there is probably a reasonable way to have a kernel random source get fed entropy from userspace drivers.

Great!

> As for CI, it is pretty simple with the redoxer tool. The readme describes its use. To test this PR, for example, use the following:
>
> ```
> git clone https://github.com/AdminXVII/ring -b fix-redox ring_redoxer
> cd ring_redoxer/
> cargo install redoxer
> redoxer toolchain
> redoxer test
> ```

Excellent!

briansmith:
This should go in mk/install-build-tools.sh when the target is redox:
> ```
> git clone https://github.com/AdminXVII/ring -b fix-redox ring_redoxer
> cd ring_redoxer/
> [check out a specific revision here]
> cargo install redoxer
> redoxer toolchain
> redoxer test
> ```

Change mk/cargo.sh to do something like this for redox:
> ```
>    export CC_x86_64_unknown_linux_musl=clang-10
>    export AR_x86_64_unknown_linux_musl=llvm-ar-10
>    export CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_LINKER=clang-10
> ```

Update .github/workflows/ci.yml to add the redox target. aarch64-linux-musl is probably a good example to follow.

This PR is old and uses `lazy_static!` so the code changes will also need to be rebased.

briansmith (changes requested):
I've unfortunately kind of given up on the idea of having CI/CD for every supported target. Similarly, I've also relaxed my stance on "/dev/urandom" and similar. 

Consequently, if this gets rebased on top of main and the minor issues I commented on are fixed, I'll merge it.

I'll file an issue against Redox suggesting Redox adopt an API more like Fuchsia's.

briansmith:
> I'll file an issue against Redox suggesting Redox adopt an API more like Fuchsia's.

An issue of that sort was already filed in the issue tracker. I added a comment to it: https://gitlab.redox-os.org/redox-os/redox/-/issues/816#note_22234

jackpot51:
https://github.com/briansmith/ring/pull/1341 fixed this already. Good to hear you will be using getrandom.

PR #903, opened 2019-10-08
Fix iOS builds for 32bit arm targets (armv7/armv7s)

kylefleming:
This PR does the following:
* Fixes the `AES` and `PMULL` CPU feature configuration for 32-bit iOS arm targets (namely, armv7 and armv7s).
* Fixes 3 "unused" errors when compiling for `ios` targets

The `AES` and `PMULL` CPU features are only supported on Armv8 (target_arch `aarch64`) and above. These features should be declared as not supported for Armv7 and Armv7s (target_arch `arm`). This is self-documented by the preceeding comment which references `ARMV8_AES` and `ARMV8_PMULL`, respectively.

### DIFF
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -206,6 +206,7 @@ pub(crate) mod arm {
     }
 
     // Keep in sync with `ARMV7_NEON`.
+    #[cfg_attr(not(target_arch = "arm"), allow(dead_code))]
     #[cfg(any(target_arch = "aarch64", target_arch = "arm"))]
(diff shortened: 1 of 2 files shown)


kylefleming:
Also, I'm not really sure how to increase code coverage for cross-compilation logic. Is there anything specific you'd like to see?

briansmith:
@kylefleming Sorry I never responded to this. I think 32-bit ARM iOS is dead and I don't think there is interest any more in it so I'm going to just close this. Thanks for the PR though.

PR #904, opened 2019-10-11
Feature gate web-sys

Jake-Shadle:
The `web-sys` crate brings in a ton of extra dependencies unconditionally, this just feature gates `web-sys` behind the `wasm` feature so that users can opt in to those extra dependencies, if they are even targeting wasm at all.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -314,7 +314,7 @@ libc = { version = "0.2.48", default-features = false }
 lazy_static = { version = "1.3", default-features = false, optional = true }
 
 [target.'cfg(all(target_arch = "wasm32", target_vendor = "unknown", target_os = "unknown", target_env = ""))'.dependencies]
-web-sys = { version = "0.3.25", default-features = false, features = ["Crypto", "Window"] }
+web-sys = { version = "0.3.25", default-features = false, features = ["Crypto", "Window"], optional = true }
 
 [target.'cfg(target_os = "windows")'.dependencies]
 winapi = { version = "0.3.7", default-features = false, features = ["ntsecapi", "wtypesbase"] }
@@ -338,6 +338,7 @@ internal_benches = []
 slow_tests = []
 std = ["alloc"]
 test_logging = []
+wasm = ["web-sys"]
 
 # XXX: debug = false because of https://github.com/rust-lang/rust/issues/34122
 
--- a/build.rs
+++ b/build.rs
@@ -636,11 +636,13 @@ where
 
 fn run_command(mut cmd: Command) {
     println!("running {:?}", cmd);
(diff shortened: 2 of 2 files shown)


Jake-Shadle:
Probably makes sense to rename the feature from `wasm` to `wasm-web` or something to take into account the WASI use case coming.

briansmith:
@Jake-Shadle Thanks for filing the issue. However, I don't understand the problem.

You wrote:

> The web-sys crate brings in a ton of extra dependencies unconditionally, this just feature gates web-sys behind the wasm feature so that users can opt in to those extra dependencies, if they are even targeting wasm at all.

The Cargo.toml already has:

> ```
> [target.'cfg(all(target_arch = "wasm32", target_vendor = "unknown", target_os = "unknown", target_env = ""))'.dependencies]	[target.'cfg(all(target_arch = "wasm32", target_vendor = "unknown", target_os = "unknown", target_env = ""))'.dependencies]
> web-sys = { version = "0.3.25", default-features = false, features = ["Crypto", "Window"] }
> ```

So, doesn't the fact that this is a target-specific dependency already do what is proposed here?

Jake-Shadle:
It doesn't build them unless you build for wasm, but it does impact dependency resolution and is included as a part of the lock file.

briansmith:
> It doesn't build them unless you build for wasm, but it does impact dependency resolution and is included as a part of the lock file.

OK, but why does that matter? I would prefer a messy Cargo.lock over messy code/configuration logic, if all other considerations are equal.

In another project I do have an issue where we're trying to generate "dependency reports" for security auditing purposes and we have to be careful that we are doing that kind of stuff on a per-target (per-configuration, really) basis instead of just using a tool that reads Cargo.lock and assumes every dependency mentioned there is actually used. I think some tools need to improve in this regard. Is your issue similar to this?

Jake-Shadle:
Yes, our situation is basically the same with [cargo-deny](https://github.com/EmbarkStudios/cargo-deny), where features are the cleanest way to keep your dependency graph simple. However, the target configuration will be accessible [more readily](https://github.com/rust-lang/cargo/pull/7132) from `cargo metadata` in 1.41, so we'll be fine once that lands.

briansmith:
> Yes, our situation is basically the same with [cargo-deny](https://github.com/EmbarkStudios/cargo-deny), where features are the cleanest way to keep your dependency graph simple. However, the target configuration will be accessible [more readily](https://github.com/rust-lang/cargo/pull/7132) from `cargo metadata` in 1.41, so we'll be fine once that lands.

Great. I commented in the cargo-deny issue: https://github.com/EmbarkStudios/cargo-deny/issues/63.

I'm going to close this on the assumption that the improvements to cargo and other tools will make this unnecessary to do.

Issue #752, opened 2019-01-04
Switch from Yasm to NASM on Windows


briansmith:
See https://boringssl.googlesource.com/boringssl/+/73d69f4d51086da93ea10c998b14e05611910972.

briansmith:
This was done in PR #1123.

PR #905, opened 2019-10-12
Add support for nasm

Jake-Shadle:
I was having issues with yasm on Windows, and noticed #752, so I just tried out nasm instead and everything seemed to work.

Resolves: #752

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -520,18 +520,13 @@ fn compile(
     if ext == "obj" {
         p.to_str().expect("Invalid path").into()
     } else {
-        let use_nasm = std::env::var("CARGO_FEATURE_NASM").map(|s| s == "1").unwrap_or(false);
         let mut out_path = out_dir.clone().join(p.file_name().unwrap());
         assert!(out_path.set_extension(target.obj_ext));
         if need_run(&p, &out_path, includes_modified) {
             let cmd = if &target.os != WINDOWS || ext != "asm" {
                 cc(p, ext, target, warnings_are_errors, &out_path)
             } else {
-                if use_nasm {
-                    nasm(p, &target.arch, &out_path)
-                } else {
-                    yasm(p, &target.arch, &out_path)
-                }
+                nasm(p, &target.arch, &out_path)
             };
 
             run_command(cmd);
@@ -642,7 +637,6 @@ fn nasm(file: &Path, arch: &str, out_file: &Path) -> Command {
     let _ = c
         .arg("-X")
         .arg("vc")
-        //.arg("--dformat=cv8")
         .arg(oformat)
         .arg("-o")
         .arg(out_file.to_str().expect("Invalid path"))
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -331,13 +331,15 @@ cc = { version = "1.0.37", default-features = false }
 
 [features]
 # These features are documented in the top-level module's documentation.
-default = ["alloc", "dev_urandom_fallback"]
+default = ["alloc", "dev_urandom_fallback", "nasm"]
 alloc = []
 dev_urandom_fallback = ["lazy_static"]
 internal_benches = []
+nasm = []
```
briansmith: There's no need to add an option to use `nasm`. We should just require `nasm` unconditionally instead of `yasm`.

Cargo.toml
```diff
@@ -331,13 +331,15 @@ cc = { version = "1.0.37", default-features = false }
...
 slow_tests = []
 std = ["alloc"]
 test_logging = []
+wasm = ["web-sys"]
```
briansmith: Please rebase this MR on master directly, instead of basing it on the other PR you submitted. Then we can deal with them independently.

mk/appveyor.bat
```diff
@@ -36,6 +36,16 @@ if %ERRORLEVEL% NEQ 0 (
   exit 1
 )
 
+echo Downloading Nasm...
```
briansmith: Let's remember to delete the downloading of yasm.

briansmith:
@Jake-Shadle Are you interested in updating this based on the above feedback?

build.rs
```diff
@@ -634,11 +632,13 @@ where
 
 fn run_command(mut cmd: Command) {
     println!("running {:?}", cmd);
-    let status = cmd.status().unwrap_or_else(|e| {
+    let output = cmd.output().unwrap_or_else(|e| {
```
briansmith: Why did you switch from `status()` to `output()`?

Jake-Shadle: I think that was in there from when I was debugging why yasm was failing, I'll revert.

briansmith:
Thanks for the PR and thanks for updating it in response to my previous comment. I haven't had time to test it out but I will do so if/when you rebase it.

I think we should check to see which version of nasm BoringSSL is using and use the same version.

Jake-Shadle:
I updated the version to the one that boringssl ostensibly is using as seen https://boringssl.googlesource.com/boringssl/+/HEAD/util/bot/UPDATING#56

briansmith:
@Jake-Shadle Thanks for the PR. I had to rewrite the CI/CD to replace Travis CI and AppVeyor with GitHub Actions. That makes your PR unmergable. I'm going to close this. I have written a new PR #1123 that attempts to accomplish the same thing yours does. I would appreciate it if you could take a look.

PR #912, opened 2019-10-26
Added TestRandom: Deterministic random for testing.

realcr:
Added TestRandom: Deterministic test random, relying on calling an inner `Fn` function.
This is a proposal to solve issue https://github.com/briansmith/ring/issues/661

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -570,9 +570,15 @@ mod tests {
         }
     }
 
+    /// Make sure that a given argument implements Debug
+    fn check_debug(_d: &impl core::fmt::Debug) {}
+
     #[test]
     fn test_test_random() {
         let my_test_random = MyTestRandom;
+
+        // Use the Debug trait, to make test coverage happy:
+        check_debug(&my_test_random);
 
         let mut array = [0u8; 16];
         my_test_random.fill(&mut array[..]).unwrap();
--- a/src/test.rs
+++ b/src/test.rs
@@ -537,11 +537,51 @@ pub mod rand {
             assert_eq!(unsafe { *self.current.get() }, self.bytes.len());
         }
     }
+
+    /// A trait for a testing random generator. May only be used for testing
+    pub trait TestRandom {
+        fn fill_impl(&self, dest: &mut [u8]) -> Result<(), error::Unspecified>;
+    }
+
+    impl<T: TestRandom + core::fmt::Debug> rand::sealed::SecureRandom for T {
+        fn fill_impl(&self, dest: &mut [u8]) -> Result<(), error::Unspecified> {
+            self.fill_impl(dest)
+        }
+    }
 }
 
 #[cfg(test)]
 mod tests {
+    use super::rand::TestRandom;
+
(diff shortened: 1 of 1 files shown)


realcr:
I think that the coverage check did good job here. I tried to add a test, but never managed to do it. It seems like SecureRandom requires that the underlying structure implements core::fmt::Debug, and it seems to be more difficult than I thought to implement it for function closures, or usual functions.

I tried to implement `Debug` directly like this:
```rust
    /// An implementation of `SecureRandom` that fills the output slice
    /// by calling an inner function.
    pub struct TestRandom<F>(pub F);

    impl<F> core::fmt::Debug for TestRandom<F> {
        fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
            write!(f, "TestRandom")
        }
    }

    impl<F> rand::sealed::SecureRandom for TestRandom<F>
    where
        F: Fn(&mut [u8]) -> Result<(), error::Unspecified> + core::fmt::Debug,
    {
        fn fill_impl(&self, dest: &mut [u8]) -> Result<(), error::Unspecified> {
            self.0(dest)
        }
    }
```

But it seems like it is still not enough to make the compiler happy:

```rust
$ cargo test
   Compiling ring v0.16.9 (/home/real/external/ring)
error[E0277]: `[closure@src/test.rs:587:24: 592:10]` doesn't implement `core::fmt::Debug`
   --> src/test.rs:596:30
    |
572 |     fn assert_secure_random<F: SecureRandom>(f: &F) 
    |        --------------------    ------------ required by this bound in `test::tests::assert_secure_random`
...
596 |         assert_secure_random(&test_random);
    |                              ^^^^^^^^^^^^ `[closure@src/test.rs:587:24: 592:10]` cannot be formatted using `{:?}` because it doesn't implement `core::fmt::Debug`
    |
    = help: the trait `core::fmt::Debug` is not implemented for `[closure@src/test.rs:587:24: 592:10]`
    = note: required because of the requirements on the impl of `rand::sealed::SecureRandom` for `test::rand::TestRandom<[closure@src/test.rs:587:24: 592:10]>`
    = note: required because of the requirements on the impl of `rand::SecureRandom` for `test::rand::TestRandom<[closure@src/test.rs:587:24: 592:10]>`
```

Strange. Do you have an idea around this issue?

realcr:
- Updated implementation to use a `TestRandom` trait instead of a `TestRandom` struct. This should allow more flexibility for the user when designing a deterministic random generator. 
- Added a basic test that uses the `TestRandom` trait.

realcr:
There is something very confusing about the output of coveralls here. Do you have an idea of how to solve it?

briansmith:
@realcr I think you don't need this anymore, and I want to delete the old "master" branch today, so I'm closing this.

realcr:
Thanks. Just to make sure, this pull request is indeed outdated, but I am still very much interested in solving issue https://github.com/briansmith/ring/issues/661
Having deterministic tests is so important for the codebase I am working on, that I had to opt out of using ring for the time being.

PR #913, opened 2019-10-30
Use slice::iter

lzutao:
See clippy::into_iter lint and https://github.com/rust-lang/rust/pull/65819

PR #914, opened 2019-11-01
Remove `#[inline]` attributes on function prototypes

LukasKalbertodt:
These attributes now lead to a compile error on nightly. See [this Rust PR](https://github.com/rust-lang/rust/pull/65294) for more information.

### DIFF
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -57,7 +57,6 @@ pub trait BoundKey<N: NonceSequence>: core::fmt::Debug {
     fn new(key: UnboundKey, nonce_sequence: N) -> Self;
 
     /// The key's AEAD algorithm.
-    #[inline]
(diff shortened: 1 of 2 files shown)


LukasKalbertodt:
Could this and #913 be merged soon? Both are just tiny fixes. Though I saw that this crate seems to be unmaintained in the recent months? Is @briansmith the only one who can merge? If you read that, @briansmith, could you consider getting [the Rust Bus](https://users.rust-lang.org/t/bus-factor-1-for-crates/17046) on board? I understand very well if you can't maintain this crate at the moment; life can always get in the way. But since `ring` has become a pretty important part of the Rust ecosystem, it's not really healthy if it's unmaintained for too long.

(The CI failure in this case is just from coveralls. But obviously, this change did not actually decrease or increase coverage.)

briansmith:
Thanks!

PR #915, opened 2019-11-13
Add sign_digest to EcdsaKeyPair to allow signing of pre-digests messages.

IanLuites:
I need to sign a relatively large message and was using `digest::Context` to piece wise stream and update the digest.

I did however find out that `sign` does not accept a `digest`, but only an undigested `message`.

This PR would add a `sign_digest` sibling, which verifies the digest is generated with the correct (matching) algorithm, and then signs it.


src/ec/suite_b/ecdsa/signing.rs
```diff
@@ -163,6 +165,23 @@ impl EcdsaKeyPair {
...
+    /// generated by `rng`.
+    ///
+    /// The `digest` algorithm must match that of the signing algorithm.
+    pub fn sign_digest(
+        &self,
+        rng: &dyn rand::SecureRandom,
+        digest: digest::Digest,
+    ) -> Result<signature::Signature, error::Unspecified> {
+        // Step 4 (out of order, already performed by caller).
+        if digest.algorithm() == self.alg.digest_alg {
```
briansmith, line 178: Please use the early return style: `if ... != ... { return Err(...); }`

src/ec/suite_b/ecdsa/signing.rs
```diff
@@ -435,6 +454,44 @@ mod tests {
         );
     }
 
+    #[test]
```
briansmith, line 457: Tests of the public API should be done in tests/ecdsa_tests.rs.

I suggest that you add a constructor `pub(crate) try_from_test_vector` to `ring::digest::Digest` that accepts a digest algorithm and a precomputed value. Then you can rewrite the existing tests to use your new function instead of calling `sign_` directly as they currently do.

src/ec/suite_b/ecdsa/signing.rs
```diff
@@ -163,6 +165,23 @@ impl EcdsaKeyPair {
         self.sign_(rng, h)
     }
 
+    /// Returns the signature of the `digest` using a random nonce
+    /// generated by `rng`.
+    ///
+    /// The `digest` algorithm must match that of the signing algorithm.
+    pub fn sign_digest(
```
briansmith, line 172: Please name this `sign_digest_less_safe`.

Please add a note like the following to the documentation: "In general, it is not safe to sign an arbitrary digest. Ensure that you only sign digests that you have computed yourself, or that you otherwise know are safe to sign. It could be a bad mistake to sign an attacker-controlled digest."

briansmith (changes requested):
Thanks for doing this. I'll get this merged right away if you make the changes I requested.

PR #917, opened 2019-11-23
SystemRandom: Fix #326 - Prefer getrandom(2) on FreeBSD

### LINKED ISSUE #326: Use something better than /dev/urandom on FreeBSD
The main drivers here are performance and the ability to work correctly and automatically in a chroot/jail.

This was split off from #316, which is now OpenBSD-only. Note that iOS and MacOS are issue #149.

See:
* https://github.com/briansmith/ring/issues/316#issuecomment-257479426
* https://github.com/briansmith/ring/issues/316#issuecomment-257484172, which among other things references https://paragonie.com/blog/2016/05/how-generate-secure-random-numbers-in-various-programming-languages.
* What LibreSSL does and why.
* What libsodium does and why.
* CVE-2014-2970
* https://lists.freebsd.org/pipermail/freebsd-security/2014-July/007849.html
* https://lists.freebsd.org/pipermail/freebsd-security/2014-July/007853.html
* https://lists.freebsd.org/pipermail/freebsd-security/2014-July/007868.html

From reading various FreeBSD mailing list messages, it seems like FreeBSD doesn't have a good way to guarantee fork-safety, which is the same problem that Linux has. Therefore, it seems like we should always be getting random values from the OS. That means, AFAICT, either reading from `/dev/[u]random` or the `KERN_ARND` `sysctl`. Note the potential problems with `KERN_ARND` mentioned in the linked-to comments above.

I don't understand this issue fully, but it seems that FreeBSD has a RANDOM kernel module that can be disabled, and in that disabled state it may be problematic to do anything better than reading from `/dev/[u]random`: "RW mentioned kernels without RANDOM, being an awkward situation for which it seems necessary to fall back to the PRNG in userland." - https://lists.freebsd.org/pipermail/freebsd-security/2014-July/007869.html.

cemeyer:
FreeBSD added the Linux-compatible getrandom(2) API in 12.0.  Like Linux,
prefer it to the /dev/urandom device, but fallback to the earlier method if
the syscall is not available.

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -165,7 +165,7 @@ impl crate::sealed::Sealed for SystemRandom {}
 
 #[cfg(any(
     all(
-        any(target_os = "android", target_os = "linux"),
+        any(target_os = "android", target_os = "freebsd", target_os = "linux"),
         not(feature = "dev_urandom_fallback")
     ),
     target_arch = "wasm32",
@@ -174,13 +174,12 @@ impl crate::sealed::Sealed for SystemRandom {}
 use self::sysrand::fill as fill_impl;
 
 #[cfg(all(
-    any(target_os = "android", target_os = "linux"),
+    any(target_os = "android", target_os = "freebsd", target_os = "linux"),
     feature = "dev_urandom_fallback"
 ))]
 use self::sysrand_or_urandom::fill as fill_impl;
 
 #[cfg(any(
-    target_os = "freebsd",
     target_os = "netbsd",
     target_os = "openbsd",
     target_os = "solaris"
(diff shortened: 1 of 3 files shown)


cemeyer:
(Of course, it builds and tests pass on FreeBSD CURRENT.  Please consider this contribution licensed under the project's preferred ISC license for new code.)

briansmith:
@Thomasdezeeuw Could you please review this? I think this may be a better solution than PR #1119.

src/rand.rs
```diff
@@ -242,6 +241,42 @@ mod sysrand_chunk {
...
+
+        // As long as we're using libc, it would be preferable if the libc crate provided
+        // getrandom() directly.  It would also be preferable to use __syscall rather than
+        // syscall().  However, at this time the Rust libc crate provides neither of these
+        // functions on FreeBSD.
+
+        // Clamp request size to INT_MAX due to limitation of the return type of libc::syscall()
+        // (c_int).
+        let chunk_len: c::size_t = dest.len().min(c_int::max_value() as usize);
+        let r = unsafe { libc::syscall(SYS_GETRANDOM, dest.as_mut_ptr(), chunk_len, 0) };
```
briansmith, line 264: Does FreeBSD have a stable syscall ABI, or is it one of those platforms where only libc has a stable ABI?

Thomasdezeeuw: I could only find https://wiki.freebsd.org/Releng/ABI, which says:

> We also try to maintain ABI compatibility across .0 releases, but they are not strictly enforced except for libraries that already implements versioned symbols.

Which I read as best effort, but not guaranteed.

Thomasdezeeuw: I'll see if I can add it to libc.

Thomasdezeeuw: PR for libc https://github.com/rust-lang/libc/pull/1982.

cemeyer: It's one of those platforms where only libc has a stable ABI.  The plan is to provide a stable `libsystem` ABI at some point in the near future for non-C programming languages, but I'm afraid we're not there yet.

cemeyer:
Thanks for looking at this.  (Or a better avenue through the libc crate — works for me either way!)

src/rand.rs
briansmith: > PR for libc rust-lang/libc#1982.

Great. It seems like we could just call `getrandom` directly from *ring* through FFI, without going through the libc crate, couldn't we? It seems like maybe we should go that route.

cemeyer: Sure, you could link FreeBSD libc directly in *ring* without using the libc crate.

briansmith:
Thanks for the PR. I'm planning to switch the implementation to use the `getrandom` crate as part of the goal of encapsulating/minimizing operating-system-specific code in *ring*; see draft PR #1531. I'm closing this in favor of that approach.

PR #919, opened 2019-12-05
Support VxWorks

BaoshanPang:
r? @n-salim
cc @UmeshKalappa


briansmith:
Interesting. How could I test this on the target? (I've never used vxworks.)

BaoshanPang:
@briansmith 
Sorry, I didn't see your comment until now.
>> Interesting. How could I test this on the target? (I've never used vxworks.)
Yes, this is always a problem when we start to work on rust with community.
We have test environment, but it can only be accessed inside Wind River company.

PR #920, opened 2019-12-10
Supprt VxWorks

BaoshanPang:


BaoshanPang:
These test seems not related to my change.

src/rand.rs
```diff
@@ -432,3 +435,32 @@ mod fuchsia {
         fn zx_cprng_draw(buffer: *mut u8, length: usize);
     }
 }
+
+#[cfg(target_os = "vxworks")]
```
BaoshanPang: The supporting for VxWorks have been added in crate getrandom. The code here is copied from https://github.com/rust-random/getrandom/blob/master/src/vxworks.rs.

briansmith:
Because we switched to using `getrandom`, it is now very easy to add vxworks support. Just add it to the list of supported operating systems in rand.rs and Cargo.toml.

PR #922, opened 2019-12-23
Fix warnings emitted by rustc 1.40.

briansmith:
Remove `#[inline]` from trait definitions.

During review, we found one `#[inline]` that should be `#[inline(always)]` in
sha2.rs. Fix that too.


briansmith:
Closing in favor of #914.

PR #924, opened 2019-12-23
Replace `lazy_static` and `spin` dependencies with `conquer-once`.

briansmith:


CryZe:
Is this not happening anymore, or why was this closed?

briansmith:
> Is this not happening anymore, or why was this closed?

I have other projects that use `once_cell` and so I'm using `once_cell` instead of `lazy_static`. The issue for replacing `spin-rs` is still open.

Issue #911, opened 2019-10-19
There is no way to construct hkdf::Algorithm with HMAC_SHA1_FOR_LEGACY_USE_ONLY


zonyitoo:
In #754 has decided to deprecate SHA1 in ring. But for legacy crates, we still need SHA1 for compatibility.

I hope SHA1 could be enabled by a feature.

briansmith:
@zonyitoo I would accept an MR to add this.

briansmith:
PR #926 addressed this.

PR #926, opened 2019-12-24
Add HKDF HMAC-SHA1 for lagacy use

zonyitoo:
Ref: #911

### DIFF
--- a/src/hkdf.rs
+++ b/src/hkdf.rs
@@ -32,6 +32,10 @@ impl Algorithm {
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/hkdf.rs
```diff
@@ -32,6 +32,10 @@ impl Algorithm {
     }
 }
 
+/// HKDF using HMAC-SHA-1
```
briansmith: Please append a period to be consistent with the other descriptions.

PR #930, opened 2020-01-02
Switch `dev_urandom_fallback` to use `once_cell` instead of `lazy_static!`

briansmith:
This is a step towards removing the spin-rs dependency. `lazy_static!`'s
conditial use of spin-rs based on a feature flag was particularly problematic
because often other dependencies would enable that feature flag even in cases
where the `std::sync::Once` implementation was preferable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -311,7 +311,7 @@ spin = { version = "0.5.2", default-features = false }
 libc = { version = "0.2.48", default-features = false }
 
 [target.'cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux", target_os = "netbsd", target_os = "openbsd", target_os = "solaris"))'.dependencies]
-lazy_static = { version = "1.3", default-features = false, optional = true }
+once_cell = { version = "1.3.1", default-features = false, features=["std"], optional = true }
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -311,7 +311,7 @@ spin = { version = "0.5.2", default-features = false }
 libc = { version = "0.2.48", default-features = false }
 
 [target.'cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux", target_os = "netbsd", target_os = "openbsd", target_os = "solaris"))'.dependencies]
-lazy_static = { version = "1.3", default-features = false, optional = true }
+once_cell = { version = "1.2.0", default-features = false, features=["std"], optional = true }
```
josephlr: Is there any reason to do this when the only `default-features` of once_cell are `std`.

You could just do `once_cell = { version = "1.2.0", optional = true }`

briansmith: I always use `default-features = false` since doing this in almost every project I join cuts the dependencies by 10-25%. Although, the issues with `spin-rs` and similar when disabling crates' `std` default features may cause me to reconsider this policy.

briansmith:
I updated this PR to use once_cell 1.3.1.

I saw https://github.com/matklad/once_cell/issues/1, which is about auditing once_cell. It seems like that kind of audit needs to be completed before we can rely on it. This is blocked on that happening.

PR #933, opened 2020-01-04
Remove deprecated Error::description and Error::cause

AnderEnder:
`Error::description` has been documented as soft-deprecated since 1.27.0 (17 months ago). It is going to be hard-deprecated soon. `Error::cause` has been documented as deprecated since 1.33.0.

This PR:
- Removes all implementations of `description` and `cause` in all error types

Related PR: https://github.com/rust-lang/rust/pull/66919

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -92,16 +92,7 @@ impl core::fmt::Display for Unspecified {
 }
 
 #[cfg(feature = "std")]
-impl std::error::Error for Unspecified {
-    #[inline]
-    fn cause(&self) -> Option<&dyn std::error::Error> {
-        None
-    }
-
-    fn description(&self) -> &str {
-        Self::description_()
-    }
-}
+impl std::error::Error for Unspecified {}
 
 impl From<untrusted::EndOfInput> for Unspecified {
     fn from(_: untrusted::EndOfInput) -> Self {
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -149,7 +140,7 @@ impl From<core::array::TryFromSliceError> for Unspecified {
 pub struct KeyRejected(&'static str);
 
 impl KeyRejected {
-    /// The value returned from <Self as std::error::Error>::description()
+    /// The value is used in Display
     pub fn description_(&self) -> &'static str {
```
briansmith: Again, I think `description_()` should be removed; its code should be inlined into the `Display` implementation.

src/error.rs
```diff
@@ -92,16 +92,7 @@ impl core::fmt::Display for Unspecified {
...
 
 #[cfg(feature = "std")]
-impl std::error::Error for Unspecified {
-    #[inline]
-    fn cause(&self) -> Option<&dyn std::error::Error> {
-        None
-    }
-
-    fn description(&self) -> &str {
-        Self::description_()
```
briansmith: `description_()` should be removed and inlined into the `Display` implementation.

briansmith:
@AnderEnder Thanks for the PR. I created a new PR that implements the changes I requested: #938, so I'm going to close this one.

PR #935, opened 2020-01-11
Add musl targets to CI.

cbiffle:
I've added musl-gcc here, and not musl-clang, because I can't figure out
what the package name for musl-clang is on Ubuntu.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -109,6 +109,28 @@ matrix:
           sources:
             - ubuntu-toolchain-r-test
 
+    - env: TARGET_X=x86_64-unknown-linux-musl CC_X=musl-gcc FEATURES_X= MODE_X=DEBUG KCOV=0 RUST_X=stable
+      rust: stable
+      os: linux
+      dist: trusty
+      addons:
+        apt:
+          packages:
+            - musl
+            - musl-dev
+            - musl-tools
+
+    - env: TARGET_X=x86_64-unknown-linux-musl CC_X=musl-gcc FEATURES_X= MODE_X=RELWITHDEBINFO KCOV=0 RUST_X=stable
+      rust: stable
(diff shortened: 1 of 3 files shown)


briansmith:
@cbiffle Your PR is still marked as a draft. Is it ready to be reviewed?

briansmith:
OK, I looked at this and it seems OK for now. However, I would really appreciate it if somebody could get it working with clang as the toolchain instead of musl-gcc. At least the last time I tried it, GCC didn't do a good job optimizing some of the multi-precision arithmetic. My understanding is that to build with clang, we "just" need to use `TARGET_CC=clang`

```sh
$ sudo apt remove musl musl-dev musl-tools
$ sudo apt install musl-dev
$ TARGET_CC=clang cargo test --target=x86_64-unknown-linux-musl
```

Actually, for me it worked even without the `sudo apt install musl-dev` step.

briansmith:
> I've added musl-gcc here, and not musl-clang, because I can't figure out
> what the package name for musl-clang is on Ubuntu.

My understanding is there is no "musl-clang". Instead, clang always works as a cross compiler, so installing any(?) version of clang will install one that can target musl. Even this ancient one was able to do it, as this is the version I used in my previous comment:

```
$ clang --version
clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

briansmith:
Given the aforementioned comments and also the fact that I'm trying to remove the `master` branch today, I'm closing this MR. I would love to see a new, simpler, version of this that uses clang instead of GCC.

briansmith:
Also, @cbiffle, thanks for your work on this!

PR #937, opened 2020-01-23
Fix RSA verification panic for modulus with length not evenly divisible by `LIMB_BYTES`

briansmith:


src/rsa/verification.rs
```diff
@@ -282,8 +283,15 @@ pub(crate) fn verify_rsa_(
...
-    m.fill_be_bytes(decoded);
+    let decoded = {
+        let n_bytes = n_bits.as_usize_bytes_rounded_up();
+        let n_bytes_padded = ((n_bytes + (LIMB_BYTES - 1)) / LIMB_BYTES) * LIMB_BYTES;
+        let decoded = &mut decoded[..n_bytes_padded];
+        m.fill_be_bytes(decoded);
+        let (padding, decoded) = decoded.split_at(n_bytes_padded - n_bytes);
+        assert!(padding.iter().all(|&b| b == 0));
+        decoded
+    };
```
briansmith: I'm not particularly happy with how ugly this is or how leaky the abstractions are. For example, this code is only correct assuming we've done various bounds checks that indicate the RSA modulus length is reasonable.

OTOH, for validation reasons it is important to avoid making the code in `arithmetic` and `limbs` more complicated than it is.

In the future, after the release that contains this fix, we should clean this up.

briansmith:
I posted this to get early feedback. Tests should arrive soon, followed by the release.

andrewtj:
I've tried this out with some samples generated using Go's crypto package and have had no issues.

briansmith:
> I posted this to get early feedback. Tests should arrive soon, followed by the release.

I did test this but I decided not to include the tests in the PR because there is a refactoring of the tests in progress and I don't want to interfere with that refactoring. The tests should appear later in a separate PR. Thanks @andrewtj for reporting this and for testing it.

PR #938, opened 2020-01-24
Remove definitions of deprecated `Error::description()` and `Error::cause()`.

briansmith:


briansmith:
This seems like a SemVer breaking change so I'll defer to the next major version bump to do this.

akhilles:
I don't think this is a breaking change at the API-level since `Error::source()` was introduced in Rust 1.30.0 (I'm assuming the MSRV is > 1.30). Also, `std::error::Error` still provides default implementations for `Error::description()` and `Error::cause()`:

```rust
fn description(&self) -> &str {
    "description() is deprecated; use Display"
}

fn cause(&self) -> Option<&dyn Error> {
    self.source()
}
```

The behavior of `Error::cause()` would be unaffected, but the description string would change from `"ring::error::Unspecified"` to `"description() is deprecated; use Display"`. Would that be considered a breaking change?

briansmith:
@akhilles Thanks for your help!

PR #939, opened 2020-01-24
Speed up SHA1 computation

LingMan:
The first commit is just a minor style fix that doesn't really warrant it's own PR.

The second commit speeds up SHA1 computation by ~50-60%. See the commit message for details. Unfortunately it does make the code a bit uglier than I initially expected because Rust doesn't support [destructuring assignments](https://github.com/rust-lang/rfcs/issues/372) (yet?). Still, even for an implementation like ring's, that favors simplicity over speed, the trade-off  is worth it in my opinion.

There's an alternative with much nicer code and the same speedup in [this branch](https://github.com/LingMan/ring/commits/sha1-speedup-crunchy). However, that turned out to increase the binary size by almost 390 KiB. Since ring is targeting small devices, that seamed unacceptable. Therefore I'm PRing this solution instead.

Before:
test digest::sha1::_1000     ... bench:       4,471 ns/iter (+/- 39) = 223 MB/s
test digest::sha1::_16       ... bench:         315 ns/iter (+/- 2) = 50 MB/s
test digest::sha1::_2000     ... bench:       8,903 ns/iter (+/- 208) = 224 MB/s
test digest::sha1::_256      ... bench:       1,452 ns/iter (+/- 14) = 176 MB/s
test digest::sha1::_8192     ... bench:      35,563 ns/iter (+/- 348) = 230 MB/s
test digest::sha1::block_len ... bench:         622 ns/iter (+/- 16) = 102 MB/s

After:
test digest::sha1::_1000     ... bench:       2,801 ns/iter (+/- 22) = 357 MB/s
test digest::sha1::_16       ... bench:         215 ns/iter (+/- 20) = 74 MB/s
test digest::sha1::_2000     ... bench:       5,556 ns/iter (+/- 86) = 359 MB/s
test digest::sha1::_256      ... bench:         922 ns/iter (+/- 35) = 277 MB/s
test digest::sha1::_8192     ... bench:      22,141 ns/iter (+/- 270) = 369 MB/s
test digest::sha1::block_len ... bench:         403 ns/iter (+/- 10) = 158 MB/s

### DIFF
--- a/src/digest/sha1.rs
+++ b/src/digest/sha1.rs
@@ -59,40 +59,49 @@ fn block_data_order_(mut H: State, M: &[[<W32 as Word>::InputBytes; 16]]) -> Sta
         }
 
         // FIPS 180-4 6.1.2 Step 2
-        let mut a = H[0];
-        let mut b = H[1];
-        let mut c = H[2];
-        let mut d = H[3];
-        let mut e = H[4];
+        let a = H[0];
+        let b = H[1];
+        let c = H[2];
+        let d = H[3];
+        let e = H[4];
 
-        // FIPS 180-4 6.1.2 Step 3
-        for t in 0..ROUNDS {
-            // FIPS 180-4 {4.1.1, 4.2.1}
-            let (k, f) = match t {
-                0..=19 => (Wrapping(0x5a827999), ch(b, c, d)),
-                20..=39 => (Wrapping(0x6ed9eba1), parity(b, c, d)),
-                40..=59 => (Wrapping(0x8f1bbcdc), maj(b, c, d)),
-                60..=79 => (Wrapping(0xca62c1d6), parity(b, c, d)),
-                _ => unreachable!(),
-            };
-
-            let T = rotl(a, 5) + f + e + k + W[t];
-            e = d;
-            d = c;
-            c = rotl(b, 30);
-            b = a;
-            a = T;
-        }
+        // FIPS 180-4 6.1.2 Step 3 with constants and functions from FIPS 180-4 {4.1.1, 4.2.1}
+        let (a, b, c, d, e) = step3(a, b, c, d, e, W[ 0..20].try_into().unwrap(), Wrapping(0x5a827999), ch);
+        let (a, b, c, d, e) = step3(a, b, c, d, e, W[20..40].try_into().unwrap(), Wrapping(0x6ed9eba1), parity);
+        let (a, b, c, d, e) = step3(a, b, c, d, e, W[40..60].try_into().unwrap(), Wrapping(0x8f1bbcdc), maj);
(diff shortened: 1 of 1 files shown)


src/digest/sha1.rs
```diff
@@ -66,15 +66,48 @@ fn block_data_order_(mut H: State, M: &[[<W32 as Word>::InputBytes; 16]]) -> Sta
...
-                0..=19 => (Wrapping(0x5a827999), ch(b, c, d)),
-                20..=39 => (Wrapping(0x6ed9eba1), parity(b, c, d)),
-                40..=59 => (Wrapping(0x8f1bbcdc), maj(b, c, d)),
-                60..=79 => (Wrapping(0xca62c1d6), parity(b, c, d)),
-                _ => unreachable!(),
-            };
+
+        // The following four loops are identical except for the values of k
+        // and f, which are as specified in FIPS 180-4 {4.1.1, 4.2.1}
+        for t in 0..20 {
```
briansmith: I like the idea but I'd still like to find a way to keep things more in 1:1 correspondence with the spec. Please try factoring out the commonality into a callable `#[inline]` or `#[inline(always)]` function that gets called four times in succession:

```
fn step3(...) {
}

let (....) = step3(0x5a827999, ch)
let (....) = step3(0x6ed9eba1, parity);
let (....) = step3(0x8f1bbcdc, maj);
let (....) = step3(0xca62c1d6, parity);
```

LingMan: Ah, yes. Shadowing the variables is a good idea. Looks much better now and a bit of extra performance fell out as well. Will clean up the patches and push.

LingMan:
Pushed a new version addressing the review. Having the function loop over a complete 20-element array allowed the compiler to further optimize the code and gave a free ~2% speedup over my initial submission. Not exactly sure why but presumably it wasn't able to fully remove bounds checks on W before.

Added another commit that avoids needlessly initiallizing W with zeros for each block. Gives a boost of 0.8-1.3%. For some reason sha1::block_len is the only benchmark which regresses though (~-1.8%). If anyone can explain that, I'd be all eyes.

briansmith:
Thanks @LingMan. Please squash all your commits into one commit and append the contributor agreement from https://github.com/briansmith/ring#contributing to the end of the commit message.

briansmith:
> For some reason sha1::block_len is the only benchmark which regresses though (~-1.8%). If anyone can explain that, I'd be all eyes.

We should choose the version that is faster for HMAC and in particular PBKDF2, instead of choosing the version that would be faster for longer inputs.

LingMan:
Done and dropped the commit with mixed results.

briansmith:
Thanks for doing this!

Are you interested in doing the same for the SHA-2 implementation?

LingMan:
What exactly do you have in mind for SHA2? From a quick glance it doesn't appear to suffer from the same problem.

PR #941, opened 2020-01-26
AES-GCM & RSA: Work around "Illegal instruction" in 32-bit ARM Android emulator.

briansmith:
The workaround is to be call `&[u8]::len()` before calling `&[u8]::as_mut_ptr()`.
It isn't clear why this is necessary to get the ARM Android target to do the right
thing. Our current guess is that the pointer returned from `as_mut_ptr()` is being
marked as `noalias` within the compiler, which is causing the compiler to make the
wrong assumptions.


briansmith:
This didn't work as expected.

PR #943, opened 2020-01-28
Initial little-endian MIPS64 port

xen0n:
Performance is dog slow but at least things build and tests pass.

*NOTE: Constant-time-ness of various ops not thoroughly verified, beware!*

* [x] CI support
* [x] `SYS_getrandom` for MIPS n64
* [x] `crypto/bn` MIPS assembly from OpenSSL
* [x] `crypto/poly1305` MIPS assembly from OpenSSL
* [x] correctly fallback for `crypto/aes`
* [x] re-introduce `aes_nohw` bits
* [x] correctly fallback for `gcm`
* [x] `crypto/chacha` -- added upstream C code
* [x] `crypto/nistz256` -- added many fallback code mostly inspired by the generic p384 impl


xen0n:
Marking as ready for review. The CI timeouts are a little strange because the `if` clause leading to `travis_wait` should be fine...

build.rs
```diff
@@ -109,6 +110,11 @@ const RING_SRCS: &[(&[&str], &str)] = &[
     (&[AARCH64], "crypto/fipsmodule/modes/asm/ghash-neon-armv8.pl"),
     (&[AARCH64], "crypto/poly1305/asm/poly1305-armv8.pl"),
     (&[AARCH64], SHA512_ARMV8),
+
+    (&[MIPS64], "crypto/fipsmodule/bn/asm/mips-mont.pl"),
+    (&[MIPS64], "crypto/poly1305/asm/poly1305-mips.pl"),
```
briansmith, line 115: Let's drop these two assembly modules.

I don't have time right now to review  the MIPS assembly code. mips-mont is no longer necessary to get things working because we now have Rust/C code that does the same thing. The C/Rust code isn't as fast as the assembly code, but we can make it faster than it is and maybe even faster than the assembly code, with not too much effort.

For poly1305-mips.pl, we don't have C/Rust code for poly1305 in *ring* yet, but we can get C code from BoringSSL, and that's what we should do. Doing that would be more time-efficient than trying to maintain the poly1305-mips code.

build.rs
```diff
@@ -109,6 +110,11 @@ const RING_SRCS: &[(&[&str], &str)] = &[
...
+    (&[MIPS64], "crypto/chacha/chacha_enc.c"),
+    (&[MIPS64], "crypto/fipsmodule/aes/aes.c"),
```
briansmith, line 117: aes/aes.c is no longer needed and should be dropped.

build.rs
```diff
@@ -239,6 +246,7 @@ const ASM_TARGETS: &[(&str, Option<&str>, &str)] = &[
     ("x86", None, "elf"),
     ("arm", Some("ios"), "ios32"),
     ("arm", None, "linux32"),
+    ("mips64", None, "linux64"),
```
briansmith, line 249: You'll need to change this to `Some("linux64")` when you rebase onto master.

build.rs
```diff
@@ -109,6 +110,11 @@ const RING_SRCS: &[(&[&str], &str)] = &[
     (&[AARCH64], "crypto/fipsmodule/modes/asm/ghash-neon-armv8.pl"),
     (&[AARCH64], "crypto/poly1305/asm/poly1305-armv8.pl"),
     (&[AARCH64], SHA512_ARMV8),
+
+    (&[MIPS64], "crypto/fipsmodule/bn/asm/mips-mont.pl"),
+    (&[MIPS64], "crypto/poly1305/asm/poly1305-mips.pl"),
+    (&[MIPS64], "crypto/chacha/chacha_enc.c"),
```
briansmith, line 116: Let's import chacha.c from BoringSSL in a separate MR.

crypto/chacha/chacha_enc.c
```diff
@@ -0,0 +1,119 @@
+/*
+ * Copyright 2015-2016 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
```
briansmith, line 7: I prefer the BoringSSL code with the BoringSSL license.

crypto/fipsmodule/ec/ecp_nistz256.c
```diff
@@ -45,6 +45,7 @@ static const Limb ONE[P256_LIMBS] = {
 /* Precomputed tables for the default generator */
 #include "ecp_nistz256_table.inl"
 
+#ifndef OPENSSL_MIPS64
```
briansmith, line 48: Instead of this condition, let's define a symbol `GFp_P256_ASM` or similar that indicates whether we have assembly code or not, and use that. There should be no mention of MIPS in this file. Instead, I would expect all the conditions to be `#if defined(GFp_P256_ASM)` or `#if !defined(GFp_P256_ASM)`

crypto/fipsmodule/ec/ecp_nistz256.c
```diff
@@ -196,6 +200,143 @@ void GFp_nistz256_point_add(P256_POINT *r, const P256_POINT *a, const P256_POINT
   limbs_copy(r->Y, res_y, P256_LIMBS);
   limbs_copy(r->Z, res_z, P256_LIMBS);
 }
+
+/* Include more reference implementation code for arches lacking assembly
+ * optimizations */
```
briansmith, line 205: Whatever code you copied out of gfp_p384.c must go into gfp_p256. Or, put another way, when you copy code out of a file with license X, you must copy it into a file that also has license X. ecp_nistz256.c and gfp_*.c have different licenses, and that's why they are split up in the weird way they are.

Probably the best thing to do w.r.t. this NIST curve code is to refactor it (in a separate PR) so that the P-256 code has the same shape as the P-384 code: rename ecp_nistz256.c to ecp_nistz256.inl, `#include "ecp_nistz256.inl"` in gfp_p256.c, etc.

include/GFp/mips_arch.h
```diff
@@ -0,0 +1,40 @@
...
+ * Copyright 2011-2016 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#ifndef OSSL_CRYPTO_MIPS_ARCH_H
+# define OSSL_CRYPTO_MIPS_ARCH_H
```
briansmith, line 11: If we can drop the assembly code then we can drop this file too, I hope.

src/aead/aes.rs
```diff
@@ -292,6 +298,12 @@ pub enum Implementation {
     Fallback = 4,
 }
 
+#[cfg(target_arch = "mips64")]
+fn detect_implementation(_cpu_features: cpu::Features) -> Implementation {
```
briansmith, line 302: I believe the new version of this function in `master` will do the right thing automatically for you.

src/aead/gcm.rs
```diff
@@ -37,6 +37,7 @@ impl Key {
                 }
             }
 
+            #[cfg(not(target_arch = "mips64"))]
```
briansmith, line 40: I believe the new version of this file in master will do the right thing for you automatically.

src/cpu.rs
```diff
@@ -66,6 +66,10 @@ pub(crate) fn features() -> Features {
     Features(())
 }
 
+#[cfg_attr(
+    not(any(target_arch = "aarch64", target_arch = "arm")),
+    allow(dead_code)
+)]
```
briansmith, line 72: I think the new version of this file in master will work without changes now.

briansmith:
Awesome work here. However, the main issue is that I just have very few cycles to review MIPS-specific code. Thus, in the last couple of weeks, as part of getting things working in WebAssembly, I tried to make the code more generic so that less MIPS-specific logic would be necessary. Please rebase this on top of master and drop as much of the MIPS-specific logic as you can.

xen0n:
> Awesome work here. However, the main issue is that I just have very few cycles to review MIPS-specific code. Thus, in the last couple of weeks, as part of getting things working in WebAssembly, I tried to make the code more generic so that less MIPS-specific logic would be necessary. Please rebase this on top of master and drop as much of the MIPS-specific logic as you can.

Thanks for reviewing this! I'll rebase soon. (Possibly this or the next weekend, depends on my free time.)

xen0n:
This is getting very outdated, yet I've shifted focus since then, no time for actual work. Might revisit this before 2021 ends, but let's close for now...

PR #961, opened 2020-03-27
Fix intra-doc links (fixes #958)

### LINKED ISSUE #958: docs.rs doc build error
see https://docs.rs/crate/ring/0.17.0-alpha.1/builds/220789

djc:

### DIFF
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -268,7 +268,7 @@ impl<N: NonceSequence> core::fmt::Debug for SealingKey<N> {
 }
 
 impl<N: NonceSequence> SealingKey<N> {
-    /// Deprecated. Renamed to [`seal_in_place_append_tag()`].
+    /// Deprecated. Renamed to [`seal_in_place_append_tag()`](#method.seal_in_place_append_tag).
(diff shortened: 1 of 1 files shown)


src/aead.rs
```diff
@@ -269,6 +269,8 @@ impl<N: NonceSequence> core::fmt::Debug for SealingKey<N> {
 
 impl<N: NonceSequence> SealingKey<N> {
     /// Deprecated. Renamed to [`seal_in_place_append_tag()`].
+    ///
+    /// [`seal_in_place_append_tag()`]: #method.seal_in_place_append_tag
```
briansmith: There must be a more compact, less repetitive, way to do this? This isn't a sustainable way to do it.

briansmith:
> I agree to license my contributions to each file under the terms given at the top of each file I changed.

Please put this in the actual commit message.

Thanks!

src/aead.rs
djc: Yup. Fixed.

briansmith:
Thanks. I was hoping it could be even more compact than that, but I guess we'd have to wait until the actual intra-rustdoc link feature is stable. In the meantime, I addressed the problem in #962. Thanks for your help!

PR #962, opened 2020-03-27
Stop using intra-rustdoc links until they are stable.

briansmith:


briansmith:
@djc Please take a look at this. It doesn't make sense to spend more than minimal effort on these comments for deprecated functions. If we can't use intra-rustdoc links yet for more useful stuff then no reason to try to emulate them for these things.

djc:
I still think the links make life slightly easier for someone who peruses the docs. The new version that I just did seems sufficiently compact to me, though.

(I'm not sure about your "use .. yet" -- I don't think any of this is about changes to the intra-rustdoc link functionality, just that nightly introduces new lints that warn on dangling intra-rustdoc links.)

briansmith:
> I still think the links make life slightly easier for someone who peruses the docs. The new version that I just did seems sufficiently compact to me, though.

Yes, it is fine, but I don't want to establish that as a pattern for people writing documentation in the crate.

> (I'm not sure about your "use .. yet" -- I don't think any of this is about changes to the intra-rustdoc link functionality, just that nightly introduces new lints that warn on dangling intra-rustdoc links.)

It seems like intra-rustdoc links are still unsable according to the Rust tracking issue. If we can't use intra-rustdoc links then I'd rather just avoid linkifying things for now, to make the documentation easier to read/write for contributors.
 
> But obviously, this is up to you. The important part for me is having the docs.rs docs work again.

I agree.

PR #965, opened 2020-04-24
[cargo] Remove no-op 'exclude' field

smklein:
From
https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields

"The options are mutually exclusive: setting include will override an
exclude."


briansmith:
Thanks!

Issue #963, opened 2020-04-15
Build failure if `cargo` environment variable exists


Nemo157:
```console
> cargo=test cargo build
    Updating crates.io index
   Compiling libc v0.2.69
   Compiling cc v1.0.50
   Compiling lazy_static v1.4.0
   Compiling untrusted v0.7.0
   Compiling spin v0.5.2
   Compiling ring v0.17.0-alpha.1 (/tmp/tmp.bJ19ZyaNOk/ring-0.16.12/ring-0.17.0-alpha.1)
error: Wrong output in build script of `ring v0.17.0-alpha.1 (/tmp/tmp.bJ19ZyaNOk/ring-0.16.12/ring-0.17.0-alpha.1)`: `cargo: test`
```

pietro:
What's the output of `cargo=test cargo build -vv`?

Nemo157:
The bug is in this loop, it will print `cargo: test` and presumably cargo attempts to parse any lines starting with `cargo.*:`.

https://github.com/briansmith/ring/blob/00c21e253ba9cd3b66ab41155414b0d0e91b6c95/build.rs#L263-L265

HeapUnderfl0w:
i can confirm this: https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

briansmith:
:sigh: Thanks for pointing this out. I suggest we change the line from:
```
     println!("{}: {}", key, value); 
```
to
```
     println!("ENV {}={}", key, value); 
```

briansmith:
Also, maybe all this output should be `eprintln!` and/or `log!` output.

briansmith:
Fixed in #968.

PR #968, opened 2020-04-27
Fix the case where a `cargo` environment variable is set in build.rs.

briansmith:
Previously, build.rs. would print "cargo:..." to stdout if a variable
`$cargo` was set. This conflicted with Cargo's interpretation of
"cargo:..." to direct it to do things; see [1].

Instead, write to stderr to avoid the conflict. Also change the
formatting of the the environment variable logging so that even if
stderr gets redirected to stdout, such conflicts won't occur.

[1]
https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -14,6 +14,11 @@
 
 //! Build the non-Rust components.
 
+// It seems like it would be a good idea to use `log!` for logging, but it
+// isn't worth having the external dependencies (one for the `log` crate, and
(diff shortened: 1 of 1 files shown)


briansmith:
This fixes #963. @HeapUnderfl0w @Nemo157 please take a look. Thanks for reporting!

PR #973, opened 2020-04-30
Merge BoringSSL 9855c1c: Add a constant-time fallback GHASH implementation.

briansmith:
*ring* tries to work without type-punning `memcpy`, so the use of that
in `GFp_gcm_ghash_nohw` was replaced by the use of `u64_from_be_bytes`.
This will (I hope) also help with the eventual support for big-endian
targets. Here's the diff from BoringSSL in that function:

```diff
-void gcm_ghash_nohw(uint64_t Xi[2], const u128 Htable[16], const uint8_t *inp,
-                    size_t len) {
+void GFp_gcm_ghash_nohw(uint64_t Xi[2], const u128 Htable[16], const uint8_t *inp,
+                        size_t len) {
   uint64_t swapped[2];
   swapped[0] = CRYPTO_bswap8(Xi[1]);
   swapped[1] = CRYPTO_bswap8(Xi[0]);

   while (len >= 16) {
-    uint64_t block[2];
-    OPENSSL_memcpy(block, inp, 16);
-    swapped[0] ^= CRYPTO_bswap8(block[1]);
-    swapped[1] ^= CRYPTO_bswap8(block[0]);
+    swapped[0] ^= u64_from_be_bytes(&inp[8]);
+    swapped[1] ^= u64_from_be_bytes(inp);
     gcm_polyval_nohw(swapped, &Htable[0]);
     inp += 16;
     len -= 16;
```

During the merge, I found that `GFp_gcm_gmult_clmul` had its
`.cfi_startproc` on the wrong line. I fixed that as part of the merge.

During my review of the BoringSSL changes, I noticed that BoringSSL had
left some of the dead code in ghash-x86_64.pl, which had previously been
removed in *ring*. That removal is being done in BoringSSL in [1].

[1] https://boringssl-review.googlesource.com/c/boringssl/+/41144


crypto/fipsmodule/modes/asm/ghash-x86_64.pl
```diff
@@ -114,10 +117,27 @@
...
+$nhi="%rbx";
+$Zlo="%r8";
+$Zhi="%r9";
+$tmp="%r10";
+$rem_4bit = "%r11";
+
+$Xi="%rdi";
+$Htbl="%rsi";
+
+
```
briansmith, line 131: I added back this chunk to stay in sync with BoringSSL. I believe it is dead code and I left a comment on the BoringSSL CL to let them know that it is dead code. I hope they will remove it in the future.

PR #977, opened 2020-05-04
AES: Replace "aliasing pointer" pattern with `MaybeUninit` pattern.

briansmith:
The code was using the "aliasing pointer" pattern to avoid needing to
deal with `core::mem::uninitialized()`. Now we can use
`core::mem::MaybeUninit` instead of either.

### DIFF
--- a/src/aead/aes.rs
+++ b/src/aead/aes.rs
@@ -109,42 +109,43 @@ impl Key {
     }
 
     #[inline]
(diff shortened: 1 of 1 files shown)


briansmith:
This raised the minimum Rust version to 1.36.

briansmith:
> This raised the minimum Rust version to 1.36.

Actually, we were using `alloc` for quite a while, so we've been requiring 1.36+ for anybody using RSA already.

PR #982, opened 2020-05-05
ChaCha: use `slice::copy_within` instead of `core::ptr::copy`.

briansmith:
Use less `unsafe`.


briansmith:
This will bump the MSRV to 1.37.

PR #1007, opened 2020-06-10
Add arm feature detection for FreeBSD

MikaelUrankar:


briansmith:
@MikaelUrankar Thanks for the PR. I noticed that this looks almost exactly like the Linux version. Would it be possible to refactor this so that the Linux and FreeBSD version of the setup function use the same code, but with different values of the `HWCAP_*` constants and different `getauxval` function names?

MikaelUrankar:
> @MikaelUrankar Thanks for the PR. I noticed that this looks almost exactly like the Linux version. Would it be possible to refactor this so that the Linux and FreeBSD version of the setup function use the same code, but with different values of the `HWCAP_*` constants and different `getauxval` function names?

it's done, the CI is broken though, don't know why.

MikaelUrankar:
ping

MikaelUrankar:
That's still an issue for FreeBSD users...

briansmith:
@MikaelUrankar Thanks for your PR. I have an alternate approach to consider in PR #1542 that allows us to avoid adding new `unsafe` code to *ring* but still get the same effect. Please take a look at that and let me know what you think.

PR #1016, opened 2020-06-21
Remove unsafe uses of core::mem::transmute

unseddd:
Replace unsafe uses of `core::mem::transmute` with safe conversions
from `u8` arrays into `u64` arrays.

### DIFF
--- a/src/aead/block.rs
+++ b/src/aead/block.rs
@@ -13,6 +13,7 @@
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 use crate::{endian::*, polyfill, polyfill::convert::*};
+use core::convert::TryInto;
 
 /// An array of 16 bytes that can (in the x86_64 and AAarch64 ABIs, at least)
 /// be efficiently passed by value and returned by value (i.e. in registers),
@@ -78,14 +79,36 @@ impl Block {
 impl From<&'_ [u8; BLOCK_LEN]> for Block {
     #[inline]
     fn from(bytes: &[u8; BLOCK_LEN]) -> Self {
-        unsafe { core::mem::transmute_copy(bytes) }
+        Self {
+            subblocks: [
+                // unwrap use is safe, only errors on length mismatch
+                // lengths are guaranteed to match here, see array::TryFrom docs
+                u64::from_be_bytes(bytes[..8].try_into().unwrap()),
+                u64::from_be_bytes(bytes[8..].try_into().unwrap()),
+            ],
+        }
     }
 }
 
 impl From_<&'_ [u8; 2 * BLOCK_LEN]> for [Block; 2] {
     #[inline]
     fn from_(bytes: &[u8; 2 * BLOCK_LEN]) -> Self {
-        unsafe { core::mem::transmute_copy(bytes) }
+        [
+            Block {
+                subblocks: [
+                    // unwrap use is safe, only errors on length mismatch
+                    // lengths are guaranteed to match here, see array::TryFrom docs
+                    u64::from_le_bytes(bytes[..8].try_into().unwrap()),
+                    u64::from_le_bytes(bytes[8..16].try_into().unwrap()),
+                ],
+            },
+            Block {
+                subblocks: [
+                    u64::from_le_bytes(bytes[16..24].try_into().unwrap()),
+                    u64::from_le_bytes(bytes[24..].try_into().unwrap()),
+                ],
+            },
+        ]
     }
 }
 


unseddd:
Thought this would be helpful after reading [the nomicon entry on mem::transmute](https://doc.rust-lang.org/nomicon/transmutes.html).

Also a bit of defensive programming from the [safety-dance project](https://github.com/rust-secure-code/safety-dance).

Still haven't figured out how to convert going the other direction for the `AsRef` impl. `u64` has a `to_le_bytes`, but it returns a slice by copy (not array). May still require unsafe code, but perhaps safer than `mem::transmute`.

Will keep trying. If you have some pointers, happy to implement (maybe in separate PR).

src/aead/block.rs
```diff
@@ -78,14 +79,56 @@ impl Block {
 impl From<&'_ [u8; BLOCK_LEN]> for Block {
     #[inline]
     fn from(bytes: &[u8; BLOCK_LEN]) -> Self {
-        unsafe { core::mem::transmute_copy(bytes) }
+        Self {
+            subblocks: [
+                u64::from_le_bytes(
+                    bytes[..8]
+                        .try_into()
+                        .expect("error converting leading bytes"),
```
unseddd, line 87: Use `expect` here and other places, because the conversion (`[u8] -> [u8; 8]`) should never fail. If conversion does fail, panicking is IMHO the correct behavior, rather than silently using such corrupted data or default values (i.e. `[0u8; 8]`).

unseddd: From the [TryFrom](https://doc.rust-lang.org/nightly/src/core/array/mod.rs.html#157) docs:

```
    fn try_from(slice: &[T]) -> Result<&[T; N], TryFromSliceError> {
        if slice.len() == N {
            let ptr = slice.as_ptr() as *const [T; N];
            // SAFETY: ok because we just checked that the length fits
            unsafe { Ok(&*ptr) }
        } else {
            Err(TryFromSliceError(()))
        }
    }
```

So, it only fails when the lengths don't match up. Here, they are guaranteed to match.

unseddd: From the [TryFrom](https://doc.rust-lang.org/src/core/array/mod.rs.html#157) docs:

```
    fn try_from(slice: &[T]) -> Result<&[T; N], TryFromSliceError> {
        if slice.len() == N {
            let ptr = slice.as_ptr() as *const [T; N];
            // SAFETY: ok because we just checked that the length fits
            unsafe { Ok(&*ptr) }
        } else {
            Err(TryFromSliceError(()))
        }
    }
```

So, it only fails when the lengths don't match up. Here, they are guaranteed to match.

src/aead/block.rs
```diff
@@ -78,14 +79,56 @@ impl Block {
 impl From<&'_ [u8; BLOCK_LEN]> for Block {
     #[inline]
     fn from(bytes: &[u8; BLOCK_LEN]) -> Self {
-        unsafe { core::mem::transmute_copy(bytes) }
+        Self {
+            subblocks: [
+                u64::from_le_bytes(
```
briansmith, line 84: Why do you use `u64::from_le_bytes` instead of `u64::from_be_bytes`?

src/aead/block.rs
briansmith: Please just use `unwrap()`. I don't want to have the strings passed to `expect()` included in the resulting library.

src/aead/block.rs
```diff
@@ -78,14 +79,56 @@ impl Block {
...
+                        .expect("error converting trailing bytes"),
+                ),
+            ],
+        }
     }
 }
 
 impl From_<&'_ [u8; 2 * BLOCK_LEN]> for [Block; 2] {
     #[inline]
     fn from_(bytes: &[u8; 2 * BLOCK_LEN]) -> Self {
```
briansmith, line 101: My plan is to eliminate the use of `Block` within `Poly1305`. Then I'm planning to experiment with split the `Block` code into `gcm::Block` and `aes::Block` types, or similar, that each only implement what AES and GCM use.

The original motivation for creating the `Block` type the way it exists today no longer applies and it just makes the code worse. I.e. basically this whole file is now a mistake and I want it to go away.

If you want to submit some PRs towards that end, I would appreciate them. The smaller the better.

src/aead/block.rs
unseddd: > Please just use unwrap()

Sounds good. Would comments about the safety of using `unwrap()` here also be good (to the effect of "lengths guaranteed to match, see the docs...")?

src/aead/block.rs
unseddd: Saw both `Block::from_le_bytes` and `Block::from_be_bytes` above, so coin flip. Do you prefer `u64::from_be_bytes` / is that more correct?

src/aead/block.rs
unseddd: > If you want to submit some PRs towards that end, I would appreciate them. The smaller the better.

Happy to work on improving things. Would need maybe some more of your design ideas, what you would like to see instead.

src/aead/block.rs
unseddd: This was my mistake. After looking at the BoringSSL AES impl, see it uses big-endian for keys. Updated with a fix.

briansmith:
@unseddd Thanks for your work on this. I filed issue #1021 for removing `Block` completely and I made PR #1020 for removing it from the ChaCha20-Poly1305 code. I'm going to close this so we can just focus on removing the `Block` type. I appreciate your effort here and you are definitely right that, especially given the current state of the code, it is unlikely worth using `unsafe` `transmute()`/`transmute_copy()` any more.

briansmith:
> Happy to work on improving things. Would need maybe some more of your design ideas, what you would like to see instead.

It would be awesome if you could remove the remove the remaining functions from `Block` that use `transmute` by replacing the code that uses it with new code that doesn't use `Block`.

unseddd:
> It would be awesome if you could remove the remove the remaining functions from Block that use transmute by replacing the code that uses it with new code that doesn't use Block.

Definitely! Will start work on removing the remaining uses of `Block`.

PR #1020, opened 2020-06-25
Remove use of `Block` from Chacha20-Poly1305.

briansmith:
This is a step towards getting rid of the `Block` type completely.


src/aead/quic.rs
```diff
@@ -170,7 +170,7 @@ pub static CHACHA20: Algorithm = Algorithm {
 };
 
 fn chacha20_init(key: &[u8], _todo: cpu::Features) -> Result<KeyInner, error::Unspecified> {
-    let chacha20_key: &[u8; chacha::KEY_LEN] = key.try_into()?;
+    let chacha20_key: [u8; chacha::KEY_LEN] = key.try_into()?;
```
unseddd, line 173: Is removing the reference here to consume (rather than copy) the converted slice?

briansmith: > Is removing the reference here to consume (rather than copy) the converted slice?

Yes, I changed the key constructor to take the key by (moved) value, so I had to change this too.

Issue #1021, opened 2020-06-25
Remove the internal `Block` type


briansmith:
The `Block` type was created a while back because *ring* needed a way  to pass 16-byte arrays to/from C in registers for efficiency. That is no longer done in *ring*, so it has outlived its purpose.

The `Block` type is confusing and uses `transmute` and `transmute_copy`, which we'd rather not use because they are `unsafe`. In particular, the endianness of the value within the `Block` is confusing and not enforced by the type system; depending on the use, the endianness is different. This was done to cope with some C code that we no longer have to cope with.

PR #1020 removes the usage of the `Block` type and can serve as an example for other cases.

briansmith:
I decided to keep `Block` for now, after minimizing its usage and removing all the `unsafe`. #1190 is the last step I'm planning to take regarding this. If there's more that should be done along these lines, I'd be happy to take a PR for further improvements.

PR #1022, opened 2020-06-30
WIP: Remove `Block` from AES modules

unseddd:
Replace uses of `Block` in `aead/aes*` modules to eventually remove the `Block` type completely.

Related AEAD modules adapted to use new interfaces

Related to #1021

### DIFF
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -23,7 +23,7 @@
 
 use self::block::{Block, BLOCK_LEN};
 use crate::{constant_time, cpu, error, hkdf, polyfill};
-use core::ops::RangeFrom;
+use core::{convert::TryInto, ops::RangeFrom};
 
 pub use self::{
     aes_gcm::{AES_128_GCM, AES_256_GCM},
@@ -635,7 +635,7 @@ impl Eq for Algorithm {}
 /// An authentication tag.
 #[must_use]
 #[repr(C)]
-pub struct Tag(Block);
+pub struct Tag([u8; BLOCK_LEN]);
 
 impl AsRef<[u8]> for Tag {
     fn as_ref(&self) -> &[u8] {
@@ -664,6 +664,76 @@ enum Direction {
     Sealing,
 }
 
+trait BitXor {
+    fn bitxor_assign(&mut self, input: [u8; BLOCK_LEN]);
+}
+
+impl BitXor for [u8; BLOCK_LEN] {
+    fn bitxor_assign(&mut self, input: [u8; BLOCK_LEN]) {
+        for (r, a) in self.iter_mut().zip(input.iter()) {
+            *r ^= *a;
+        }
+    }
+}
+
+trait ConvertEndian {
+    fn u64s_be_to_native(&self) -> [u64; 2];
+    fn from_be_u64s(u64s: [u64; 2]) -> Self;
+}
+
+impl ConvertEndian for [u8; BLOCK_LEN] {
+    fn u64s_be_to_native(&self) -> [u64; 2] {
+        [
+            // unwrap safe to use, only errors on length mismatch
+            // lengths guaranteed to match, see [u8]::TryFrom docs
+            u64::from_be_bytes(self[..8].try_into().unwrap()),
+            u64::from_be_bytes(self[8..].try_into().unwrap()),
+        ]
+    }
+
+    fn from_be_u64s(u64s: [u64; 2]) -> Self {
+        let mut out = u64s[0].to_be_bytes().to_vec();
+        out.extend_from_slice(&u64s[1].to_be_bytes());
+
+        // unwrap safe to use, only errors on length mismatch
+        // lengths guaranteed to match, see [u8]::TryFrom docs
+        out.as_slice().try_into().unwrap()
+    }
+}
+
+trait ZeroFrom {
+    fn zero_from(&mut self, index: usize);
+}
+
+impl ZeroFrom for [u8; BLOCK_LEN] {
+    fn zero_from(&mut self, index: usize) {
+        // panics when index > self.len()
+        // TODO: change the API to return Result<(), Error>, and bubble up?
+        let (_, zero) = self.split_at_mut(index);
+        for elem in zero.iter_mut() {
+            *elem = 0u8;
+        }
+    }
+}
+
+trait Overwrite {
+    /// Overwrite part of a slice
+    ///
+    /// Panics:
+    ///
+    /// Panics if index > self.len()
+    // TODO: change API to return Result<(), Error>, and bubble up?
+    fn overwrite_part_at(&mut self, index: usize, a: &[u8]);
+}
+
+impl Overwrite for [u8; BLOCK_LEN] {
+    fn overwrite_part_at(&mut self, index: usize, a: &[u8]) {
+        let (_, part) = self.split_at_mut(index);
+        let (fb, _) = part.split_at_mut(a.len());
+        fb.copy_from_slice(a);
+    }
+}
+
 mod aes;
 mod aes_gcm;
 mod block;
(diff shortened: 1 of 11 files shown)


unseddd:
Attempted to keep the diff as small as possible. Some things unrelated to `aead/aes*` got pulled in because of change in interfaces.

src/aead.rs
```diff
@@ -664,6 +664,76 @@ enum Direction {
     Sealing,
 }
 
+trait BitXor {
+    fn bitxor_assign(&mut self, input: [u8; BLOCK_LEN]);
```
briansmith, line 668: Isn't this trait the same as `core::ops::BitXorAssign<[u8; BLOCK_LEN]`?

src/aead.rs
```diff
@@ -664,6 +664,76 @@ enum Direction {
...
+}
+
+impl BitXor for [u8; BLOCK_LEN] {
```
briansmith, line 671: I understand you can't implement `core::ops::BitXorAssign` for [u8; BLOCK_LEN]` due to the coherence rules. But, maybe we just need a function: `bitxor_assign(r: &mut [u8; BLOCK_LEN], a: [u8; BLOCK_LEN]) { ... }` instead of a whole trait?

src/aead/gcm.rs
```diff
@@ -250,16 +253,16 @@ struct u128 {
 const HTABLE_LEN: usize = 16;
 
 #[repr(transparent)]
-pub struct Xi(Block);
+pub struct Xi([u8; BLOCK_LEN]);
```
briansmith, line 256: It seems like we need types like this?:

```
struct H([BigEndian<u64>; 2);
struct Xi([BigEndian<u64>; 2);
```

or maybe:

```
struct H([u64; 2);
struct Xi([u64; 2);
```

depending on what the actual endianness is supposed to be.

In particular, doesn't the assembly code require these to have the same ABI (alignment, etc.) as u64?

src/aead/gcm.rs
```diff
@@ -272,7 +275,7 @@ impl From<Xi> for Block {
 #[repr(C, align(16))]
 pub(super) struct ContextInner {
     Xi: Xi,
-    _unused: Block,
+    _unused: [u8; BLOCK_LEN],
```
briansmith, line 278: This should have type `H` as I suggested above.

src/aead/nonce.rs
```diff
@@ -99,7 +102,7 @@ where
 
     fn new(Nonce(nonce): Nonce, initial_counter: u32) -> Self {
         let mut r = Self {
-            block: Block::zero(),
+            block: [0u8; BLOCK_LEN],
         };
         let block = unsafe { &mut r.block };
         block.overwrite_part_at(U32::NONCE_BYTE_INDEX, nonce.as_ref());
```
briansmith, line 108: I understand what you are trying to do and this is a tricky module w.r.t. this effort.

This counter code is weird in large part because of my attempt to use `Block`. Once we get rid of `Block` it doesn't need to be as weird any more. Check out https://github.com/briansmith/ring/pull/1028 to see what I mean.

src/aead/quic.rs
```diff
@@ -82,7 +81,7 @@ const SAMPLE_LEN: usize = super::TAG_LEN;
 pub struct Algorithm {
     init: fn(key: &[u8], cpu_features: cpu::Features) -> Result<KeyInner, error::Unspecified>,
 
-    new_mask: fn(key: &KeyInner, sample: Block) -> [u8; 5],
+    new_mask: fn(key: &KeyInner, sample: [u8; BLOCK_LEN]) -> [u8; 5],
```
briansmith, line 84: We have a constant `SAMPLE_LEN` that we should use instead. Ditto for the other `sample:` stuff.

src/aead.rs
```diff
@@ -664,6 +664,76 @@ enum Direction {
...
+}
+
+trait Overwrite {
+    /// Overwrite part of a slice
+    ///
+    /// Panics:
+    ///
+    /// Panics if index > self.len()
+    // TODO: change API to return Result<(), Error>, and bubble up?
+    fn overwrite_part_at(&mut self, index: usize, a: &[u8]);
```
briansmith, line 726: I understand that you factored each part of `Block` into a separate trait and then implemented them all for `[u8; BLOCK_LEN]` but I think that made things more complicated than necessary. When we get rid of uses of `Block` I expect the code should actually get simpler, not bigger.

For example, I suspect that the right thing to do is just completely rewrite the code that uses `overwrite_part_at` to be simpler/smarter.

src/aead.rs
```diff
@@ -664,6 +664,76 @@ enum Direction {
...
+    fn bitxor_assign(&mut self, input: [u8; BLOCK_LEN]) {
+        for (r, a) in self.iter_mut().zip(input.iter()) {
+            *r ^= *a;
+        }
+    }
+}
+
+trait ConvertEndian {
+    fn u64s_be_to_native(&self) -> [u64; 2];
+    fn from_be_u64s(u64s: [u64; 2]) -> Self;
```
briansmith, line 681: I think the right thing to do is to move all the wierd endianness logic for GCM into the GCM submodule, and all the wierd endianness logic for AES into the AES submodule, with the hope that we can then simplify and remove a lot of the wierdness. In particular, I don't think this higher-level `aead` module should be exposing this wierdness to all its submodules, because our goal should be to eliminate the weirdness.

unseddd:
> Thanks!

No problem :)

> I suggest the following course of action:
> 
>     * Review and merge PR #1028.
>     * Create a new PR that removes `Block` from the GCM code and get that reviewed and merged.
>     * Do the same for the AES code and that reviewed and merged.
>     * Do whatever else we need to do to remove `Block`.
> 
> 
> WDYT.

Think that sounds like a good plan. Can leave this PR open, and rebase/edit as the parts get merged.

src/aead.rs
unseddd: It very likely is, can double check, and just use `core::ops::BitXorAssign<[u8; BLOCK_LEN]>` if possible. Will report any blockers. Was basically just re-implementing functions from `Block` as needed.

src/aead.rs
unseddd: > maybe we just need a function:bitxor_assign(r: &mut [u8; BLOCK_LEN], a: [u8; BLOCK_LEN]) { ... }` instead of a whole trait?

Probably just the function would be fine. My idea was to separate into traits for modularity/reuse (maybe something else would need it). Another option would be once `Block` struct is removed, create a new trait with the good bits pulled from `Block` that we still want to keep around.

Wasn't sure what grouping made sense, though, so separate traits it was.

src/aead/gcm.rs
unseddd: I didn't dig into the assembly, and I'm sure you're right. Will take a closer look on the next round of refactors.

src/aead/nonce.rs
unseddd: Keep crypto weird ~.~ (jk)

Will check out #1028

src/aead/quic.rs
unseddd: Alright, was following patterns here. Will use `SAMPLE_LEN`.

src/aead.rs
unseddd: > For example, I suspect that the right thing to do is just completely rewrite the code that uses overwrite_part_at to be simpler/smarter.

Think you're right. Going from the bottom up (replacing in `aead/aes`, breaking things, fixing things), overall design got lost. Think I will have a better feel for design going from top uses down. Will keep the simplification factor in mind.

src/aead/nonce.rs
unseddd: > Once we get rid of Block it doesn't need to be as weird any more. Check out #1028 to see what I mean.

Will definitely check out #1028. I thought I was reproducing the previous behaviour, but can see your point about weirdness, too.

src/aead.rs
unseddd: > I think the right thing to do is to move all the wierd endianness logic for GCM into the GCM submodule...

Yeah, wasn't sure where to put the endianness stuff, but that definitely makes sense.

> our goal should be to eliminate the weirdness

Unweird all the things! :)

PR #1033, opened 2020-07-18
Remove the use of `Block` from `aead::Tag`.

unseddd:
Related to #1021

### DIFF
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -635,7 +635,7 @@ impl Eq for Algorithm {}
 /// An authentication tag.
 #[must_use]
 #[repr(C)]
-pub struct Tag(Block);
+pub struct Tag([u8; TAG_LEN]);
 
 impl AsRef<[u8]> for Tag {
     fn as_ref(&self) -> &[u8] {
--- a/src/aead/poly1305.rs
+++ b/src/aead/poly1305.rs
@@ -17,7 +17,7 @@
 
 use super::{
     block::{Block, BLOCK_LEN},
-    Tag,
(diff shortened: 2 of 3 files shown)


src/aead.rs
```diff
@@ -635,11 +635,11 @@ impl Eq for Algorithm {}
 /// An authentication tag.
 #[must_use]
 #[repr(C)]
-pub struct Tag(Block);
+pub struct Tag([u8; BLOCK_LEN]);
```
briansmith, line 638: We have `TAG_LEN` that we should use instead.

src/aead.rs
```diff
@@ -635,11 +635,11 @@ impl Eq for Algorithm {}
...
 
 impl AsRef<[u8]> for Tag {
     fn as_ref(&self) -> &[u8] {
-        self.0.as_ref()
+        &self.0[..]
```
briansmith, line 642: `self.0.as_ref()`?

src/aead/poly1305.rs
```diff
@@ -160,7 +160,7 @@ impl Funcs {
 
     #[inline]
     fn emit(&self, state: &mut Opaque, nonce: &Nonce) -> Tag {
-        let mut tag = Tag(Block::zero());
+        let mut tag = Tag([0u8; BLOCK_LEN]);
```
briansmith, line 163: Ditto: Use `TAG_LEN`.

unseddd:
Let me know if you want me to rebase on current `main`.

briansmith:
Thanks!

PR #1036, opened 2020-07-23
Remove `Block` from `aead::{aes,chacha}::new_mask`

unseddd:
Remove use of `Block` from `aead::{aes,chacha}::new_mask` API.

Leave internal use of `Block` in `aead::aes::new_mask`

### DIFF
--- a/src/aead/aes.rs
+++ b/src/aead/aes.rs
@@ -12,7 +12,7 @@
 // OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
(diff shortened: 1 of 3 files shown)


src/aead/aes.rs
```diff
@@ -291,8 +291,8 @@ impl Key {
         }
     }
 
-    pub fn new_mask(&self, sample: Block) -> [u8; 5] {
-        let block = self.encrypt_block(sample);
+    pub fn new_mask(&self, sample: [u8; BLOCK_LEN]) -> [u8; 5] {
```
briansmith, line 294: I think we move put `SAMPLE_LEN` in `aead::quic`. I thinki we should also define `pub type Sample = [u8; SAMPLE_LEN]` there. Then all these uses of `Block` would be replaced by `Sample`.

briansmith:
Thanks!

PR #1037, opened 2020-07-25
add support for the s390x arch

knarz:
Hello,

this PR adds support for the s390x architecture, using available openssl libraries. `cargo test` runs without failures. (I tried running `cargo test --no-default-features` but that doesn't seem to run on `main` or my branch?).

Please let me know, if you'd need anything else in order to merge this.

This work is based on a patch by Ulrich Weigand <uweigand@de.ibm.com>.


vincenthz:
Just tested this on s390x and this is working great: passing tests and able to give rustls a go. I haven't reviewed any of the asm, but the rest of the integration looks sensible otherwise.

briansmith:
>  (I tried running cargo test --no-default-features but that doesn't seem to run on main or my branch?).

That is fixed by #1652.

briansmith:
Closing this since we do support s390x now. I understand that in some ways this is a better approach since it enables the optimized assembly language code for s390x. However, we don't have resources to QA that assembly code. Please reach out if you're interested in helping with that problem.

uweigand:
I'm about to go on vacation, but once I'm back, I'd be happy to have another look into adding optimized implementations on top of the current generic support for s390x.  I agree it would be good to have QA for those.  qemu does not emulate the crypto instructions that these assembler implementations use, so the qemu-based QA setup in mainline wouldn't test these.

Now, we actually do also run an IBM internal daily ring regression test on native machines, which *would* test the crypto instructions if used.  Maybe that would already be sufficient.  In the alternative, we could set up a github runner on an IBM Z machine (e.g. on the LinuxONE Community Cloud) which you could integrate into this repo to run QA there.

PR #1040, opened 2020-07-30
GCM: Remove use of `Block`.

briansmith:
This is a step towards removing `Block`.

### DIFF
--- a/src/aead/aes_gcm.rs
+++ b/src/aead/aes_gcm.rs
@@ -16,7 +16,7 @@ use super::{
     aes::{self, Counter},
     gcm, shift, Aad, Block, Direction, Nonce, Tag, BLOCK_LEN,
 };
-use crate::{aead, cpu, endian::*, error, polyfill};
+use crate::{aead, cpu, error, polyfill};
 
 /// AES-128 in GCM mode with 128-bit tags and 96 bit nonces.
 pub static AES_128_GCM: aead::Algorithm = aead::Algorithm {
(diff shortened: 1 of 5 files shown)


briansmith:
@unseddd PTAL.

unseddd:
LGTM :+1:

src/aead/aes_gcm.rs
```diff
@@ -167,29 +167,26 @@ fn aead(
...
-    gcm_ctx.update_block(Block::from_u64_be(
-        BigEndian::from(aad_bits),
-        BigEndian::from(ciphertext_bits),
-    ));
+    gcm_ctx.update_block([aad_bits.into(), ciphertext_bits.into()]);
 
     // Finalize the tag and return it.
     gcm_ctx.pre_finish(|pre_tag| {
         let bytes = tag_iv.into_bytes_less_safe();
         let mut tag = aes_key.encrypt_block(Block::from(&bytes));
```
unseddd, line 188: WDYT about using `aes_key.encrypt_block((&bytes).into())` here?

src/aead/aes_gcm.rs
```diff
@@ -167,29 +167,26 @@ fn aead(
...
-        tag.bitxor_assign(pre_tag.into());
+        tag.bitxor_assign(Block::from(&pre_tag));
```
unseddd, line 189: WDYT about using `tag.bitxor_assign((&pre_tag).into());` here?

unseddd (changes requested):
Couple suggestions for removing `Block` stragglers in `aes_gcm`.

Otherwise LGTM :+1:

src/aead/aes_gcm.rs
briansmith: Since the end goal is to remove `Block`, I think using `Block::from` here is better, because that helps us see what we need to change. If we use `Into::into()` instead then the code maybe would still compile even after we remove `Block` even if we don't remove the `.into()`.

src/aead/aes_gcm.rs
briansmith: Ditto.

unseddd (approved):

briansmith:
I'm going to avoid removing the use of `Block` from AES and GCM for now. Instead, I've posted PR #1189 which removes the use of `Block` from non-AES/GCM parts of *ring*. I intend to follow that up with another PR that cleans up `aead::block`, removing all its `unsafe` blocks, for use by AES and GCM, at least for now.

PR #1042, opened 2020-08-04
Fix alignment of ChaCha20 keys.

briansmith:
When we removed the use of `Block` from ChaCha20, we didn't preserve the alignment. Fix that.

BoringSSL uses this declaration of `ChaCha20_ctr32`:

```
void ChaCha20_ctr32(uint8_t *out, const uint8_t *in, size_t in_len,
                    const uint32_t key[8], const uint32_t counter[4]);
```

### DIFF
--- a/src/aead/chacha.rs
+++ b/src/aead/chacha.rs
@@ -16,13 +16,13 @@
 use super::{counter, iv::Iv, quic::Sample, BLOCK_LEN};
 use crate::{c, endian::*};
 
-#[repr(C)]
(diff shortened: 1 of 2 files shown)


briansmith:
@rugo Do you know how to make QEMU check alignment.

Is https://stackoverflow.com/questions/51520635/how-to-emulate-arm-unaligned-memory-access-exceptions (no alignment checking) still the current state of QEMU?

rugo:
No clue to be honest, sorry.

Issue #1003, opened 2020-06-08
Ed25519: Please generalize exporting a key into PKCS#8 format


triska:
In [`library(crypto)`](https://github.com/mthom/scryer-prolog/blob/master/src/lib/crypto.pl), I would like to provide the new predicate `ed25519_seed_keypair(+Seed, -Pair)`, which takes `Seed` (a list of bytes) as input and relates it to a new key&nbsp;pair `Pair` that is generated from `Seed`.

In *ring*, `Ed25519KeyPair::from_seed_unchecked` seems ideally suited to implement this. However, once I have generated a key in this way, is there any way to make it available to the Prolog&nbsp;level in PKCS#8 format? I would like to use this format because this is also what the existing predicate `ed25519_new_keypair/1` uses. Internally, this predicate uses `Ed25519KeyPair::generate_pkcs8`.

Could you please consider generalizing exporting a key into PKCS#8 format to cases where the key was generated from a seed?

Thank you a lot!

briansmith:
> However, once I have generated a key in this way, is there any way to make it available to the Prolog level in PKCS#8 format?

*ring* intentionally doesn't allow a `KeyPair` to be serialized and intentionally doesn't allow the private key of the keypair to be accessed using its API.

If I understand correctly, you are looking for an API that converts a "raw" seed to PKCS#8 format, right?

Check out this:

```
    pub fn generate_pkcs8(
        rng: &dyn rand::SecureRandom,
    ) -> Result<pkcs8::Document, error::Unspecified> {
        let seed: [u8; SEED_LEN] = rand::generate(rng)?.expose();
        let key_pair = Self::from_seed_(&seed);
        Ok(pkcs8::wrap_key(
            &PKCS8_TEMPLATE,
            &seed[..],
            key_pair.public_key().as_ref(),
        ))
    }
```

IIUC, What you need is a new API like this, right?:

```
    pub fn seed_to_pkcs8(
        seed: &Seed,
    ) -> Result<pkcs8::Document, error::KeyRejected> {
        let key_pair = Self::from_seed_(seed);
        Ok(pkcs8::wrap_key(
            &PKCS8_TEMPLATE,
            &seed[..],
            key_pair.public_key().as_ref(),
        ))
   }
```

If somebody makes a PR that adds this and adds test for it, I'm happy to add it.

triska:
Thank you a lot, I have filed PR #1044 for this!

justprosh:
[...]

PR #1044, opened 2020-08-08
ADDED: seed_to_pkcs8, deriving an Ed25519 key pair from a seed.

triska:
This is useful for example to generate keys from passwords, so
that data can be signed without having to store the key.

This addresses #1003.

### DIFF
--- a/tests/ed25519_tests.rs
+++ b/tests/ed25519_tests.rs
@@ -46,6 +46,14 @@ fn test_signature_ed25519() {
         let public_key = test_case.consume_bytes("PUB");
         assert_eq!(32, public_key.len());
 
+        // Test deterministic creation of key pair from seed.
+        let mut seed_bytes = [0;32];
+        seed_bytes.copy_from_slice(&seed);
+
+        let pkcs8_from_seed = Ed25519KeyPair::seed_to_pkcs8(&seed_bytes).unwrap();
+        let key_pair_from_seed = Ed25519KeyPair::from_pkcs8(pkcs8_from_seed.as_ref()).unwrap();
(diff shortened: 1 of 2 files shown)


tests/ed25519_tests.rs
```diff
@@ -46,6 +46,14 @@ fn test_signature_ed25519() {
         let public_key = test_case.consume_bytes("PUB");
         assert_eq!(32, public_key.len());
 
+        // Test deterministic creation of key pair from seed.
+        let mut seed_bytes = [0;32];
```
unseddd, line 50: Small nit: what do you think about moving this chunk into its own test function?

unseddd (changes requested):
One small nit, otherwise LGTM.

tests/ed25519_tests.rs
triska: Thank you a lot for taking a look, and for your comments!

To me, this seemed a natural place to add this test case with few modifications to existing code, since this is the place where the public key is read and used, and where both the seed and the public key are already available. The existing function already tests many different features that all depend on the read values and can therefore be naturally tested as a logical unit in this function.  Would there be any benefit in adding a new function just for testing the relation between seed and public key? I will do it if that is necessary to merge this pull&nbsp;request.

unseddd: > Would there be any benefit in adding a new function just for testing the relation between seed and public key?

Main benefit would be isolation of functionality, but is outweighed by the points you brought up.

> I will do it if that is necessary to merge this pull request.

I'm just a contributor with no say over merge-ability.

unseddd (approved):

PR #1051, opened 2020-09-02
Fix a PBKDF2 documentation comment typo

BlackHoleFox:
One of the documentation comments for the predefined PBKDF2 algorithms (HMAC-SHA256) was incomplete so this fixes it to be like the others.

This is a very small change, so feel free to take care of this as part of a bigger fix later.


briansmith:
Thanks!

PR #1053, opened 2020-09-09
Move from `spin` to `conquer-once`

ErichDonGubler:
This resolves [RUSTSEC-2019-0031][0].

[0]: https://rustsec.org/advisories/RUSTSEC-2019-0031.html

---

The advisory page suggests several alternatives to the now-deprecated `spin` crate. The reason I picked `conquer-once` was because it seemed to have the most similar (and limited) API to how `spin` was being used -- it seemed a good small dependency to trade for.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -297,7 +297,7 @@ name = "ring"
 untrusted = { version = "0.7.1" }
 
 [target.'cfg(all(any(target_arch = "aarch64", target_arch = "arm", target_arch = "x86", target_arch = "x86_64"), not(target_os = "ios")))'.dependencies]
-spin = { version = "0.5.2", default-features = false }
+conquer-once = { version = "0.2.1", default-features = false }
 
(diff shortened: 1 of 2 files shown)


ErichDonGubler:
The CI failures here don't seem related, esp. since they're occurring in other PRs right now (for example, [this one](https://ci.appveyor.com/project/briansmith/ring/builds/34998935)).

kellpossible:
Seems like `spin` is being maintained again now with a new `0.6` release?

ErichDonGubler:
@kellpossible: I've [asked in upstream `spin`](https://github.com/mvdnes/spin-rs/issues/79#issuecomment-706301169) for clarification around the current maintainership of the project.

briansmith:
Thanks for the PR. I also saw your most recent comment in the issue. I do think we should just close this. I don't know if conquer-once or spin-rs is better but also I don't want to spend a lot of time on the issue. As was noted in the issue, `cargo deny` is no longer warning about spin-rs so this isn't urgent. Thanks again for the PR though!

PR #1055, opened 2020-09-14
Added build configuration for Apple Silicon based Macs.

chadseld:
Added the required build configuration changes to support compilation on Apple Silicon Macs. 

Tested on a Developer Transition Kit Mac.

> cargo build
> cargo test

For now, the flavor portion of the ASM_TARGETS triple is left as ios64. With some additional changes to arm-xlate.pl, we could insert a custom string here (e.g. macArm64).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -235,6 +235,7 @@ const ASM_TARGETS: &[(&str, Option<&str>, Option<&str>)] = &[
     ("x86_64", Some(WINDOWS), Some("nasm")),
     ("x86_64", None, Some("elf")),
     ("aarch64", Some("ios"), Some("ios64")),
+    ("aarch64", Some("macos"), Some("ios64")),
     ("aarch64", None, Some("linux64")),
     ("x86", Some(WINDOWS), Some("win32n")),
     ("x86", Some("ios"), Some("macosx")),
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -33,7 +33,9 @@ pub(crate) fn features() -> Features {
             target_arch = "x86",
             target_arch = "x86_64"
         ),
-        not(target_os = "ios")
+        not(
+            any(target_os = "ios",
+                target_os = "macos"))
     ))]
     {
         static INIT: spin::Once<()> = spin::Once::new();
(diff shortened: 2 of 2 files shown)


repi:
Linking together with #1063

Darkspirit:
https://github.com/alexcrichton/openssl-src-rs/pull/77 added aarch64-apple-darwin to openssl CI.

src/cpu.rs
```diff
@@ -33,7 +33,9 @@ pub(crate) fn features() -> Features {
             target_arch = "x86",
             target_arch = "x86_64"
         ),
-        not(target_os = "ios")
+        not(
+            any(target_os = "ios",
+                target_os = "macos"))
```
briansmith, line 38: AFAICT, this would have disabled CPU feature detection on x86-64 Macs, so I didn't take this change.

Also, when I was trying to understand your PR, I noticed the bug that PR #1100 fixes. (That bug is just a performance bug, not a correctness bug, but it is a pretty severe performance bug.)

build.rs
```diff
@@ -235,6 +235,7 @@ const ASM_TARGETS: &[(&str, Option<&str>, Option<&str>)] = &[
     ("x86_64", Some(WINDOWS), Some("nasm")),
     ("x86_64", None, Some("elf")),
     ("aarch64", Some("ios"), Some("ios64")),
+    ("aarch64", Some("macos"), Some("ios64")),
```
briansmith, line 238: I agree that this is the right thing to do; this is what OpenSSL upstream does and this is what PR #1101 now does.

src/cpu.rs
```diff
@@ -173,21 +175,22 @@ pub(crate) mod arm {
         #[cfg_attr(
             any(
                 target_os = "ios",
+                target_os = "macos",
```
briansmith, line 178: A lot of my recent changes to cpu.rs were to clean up this repetitive logic that existed just to avoid dead code warnings.

briansmith:
Thanks for this PR. I used this as input for making PR #1100 and PR #1101. It was also motivation for cleaning up cpu.rs. I hope to complete more cleanup after the release.

PR #1056, opened 2020-09-15
Update doc links to use correct branch

josephlr:
It looks like *ring* is using `main` now.

Signed-off-by: Joe Richey <joerichey@google.com>


briansmith:
Thanks!

PR #1057, opened 2020-09-18
Initial ppc64le Support

erichte-ibm:
Revive PR #819 (by q66) in reduced-scope by implementing powerpc64le support in ring. This effort supports only little endian machines using POWER8 and later processors for now. The goal is to enable ring to target modern powerpc64le machines.

Only the bare minimum needed to build ring and pass ring tests is included; there is no new optimized code for algorithms already implemented in C/Rust. This has been tested on a POWER9 Blackbird, and on a POWER8 machine.

### DIFF
--- a/src/cpu/arm.rs
+++ b/src/cpu/arm.rs
@@ -12,6 +12,11 @@
 // OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
+#![cfg_attr(
+    not(any(target_arch = "aarch64", target_arch = "arm")),
+    allow(dead_code)
+)]
+
 #[cfg(all(
     any(target_os = "android", target_os = "linux"),
     any(target_arch = "aarch64", target_arch = "arm")
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,6 +180,7 @@ jobs:
           - i686-pc-windows-msvc
           - i686-unknown-linux-gnu
           - i686-unknown-linux-musl
+          - powerpc64le-unknown-linux-gnu
           - riscv64gc-unknown-linux-gnu
           - x86_64-pc-windows-gnu
           - x86_64-pc-windows-msvc
@@ -246,6 +247,9 @@ jobs:
             host_os: ubuntu-22.04
 
           - target: i686-unknown-linux-musl
+            host_os: ubuntu-22.04
+
+          - target: powerpc64le-unknown-linux-gnu
             host_os: ubuntu-22.04
 
           - target: riscv64gc-unknown-linux-gnu
@@ -337,6 +341,7 @@ jobs:
         target:
           - aarch64-unknown-linux-musl
           - i686-pc-windows-msvc
+          - powerpc64le-unknown-linux-gnu
           - x86_64-unknown-linux-gnu
 
         mode:
@@ -353,6 +358,9 @@ jobs:
 
           - target: i686-pc-windows-msvc
             host_os: windows-latest
+
+          - target: powerpc64le-unknown-linux-gnu
+            host_os: ubuntu-22.04
 
           - target: x86_64-unknown-linux-gnu
             host_os: ubuntu-22.04
@@ -487,6 +495,7 @@ jobs:
         target:
           - aarch64-unknown-linux-gnu
           - i686-unknown-linux-gnu
+          - powerpc64le-unknown-linux-gnu
           - riscv64gc-unknown-linux-gnu
           - x86_64-unknown-linux-musl
 
@@ -509,6 +518,9 @@ jobs:
           # https://github.com/rust-lang/rust/issues/79556 and
           # https://github.com/rust-lang/rust/issues/79555 are fixed.
           - target: i686-unknown-linux-gnu
+            host_os: ubuntu-22.04
+
+          - target: powerpc64le-unknown-linux-gnu
             host_os: ubuntu-22.04
 
           - target: riscv64gc-unknown-linux-gnu
--- a/include/ring-core/target.h
+++ b/include/ring-core/target.h
@@ -40,6 +40,9 @@
 #elif defined(__MIPSEL__) && defined(__LP64__)
 #define OPENSSL_64_BIT
 #define OPENSSL_MIPS64
+#elif (defined(__PPC64__) || defined(__powerpc64__)) && defined(_LITTLE_ENDIAN)
+#define OPENSSL_64_BIT
+#define OPENSSL_PPC64LE
 #elif defined(__riscv) && __SIZEOF_POINTER__ == 8
 #define OPENSSL_64_BIT
 #define OPENSSL_RISCV64
--- a/mk/cargo.sh
+++ b/mk/cargo.sh
@@ -21,6 +21,7 @@ rustflags_self_contained="-Clink-self-contained=yes -Clinker=rust-lld"
 qemu_aarch64="qemu-aarch64 -L /usr/aarch64-linux-gnu"
 qemu_arm="qemu-arm -L /usr/arm-linux-gnueabihf"
 qemu_mipsel="qemu-mipsel -L /usr/mipsel-linux-gnu"
+qemu_powerpc64le="qemu-ppc64le -L /usr/powerpc64le-linux-gnu"
 qemu_riscv64="qemu-riscv64 -L /usr/riscv64-linux-gnu"
 
 # Avoid putting the Android tools in `$PATH` because there are tools in this
@@ -99,6 +100,13 @@ case $target in
     export AR_mipsel_unknown_linux_gnu=mipsel-linux-gnu-gcc-ar
     export CARGO_TARGET_MIPSEL_UNKNOWN_LINUX_GNU_LINKER=mipsel-linux-gnu-gcc
     export CARGO_TARGET_MIPSEL_UNKNOWN_LINUX_GNU_RUNNER="$qemu_mipsel"
+    ;;
+  powerpc64le-unknown-linux-gnu)
+    export CC_powerpc64le_unknown_linux_gnu=clang-$llvm_version
+    export AR_powerpc64le_unknown_linux_gnu=llvm-ar-$llvm_version
+    export CFLAGS_powerpc64le_unknown_linux_gnu="--sysroot=/usr/powerpc64le-linux-gnu"
+    export CARGO_TARGET_POWERPC64LE_UNKNOWN_LINUX_GNU_LINKER=powerpc64le-linux-gnu-gcc
+    export CARGO_TARGET_POWERPC64LE_UNKNOWN_LINUX_GNU_RUNNER="$qemu_powerpc64le"
     ;;
   riscv64gc-unknown-linux-gnu)
     export CC_riscv64gc_unknown_linux_gnu=clang-$llvm_version
--- a/mk/install-build-tools.sh
+++ b/mk/install-build-tools.sh
@@ -91,6 +91,13 @@ case $target in
     libc6-dev-mipsel-cross \
     qemu-user
   ;;
+--target=powerpc64le-unknown-linux-gnu)
+  use_clang=1
+  install_packages \
+    gcc-powerpc64le-linux-gnu \
+    libc6-dev-ppc64el-cross \
+    qemu-user
+  ;;
 --target=riscv64gc-unknown-linux-gnu)
   use_clang=1
   install_packages \


erichte-ibm:
Updated the branch on top of current `main`, fixed some issues reported by CI.

erichte-ibm:
Updated the branch on top of current `main`. Removed algorithm-specific assembly to match commit 501fc4eeaa , only the minimum required BN assembly from OpenSSL is included.

erichte-ibm:
Rebased branch on top of current `main`. 

Please let us know if there is anything you need from us to aid you in moving this forward.

a-m-joseph:
Did you ever hear from @briansmith?

runlevel5:
@erichte-ibm @briansmith is there anything I could help to get this PR wrap up?

erichte-ibm:
> Did you ever hear from @briansmith?

We are still waiting to hear back, though we have be actively rebasing every week or so to try to keep this PR up to date.


> @erichte-ibm @briansmith is there anything I could help to get this PR wrap up?

Reviews and testing would be wonderful! So far, we have tested on little endian P8/P9/P10. We'd love to know your use case and platform so that we can work together on any issues you might find.

a-m-joseph:
I've added [support](https://github.com/NixOS/nixpkgs/pull/168199) for powerpc64le to [nixpkgs](https://github.com/NixOS/nixpkgs)[1].

Unfortunately Rust's recursive hashing scheme (`Cargo.lock`) makes it really awkward to use out-of-tree patches to a deep library dependency like Ring.  You need to use the cargo `[patch]` stanza, but unfortunately this requires recalculating the entire `Cargo.lock`, which requires network access to crates.io, which unfortunately is a non-starter for our reproducible/auditable/deterministic build framework.

That framework, and "no network access inside a build" is deeply fundamental to the breakthroughs that make it possible for nix to manage complexity with drastically less effort than any other packaging scheme.

[1] Nixpkgs is sort of like Gentoo Linux, except that:

* It can run without root or needing to take over the whole machine, on top of any other Linux distribution
* It natively supports installing multiple different versions of a package or library, with zero additional effort or complexity 
* It uses a very simple and elegant lazy functional language instead of Python
* It keeps all generated files (compiled binaries, etc) in an immutable-once-written garbage collected "heap on the filesystem" (`/nix/store`)

Gentoo can do the first two things, sort of, sometimes, with Gentoo Prefixes, but those only work for certain packages and are still sort of an experimental feature.

erichte-ibm:
> You need to use the cargo `[patch]` stanza, but unfortunately this requires recalculating the entire `Cargo.lock` during builds, which requires network access to https://crates.io. 

Does adding a feature still require a `[patch]` to enable that feature for ring somewhere in the downstream dependency chain? (e.g. projects that depend on something like `rustls` and not ring directly)


> 
> I would really like to see this merged, even if it's hidden behind an off-by-default "yes I am crazy and I know what I'm doing" feature flag. We would expose it via an off-by-default flag with a similarly-scary name to our own users.

A feature would be somewhat redundant since we check the `target_arch` in most places, however it would make it more explicit that we wish to use a "less supported" platform.

@briansmith Would adding an explicit feature for POWER support help get this reviewed/merged?

a-m-joseph:
> > You need to use the cargo `[patch]` stanza, but unfortunately this requires recalculating the entire `Cargo.lock` during builds, which requires network access to https://crates.io.
> 
> Does adding a feature still require a `[patch]` to enable that feature for ring somewhere in the downstream dependency chain? (e.g. projects that depend on something like `rustls` and not ring directly)

I don't think features are covered by the `Cargo.lock` `checksum` field, but I've had an extraordinarily hard time finding a formal specification of that file format (precisely how the checksum is computed, an exhaustive list of the fields, grammar for the `source` quasi-urls, etc...).  Do you know of one?

But even if it is covered by the checksum, a `[patch]` stanza will not be required, because feature flags are "[unioned up the dependency tree](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification)".  Cargo will only build one copy of `ring` per binary, even if different parts of the binary express dependencies upon `ring` with different feature sets.  Cargo will union all of those feature sets and build `ring` only once.

This means that it is enough to add an extra (possibly redundant) `ring` dependency **at the top level** with the scarily-named feature flag enabled.  That is enough for every package throughout the entire dependency tree of that particular binary to be built with the feature enabled.

The `Cargo.lock` file contains an entry for the top-level package, but that entry is special: it has no `checksum` field.  This makes sense, because `Cargo.lock` is distributed alongside the top-level package's source code.  Including a checksum would be redundant.  So we can add feature flags there without breaking any checksums.

erichte-ibm:
> Cargo will union all of those feature sets and build ring only once.

Good to know!

> This means that it is enough to add an extra (possibly redundant) ring dependency at the top level with the scarily-named feature flag enabled

With that knowledge, this makes sense to us at least. Hopefully we can hear back from the maintainer at some point if they agree this is an acceptable approach.

bkeys:
I would like to have a rust desk build for ppc64le; it would be appreciated if someone reviewed this pull request if there is nothing wrong with it.

rjzak:
I'm thankful for this work, it's enabled me to listen to Spotify with [Psst](https://github.com/jpochyla/psst) on my Talos II. So far, so good! I've tested it with some other projects as well.

justinlynn:
Confirmed that this project builds with hid-io/hid-io-core#5394d79da0b9fbc4a56bc104ca468e992be1241e using cargo patch stanza pointed to IBM/ring#3f8f04a1057e286f3b8c5b9e39c9f795b15743cc (ppc-0.16.20), but required reverting f06811a150fcded1555911678bbca6dcb5440cda due to API usage in rcgen 0.9. Additionally, PR branch proposed builds with multiple packages using latest ring HEAD. Thank you to @erichte-ibm and the rest of the maintainers of the ppc64le ring port for unblocking so many packages with rcgen, webpki, and rustls dependencies!

justinlynn:
Note for other users requiring ppc64le support: IBM/ring also implements a patchset for ring-0.16.20 ( https://github.com/IBM/ring/tree/ppc-0.16.20 ) and generally tracks main on ( https://github.com/IBM/ring/tree/main-ppc ) but the branch with which this PR is concerned ( https://github.com/IBM/ring/tree/initial-ppc64le-support-pr ) is kept fully up to date and thus also tracks IBM/ring@main-ppc .

runlevel5:
Wondering if the core team could help review this PR. It would mean a world for the POWER community

amjoseph-nixpkgs:
Apparently @briansmith [replies to Hacker News comments about ring within minutes](https://news.ycombinator.com/item?id=33386439), but waits multiple years before acknowledging pull requests here on github.

Perhaps Hacker News would be a better place to get his attention?

Seriously, I've run out of ideas here; that's all I can come up with.

briansmith:
First, I think that with the changes to base.h, PR #1558 will get PowerPC working.

In order to have PowerPC-specific code paths (optimizations), at a minimum we need to add PoewrPC to the `coverage` jobs in GitHub Actions to ensure that all the new code paths are actually being tested. This probably requires modifying the Rust toolchain to add support for the profiler builtins to the Linux PowerPC targets, like https://github.com/rust-lang/rust/pull/104304 does for s390x. Also, we should get some assurance that somebody (Google?) is regularly fuzzing the PowerPC code. We also need to verify that this PowerPC code matches exactly what's in the latest BoringSSL.

We also should have some performance justification; i.e. we should have the ability to benchmark the before/after. I can help somebody with this. In this case, we should add benchmarks for `bn_mul_mont`.

Given all that, I think we should take the baby step of an initial PR that just changes base.h so that things will work after #1558 is merged.

amjoseph-nixpkgs:
> In order to have PowerPC-specific code paths (optimizations)

> We also should have some performance justification

The justification is **side channel leakage**, not performance.

Cryptographic primitives are written in assembler to prevent a compiler from optimizing your constant-time primitives into faster variable-time instruction sequences.

briansmith:
> The justification is **side channel leakage**, not performance.

We need to find solutions to these challenges that don't rely on PerlAsm. I do think there is a role for assembly language to play, and I'm experimenting with using the Rust compiler's inline assembly mechanism. I highly recommend people do the work to get POWER/PowerPC inline assembly working and in Stable Rust so we can use it, as PerlAsm isn't a viable long-term solution for any target.

rjzak:
> I highly recommend people do the work to get POWER/PowerPC inline assembly working and in Stable Rust

What's required for this to happen?

briansmith:
> What's required for this to happen?

I don't know. Maybe it is already done? It would be great if somebody could check with the rust-lang/rust project and report back. (Also for `std::simd` for POWER/PowerPC.)

programmerjake:
> > What's required for this to happen?
> 
> I don't know. Maybe it is already done? It would be great if somebody could check with the rust-lang/rust project and report back. (Also for `std::simd` for POWER/PowerPC.)

std::simd works fine on PowerPC iirc, no special rustc support is required because it just generates generic llvm ir, no arch-specific ir.

erichte-ibm:
> Maybe it is already done?

Just tested, inline assembly appears to work with a trivial test in nightly with `#![feature(asm_experimental_arch)]`, and there is a tracking issue [here for the experimental archs](https://github.com/rust-lang/rust/issues/93335) .

programmerjake:
> > Maybe it is already done?
> 
> Just tested, inline assembly appears to work with a trivial test in nightly with `#![feature(asm_experimental_arch)]`, and there is a tracking issue [here for the experimental archs](https://github.com/rust-lang/rust/issues/93335) .

last I checked PowerPC inline assembly mostly works but is missing support for some clobbers and has no vmx/vsx support. it may also have the wrong set of default clobbers.

erichte-ibm:
> We also need to verify that this PowerPC code matches exactly what's in the latest BoringSSL.

Currently, this PR pulled code from OpenSSL, as BoringSSL does not appear to have support for POWER. I'm double checking now, but I think it was only the `bn_mul_mont` that was actually needed to support the build, as there wasn't a fallback before to use.

> Given all that, I think we should take the baby step of an initial PR that just changes base.h so that things will work after https://github.com/briansmith/ring/pull/1558 is merged.

Agreed, and since it appears to have just been merged, I will work on rebasing onto current upstream and see if we can drop the perlasm.

> In order to have PowerPC-specific code paths (optimizations), at a minimum we need to add PoewrPC to the coverage jobs in GitHub Actions to ensure that all the new code paths are actually being tested.

Is this needed for the initial baseline support, or specifically for any future optimizations?

pkubaj:
> Currently, this PR pulled code from OpenSSL, as BoringSSL does not appear to have support for POWER.

Last I checked, BoringSSL definitely supported ppc64le, although not ppc64 or ppc.

erichte-ibm:
> > Currently, this PR pulled code from OpenSSL, as BoringSSL does not appear to have support for POWER.
> 
> Last I checked, BoringSSL definitely supported ppc64le, although not ppc64 or ppc.

Ah apologies, got ahead of myself there with the wording. It builds on POWER, but it appears to only include optimized assembly for x86 and ARM. Doing a quick grep, BoringSSL does has POWER assembly for AES/GHASH, but not for ECC or chacha/poly1305. For previous versions of this PR, those were sourced from OpenSSL.

runlevel5:
@erichte-ibm what's the progress of this PR please? AFAIK there is also another branch https://github.com/IBM/ring/commits/ppc-0.16.20 downstream.

erichte-ibm:
To be honest, I don't know. Last I checked, ppc64le support still relies on some perlasm to work in ring, albeit less than the original pull request. If that is a blocker for initial support, then we can't make much progress here. There have been some changes though since the last time I really looked at this PR though, so when I get the chance in the next few days, I'll dive into refreshing this PR branch.

The other note was CI -- I've coincidentally be doing work on testing ppc64le code in Actions via docker/qemu, so if running tests / getting coverage is an issue as well, I can also look into adapting that here as well.

As for the IBM/ring `ppc-0.16.20` branch, that exists for other projects to patch support so that they can compile on ppc64le. Right now, it's actually a cobbled mess of newer commits that causes some build issues. I will be pushing a revised version of the branch soon that is based directly on `0.16.20` that should fix it.

src/cpu/arm.rs
```diff
@@ -12,6 +12,11 @@
 // OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
+#![cfg_attr(
+    not(any(target_arch = "aarch64", target_arch = "arm")),
+    allow(dead_code)
+)]
```
briansmith, line 18: This doesn't seem right to me. Why is it here?

erichte-ibm: There is a similar line at the top of `src/cpu/intel.rs` [here](https://github.com/briansmith/ring/blob/main/src/cpu/intel.rs#L15). Without it, building on ppc64le yields dead code warnings, e.g.:
```
warning: method `available` is never used
   --> src/cpu/arm.rs:206:12
    |
204 | impl Feature {
    | ------------ method in this implementation
205 |     #[inline(always)]
206 |     pub fn available(&self, _: super::Features) -> bool {
    |            ^^^^^^^^^
```

These warnings will probably appear on any non-ARM or non-Intel platform. They don't appear on Intel because they are referenced in functions like `detect_implementation()` in [`src/aead/aes.rs`](https://github.com/briansmith/ring/blob/main/src/aead/aes.rs#L404) -- however in this example, default `_cpu_features` are used on any non-Intel or non-ARM platform, and skip the calls to the cpu-specific functions.

briansmith:
Could you please rebase on top of `main` and then resubmit this so that the CI run will kick in?

I think that we need to adapt the CI changes from #1297 to also test this target in CI, especially because I believe this will be the first 64-bit little-endian target that uses some of the fallback code, so we need to make sure it is tested in CI. Originally we had some MIPS CI coverage for this code, but I had to remove it because that MIPS target got downgraded by the Rust project and broke our build.

Here is how I am going to test this PR locally:
```
mk/install-build-tools.sh --target=WHATEVER
mk/cargo.sh test --target=WHATEVER
```

Roughly you need to copy and adapt this logic in `mk/install-build-tools.sh`:
```
--target=aarch64-unknown-linux-musl|--target=armv7-unknown-linux-musleabihf)
  use_clang=1
  install_packages \
    qemu-user
  ;;
```
And this logic in `mk/cargo.sh`:
```
  aarch64-unknown-linux-gnu)
    export CC_aarch64_unknown_linux_gnu=clang-$llvm_version
    export AR_aarch64_unknown_linux_gnu=llvm-ar-$llvm_version
    export CFLAGS_aarch64_unknown_linux_gnu="--sysroot=/usr/aarch64-linux-gnu"
    export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc
    export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER="$qemu_aarch64"
    ;;
```
Then in .github/ci.yml you need to add a line to the target matrix for this target. See #1297.

(Please use clang and not GCC for this CI integration.)

briansmith:
I am intending to do a release soon, probably this week. If this PR is ready to merge before the release, I will include it. Otherwise, I am happy to do a 0.17.1 release with this.

briansmith:
> To be honest, I don't know. Last I checked, ppc64le support still relies on some perlasm to work in ring, albeit less than the original pull request. If that is a blocker for initial support, then we can't make much progress here. There have been some changes though since the last time I really looked at this PR though, so when I get the chance in the next few days, I'll dive into refreshing this PR branch.

Since we got MIPS working at one point in CI, I believe that PowerPC should work now as well.

erichte-ibm:
Thanks for the response! I've pushed a rebased branch with the CI support added, though I have a couple comments/questions:


1. There seems to be an issue in the fallback code when compiling without the `"alloc"` feature -- `bn_mul_mont_fallback.rs` is not linked in, as the `bigint` mod is not included without `"alloc"`, however the C files `gfp_p256.c` and `gfp_p384.c` are still compiled and expect the prefixed symbol exported. Do you have a preferred method for resolving this?
2. Like the s390x PR, this still uses the gcc linker -- for some reason I am getting linking errors with the LLVM linker, I'm going to keep working on that. It's possible it's either my test environment, or I am calling the wrong thing.

briansmith:
> There seems to be an issue in the fallback code when compiling without the "alloc" feature -- bn_mul_mont_fallback.rs is not linked in, as the bigint mod is not included without "alloc", however the C files gfp_p256.c and gfp_p384.c are still compiled and expect the prefixed symbol exported. Do you have a preferred method for resolving this?

I will post a PR to fix this today.

> Like the s390x PR, this still uses the gcc linker -- for some reason I am getting linking errors with the LLVM linker, I'm going to keep working on that. It's possible it's either my test environment, or I am calling the wrong thing.

Don't worry about it for now.

briansmith:
Please rebase on main and resubmit. Thanks!

erichte-ibm:
> Please rebase on main and resubmit. Thanks!

Done!

runlevel5:
It's seems to me the coverage ci build failed due to missing token

ref: https://github.com/briansmith/ring/actions/runs/6321304636/job/17166234616?pr=1057#step:7:121

I think it is unrelated to changes of this PR

@briansmith wondering if you could manually retry that failed build. Many thanks

briansmith:
Good news:
The refactoring to fix the --no-default-features build was merged into main.

Less good news:
There's a merge conflict in crypto/internal.h from the latest BoringSSL merge. Also, PR #1663 moved the stuff in base.h to target.h. I will merge PR #1663 ASAP so you can rebase this on top of main for the release. I am hoping to finish the BoringSSL merge today so there shouldn't be any new conflicts.

erichte-ibm:
Rebased on top of new main, moved definitions to `target.h`.

briansmith:
Sorry, this is a nightmare for merge conflicts because the merges I have are riscv (done), ppc64le, and s390x, which alphabetically are adjacent, so every change conflicts.

Please rebase one more time and I will do my best to merge this before I merge another conflict. Then I will merge s390x after that is rebased.

Sorry for the inconveniences.

erichte-ibm:
No worries! Rebased, and fixed the conflicts, preserving the alphabetic order of the targets.

briansmith (approved):

briansmith:
Thank you!

PR #1060, opened 2020-10-04
Re-enable debug = true in Cargo.toml

bbqsrc:
https://github.com/rust-lang/rust/issues/28233 seems to have been fixed for a while now, and I could not re-produce the original build fails that caused this to be disabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -334,11 +334,9 @@ std = ["alloc"]
 test_logging = []
 wasm32_c = []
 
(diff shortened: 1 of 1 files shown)


briansmith:
Thanks. I am glad to see this. I will need to verify that this doesn't have any performance or size impact before merging.

briansmith:
I think instead we should just remove the outdated comment and stop overriding the default values. The profiles are really only used during development of *ring* and don't affect users' builds, IIUC.

briansmith:
I'm going to close this as not planned but I appreciate the thought.

Issue #713, opened 2018-10-30
Add support for statically-linked musl libc Linux targets (-unknown-linux-musl)


briansmith:

gnzlbg:
Why is removing the `libc` crate dependency a goal ?

cbiffle:
FYI - from updates on rust-lang/libc#659, `getrandom` is now available in `libc` on *most* Linux platforms. The holdout is not `musl`, but `android`/Bionic. I suggest decoupling the two parts of this issue an moving the `getrandom` portion into a separate bug.

briansmith:
More discussion in https://github.com/briansmith/ring/pull/935

briansmith:
@aig787's PR #1067 added the -musl targets for Linux x86_64 and i686 to CI/CD.

Still to do:

- [ ] https://github.com/rust-lang/rust/pull/70740/ indicates that we need to verify that we're doing the right thing w.r.t. PIE (position independent code), especially when compiling the C and assembly code.
- [ ] Add Aarch64 musl targets to CI/CD.
- [ ] Minimize the dependencies installed. Currently the i686 job installs "libc6-dev-i386" and "gcc-multilib", but neither of those should be needed for -musl.

briansmith:
- [ ] https://github.com/alexcrichton/cc-rs/issues/436 indicates that we should be doing more sanity checks to make sure we're not using the wrong header files. e.g. statically asserting that `sizeof(uint64_t) == 8`.

> rust-lang/rust#70740 indicates that we need to verify that we're doing the right thing w.r.t. PIE (position independent code), especially when compiling the C and assembly code.

See also https://github.com/alexcrichton/cc-rs/issues/165. We should have a step in CI/CD that verifies that we support PIE correctly.

briansmith:
> * Minimize the dependencies installed. Currently the i686 job installs "libc6-dev-i386" and "gcc-multilib", but neither of those should be needed for -musl.

I attempted to remove the gcc-multilib dependency but the build fails with:
```
[...]

PR #1067, opened 2020-10-27
Added musl targets

aig787:
Building off of #935, I added musl targets for i686 and x86_64 using clang as the compiler. Addresses #713

### DIFF
--- a/mk/travis.sh
+++ b/mk/travis.sh
@@ -73,6 +73,12 @@ if [[ ! "$TARGET_X" =~ "x86_64-" ]]; then
   fi
 fi
 
+# This stanza is separate from the above because MUSL requires a target install,
+# but does *not* require special linking / multilib.
+if [[ "$TARGET_X" =~ .*"-unknown-linux-musl" ]]; then
+  rustup target add "$TARGET_X"
+fi
+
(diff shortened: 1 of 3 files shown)


mk/travis.sh
```diff
@@ -73,6 +73,12 @@ if [[ ! "$TARGET_X" =~ "x86_64-" ]]; then
   fi
 fi
 
+# This stanza is separate from the above because MUSL requires a target install,
+# but does *not* require special linking / multilib.
```
briansmith, line 77: Let's remove this comment and...

mk/travis.sh
```diff
@@ -73,6 +73,12 @@ if [[ ! "$TARGET_X" =~ "x86_64-" ]]; then
...
+if [[ "$TARGET_X" =~ .*"-unknown-linux-musl" ]]; then
```
briansmith, line 78: Let's move the `rustup target add` from up above to here. We can change the condition to `[` "$TARGET_X" =~ .*"-unknown-linux-musl" || ! "$TARGET_X" =~ "x86_64-" ]]` or whatever Bashism.

That way, we have only one place where we do `rustup target add`. This will be important for other refactoring I am doing.

PR #1088, opened 2020-11-11
Clean up and fix .gitattributes files.

briansmith:
GitHub is reporting some include files as C++ instead of C, and it is reporting
some .pl files as Perl instead of Assembly. Attempt to fix that.

Remove superfluous entries in ./.gitattributes and consolidate some of the
.gitattributes files into ./.gitattributes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,10 +1,7 @@
 * text=auto !eol
-*.sln eol=crlf
-*.vcxproj eol=crlf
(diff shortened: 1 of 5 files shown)


briansmith:
This did change GitHub's reporting of the .inl/.h files as C++; they are now reported as C. The PerlAsm source files are still being reported as Perl though.

PR #1097, opened 2020-11-17
Switch Poly1305 implementation to the BoringSSL implementation.

briansmith:
Previously the OpenSSL implementation was being used. Switch to the BoringSSL
version.

Switching to the BoringSSL implementation will make it easier to refactor the CPU feature
detection, which is important for upcoming ports.

This switch will also implicitly add support for BTI and pointer authentication for
Poly1305.

This is based on BoringSSL 63d06626d3a104868eee622e8e56d9f2dd643366.


briansmith:
A note on how to review changes like this:

* Review the diff between *ring* and *ring* (i.e. what is shown in the GitHub UI or the `git diff` output).
* Review the diff between *ring* and BoringSSL. In this case, compare BoringSSL's crypto/poly1305/* to *ring*'s crypto/poly1305/*.

PR #1098, opened 2020-11-17
Chacha20: Move AAarch64 NEON dispatching logic to Rust.

briansmith:
Move away from using `GFp_armcap_P`, starting with the AAarch64 code. A similar change has already been made to the Poly1305 code for AAarch64.

After AAarch64 is done, the dispatching logic for other targets will likely be changed similarly.


briansmith:
Maybe this is a good idea but I'm going to put a hold on it for now. I think there is an alternate approach that might work better and I want to see if I can get some agreement with the BoringSSL team about how to pull the dispatching logic out of the assembly language before I continue down this path.

PR #1108, opened 2020-11-18
Revert "Remove definitions of deprecated `Error::description()` ..."

briansmith:
This reverts commit f06811a150fcded1555911678bbca6dcb5440cda because it changed the
pulic API in a non-backward-compatible way.


itmekriss (commented):
P

PR #1109, opened 2020-11-19
Fix linker errors when *ring* is incorporated into a (32-bit Android) ARM shared library.

briansmith:
commit 8a90093cccf60bd2aaf582fe7e56c0b0eff9b002 regressed the ability to incorporate *ring* into a 32-bit ARM shared library, on 32-bit Android at least.

The symbol needs to be declared `.protected`. Not all assemblers (notably, Apple's) understand `.protected`, so use `.hidden`, which implies `.protected`.

Additionally explicitly declare them as `.extern` even though this isn't required by the assembler.

### DIFF
--- a/crypto/chacha/asm/chacha-armv4.pl
+++ b/crypto/chacha/asm/chacha-armv4.pl
@@ -197,6 +197,8 @@ $code.=<<___;
 .Lone:
 .long	1,0,0,0
 #if __ARM_MAX_ARCH__>=7
+.extern GFp_armcap_P
+.hidden GFp_armcap_P
 .LOPENSSL_armcap:
 .word   GFp_armcap_P-.LChaCha20_ctr32
 #else
--- a/crypto/fipsmodule/bn/asm/armv4-mont.pl
+++ b/crypto/fipsmodule/bn/asm/armv4-mont.pl
@@ -112,6 +112,8 @@ $code=<<___;
 #endif
 
 #if __ARM_MAX_ARCH__>=7
+.extern GFp_armcap_P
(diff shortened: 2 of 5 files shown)


ghedo:
FWIW, I [updated quiche](https://github.com/cloudflare/quiche/tree/ring-arm32-build) to use ring from this branch, and the changes here do appear to fix the Android arm32 targets. Can't really say whether they cause other problems though, and we also don't run quiche's test suite on those platforms.

Before:
* https://github.com/cloudflare/quiche/runs/1442631264
* https://github.com/cloudflare/quiche/runs/1442631195

After:
* https://github.com/cloudflare/quiche/runs/1442710727
* https://github.com/cloudflare/quiche/runs/1442710620
* https://github.com/cloudflare/quiche/runs/1442710640

Other builds fail due to ring versions not matching, but that's a side-effect of pointing one crate to ring git, while not the others, so they can be ignored.

Somewhat curiously the Windows i686 build also failed https://github.com/cloudflare/quiche/runs/1442709312, but I can't really say whether that's because of these changes or not.

briansmith:
> Can't really say whether they cause other problems though, and we also don't run quiche's test suite on those platforms.

The main thing I need to check is whether there are now multiple `GFp_armcap_P`s being used, and whether the assembly code is using one that is always zero so that none of the CPU-feature-codepaths would be used. Once I've verified that, this should be good to go.

Frankly, I also just want to make sure I understand what the root cause of this problem is. I don't understand how the chances in the source code translated into the wrong kind of relocations being present in the library. I also simply need to learn what `.comm` does.

PR #1111, opened 2020-11-19
Pull the integrated assembly x86-64 ChaCha20-Poly1305  implementation from BoringSSL

vkrasnov:
It is significantly faster on Intel CPUs, especially for smaller buffers.

### DIFF
--- a/crypto/cipher_extra/asm/chacha20_poly1305_x86_64.pl
+++ b/crypto/cipher_extra/asm/chacha20_poly1305_x86_64.pl
@@ -1,0 +1,2473 @@
+#!/usr/bin/env perl
+
+# Copyright (c) 2015, CloudFlare Ltd.
+#
+# Permission to use, copy, modify, and/or distribute this software for any
+# purpose with or without fee is hereby granted, provided that the above
+# copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
+# SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+# OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+# CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE. */
+
+##############################################################################
+#                                                                            #
+# Author:  Vlad Krasnov                                                      #
+#                                                                            #
+##############################################################################
+
+$flavour = shift;
+$output  = shift;
+if ($flavour =~ /\./) { $output = $flavour; undef $flavour; }
+
+$win64=0; $win64=1 if ($flavour =~ /[nm]asm|mingw64/ || $output =~ /\.asm$/);
+
+$0 =~ m/(.*[\/\\])[^\/\\]+$/; $dir=$1;
+( $xlate="${dir}x86_64-xlate.pl" and -f $xlate ) or
+( $xlate="${dir}../../perlasm/x86_64-xlate.pl" and -f $xlate) or
+die "can't locate x86_64-xlate.pl";
+
+open OUT,"| \"$^X\" \"$xlate\" $flavour \"$output\"";
+*STDOUT=*OUT;
+
+$avx = 2;
+
+$code.=<<___;
+.text
+.extern GFp_ia32cap_P
+
+.align 64
+.Lchacha20_consts:
+.byte 'e','x','p','a','n','d',' ','3','2','-','b','y','t','e',' ','k'
+.byte 'e','x','p','a','n','d',' ','3','2','-','b','y','t','e',' ','k'
+.Lrol8:
+.byte 3,0,1,2, 7,4,5,6, 11,8,9,10, 15,12,13,14
+.byte 3,0,1,2, 7,4,5,6, 11,8,9,10, 15,12,13,14
+.Lrol16:
+.byte 2,3,0,1, 6,7,4,5, 10,11,8,9, 14,15,12,13
+.byte 2,3,0,1, 6,7,4,5, 10,11,8,9, 14,15,12,13
+.Lavx2_init:
+.long 0,0,0,0
+.Lsse_inc:
+.long 1,0,0,0
+.Lavx2_inc:
+.long 2,0,0,0,2,0,0,0
+.Lclamp:
+.quad 0x0FFFFFFC0FFFFFFF, 0x0FFFFFFC0FFFFFFC
+.quad 0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF
+.align 16
+.Land_masks:
+.byte 0xff,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00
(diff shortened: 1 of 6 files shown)


crypto/cipher_extra/asm/chacha20_poly1305_x86_64.pl
```diff
@@ -426,14 +424,16 @@ sub emit_body {
...
+# {u64, u64} chacha20_poly1305_open_asm(uint8_t *in_out, size_t out_offset, size_t len_in, uint8_t *ad, size_t len_ad, uint8_t keyp[48], uint64 avx2_and_bmi2_capable);
 $code.="
-.globl chacha20_poly1305_open
-.type chacha20_poly1305_open,\@function,2
+
+.globl chacha20_poly1305_open_asm
+.type chacha20_poly1305_open_asm,\@function,2
 .align 64
-chacha20_poly1305_open:
+chacha20_poly1305_open_asm:
```
briansmith, line 433: Every symbol must start with `GFp_` in *ring*

src/aead/chacha20_poly1305.rs
```diff
@@ -50,6 +50,34 @@ fn chacha20_poly1305_seal(
     in_out: &mut [u8],
     cpu_features: cpu::Features,
 ) -> Tag {
+    #[cfg(target_arch = "x86_64")]
+    if has_sse41(cpu_features) {
```
briansmith, line 54: Replace
```
#[cfg(...)] 
if ...
```
with
```
#[cfg(...)]
{
    if ...
}
```

And run rustfmt.

vkrasnov:
Some benchmarks:

  | 16" MacBook Pro 2.3 GHz Intel Core i9 |   |  
-- | -- | -- | --
  | 0.16.17 | This | Speedup
60B | 172.6 | 396.1 | 2.29
128B | 316.7 | 816.9 | 2.58
1280B | 1329.4 | 1874.2 | 1.41
8192B | 1809.5 | 2320.1 | 1.28
  |   |   |  
  | Linux AMD EPYC 7642 |   |  
  | 0.16.17 | This | Speedup
60B | 161.9 | 370.5 | 2.29
128B | 249.7 | 661.4 | 2.65
1280B | 918.7 | 1584.2 | 1.72
8192B | 1242.4 | 2046.9 | 1.65
  |   |   |  
  | Windows msvc i7-7700HQ |   |  
60B | 156.2 | 353.0 | 2.26
128B | 264.4 | 716.0 | 2.71
1280B | 680.4 | 1622.9 | 2.39
8192B | 783.1 | 2021.6 | 2.58
  |   |   |  
  | Linux E5-2630|   |  
60B | 136.1 | 319.1 | 2.34
128B | 215.4 | 568.6 | 2.64
1280B | 735.2 | 1175.7 | 1.60
8192B | 958.3 | 1573.8 | 1.64

src/aead/chacha20_poly1305.rs
```diff
@@ -51,31 +51,35 @@ fn chacha20_poly1305_seal(
...
+
+            extern "C" {
+                // This function stores the calculated Tag in keyp[8..12] and returns a reference to it
+                fn GFp_chacha20_poly1305_seal<'ctx>(
+                    in_out: *mut u8,
+                    len_in: usize,
+                    ad: *const u8,
+                    len_ad: usize,
+                    keyp: &'ctx mut [u32; 12],
+                ) -> &'ctx [u8; 16];
```
briansmith, line 66: Should we use `__m128i` as the return type? Then it would get returned in xmm0 and we could avoid copying it around.

vkrasnov: Returning `__m128i` from FFI is an experimental feature.

crypto/cipher_extra/asm/chacha20_poly1305_x86_64.pl
```diff
@@ -0,0 +1,2468 @@
...
+    movaps %xmm14,16*8+$xmm_store
+    movaps %xmm15,16*9+$xmm_store\n" if ($win64);
+$code.="
+    mov $adl, 0+$len_store
+    mov $inl, 8+$len_store\n";
+$code.="
+    mov GFp_ia32cap_P+8(%rip), %eax
+    and \$`(1<<5) + (1<<8)`, %eax # Check both BMI2 and AVX2 are present
+    xor \$`(1<<5) + (1<<8)`, %eax
+    jz chacha20_poly1305_open_avx2\n" if ($avx>1);
```
briansmith, line 483: In the calling Rust code, we already do feature detection to get this far. It seems like, especially since we're already changing the API of the assembly code, we should change the assembly API so that the Rust code calls `chacha20_poly1305_open_sse41_avx2` if BMI2 and AVX2 are present, and otherwise calls `GFp_chacha20_poly1305_open`, which should be renamed to `GFp_chacha20_poly1305_open_sse41`.

Also, how important it is to retain the non-AVX2+BMI2 SSE4.1 implementation in the first place? Why not just have the Rust code check for AVX2+BMI2+SSE4.1 and then call chacha20_poly1305_open_avx2, or else fall back to the previously-existing code? Are machines that have SSE4.1 but don't have AVX2 important for what you're doing?

vkrasnov: > we should change the assembly API so that the Rust code calls `chacha20_poly1305_open_sse41_avx2` if BMI2 and AVX2 are present

That is doable, but will result in some code duplication. Also the avx2 function returns to the use function to handle the remainder, which I would also have to duplicate. In addition checking `GFp_ia32cap_P` is performed in all of the current asm files: `aesni-x86_64.pl`, `x86_64-mont5.pl`, `chacha-x86_64.pl`, `p256-x86_64-asm.pl` and others.

>  how important it is to retain the non-AVX2+BMI2 SSE4.1 implementation in the first place

To my knowledge Intel still produces processors without AVX2 support. I would prefer to support low end devices as well. I don't have access to one right now to measure the performance delta, but I suspect it is even more significant than for AVX2 enabled devices.

briansmith: Thanks. I think it makes sense to leave it as-is for now.

vkrasnov:
The Windows fixes were now merged to upstream boringssl, and I update the PR to use the latest asm, so the changes are now minimal.

briansmith:
> The Windows fixes were now merged to upstream boringssl, and I update the PR to use the latest asm, so the changes are now minimal.

In the BoringSSL CL review, there were some changes that were reverted so that, IIUC, they could be split into separate CLs. are you going to submit more CLs related to this?

vkrasnov:
I am not sure what do you mean? The current PR represents the final asm from BoringSSL after the CL was merged. I can split into to commits if that is easier.

vkrasnov:
beep boop

briansmith:
> I am not sure what do you mean? The current PR represents the final asm from BoringSSL after the CL was merged. I can split into two commits if that is easier.

I created a branch  b/merge-boringssl-integrated-chacha-x86_64 to compare the BoringSSL code to the code in this PR and I noticed there were differences that weren't clear to me . At the minimum they should be in a separate commit. Please rebase this branch on top of that branch. (I already set the base branch of this PR to that branch.)

It is true that in *ring* we can currently use an `in_out` parameter since the current *ring* API only supports in-place (with offset) operation. However, in the near future I want to add out-of-place seal/open too. So, let's keep the separate ciphertext and plaintext parameters. Also make sure that the prototypes in the comments in the *.pl file are correct.

I see the BoringSSL's seal code uses a structure like this:
```
  struct {
    alignas(16) uint8_t key[32];
    uint32_t counter;
    uint8_t nonce[12];
    const uint8_t *extra_ciphertext;
    size_t extra_ciphertext_len;
  } in;
```
*ring* currently doesn't have the "extra_ciphertext" and it looks like you've simplified the code to account for the fact that *ring* doesn't have the "extra ciphertext". That seems fine to me but we need to add some comments about this in the comment atop the `_seal` function. OTOH, isn't the way BoringSSL does this actually pretty smart? I'm leaning towards just retaining this capability in the assembly code with the expectation that we'd expose such a capability in the *ring* API in the future.

There are other changes compared to the BoringSSL code that weren't immediately obvious. I think maybe this PR just got out of sync with what was actually merged into BoringSSL? Let's revert any non-essential differences.

vkrasnov:
I am opening a new PR, thanks.

vkrasnov:
Opened https://github.com/briansmith/ring/pull/1162

PR #1119, opened 2020-11-24
Fix BSD `--no-default-features` build.

briansmith:
`once_cell` is a required, not optional, dependency, on these platforms.


Thomasdezeeuw:
This build with `--no-default-features` on FreeBSD and tests pass without it. Thanks @briansmith

PR #1122, opened 2020-11-24
build.rs: Keep going when there is no relevant assembly.

briansmith:
Take a step towards supporting (non-wasm32) platforms without assembly code.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -230,20 +230,19 @@ const LD_FLAGS: &[&str] = &[];
 
 // None means "any OS" or "any target". The first match in sequence order is
 // taken.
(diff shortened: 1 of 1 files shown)


LinusU (approved):
Nice! 👍

briansmith:
I want to do more work before I merge this. Right now, every `target_arch` has a default PerlAsm flavor for operating systems that aren't known by build.rs. However, that's presumptuous; there's no guarantee that some arbitrary operating system is going to be ABI-compatible or source compatible with the assembly code. So, we should continue having an allowlist approach for using the assembly code, where we list explicitly list operating systems that we know are binary compatible. Then we can make this change on top of that change.

briansmith:
Closing this in favor of #1174.

PR #1125, opened 2020-11-26
Fix inline syntax highlighting in documentation.

frewsxcv:


briansmith:
Thanks for pointing this out. I fixed this in PR #1126.

PR #1133, opened 2020-11-30
CI/CD: Use self-contained linking for all -musl targets.

briansmith:


briansmith:
https://github.com/rust-lang/rust/issues/79555

PR #1147, opened 2020-12-05
AEAD: Remove all the `unsafe` from `Block` and clean up its API

briansmith:
Remove all (direct) uses of `unsafe` from `Block` by re-implementing it in terms of `ring::endian`.

Simplify its API from the caller's perspective.


briansmith:
I will post a replacement for this PR after PR ##1189 is merged.

PR #1148, opened 2020-12-05
Remove now-superfluous uses of `ring::aead::block`

briansmith:
After recent refactorings, especailly the switch to BoringSSL's Poly1305, the use of `ring::aead::block` outside of AES and GCM doesn't make so much sense. Clean up those uses.


briansmith:
Replaced by PR #1189.

PR #1152, opened 2020-12-16
Make aead::Nonce implement RandomlyConstructable

ab-cgh:
This adds the ability to use `ring::rand::random<T>` to generate AEAD nonces with the function by implementing `RandomlyConstructable` for it.

### DIFF
--- a/src/aead/nonce.rs
+++ b/src/aead/nonce.rs
@@ -40,6 +40,10 @@ impl Nonce {
     pub fn assume_unique_for_key(value: [u8; NONCE_LEN]) -> Self {
         Self(value)
     }
+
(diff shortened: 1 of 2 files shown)


briansmith:
Thanks for the PR. It is intentional that `Nonce` does not implement `RandomlyConstructable`. Although some things do use AES-GCM with 96-bit random nonces, I didn't (and don't) want to make it trivial to do that.

My plan so far has been to implement an API similar to `ring::aead` that supports misuse-resistant AEADs and for *that* API to support random nonces.

WDYT?

ab-cgh:
I think that sounds fine. I wasn't aware that the lack of easily random generation was intended, but figured it was a small oversight instead. 

I'll close this PR as "working as intended" and use the workaround that I'm currently using for now.

Thanks for your time!

PR #1154, opened 2020-12-18
Merge BoringSSL through 41a14304d7e0cfcb8afa82ee8735f0bd6763e415

briansmith:


briansmith:
This MR will have to be undone when we add big-endian support for each architecture. That will require larger changes.

PR #1157, opened 2020-12-20
Use non-zero value when initializing test buffers

caio:
Hi!

Picking up from PRs #564 and #600:

* The constant is in the test module now as `UNINITIALIZED_U8` with
  value `0xDE`
* Left the lone zeroed usize buffer untouched

I tried to find every candidate in `tests/` and found a few tests from
inside `src/` too.

And I've opted to leave the buffers from `max_input_tests!` in
`src/digest.rs` untouched as they are about the input length, but let me
know if you prefer otherwise.

Aside: the README recommends running `cargo test --no-default-features`
prior to a PR but it fails to compile for me (Windows, WSL2 with Ubuntu
20.04) because the `test_file!` macro is gated by the `alloc` feature.
Running `cargo test --no-default-features --features alloc` works fine.


caio:
Force-pushed an update to the branch due to a failed `rustfmt` check. That one should pass now :)

Issue #859, opened 2019-06-20
impl Clone and PartialEq for keypair types


jimmycuadra:
`ring::signature::Ed25519KeyPair` is used in [ruma-signatures](https://github.com/ruma/ruma-signatures) in a type that I'd like to implement `Clone` and `PartialEq`. I'm wondering if there's a specific reason `Ed25519KeyPair` doesn't implement these traits, or if it was just never needed before.

I can work around this by keeping the public and private keys and just creating a new `Ed25519KeyPair` in a manual `Clone` impl, and comparing the public and private keys in a manual implementation of `PartialEq`, but I'm not sure if 1) there is some security implication to this 2) there is any sort of random state held in `Ed25519KeyPair` that would make two values not equal even if they were generated from the same key pair.

If it's in fact okay to implement these traits in *ring*, I'd be happy to submit a PR.

est31:
:+1: this would be great for rcgen as well: https://github.com/est31/rcgen/issues/17

Optimally, all KeyPair structs would get Clone and PartialEq (maybe Eq even)

Ralith:
Possibly related prior discussion: https://github.com/briansmith/ring/pull/727

briansmith:
I'm still on the fence regarding `Clone`, for the reasons I stated in #727. I want to support keys that are stored in places that don't allow cloning. Also, going forward the keypairs might store state (especially `EcdsaKeyPair`s) that shouldn't be cloned for semantic reasons related to randomization. Most users can use `Arc<XxxKeyPair>` to get a good enough approximation to a clonable `XxxKeyPair`.

Regarding PartialEq, maybe it is fine to implement `PartialEq` for public keys and then implement `PartialEq` for keypairs by saying that two keypairs are partially equal if the public keys are partially equal. Presumably this would allow people to look up a key pair by the public key in a map, so that might be useful.

I think that we're better off waiting to do this until after the big refactoring of the KeyPair types is merged; I've started working on it for RSA and I hope to have it merged soon.

PR #1158, opened 2020-12-22
derive Copy and Clone for UnboundKey

oconnor663:
OpeningKey and SealingKey intentionally avoid implementing Copy and
Clone, because they're attached to a nonce sequence that should be
unique. UnboundKey isn't attached to a nonce sequence, though, and in
the case of AES it also takes some nontrivial key setup work to
construct from raw bytes. Making it Copy and Clone allows lets callers
avoid repeating that work in cases where a key is reused without a fixed
nonce sequence.

---

Copying keys around is a risky business, which is why this hierarchy of types exists in the first place, and I wouldn't be surprised if there were some objections here. Maybe some folks would prefer to implement `Clone` but not `Copy`, for example. I figured it would be easier to start that conversation with some code than to just open an issue, so here's a first draft PR :)

### DIFF
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -384,6 +384,7 @@ impl Aad<[u8; 0]> {
 }
 
 /// An AEAD key without a designated role or nonce sequence.
+#[derive(Copy, Clone)]
 pub struct UnboundKey {
     inner: KeyInner,
     algorithm: &'static Algorithm,
@@ -399,6 +400,7 @@ impl core::fmt::Debug for UnboundKey {
 }
 
 #[allow(clippy::large_enum_variant, variant_size_differences)]
+#[derive(Copy, Clone)]
 enum KeyInner {
     AesGcm(aes_gcm::Key),
     ChaCha20Poly1305(chacha20_poly1305::Key),
--- a/src/aead/aes.rs
+++ b/src/aead/aes.rs
@@ -15,6 +15,7 @@
 use super::{counter, iv::Iv, quic::Sample, Block, Direction, BLOCK_LEN};
 use crate::{bits::BitLength, c, cpu, endian::*, error, polyfill};
 
+#[derive(Copy, Clone)]
 pub(crate) struct Key {
     inner: AES_KEY,
     cpu_features: cpu::Features,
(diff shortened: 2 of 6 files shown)


oconnor663:
Looks like something like this has been discussed before in https://github.com/briansmith/ring/pull/727 and https://github.com/briansmith/ring/issues/859, but some of that discussion was before `LessSafeKey` was added to the API, and I'm not sure whether circumstances might've changed.

One thing @briansmith mentioned is that *ring* wants to abstract over keys contained in keyrings and things. In that world, presumably we'd still like to give callers the option to avoid repeating the work of AES key setup? Where in the type hierarchy would it make sense to this sort of thing?

briansmith:
I think it might make sense for `LessSafeKey` to implement `Clone`. I don't see a need to implement `Copy`. Like you said, it seems wrong for `UnboundKey` to be either `Clone` or `Copy` since the `UnboundKey` may be converted into `OpeningKey` or `SealingKey`, which make assumptions about the uniqueness of the key w.r.t. the nonce sequence.

briansmith:
Just to clarify further, `UnboundKey` exists only for the purpose of eventually converting it to an `OpeningKey` or `SealingKey` or `LessSafeKey`. In the event that it would be converted into an `OpeningKey` or `SealingKey`, it would be wrong to clone it, because then the nonce protection mechanism wouldn't work right. So, I think the only thing that would make sense would be to change `LessSafeKey` to implement `Clone`. However, I think applications can just use `Arc<LessSafeKey>` instead, and that would be better for almost all applications.

If somebody thinks that we need to support cloning `OpeningKey`/`SealingKey` (or `UnboundKey`) then I suggest we discuss that in an issue that clearly explains why it is important to support, and more important than our attempts to prevent nonce reuse.

I'm going to close this. If I've overlooked something, please let me know. Thanks!

oconnor663:
> However, I think applications can just use `Arc<LessSafeKey>` instead, and that would be better for almost all applications.

Is it a downside to require heap allocation in this case? Might not be ideal for libraries planning to support `no_std`. (Though I'll admit I don't personally plan to support `no_std` in the use case that led me to file this bug.)

briansmith:
> Is it a downside to require heap allocation in this case? Might not be ideal for libraries planning to support `no_std`. (Though I'll admit I don't personally plan to support `no_std` in the use case that led me to file this bug.)

Given that `Arc<LessSafeKey>` works and would be effectively equivalent to making `LessSafeKey` clonable, I I think it would be fine to implement `Clone` for `LessSafeKey`, especially since semantically it's no different than `Arc<LessSafeKey>` AFAICT.

oconnor663:
Cool, I just opened a separate PR for that at https://github.com/briansmith/ring/pull/1175.

PR #1162, opened 2020-12-23
Enable the integrated assembly x86-64 ChaCha20-Poly1305 implementation from BoringSSL

vkrasnov:
Which is significantly faster.

This new PR keeps the original BoringSSL asm interface.

### DIFF
--- a/src/aead/chacha20_poly1305.rs
+++ b/src/aead/chacha20_poly1305.rs
@@ -15,7 +15,7 @@
 use super::{
     chacha::{self, Counter},
     iv::Iv,
-    poly1305, Aad, Block, Direction, Nonce, Tag, BLOCK_LEN,
+    poly1305, Aad, Block, Direction, Nonce, Tag, BLOCK_LEN, TAG_LEN,
 };
 use crate::{aead, cpu, endian::*, error, polyfill};
 use core::convert::TryInto;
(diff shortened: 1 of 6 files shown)


briansmith:
@vkrasnov I have reviewed this and I have some changes to it that I've already got in a local branch. Those changes require PR #1177. Once I merge that PR I'll post my change to your PR as a PR based on your branch for you to look at.

src/aead/chacha20_poly1305.rs
```diff
@@ -61,6 +121,49 @@ fn chacha20_poly1305_open(
     in_out: &mut [u8],
     cpu_features: cpu::Features,
 ) -> Tag {
+    #[cfg(target_arch = "x86_64")]
+    {
+        if has_sse41(cpu_features) {
+            #[repr(C)]
+            #[repr(align(16))]
```
briansmith, line 128: One thing that's unclear is whether the assembly code requires the tag buffer to be 16-byte aligned or whether it really only requires the key block to be 16-byte aligned. If the tag buffer isn't 16-byte aligned, would there be any ill consequences (performance or otherwise)?

vkrasnov: I think the requirement can easily be avoided using unaligned instructions?

Other than that for repr(C) if struct is aligned then everything is aligned.

briansmith: > I think the requirement can easily be avoided using unaligned instructions?
> 
> Other than that for repr(C) if struct is aligned then everything is aligned.

Thanks! I've updated the other PR to force the alignment so we don't have to think about this; see https://github.com/briansmith/ring/pull/1179/commits/af3fa46f2906f7961c96092a32c0ab529aa5d8c7.

vkrasnov:
🤗

PR #1164, opened 2020-12-27
Use -target when building for Mac Catalyst

visigoth:
When building for Mac Catalyst, the output object files must have the correct
`LC_BUILD_VERSION`. This change ensures that the `-target` option is used with
the compiler.

The documentation for `Build::target` suggests that "this option is
automatically scraped from the `TARGET` environment variable by build scripts,"
but it appears that build scripts are required to be sensitive to it.

Note that `Build::target` seems to incorrectly format `"-target
<target>"` as a single argument to the compiler, which it does not recognize.
Therefore, the option is built as separate flags.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -379,7 +379,7 @@ fn build_c_code(target: &Target, pregenerated: PathBuf, out_dir: &Path) {
                 perlasm_format,
                 Some(includes_modified),
             );
-       }
+        }
 
         let mut asm_srcs = asm_srcs(perlasm_src_dsts);
 
@@ -415,7 +415,7 @@ fn build_c_code(target: &Target, pregenerated: PathBuf, out_dir: &Path) {
     // XXX: Ideally, ring-test would only be built for `cargo test`, but Cargo
     // can't do that yet.
     libs.iter().for_each(|&(lib_name, srcs, additional_srcs)| {
-       build_library(
+        build_library(
             &target,
             &out_dir,
             lib_name,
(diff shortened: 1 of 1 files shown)


briansmith:
Are you saying that cc-rs PR https://github.com/alexcrichton/cc-rs/pull/556 does not do the right thing? When I look at the cc-rs PR it seems like a pretty complete solution. Maybe we just need to bump our cc-rs dependency in Cargo.toml instead?

briansmith:
Also, if cc-rs isn't doing the right thing even after that PR, then isn't a better solution to fix cc-rs, and avoid needing to work around the problem in *ring*'s build.rs?

briansmith:
The most important improvement here is adding testing that this works in CI/CD, by modifying .github/workflows/ci.yml to add a build configuration for the Catalyst target.

briansmith:
@visigoth Do you have any thoughts regarding the questions I asked above? Should we replace this PR with a new PR that bumps the minimum cc-rs version instead? Am I overlooking something?

briansmith:
Unless/until I hear otherwise, I'm going to assume that PR #1186 fixes this.

visigoth:
@briansmith apologies for not responding; i had to step away from stuff for a while. indeed, i believe that PR fixed the issue. i'm now able to `cargo build --target x86_64-apple-ios-macabi` against tip of main.

briansmith:
@visigoth Thanks for confirming that!

PR #1173, opened 2021-01-21
Show error if building on unsupported platform

BenediktSeidl:
When building on a unsupported platform a cryptic error message is shown:

```
   Compiling ring v0.16.19
error: failed to run custom build command for `ring v0.16.19`

Caused by:
  process didn't exit successfully: `/home/bs/Code/ev/rust_test/hello_world/target/release/build/ring-a6c015645fcb51f3/build-script-build` (exit code: 101)
  --- stderr
  thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/bs/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.16.19/build.rs:375:10
```

Now a human readable error message is shown:

```
error: failed to run custom build command for `ring v0.16.19 (/home/bs/Code/p/ring)`

Caused by:
  process didn't exit successfully: `/home/bs/Code/p/ring/target/release/build/ring-ebe46f27c25935e4/build-script-build` (exit code: 101)                                                                                                               
  --- stderr
  thread 'main' panicked at 'Operating system 'linux' and/or architecture 'mips' is not supported', build.rs:357:10
/tmp/d/openwrt-sdk-19.07.6-ath79-generic_gcc-7.5.0_musl.Linux-x86_64/staging_dir/toolchain-mips_24kc_gcc-7.5.0_musl/bin/mips-openwrt-linux-strip: 'target/mips-unknown-linux-musl/release/hello_world': No such file                                    
```

I'm not sure about the wording though,...

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -354,7 +354,12 @@ fn build_c_code(target: &Target, pregenerated: PathBuf, out_dir: &Path) {
             let &(entry_arch, entry_os, _) = *entry;
             entry_arch == target.arch && is_none_or_equals(entry_os, &target.os)
         })
-        .unwrap();
+        .unwrap_or_else(|| {
+            panic!(
+                "Operating system '{}' and/or architecture '{}' is not supported",
+                target.os, target.arch
+            )
+        });
 
     let use_pregenerated = !target.is_git;
     let warnings_are_errors = target.is_git;


briansmith:
Thanks for the PR. build.rs isn't the right place for this logic, as we're actually actively working on allowing build.rs to keep going even when there is no assembly code for the target, so that we can use fallback Rust (or C) implementations for any targets we don't have optimized assembly for. Also, our goal is to support nearly all targets, and I'd rather focus on improving the breadth of targets we support rather than making it clearer which targets we currently, temporarily, don't support. So I'm going to close this.

briansmith:
See PR #1174 to see where we're going w.r.t. changing build.rs for currently-unsupported targets.

BenediktSeidl:
thanks for your reply, looking forward to it!

PR #1174, opened 2021-01-22
Only use assembly code on known-compatible OS+Arch combinations.

briansmith:
Assume by default that an operating system does not have an ABI compatible
with the PerlAsm sources. Add all the operating systems that we've
explicitly added support for to the allowlist. Avoid trying to build or
use the PerlAsm code for those targets.

On top of this, we can build fallback logic for using Rust (or C)
implementations for those targets that aren't compatible with the
assembly.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -216,23 +216,106 @@ const LD_FLAGS: &[&str] = &[];
 
 // None means "any OS" or "any target". The first match in sequence order is
 // taken.
-const ASM_TARGETS: &[(&str, Option<&str>, Option<&str>)] = &[
-    ("x86_64", Some("ios"), Some("macosx")),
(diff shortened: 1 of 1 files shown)


briansmith:
@LinusU PTAL at this. I think with this and a couple other PRs that are currently posted, included yours, we are very close to having MIPS working.

briansmith:
@LinusU @bltavares Could you please review this and leave comments about what needs to be improved? Thanks!

bltavares:
I'm not much of an specialist on Assembly or C - but this change was useful to unblock efforts on MIPS and LGTM

LinusU (dismissed):
I'm not too familiar with the build system, but this looks great 👍

PR #1175, opened 2021-01-22
implement Clone for LessSafeKey

oconnor663:
OpeningKey and SealingKey intentionally avoid implementing Copy and
Clone, because they're attached to a fixed nonce sequence that should be
unique. LessSafeKey isn't attached to a nonce sequence, though, and
making it Clone lets callers avoid repeating key setup work.

Coming from https://github.com/briansmith/ring/pull/1158#issuecomment-765050031.

### DIFF
--- a/tests/aead_tests.rs
+++ b/tests/aead_tests.rs
@@ -513,6 +513,52 @@ fn test_aead_key_debug() {
     );
 }
 
+fn test_aead_lesssafekey_clone_for_algorithm(algorithm: &'static aead::Algorithm) {
+    let test_bytes: Vec<u8> = (0..32).collect();
+    let key_bytes = &test_bytes[..algorithm.key_len()];
+    let nonce_bytes = &test_bytes[..algorithm.nonce_len()];
+
+    let key1: aead::LessSafeKey =
+        aead::LessSafeKey::new(aead::UnboundKey::new(algorithm, &key_bytes).unwrap());
+    let key2 = key1.clone();
+
+    // LessSafeKey doesn't support AsRef or PartialEq, so instead just check that both keys produce
+    // the same encrypted output.
+    let mut buf1: Vec<u8> = (0..100).collect();
+    let mut buf2 = buf1.clone();
+    let tag1 = key1
+        .seal_in_place_separate_tag(
+            aead::Nonce::try_assume_unique_for_key(&nonce_bytes).unwrap(),
+            aead::Aad::empty(),
+            &mut buf1,
+        )
+        .unwrap();
+    let tag2 = key2
+        .seal_in_place_separate_tag(
+            aead::Nonce::try_assume_unique_for_key(&nonce_bytes).unwrap(),
+            aead::Aad::empty(),
+            &mut buf2,
+        )
+        .unwrap();
+    assert_eq!(tag1.as_ref(), tag2.as_ref());
+    assert_eq!(buf1, buf2);
+}
+
+#[test]
+fn test_aead_lesssafekey_clone_aes_128_gcm() {
+    test_aead_lesssafekey_clone_for_algorithm(&aead::AES_128_GCM);
+}
+
+#[test]
+fn test_aead_lesssafekey_clone_aes_256_gcm() {
+    test_aead_lesssafekey_clone_for_algorithm(&aead::AES_256_GCM);
+}
+
+#[test]
+fn test_aead_lesssafekey_clone_chacha20_poly1305() {
+    test_aead_lesssafekey_clone_for_algorithm(&aead::CHACHA20_POLY1305);
+}
+
 fn make_key<K: aead::BoundKey<OneNonceSequence>>(
     algorithm: &'static aead::Algorithm,
     key: &[u8],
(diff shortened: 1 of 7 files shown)


briansmith:
Please review PR #1185 and see if rebasing this on top of that PR will simplify this. AFAICT it will.

oconnor663:
Yes it looks like the manual impl isn't needed on that branch. Everything is just `#[derive(Clone)]`: https://github.com/oconnor663/ring/commit/3275a8df3c1abb31ae52d96eeb7719ef95bdad80

briansmith:
> Yes it looks like the manual impl isn't needed on that branch. Everything is just `#[derive(Clone)]`: [oconnor663@3275a8d](https://github.com/oconnor663/ring/commit/3275a8df3c1abb31ae52d96eeb7719ef95bdad80)

Thanks! I have now merged that refactoring. If you rebase this on top of main, and add a test that encrypting some arbitrary data with the original key results in the same output as encrypting the same arbitrary data with a clone of the key (with the same nonce), then I'll prioritize merging this and ensure it is included in the next release.

briansmith:
@oconnor663 Are you still interested in this? I'm preparing a new release that could include this if you rebase it onto main.

oconnor663:
Done! Thanks for the reminder.

tests/aead_tests.rs
```diff
@@ -50,7 +50,7 @@ macro_rules! test_aead {
             $(
                 #[allow(non_snake_case)]
                 mod $alg { // Provide a separate namespace for each algorithm's test.
-                    use super::super::{*, test};
```
briansmith, line 53: I think you'll have to undo this change? I put this ugliness there because the tests wouldn't pass correctly in wasm32 targets without it.

tests/aead_tests.rs
```diff
@@ -509,6 +509,38 @@ fn test_aead_key_debug() {
     );
 }
 
+#[test]
+#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
```
briansmith, line 513: The wasm_bindgen_test thing here shouldn't be necessary anymore because of what was done in PR #1252.

tests/aead_tests.rs
```diff
@@ -509,6 +509,38 @@ fn test_aead_key_debug() {
...
+fn test_aead_lesssafekey_clone() {
```
briansmith, line 514: Please parameterize this on the `&'static aead::Algorithm` and then test it with all three AES-128-GCM, AES-256-GCM, and ChaCha20-Poly1305. Especially the AES and ChaCha20-Poly1305 code is completely different, it's good to cover all.

briansmith (changes requested):

tests/aead_tests.rs
oconnor663: Ok, I'll split it out into its own line with its own `#[allow(unused_imports)]`. But if you'd prefer to just leave the warning in place let me know.

oconnor663:
Addressed the comments above. Let me know what you think about the `allow(unused_imports)` workaround in particular. Also if you want me to combine the three separate tests into one test with a loop.

tests/aead_tests.rs
briansmith: I understand now that you're just trying to fix a warning. It turns out that our warning detection in CI was broken. I fixed it in PR #1260. So, you should be able to rebase this on top of the current `main` and you shouldn't need to make changes here any more.

tests/aead_tests.rs
```diff
@@ -509,6 +512,52 @@ fn test_aead_key_debug() {
...
+    let tag2 = key2
+        .seal_in_place_separate_tag(
+            aead::Nonce::try_assume_unique_for_key(&nonce_bytes).unwrap(),
+            aead::Aad::empty(),
+            &mut buf2,
+        )
+        .unwrap();
+    assert_eq!(tag1.as_ref(), tag2.as_ref());
+    assert_eq!(buf1, buf2);
+}
```
briansmith, line 544: I wonder why codecov is saying this line isn't hit?

briansmith:
@oconnor663 If you can rebase this and address the comment above, then I think we're good to go and I'll be able to include this in the next release.

oconnor663:
I've rebased the branch and removed the extra commit that was just fixing a warning. Not sure about the codecov message. If I put a `panic!()` at the end of that function, `cargo test` fails as expected, so those lines are certainly getting run. I haven't yet figured out how to run codecov locally, but if you think that's the best next step let me know.

It looks like some repo settings have changed, and the CI run for the rebased branch won't start until I get some new permissions. Let me know if you need anything from me for that.

tests/aead_tests.rs
briansmith: Let's now worry about the codecov thing.

briansmith:
> It looks like some repo settings have changed, and the CI run for the rebased branch won't start until I get some new permissions. Let me know if you need anything from me for that.

I started the CI run. Recently GitHub started requiring manual approval for the first PR from a new contributor to a repo, to prevent cryptocurrency mining and other abuse of GitHub Actions.

oconnor663:
Oof. As they say about nice things, and why we cannot have them...

briansmith (approved):

briansmith:
Thanks! Merged.

oconnor663:
Great! Thanks for all the feedback.

PR #1177, opened 2021-01-25
Endian: Allow `ArrayEncoding` to be implemented for larger arrays.

briansmith:
The original version avoided `unsafe` but doesn't work for older
versions of Rust that we're still trying to support. Switch to the
`unsafe` implementation that works everywhere.


briansmith:
@vkrasnov I need this change for some of the changes I'm planning to make to your ChaCha20-Poly1305 PR. PTAL.

PR #1185, opened 2021-02-01
Simplify AEAD implementation by implementing all key types in terms of `LessSafeKey`.

briansmith:
See the individual commit messages for details.


briansmith:
Some of the code test coverage issues pointed out by codecov are trivial but it does seem like we should improve the test coverage of the open/seal functions that aren't covered before we merge this.

PR #1195, opened 2021-02-09
add support for Hermit 

quangIO:
To [build it for RustyHermit](/hermitcore/rusty-hermit), .cargo/config.toml should contain something like
```toml
[unstable]
build-std = ["std", "core", "alloc", "panic_abort"]
build-std-features = ["compiler-builtins-mem"]

[build]
target = "x86_64-unknown-hermit"
```
or build with
```bash
cargo build -Z build-std=std,core,alloc,panic_abort --target x86_64-unknown-hermit
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -319,6 +319,10 @@ web-sys = { version = "0.3.37", default-features = false, features = ["Crypto", 
 [target.'cfg(target_os = "windows")'.dependencies]
 winapi = { version = "0.3.8", default-features = false, features = ["ntsecapi", "wtypesbase"] }
 
+[target.'cfg(target_os = "hermit")'.dependencies]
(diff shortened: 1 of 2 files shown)


briansmith:
Thanks for the PR. I'm planning to switch the implementation of `ring::rand` to use the getrandom crate as part of the goal of encapsulating/minimizing operating-system-specific code; see draft PR https://github.com/briansmith/ring/pull/1531. I'm closing this in favor of that approach.

I seem to remember when reading the `getrandom` crate's source code that its code for Hermit needs to be improved. If so, after those improvements are made and #1531 is merged, we should be able to get `Hermit` working in *ring* with a single line `$[cfg()]` logic in a new PR.

Thanks again!

PR #1197, opened 2021-02-09
Add a Rust ChaCha20 implementation.

briansmith:
See the comments in fallback.rs for more details on the origin.


briansmith:
I updated the PR to avoid the removal of the x86 assembly implementation. That will be deferred to another day.

briansmith:
`chacha20_test_fallback` is very slow in debug builds but OK in release builds. I guess I need to look at the compiler output to see if there's something I can do to speed it up without making things too ugly.

PR #1202, opened 2021-02-10
AES-GCM: Clarify `ghash()` functions operate over complete blocks.

briansmith:


briansmith:
Looking at the code test coverage report, the coverage failure is a false positive; this code is tested by the `wasm32_c` configurations in CI, which I verified in the log. Unfortunately we don't have coverage jobs for that target yet.

briansmith:
The one Apple failure seems to be due to Mac builders being flaky in GitHub Actions, as evidenced by the fact that there is no long and all the other build/test jobs succeeded.

PR #1210, opened 2021-02-25
AEAD: Make the internal opening API more like `copy_within`

briansmith:
See the individual commits for details.

### DIFF
--- a/src/aead/aes.rs
+++ b/src/aead/aes.rs
@@ -16,7 +16,6 @@ use super::{
     block::{Block, BLOCK_LEN},
     nonce::Nonce,
     quic::Sample,
-    Direction,
 };
 use crate::{
     bits::BitLength,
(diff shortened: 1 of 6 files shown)


src/aead/aes.rs
```diff
@@ -263,29 +258,29 @@ impl Key {
                     }
                     ctr32_encrypt_blocks!(
                         GFp_bsaes_ctr32_encrypt_blocks,
-                        &mut in_out[..(bsaes_in_out_len + in_prefix_len)],
-                        in_prefix_len,
+                        &mut in_out[src.clone()][bsaes_in_out_len..],
+                        src.clone(),
                         &bsaes_key,
                         ctr
```
goffrie, line 264: I may be mistaken but this looks incorrect - it will encrypt a suffix of the input rather than a prefix?

PR #1212, opened 2021-02-26
CI: Always use `macos-latest` and Xcode 12.4.

briansmith:
macos-11.0 is pre-production. Use macos-latest instead, to (I hope)
make macOS jobs more likely to succeed. (Presently, they frequently
fail with no logs.)

Upgrade to Xcode 12.4 in hopes that this will make things work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,8 +203,7 @@ jobs:
 
         include:
           - target: aarch64-apple-darwin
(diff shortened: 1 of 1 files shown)


briansmith:
This PR is the first one in a long time that ran during the daytime (or any time of day recently) where all the Apple jobs succeeded. So I'm hoping this will make CI much more reliably be all green.

Issue #1216, opened 2021-03-07
Ability to construct hmac::Algorithm


littledivy:
Hi :wave: 
I'm working on [Deno's WebCrypto implementation](https://github.com/denoland/deno/pull/9614). [Webcrypto allows](https://developer.mozilla.org/en-US/docs/Web/API/HmacKeyGenParams#properties) modifying length of the key, which is not possible since `hmac::Algorithm` digest field is private.

https://github.com/briansmith/ring/blob/1c36465f817f2e6697be917450cf1aded6d447ce/src/hmac.rs#L116

 Would you be interested in making this public? or are there other ways to do this? :slightly_smiling_face:  Thanks!

briansmith:
> hash A DOMString representing the name of the digest function to use. You can pass any of SHA-1, SHA-256, SHA-384, or SHA-512 here.

Your project probably has a function that maps from strings to `&;static digest::Algorithm`. Instead, change that map to be from strings to `hmac::Algorithm` or `hkdf::Algorithm`. You can get the underlying `hmac::Algorithm` from the `hkdf::Algorithm` and you can get the underlying `digest::Algorithm` from an `hmac::Algorithm`. We do this in Rustls and it works great. Check out https://github.com/ctz/rustls/pull/457.

Alternatively, less ideally, you can easily write a function that returns an HMAC algorithm from a digest algorithm. Rustls use to do this before I refactored the code. Again, check out the above PR, this time looking at the old code, to see how to do it.

I intentionally make the `hmac::Algorithm` and `hkdf::Algorithm` types opaque because in the near future (or maybe even presently; I forget) there will be optimizations that rely on each `hmac::Algorithm` knowing more details about the digest algorithm than are available from a `digest::Algorithm`. So I don't intend to make an API to construct `hmac::Algorithm` from a `digest::Algorithm`.

littledivy:
Yes, I directly map to `hmac::Algorithm` ([code](https://github.com/denoland/deno/blob/cd821f39b06bb5c3805496a147aa32060a9a7900/op_crates/crypto/key.rs#L67-L76)).

Here's a small example of my use case:
```rust
// Behaves as per WebCrypto HmacKeyGenParams. 
// https://developer.mozilla.org/en-US/docs/Web/API/HmacKeyGenParams#properties
fn generate_hmac(hash: String, length: Option<usize>) {
  let hmac_algorithm: ring::hmac::Algorithm = match hash { ... };
  // Ability to modify length (?)
[...]

PR #1218, opened 2021-03-10
feat(hmac): Add `generate_with_length` method to hmac::Key

littledivy:
#1216

### DIFF
--- a/src/hmac.rs
+++ b/src/hmac.rs
@@ -175,15 +175,31 @@ impl Key {
         algorithm: Algorithm,
         rng: &dyn rand::SecureRandom,
     ) -> Result<Self, error::Unspecified> {
-        Self::construct(algorithm, |buf| rng.fill(buf))
+        Self::construct(algorithm, algorithm.0.output_len, |buf| rng.fill(buf))
     }
 
-    fn construct<F>(algorithm: Algorithm, fill: F) -> Result<Self, error::Unspecified>
+    /// Generate an HMAC singing key using a custom length for the given
+    /// digest algorithm.
+    ///
+    /// The key will be `length` bytes long.
+    pub fn generate_with_length(
+        algorithm: Algorithm,
+        length: usize,
+        rng: &dyn rand::SecureRandom,
+    ) -> Result<Self, error::Unspecified> {
+        Self::construct(algorithm, length, |buf| rng.fill(buf))
+    }
+
+    fn construct<F>(
+        algorithm: Algorithm,
+        length: usize,
+        fill: F,
+    ) -> Result<Self, error::Unspecified>
     where
         F: FnOnce(&mut [u8]) -> Result<(), error::Unspecified>,
     {
         let mut key_bytes = [0; digest::MAX_OUTPUT_LEN];
-        let key_bytes = &mut key_bytes[..algorithm.0.output_len];
+        let key_bytes = &mut key_bytes[..length];
         fill(key_bytes)?;
         Ok(Self::new(algorithm, key_bytes))
     }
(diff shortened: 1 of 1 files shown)


src/hmac.rs
```diff
@@ -175,15 +175,31 @@ impl Key {
...
     }
 
-    fn construct<F>(algorithm: Algorithm, fill: F) -> Result<Self, error::Unspecified>
+    /// Generate an HMAC singing key using a custom length for the given
+    /// digest algorithm.
+    ///
+    /// The key will be `length` bytes long.
+    pub fn generate_with_length(
+        algorithm: Algorithm,
+        length: usize,
```
briansmith, line 187: Should this be `NonZeroUsize` at least? I don't think it makes sense to support zero-length keys.

src/hmac.rs
```diff
@@ -175,15 +175,31 @@ impl Key {
...
+    fn construct<F>(
+        algorithm: Algorithm,
+        length: usize,
+        fill: F,
+    ) -> Result<Self, error::Unspecified>
     where
         F: FnOnce(&mut [u8]) -> Result<(), error::Unspecified>,
     {
         let mut key_bytes = [0; digest::MAX_OUTPUT_LEN];
-        let key_bytes = &mut key_bytes[..algorithm.0.output_len];
```
briansmith, line 186: This doesn't work when the requested length is larger than `digest::MAX_OUTPUT_LEN`.

Please write a parameterized test that accepts an `hmac::Algorithm` and a length, and then tests your new function with that algorithm and length, for the following lengths:
1, 2, output_len - 1, output_len, output_len + 1, block_len - 1, block_len, block_len + 1, (block_len * 2) - 1, block_len * 2, (block_len * 2) + 1.

briansmith (changes requested):

src/hmac.rs
littledivy: Done

src/hmac.rs
littledivy: Updated logic and added tests.

src/hmac.rs
```diff
@@ -364,4 +396,40 @@ mod tests {
...
+                (block_len * 2) - 1,
+                block_len * 2,
+                (block_len * 2) + 1,
+            ] {
+                let _key = hmac::Key::generate_with_length(
+                    *algorithm,
+                    NonZeroUsize::new(*length).unwrap(),
+                    &rng,
+                )
+                .unwrap();
```
littledivy, line 431: Not sure how to assert the key length with the expected one. :thinking:

src/hmac.rs
briansmith: You can't add a call to `to_vec` to `construct` because `construct` needs to be able to work even in environments that doesn't have a memory allocator. Instead you need to move the allocation to your new constructor and mark that constructor `#[cfg(feature = "alloc")]`.

src/hmac.rs
```diff
@@ -364,4 +396,40 @@ mod tests {
             assert!(hmac::verify(&key, HELLO_WORLD_BAD, tag.as_ref()).is_err())
         }
     }
+
+    #[test]
+    pub fn hmac_generate_with_length() {
```
briansmith, line 401: Tests of the public HMAC API should go in `hmac_tests.rs`.

I will write a test for the original `generate()` that you can emulate.

src/hmac.rs
littledivy: Ah nice catch! Moved new `construct` logic to `generate_with_length` behind the alloc feature :+1:

briansmith (changes requested):
Are you still interested in this? Sorry for the long delay.

briansmith:
Closing this because the effect was already achievable (generate random bytes into a buffer, then use `Key::new`), and due to lack of time to get to it.

PR #1225, opened 2021-03-22
Use `Self` more.

briansmith:


briansmith:
I found these using `cargo clippy --target-dir target/clippy -- -D clippy::use_self` but I didn't enable that lint in the `clippy` job because it finds false positives.

PR #1228, opened 2021-03-23
Add support for Memory Sanitizer (msan).

briansmith:


briansmith:
TODO:
- [ ] See if we can avoid adding a feature to *ring* in favor of being able to detect memory sanitizer using some built-in `#[cfg(...)]` check.
- [ ] Run *ring*'s tests under MSAN in CI/CD.

PR #1254, opened 2021-04-23
Adding tests and fuzzing to der.rs (using cargo-fuzz)

st3fan:
Here is an _experiment_ to add fuzzing to `der.rs`. Bunch of thoughts and notes:

 * It looks like there are two popular options for fuzzing in Rust: `cargo-fuzz` and `afl.rs`. I picked the first one, not for any specific reason. It may be interesting to try to implement the same with `afl.rs` to see what the experience is.
 * This was easy to setup but it does require _Rust Nightly_ tooling.
 * I implemented a super basic fuzzer for `` which may not be the best function to fuzz. I'll add some more der.rs coverage in the coming days.
 * With `cargo-fuzz`, fuzzers are small standalone binaries. You basically implement a single function that the fuzzer tooling with turn into a binary that can then run in the fuzzing framework.
 * Simply running `cargo fuzz run fuzz_bit_string_with_no_unused_bits` is very nice and easy to automate
 *  I had hoped you could combine `test_*` and `fuzz_*` functions easily in a module's test section, but that is not possible. It is definitely possible to extract common code and share that between fuzzing and unit testing. (Although I am not familiar enough with Rust yet to understand how that would work with public/private and test namespaces.)

Let me know what you think @briansmith .. I'll play around with this a bit more. No expectation that this PR will land - consider it an experiment.

### DIFF
--- a/fuzz/fuzz_targets/fuzz_bit_string_with_no_unused_bits.rs
+++ b/fuzz/fuzz_targets/fuzz_bit_string_with_no_unused_bits.rs
@@ -1,0 +1,7 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    let mut reader = untrusted::Reader::new(untrusted::Input::from(data));
+    let _ = ring::io::der::bit_string_with_no_unused_bits(&mut reader);
+});
--- a/src/io/der.rs
+++ b/src/io/der.rs
@@ -308,4 +308,46 @@ mod tests {
             });
         }
     }
+
+    fn bytes_reader(bytes: &[u8]) -> untrusted::Reader {
+        return untrusted::Reader::new(untrusted::Input::from(bytes));
+    }
+
+    #[test]
+    fn test_bit_string_with_no_unused_bits() {
+        // Unexpected type
+        assert_eq!(
+            Err(error::Unspecified),
+            bit_string_with_no_unused_bits(&mut bytes_reader(&[0x01, 0x01, 0xff]))
+        );
+
+        // Unexpected nonexistent type
+        assert_eq!(
+            Err(error::Unspecified),
+            bit_string_with_no_unused_bits(&mut bytes_reader(&[0x42, 0xff, 0xff]))
+        );
+
+        // Unexpected empty input
+        assert_eq!(
+            Err(error::Unspecified),
+            bit_string_with_no_unused_bits(&mut bytes_reader(&[]))
+        );
+
+        // Valid input with non-zero unused bits
+        assert_eq!(
+            Err(error::Unspecified),
+            bit_string_with_no_unused_bits(&mut bytes_reader(&[0x03, 0x03, 0x04, 0x12, 0x34]))
+        );
+
+        // Valid input
+        assert_eq!(
+            untrusted::Input::from(&[0x12, 0x34]),
+            bit_string_with_no_unused_bits(&mut bytes_reader(&[0x03, 0x03, 0x00, 0x12, 0x34]))
+                .unwrap()
+        );
+    }
+
+    #[test]
+    fn fuzz_bit_string_with_no_unused_bits() {
+    }
 }
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,27 @@
+
+[package]
+name = "ring-fuzz"
+version = "0.0.0"
+authors = ["Automatically generated"]
+publish = false
+edition = "2018"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+untrusted = "0.7.1"
+
+[dependencies.ring]
+path = ".."
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
+
+[[bin]]
+name = "fuzz_bit_string_with_no_unused_bits"
+path = "fuzz_targets/fuzz_bit_string_with_no_unused_bits.rs"
+test = false
+doc = false


st3fan:
I tried to run this on an M1 Mac Mini but I don't think that is supported yet by this tooling:

> libFuzzer needs LLVM sanitizer support, so this only works on x86-64 Linux and x86-64 macOS for now. This also needs a nightly compiler since it uses some unstable command-line flags. You'll also need a C++ compiler with C++11 support.

briansmith:
Hi, I'll take a look at this really soon. I just wanted to jot down my initial thinking:

My current thoughts are to try to follow the pattern suggested in https://alastairreid.github.io/papers/hatra2020.pdf:

1. Define most tests as parameterized test functions: `fn test_foo(test_case: FooTestCase) { ... }`.

2. Have a single (probably) `#[test]` wrapper like this:
```
#[test]
fn test_foo_kat() { // `kat` is jargon for "known answer test"
    const TEST_CASES: &[FooTestCase] = &[...] {
    };
    TEST_CASES.for_each(test_foo)
}
```

3. Also use a property-based testing/fuzzing framework to generate random inputs for the parameterized test function. Depending on the testing library, this could be done by having the `FooTestCase` implement `Arbitrary` or similar.

4. Eventually replace the property-based testing with verification (proving) with minimal changes.

fuzz/fuzz_targets/fuzz_bit_string_with_no_unused_bits.rs
```diff
@@ -0,0 +1,7 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    let mut reader = untrusted::Reader::new(untrusted::Input::from(data));
+    let _ = ring::io::der::bit_string_with_no_unused_bits(&mut reader);
+});
```
briansmith, line 7: What I would expect here is that the thing inside `fuzz_target!` would be a function that can be shared between fuzz tests and unit known-answer tests.

Also, I want to clarify the goal here: The purpose of this test is to verify that `bit_string_with_no_unused_bits` never panics or crashes, right?

fuzz/fuzz_targets/read_tag_and_get_value.rs
```diff
@@ -0,0 +1,7 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    let mut reader = untrusted::Reader::new(untrusted::Input::from(data));
+    let _ = ring::io::der::read_tag_and_get_value(&mut reader);
```
briansmith, line 6: ditto.

src/io/der.rs
```diff
@@ -308,4 +308,46 @@ mod tests {
             });
         }
     }
+
+    fn bytes_reader(bytes: &[u8]) -> untrusted::Reader {
+        return untrusted::Reader::new(untrusted::Input::from(bytes));
+    }
+
+    #[test]
+    fn test_bit_string_with_no_unused_bits() {
```
briansmith, line 317: Independently of the idea of sharing the test logic between the fuzzer and the known-answer tests, it is really important that tests in *ring* are written as parameterized tests with minimal boilerplate.
   
```
    use bit_string_with_no_unused_bits_tests::TestCase;
    const TEST_CASES: &[TestCase] = &[
             // the comment for the test case.
            TestCase { 
                input: &[0x01, 0x01, 0xff],
                result: Err(error::Unspecified),
            },
            ...
        ];
        TEST_CASES.for_each(|TestCase { input, result, .. }| {
           let mut reader = ...;
           assert_eq!(bit_string_with_no_unused_bits(&mut reader), result);
        });
}

File: src/io/der.rs [hunk 4]
```diff
@@ -308,4 +308,46 @@ mod tests {
...
+        // Valid input
+        assert_eq!(
+            untrusted::Input::from(&[0x12, 0x34]),
+            bit_string_with_no_unused_bits(&mut bytes_reader(&[0x03, 0x03, 0x00, 0x12, 0x34]))
+                .unwrap()
+        );
+    }
+
+    #[test]
+    fn fuzz_bit_string_with_no_unused_bits() {
```
briansmith, line 351: I'm not sure I understand this part. Maybe just incomplete?

briansmith (commented):
LMK what you think

briansmith:
> * This was easy to setup but it does require _Rust Nightly_ tooling on _Intel_. (No M1 support - sad)

Rust nightly isn't the end of the world for an initial stab, though we should investigate options that are most likely to work on stable.

> * I implemented a super basic fuzzer for `fuzz_bit_string_with_no_unused_bits ` which may not be the best function to fuzz. I'll add some more der.rs coverage in the coming days.

Let's consider different goals of the fuzzing:
1. Verify that no matter what, the function never panics or crashes. That seems to be mostly what you're doing in this draft.
2. No matter the values of the encoded bits, any non-zero value for the "unused bits" part is rejected. For this, we'd need some scaffolding to construct an input based on two parameters: the value of the "unused bits" field (a small integer) and the value of the bitfield (a slice of bytes?). Then the fuzz tests asserts that if the unused bits field is nonzero then the result is always `Err`.

> * Unclear that corpus is thrown at this code. It was my understanding that fuzzing could also use known good input as a starting point instead of just throwing random data at it. Both strategies are useful I think and I will explore that a bit further.

Not just known good input, but known inputs with known expected results. I suggest we try to find an approach that lets us use your manually-written known-answer tests as inputs to the fuzzer.

> * With `cargo-fuzz`, fuzzers are small standalone binaries. You basically implement a single function that the fuzzer tooling with turn into a binary that can then run in the fuzzing framework.

If we have to do that, we can. But it would be better, perhaps, to use a framework that lets us define a bunch of fuzzers grouped together in a single file with less boilerplate.

> I'm going to try to be more creative with fuzzing in Ring, but I think the more interesting fuzzing should be done in WebPKI - the primitives here are probably good to be fuzzed but I have a gut feeling that it is not going to give interesting results. Fuzzing larger parsers like in WebPKI is a lot more interesting I think.

For fuzzing webpki I suggest you take a peek at what Google has done for fuzzing their X.509 library. I know they had to add a way to disable signature verification because otherwise the fuzzer doesn't make enough progress since it will literally never guess a valid signature.

briansmith:
Also, thanks for working on this!

I would suggest a humble project: Factor out the code that can construct a Tag-Length-Value encoding from a (tag, value) pair. Test that parsing such an encoding succeeds when the tag matches and the length is <= the maximum supported length, and returns `Err` otherwise, and never crashes or panics.

briansmith:
I am curious about your thoughts on https://github.com/AltSysrq/proptest for fuzzing compared to the others you found. Maybe I'm just too biased towards it do to what I've researched so far, but it does seem like a great fit.

PR #1258, opened 2021-04-27
CI/CD: Remove optimization for repository owner's PRs.

briansmith:
The optimization avoids every CI job running twice for the repository owner's
PRs. This is a pretty useful optimization because there are so many slow jobs
that are run for every commit that are duplicated for each of my PRs. However,
in the interest of making CI as bulletproof as possible, I'm going to try to
eliminate this. We'll add it back, or similar, if things get too bad.

```
$ grep "if:" .github/workflows/ci.yml | sort | uniq
      - if: ${{ !contains(matrix.host_os, 'windows') }}
      - if: ${{ contains(matrix.host_os, 'ubuntu') }}
      - if: ${{ contains(matrix.host_os, 'windows') }}
      - if: ${{ matrix.target == 'aarch64-apple-darwin' }}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,9 +6,6 @@ on:
   push:
 jobs:
   rustfmt:
-    # Don't run duplicate `push` jobs for the repo owner's PRs.
-    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
-
     runs-on: ubuntu-18.04
 
     steps:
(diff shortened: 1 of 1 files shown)


briansmith:
Unfortunately, the first CI attempt seems to have gotten hung up as pull request UI got stuck "expecting" the results from the `rustdoc` and `tests` jobs. Maybe it was a fluke. I see now, with this change, there are more than 256 total jobs. I wonder if I misunderstood the limits on jobs.

briansmith:
The documentation says "Job matrix - A job matrix can generate a maximum of 256 jobs per workflow run." I'm not sure what that means though.

briansmith:
It seems I need to avoid doing this for now, as we're apparently hitting that 256 job limit, I guess. The GitHub PR UI is looking strange during this run. I'll try to find a solution another time.

PR #1263, opened 2021-04-30
build.rs: Avoid manual dirty checking.

briansmith:
Simplify build.rs to make it easier to maintain. It seemed like the logic
being removed here wasn't always correct anyway.


briansmith:
I'll come back to this another time.

PR #1265, opened 2021-05-01
Add Set and UTF8String DER tags.

bnoordhuis:
Needed by https://github.com/briansmith/webpki/pull/226.


briansmith:
First, thank you very much for attempting to fix this.

I know how to fix the webpki bug and this isn't required for that, so I'm going to close this. I will post a PR for the webpki bug RSN.

PR #1267, opened 2021-05-03
Use a prefix different from `GFp_` so *ring* 0.17.x and an older version of *ring* can coexist.

briansmith:
See the individual commit messages for details. Future work will automate changing the prefixes as part of the release process.


briansmith:
Issue #1268 tracks the remaining related work.

crypto/cipher_extra/asm/chacha20_poly1305_x86_64.pl
```diff
@@ -879,7 +880,7 @@ sub emit_body {
     push %r14
 .cfi_push %r14
     push %r15
-.cfi_push %r15
+.cfi_push %r15
```
briansmith, line 883: I updated this file by merging the BoringSSL version to make sure they were in sync. The trailing whitespace is from BoringSSL, here and below.

crypto/cpu-intel.c
```diff
@@ -107,7 +107,7 @@ static void OPENSSL_cpuid(uint32_t *out_eax, uint32_t *out_ebx,
 //
 // See https://software.intel.com/en-us/articles/how-to-detect-new-instruction-support-in-the-4th-generation-intel-core-processor-family
 static uint64_t OPENSSL_xgetbv(uint32_t xcr) {
-#if defined(_MSC_VER) && !defined(__clang__)
+#if defined(_MSC_VER)
```
briansmith, line 110: When I started this work, I undid the manual prefixing by merging some files from BoringSSL to minimize the difference. This is one of the differences that was unrelated that was a result of that process. I decided to stop building this PR that way because there were too many of these unrelated changes; the work to minimize the diff from BoringSSL is now tracked in #1269.

crypto/fipsmodule/aes/asm/aesv8-armx.pl
```diff
@@ -276,11 +276,12 @@ ()
...
 	vst1.8	{$inout},[$out]
 	ret
-.size	GFp_${prefix}_${dir}crypt,.-GFp_${prefix}_${dir}crypt
+.size	${prefix}_${dir}crypt,.-${prefix}_${dir}crypt
 ___
 }
 &gen_block("en");
-&gen_block("de");
+# Decryption removed in *ring*.
+# &gen_block("de");
```
briansmith, line 284: If we don't do this then we'd need to prefix the generated decrypt function. But since we don't need/want it, it's better to just delete it.

Issue #1268, opened 2021-05-03
Allow multiple versions of *ring* to coexist in the same executable


briansmith:
Currently, one cannot build an app that uses more than one version of *ring* because *ring* uses the Cargo.toml "links" mechanism to prevent it. Although only the latest version of *ring* is supported at any one time, and everybody is strongly recommended to use only the latest version, this limitation makes it really difficult to incrementally upgrade dependencies to a new version of *ring*.

My plan is to use the same value of "links" in Cargo.toml for every semver-compatible version. So, for example, *ring* 0.17.x will likely use `links="ring_core_0_17"`, *ring* 0.18.x will probably use `links="ring_core_0_17"`, etc. This would still end up preventing using *ring* 0.17.1 and 0.17.2 (for example) in the same program but it will allow using any one 0.17.x version along with any one 0.18.0 version, for example. In practice, this seems to be exactly the level of flexibility that people need.

PR #1267 starts the implementation of the solution. It doesn't implement automation for choosing the prefix but it does automate (nearly) everything once the prefix is edited. It doesn't implement the check that we haven't missed any symbols as part of the renaming.

Very soon, I will implement additional automation to verify that no prefixing was skipped, that will run for each PR, and that will run as part of the release process.

In the (hopefully near) future, I hope to add automation that derives the prefix to use from the *ring* version number, but this isn't urgent. Already most of the automation other than this is done in PR #1267.

Modulo a few details, the implementation strategy is about the same as what BoringSSL natively supports and seems to be similar to the previous, aborted, attempt to solve this problem.

briansmith:
The solution in #1267 requires that the prefix is updated in multiple places:
* Cargo.toml: `links = "ring_core_dev"`.
* build.rs: `const BORINGSSL_PREFIX_VALUE: &str = "ring_core_dev_";`
* src/prefixed.rs: `{ concat!("ring_core_dev_", stringify!($name)) }`

It would be great to use `env!("CARGO_MANIFEST_LINKS") + "_"` within build.rs to eliminate needing to manually update build.rs when the prefix is updated. I haven't done that yet because that would not work for the "pregenerated_asm" mechanism. Alternatively, I suppose build.rs could just read it out of Cargo.toml itself; then it could also ensure that crate version number is consistent with the prefix. Suggestions regarding this are appreciated.

[...]

PR #1271, opened 2021-05-03
Build: Simplify assembly pregeneration mechanism.

briansmith:
This is a prerequisite for automating more of the symbol prefixing work in #1268.


src/lib.rs
```diff
@@ -45,6 +45,8 @@
 //!        requiring these hacks, and without requiring a C compiler.)
 //! </table>
 
+// When running mk/package.sh, don't actuall build any code.
```
est31, line 48: actually

PR #1276, opened 2021-05-04
Use untrusted 0.8.

briansmith:


complexspaces:
Can this PR be closed seeing as the main branch [is using `0.9`](https://github.com/briansmith/ring/blob/main/Cargo.toml#L163) of `untrusted`?

PR #1181, opened 2021-01-26
Add MIPS little-endian support

bltavares:
This is an attempt to add #562 MIPS support, starting with Little Endian platforms.

This includes the merge of several different PRs and fixes to match the latest code in the main branch, such as:

- #943
- #1174
- #1011
- #1120

Tests on other platforms (Windows, Linux) still works. There are errors when running on mips, to be fixed yet. So far, Big Endian support is not present, [as many of the BoringSSL files don't support it](https://boringssl.googlesource.com/boringssl/+/c3f4612d83f345ef0bb0024ce6086a6a773324ca).

<details>
<summary>Failed tests with `cross test --target mipsel-unknown-linux-musl` </summary>

```
test ec::suite_b::ecdsa::digest_scalar::tests::test ... ok
test ec::suite_b::ecdsa::signing::tests::signature_ecdsa_sign_asn1_test ... FAILED
test ec::suite_b::ecdsa::signing::tests::signature_ecdsa_sign_fixed_test ... FAILED
test ec::suite_b::ecdsa::verification::tests::test_digest_based_test_vectors ... FAILED
test ec::suite_b::ops::tests::p256_elem_add_test ... ok
test ec::suite_b::ops::tests::p256_elem_mul_test ... ok
test ec::suite_b::ops::tests::p256_point_double_test ... ok
test ec::suite_b::ops::tests::p256_point_mul_base_test ... ok
test ec::suite_b::ops::tests::p256_point_mul_serialized_test ... ok
test ec::suite_b::ops::tests::p256_point_mul_test ... ok
test ec::suite_b::ops::tests::p256_point_sum_test ... ok
test ec::suite_b::ops::tests::p256_q_minus_n_plus_n_equals_0_test ... ok
test ec::suite_b::ops::tests::p256_scalar_inv_to_mont_zero_panic_test ... ok
test ec::suite_b::ops::tests::p256_scalar_mul_test ... FAILED
test ec::suite_b::ops::tests::p256_scalar_square_test ... FAILED
test ec::suite_b::ops::tests::p384_elem_add_test ... ok
test ec::suite_b::ops::tests::p384_elem_div_by_2_test ... ok
test ec::suite_b::ops::tests::p384_elem_mul_test ... qemu: uncaught target signal 11 (Segmentation fault) - core dumped
```
</details>

PR #1120, opened 2020-11-24
Add mipsel to Github Actions CI setup

LinusU:
replaces #804

PR #1277, opened 2021-05-04
Add MIPS support

bltavares:
Built on top of https://github.com/briansmith/ring/pull/1120 

This commit ports the changes from #1181 into the updated codebase

The following command is working:
> cross build --target mipsel-unknown-linux-musl

But tests are still not working:
> cross test --target mipsel-unknown-linux-musl


Cargo.toml
```diff
@@ -58,6 +58,7 @@ include = [
     "crypto/fipsmodule/bn/asm/x86-mont.pl",
     "crypto/fipsmodule/bn/asm/x86_64-mont.pl",
     "crypto/fipsmodule/bn/asm/x86_64-mont5.pl",
+    "crypto/fipsmodule/bn/asm/mips-mont.pl",
```
briansmith, line 61: I understand why you're trying to do it this way, but this is exactly what I'm hoping to avoid. Let's find a way to do it without bringing in any MIPS assembly code. It's just not realistic to maintain the MIPS assembly code given current constraints. I'd rather concentrate on improving the C (in the near future, Rust) code instead.

build.rs
```diff
@@ -40,12 +42,12 @@ const RING_SRCS: &[(&[&str], &str)] = &[
     (&[], "crypto/mem.c"),
     (&[], "crypto/poly1305/poly1305.c"),
 
-    (&[AARCH64, ARM, X86_64, X86], "crypto/crypto.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/curve25519/curve25519.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/ecp_nistz.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/gfp_p256.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/gfp_p384.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/p256.c"),
+    (&[AARCH64, ARM, MIPS, MIPS64, X86_64, X86], "crypto/crypto.c"),
```
briansmith, line 45: crypto.c isn't relevant to MIPS so this should be reverted.

build.rs
```diff
@@ -40,12 +42,12 @@ const RING_SRCS: &[(&[&str], &str)] = &[
...
+    (&[AARCH64, ARM, MIPS, MIPS64, X86_64, X86], "crypto/curve25519/curve25519.c"),
+    (&[AARCH64, ARM, MIPS, MIPS64, X86_64, X86], "crypto/fipsmodule/ec/ecp_nistz.c"),
+    (&[AARCH64, ARM, MIPS, MIPS64, X86_64, X86], "crypto/fipsmodule/ec/gfp_p256.c"),
+    (&[AARCH64, ARM, MIPS, MIPS64, X86_64, X86], "crypto/fipsmodule/ec/gfp_p384.c"),
+    (&[AARCH64, ARM, MIPS, MIPS64, X86_64, X86], "crypto/fipsmodule/ec/p256.c"),
```
briansmith, line 50: Just use `&[]` for all of these. An empty list means "every platform."

build.rs
```diff
@@ -200,6 +204,20 @@ const ASM_TARGETS: &[AsmTarget] = &[
         asm_extension: "S",
         preassemble: false,
     },
+    AsmTarget {
```
briansmith, line 207: Let's remove the `AsmTarget` stuff for MIPS since my goal is to not have any assembly code for MIPS.

build.rs
```diff
@@ -322,6 +341,7 @@ fn ring_build_rs_main() {
 
     let target = Target {
         arch,
+        endian,
```
briansmith, line 344: No need to store endian in the target since no big-endian targets are supported.

build.rs
```diff
@@ -391,6 +412,10 @@ fn build_c_code(target: &Target, pregenerated: PathBuf, out_dir: &Path, ring_cor
         }
     }
 
+    if target.arch == "mips" && target.endian == "big" {
```
briansmith, line 415: No need for `target.arch == "mips"` because no big-endian targets are supported yet.

build.rs
```diff
@@ -391,6 +412,10 @@ fn build_c_code(target: &Target, pregenerated: PathBuf, out_dir: &Path, ring_cor
...
+        panic!("MIPS Big-Endian detected. Stoping compilation as BoringSSL code are not available for this platform");
```
briansmith, line 416: I would just write "Big-endian targets are not supported yet."

include/ring-core/mips_arch.h
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2011-2016 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
```
briansmith, line 8: I think (hope) we can avoid adding this file.

src/arithmetic/bigint.rs
```diff
@@ -1194,7 +1194,9 @@ fn limbs_mont_mul(r: &mut [Limb], a: &[Limb], m: &[Limb], n0: &N0) {
         target_arch = "aarch64",
         target_arch = "arm",
         target_arch = "x86_64",
-        target_arch = "x86"
+        target_arch = "x86",
+        target_arch = "mips",
+        target_arch = "mips64",
```
briansmith, line 1199: Everywhere in this file, I would expect no conditional logic for MIPS because there is a default implementation for platforms for which we don't have assembly. If those default implementations are not working for MIPS, let's investigate together why not.

briansmith (changes requested):

bltavares:
As is, the linker fails to find the `bn_mul_mont` function:

```
 = note: /target/mipsel-unknown-linux-musl/debug/deps/libring-f34856797b3553d1.rlib(gfp_p384.o): In function `elem_mul_mont':
          /project/crypto/fipsmodule/ec/gfp_p384.c:162: undefined reference to `ring_core_0_17_0_alpha_9_bn_mul_mont'
          /project/crypto/fipsmodule/ec/gfp_p384.c:162: undefined reference to `ring_core_0_17_0_alpha_9_bn_mul_mont'
          collect2: error: ld returned 1 exit status
```

Running `nm` on the output dir, we can see that the object file contains the expect function:

```
% docker run -ti --rm -v ${PWD}:/project:Z rustembedded/cross:mipsel-unknown-linux-musl-0.2.1  nm /project/target/mipsel-unknown-linux-musl/debug/build/ring-9ee2ef63756708a4/out/mips-mont-elf.o

00000000 T bn_mul_mont
00000020 t bn_mul_mont_internal
```

The output folder also contains fils for `ring_core_generated\prefix_symbols{,_asm}.h` , and both contains:

```
#define bn_mul_mont ring_core_0_17_0_alpha_9_bn_mul_mont
```

I'm not sure how to further figure out the missing link

Cargo.toml
bltavares: I'm mostly reverse-engineering the error messages which complained about `bn_mul_mont` not existing, and I found this file defines it. I'm ok with removing it if I can figure out another way to overcome the error message.

bltavares:
@briansmith I'm not sure how we'll make certain C files compile wihout adding the assembly definition for `bn_mul_mont`.
Currently, `gfp_p256.c` and `gfp_p384.c` files refer this function.

```
  = note: /target/mipsel-unknown-linux-musl/debug/deps/libring-f34856797b3553d1.rlib(gfp_p384.o): In function `elem_mul_mont':
          /project/crypto/fipsmodule/ec/gfp_p384.c:162: undefined reference to `ring_core_0_17_0_alpha_9_bn_mul_mont'
          /project/crypto/fipsmodule/ec/gfp_p384.c:162: undefined reference to `ring_core_0_17_0_alpha_9_bn_mul_mont'
          collect2: error: ld returned 1 exit status
```

briansmith:
OK, I see the problem. I was mistaken about `bn_mul_mont`. It turns out that I didn't implement a fallback for `bn_mul_mont` in the way I expected. I had implemented a fallback in the Rust wrapper, but not one that can be called by C code. Here the build fails because `bn_mul_mont` is still being used by `p256_scalar_mul_mont`, `p256_scalar_sqr_rep_mont`, and the gfp_p384.c implementation of `elem_mul_mont`.

The best thing to do is make a separate PR that transliterates those C functions to the Rust equivalent. Note that those functions are very simple. The Rust implementations of those functions should use the Rust function `limbs_mont_mul` instead of `bn_mul_mont`. For targets archs other than the big 4, `limbs_mont_mul` has a fallback implementation (in Rust, parts in C) that will work on MIPS.

Then, we can rebase this PR on top of that PR and everything should work.

briansmith:
Note that in order for the MIPS assembly to work, we'd need to also bring in the MIPS perlasm translator, and then also patch it with whatever patches would be needed. That's pretty error-prone and I'd rather focus on replacing gfp_p256.c and gfp_p384.c with Rust code, which will solve the problem with less assembly and less C and less Perl.

bltavares:
Does it mean to keep `gfp_p256.c` and `gfp_p384.c` as part of the build for the big 4 or fully replace it for all platforms? I'm not familiar with any sort of crypto code (but I'm comfortable fighting build tools hehe)

briansmith:
> Does it mean to keep `gfp_p256.c` and `gfp_p384.c` as part of the build for the big 4 or fully replace it for all platforms?

1. Delete the code for the functions in those files that call bn_mul_mont.
2. In the case of gfp_p256.c, that will be the whole file, but not for gfp_p384.c.
3. Transliterate the implementations of the deleted functions into Rust, substituting `limbs_mont_mul` for `bn_mul_mont`. The replacement for the gfp_p256.c functions goes in p256.rs and the replacements for the functions in gfp_p384.c go in p384.rs. Note that the C constant `N` is `COMMON_OPS.n` in Rust and the C constant `Q` is `COMMON_OPS.q` in Rust.

> I'm not familiar with any sort of crypto code (but I'm comfortable fighting build tools hehe)

Understandable. These functions don't implement any tricky crypto, however I understand that if you're not familiar with the code then it is going to be hard. I will try to find some time on the weekend to do it.

vDorst:
Any news?

I would like to cross compile a project for a mipsel device that is using rust-lts.
But now it fails to compile because:
```
error: failed to run custom build command for `ring v0.16.20`

Caused by:
  process didn't exit successfully: `/target/release/build/ring-fbbd3758c762a457/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.16.20/build.rs:358:10
warning: build failed, waiting for other jobs to finish...
```

target: `mipsel-unknown-linux-musl`

lancethepants:
> Any news?
> 
> I would like to cross compile a project for a mipsel device that is using rust-lts. But now it fails to compile because:
> 
> ```
> error: failed to run custom build command for `ring v0.16.20`
> 
> Caused by:
>   process didn't exit successfully: `/target/release/build/ring-fbbd3758c762a457/build-script-build` (exit status: 101)
>   --- stderr
>   thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.16.20/build.rs:358:10
>   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> warning: build failed, waiting for other jobs to finish...
> ```
> 
> target: `mipsel-unknown-linux-musl`

Also encountering this issue with mipsel-unknown-linux-uclibc

briansmith:
@bltavares Can you rebase this on top of #1558. Considering the ideas in #1555 and the work in #1297, I think the main thing to take from this PR is the changes to the scripts in `mk/` so that once the general issues with endianness are addressed, the build system will fully support the big-endian MIPS too. If you could help figure out how to get Rust toolchains for MIPS to add the profiler builtins, e.g. by cribbing from https://github.com/rust-lang/rust/pull/104304, that would also be awesome because then we could use the mipsel-unknown-linux-gnu for the little-endian code coverage measurement.

briansmith:
Thanks for your help. MIPS support was added in another PR and then temporarily removed in PR #1637; see https://github.com/briansmith/ring/issues/562#issuecomment-1741876211 for what needs to be done to add it back.

Closing this due to inactivity.

PR #1278, opened 2021-05-04
Attempt to migrate away from bn_mul_mont into Rust

bltavares:
As part of the effort to support MIPS, we would need to avoid using bn_mul_mont on C code to avoid needing to add MIPS assembly.
This commit attempts a literal port of the C code into Rust using limbs_mont_mul.

I'm opening it early [to collect feedback if this is how we should approach it](https://github.com/briansmith/ring/pull/1277#issuecomment-831660110).


bltavares:
The code currently checks in Rust, but it fails to link as `ecp_nitsz384.inl` code still refers to `elem_mul_mont` in C. Does Ring exposes Rust code as C to link back to C files?

briansmith:
Hi, thanks for doing this. You've done almost everything exactly right. Either this weekend or next weekend, I'll experiment with tweaking this to work the way I think it should work, and I'll submit a PR for you to review the changes.

bltavares:
Thank you. It's been an learning experience so far - I appreciate the feedback on these changes

briansmith:
For the P-384 stuff, we need to replace the current C code with code that uses the Fiat Crypto P-384 code that BoringSSL also uses. That should solve the remaining blocker. I will do that in a separate PR when I have some time.

Niederb:
I'm also interested in this PR. Use case would be to do EC in WebAssembly.
@briansmith Is there something we can do to help out?

src/ec/suite_b/ops.rs
```diff
@@ -240,7 +240,7 @@ pub struct ScalarOps {
     pub common: &'static CommonOps,
 
     scalar_inv_to_mont_impl: fn(a: &Scalar) -> Scalar<R>,
-    scalar_mul_mont: unsafe extern "C" fn(r: *mut Limb, a: *const Limb, b: *const Limb),
+    scalar_mul_mont: fn(r: &mut [Limb], a: &[Limb], b: &[Limb]),
```
briansmith, line 243: Ditto: We'll have to leave this as it was so that the P256 implementation can directly call the assembly/C code without any additional overhead.

src/ec/suite_b/ops/p256.rs
```diff
@@ -297,6 +298,23 @@ fn p256_scalar_inv_to_mont(a: &Scalar<Unencoded>) -> Scalar<R> {
...
+    limbs_mont_mul(r, a, b, &N_N0)
+}
+
+fn p256_scalar_sqr_rep_mont(r: &mut [Limb], a: &[Limb], rep: Limb) {
+    debug_assert!(rep >= 1);
+    p256_scalar_mul_mont(r, a, a);
+    for i in 1..rep {
+        p256_scalar_mul_mont(r, r, r);
+    }
+}
```
briansmith, line 312: Doing this would regress the performance on x86-64 and AArch64 targets that have an optimized implementation of this.

briansmith (commented):
Thanks for this PR. I'm very sorry about the long delay in reviewing it. I do absolutely love the idea of replacing gfp_p256.c with Rust code. In order for this approach to work we'd need to do the analogous thing for gfp_p384.c, which is much more tricky. I spent some time today looking into doing that but it looks like it will be a more involved process than I presently have time for this month.

I used this PR as inspiration for #1558, which uses basically the same code in `bigint` but through the `bn_mul_mont` interface, unfortunately.

Thanks again.

briansmith:
Thanks for the contribution. My understanding is that you just wanted to get MIPS and maybe other targets working. They are all working now, so I'm closing this. I appreciate your effort here.

PR #1284, opened 2021-05-11
Address `clippy::inconsistent-struct-constructor` warnings.

briansmith:


src/aead/gcm/gcm_nohw.rs
```diff
@@ -164,7 +164,7 @@ pub(super) fn init(xi: [u64; 2]) -> super::u128 {
     hi ^= carry & 0xc200000000000000;
 
     // This implementation does not use the rest of |Htable|.
-    super::u128 { lo, hi }
+    super::u128 { hi, lo }
```
briansmith, line 167: This is covered by the wasm32 tests, but this isn't measured by codecov currently.

PR #1285, opened 2021-05-11
Reduce the use of slice indexing

briansmith:
Reduce the use of slice indexing and stop disabling some slicing-related Clippy lints.


complexspaces (dismissed):
Looks good to me, none of the operations appear mixed up 👍

Issue #986, opened 2020-05-13
"cargo build" failed on s390x platform


xieqiang2020:
Hi, I tried to build ring on s390x platform but got error: 
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', build.rs:364:34
```
From build.xs, line 364:
```
    let (_, _, perlasm_format) = ASM_TARGETS
        .iter()
        .find(|entry| {
            let &(entry_arch, entry_os, _) = *entry;
            entry_arch == &target.arch && is_none_or_equals(entry_os, &target.os)
        })
        .unwrap();
```
I searched ASM_TARGETS from build.xs:
```
const ASM_TARGETS: &[(&str, Option<&str>, &str)] = &[
    ("x86_64", Some("ios"), "macosx"),
    ("x86_64", Some("macos"), "macosx"),
    ("x86_64", Some(WINDOWS), "nasm"),
    ("x86_64", None, "elf"),
    ("aarch64", Some("ios"), "ios64"),
    ("aarch64", None, "linux64"),
    ("x86", Some(WINDOWS), "win32n"),
    ("x86", Some("ios"), "macosx"),
    ("x86", None, "elf"),
    ("arm", Some("ios"), "ios32"),
    ("arm", None, "linux32"),
];
```
It seems the error is due to no s390x support?

My question are: 
1. Does ring support s390x? 
2. If it does, how should I build it? 
3. If it doesn't, will it support s390x in the future? And if possible, at what time?

Thanks

The trace of "cargo build" is as below (RUST_BACKTRACE=1):
```
[linux1@linuxone ring]$ cargo build
   Compiling ring v0.16.13 (/home/linux1/ring)
error: failed to run custom build command for `ring v0.16.13 (/home/linux1/ring)`

Caused by:
  process didn't exit successfully: `/home/linux1/ring/target/debug/build/ring-596b2342a1654bcb/build-script-build` (exit code: 101)
--- stderr
ENV CARGO=/home/linux1/.rustup/toolchains/stable-s390x-unknown-linux-gnu/bin/cargo
ENV CARGO_CFG_TARGET_ARCH=s390x
ENV CARGO_CFG_TARGET_ENDIAN=big
ENV CARGO_CFG_TARGET_ENV=gnu
ENV CARGO_CFG_TARGET_FAMILY=unix
ENV CARGO_CFG_TARGET_OS=linux
ENV CARGO_CFG_TARGET_POINTER_WIDTH=64
[...]

PR #1297, opened 2021-05-17
Add basic support for s390x

uweigand:
This adds basic support for Linux on s390x, using the fallback implementation of all algorithms and no assembler routines.   This addresses https://github.com/briansmith/ring/issues/986 and makes "cargo test" fully pass on s390x.

There were two main changes I had to make to get this working:

- Provide a bn_mul_mont fallback implementation.  Note that while there is a fallback implementation on the Rust side for limbs_mont_mul, some elliptic curve code directly calls bn_mul_mont on the C side, where there is currently no fallback.  However, given that we have a bn_from_montgomery_in_place fallback, it is straightforward to add bn_mul_mont (along the lines of the limbs_mont_mul fallback).
- Support big-endian platforms.   This consists of adding a number of endian access primitives in internal.h, and using them where required, currently in aes_nohw.c, poly1305.c, and the p256_scalar_bytes_from_limbs routine (for some reason, the p384 version of that routine is already endian-agnostic).  I have chosen to use a plain C implementation of the primitives rather than bswap intrinsics, because that is easier to maintain, doesn't require platform #ifdef's, and still compiles to good code with any recent compiler.

@briansmith, please let me know if this approach makes sense to you or if this should be done differently.

Once this basic (unoptimized) support is in, we can then build on it by adding assembler optimizations (they're already present in OpenSSL, but would need to be copied over and adapted).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,6 +182,7 @@ jobs:
           - i686-unknown-linux-musl
           - powerpc64le-unknown-linux-gnu
           - riscv64gc-unknown-linux-gnu
+          - s390x-unknown-linux-gnu
           - x86_64-pc-windows-gnu
           - x86_64-pc-windows-msvc
           - x86_64-apple-darwin
@@ -253,6 +254,9 @@ jobs:
             host_os: ubuntu-22.04
 
           - target: riscv64gc-unknown-linux-gnu
+            host_os: ubuntu-22.04
+
+          - target: s390x-unknown-linux-gnu
             host_os: ubuntu-22.04
 
           - target: x86_64-pc-windows-gnu
@@ -342,6 +346,7 @@ jobs:
           - aarch64-unknown-linux-musl
           - i686-pc-windows-msvc
           - powerpc64le-unknown-linux-gnu
+          - s390x-unknown-linux-gnu
           - x86_64-unknown-linux-gnu
 
         mode:
@@ -361,6 +366,9 @@ jobs:
             host_os: windows-latest
 
           - target: powerpc64le-unknown-linux-gnu
+            host_os: ubuntu-22.04
+
+          - target: s390x-unknown-linux-gnu
             host_os: ubuntu-22.04
 
           - target: x86_64-unknown-linux-gnu
@@ -498,6 +506,7 @@ jobs:
           - i686-unknown-linux-gnu
           - powerpc64le-unknown-linux-gnu
           - riscv64gc-unknown-linux-gnu
+          - s390x-unknown-linux-gnu
           - x86_64-unknown-linux-musl
 
         mode:
@@ -525,6 +534,9 @@ jobs:
             host_os: ubuntu-22.04
 
           - target: riscv64gc-unknown-linux-gnu
+            host_os: ubuntu-22.04
+
+          - target: s390x-unknown-linux-gnu
             host_os: ubuntu-22.04
 
           - target: x86_64-unknown-linux-musl
--- a/crypto/fipsmodule/aes/aes_nohw.c
+++ b/crypto/fipsmodule/aes/aes_nohw.c
@@ -273,6 +273,9 @@ static inline aes_word_t aes_nohw_delta_swap(aes_word_t a, aes_word_t mask,
 // http://programming.sirrida.de/calcperm.php on smaller inputs.
 #if defined(OPENSSL_64_BIT)
 static inline uint64_t aes_nohw_compact_word(uint64_t a) {
+#if defined(RING_BIG_ENDIAN)
+  a = CRYPTO_bswap8(a);
+#endif
   // Numbering the 64/2 = 16 4-bit chunks, least to most significant, we swap
   // quartets of those chunks:
   //   0 1 2 3 | 4 5 6 7 | 8  9 10 11 | 12 13 14 15 =>
@@ -294,10 +297,16 @@ static inline uint64_t aes_nohw_uncompact_word(uint64_t a) {
   a = aes_nohw_delta_swap(a, UINT64_C(0x00000000ffff0000), 16);
   a = aes_nohw_delta_swap(a, UINT64_C(0x0000ff000000ff00), 8);
   a = aes_nohw_delta_swap(a, UINT64_C(0x00f000f000f000f0), 4);
+#if defined(RING_BIG_ENDIAN)
+  a = CRYPTO_bswap8(a);
+#endif
   return a;
 }
 #else   // !OPENSSL_64_BIT
 static inline uint32_t aes_nohw_compact_word(uint32_t a) {
+#if defined(RING_BIG_ENDIAN)
+  a = CRYPTO_bswap4(a);
+#endif
   // Numbering the 32/2 = 16 pairs of bits, least to most significant, we swap:
   //   0 1 2 3 | 4 5 6 7 | 8  9 10 11 | 12 13 14 15 =>
   //   0 4 2 6 | 1 5 3 7 | 8 12 10 14 |  9 13 11 15
@@ -316,6 +325,9 @@ static inline uint32_t aes_nohw_uncompact_word(uint32_t a) {
   // Reverse the steps of |aes_nohw_uncompact_word|.
   a = aes_nohw_delta_swap(a, 0x0000f0f0, 12);
   a = aes_nohw_delta_swap(a, 0x00cc00cc, 6);
+#if defined(RING_BIG_ENDIAN)
+  a = CRYPTO_bswap4(a);
+#endif
   return a;
 }
 
--- a/crypto/internal.h
+++ b/crypto/internal.h
@@ -401,55 +401,77 @@ static inline void *OPENSSL_memset(void *dst, int c, size_t n) {
 // endianness. They use |memcpy|, and so avoid alignment or strict aliasing
 // requirements on the input and output pointers.
 
+#if defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__)
+#if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+#define RING_BIG_ENDIAN
+#endif
+#endif
+
 static inline uint32_t CRYPTO_load_u32_le(const void *in) {
   uint32_t v;
   OPENSSL_memcpy(&v, in, sizeof(v));
+#if defined(RING_BIG_ENDIAN)
+  return CRYPTO_bswap4(v);
+#else
   return v;
+#endif
 }
 
 static inline void CRYPTO_store_u32_le(void *out, uint32_t v) {
+#if defined(RING_BIG_ENDIAN)
+  v = CRYPTO_bswap4(v);
+#endif
   OPENSSL_memcpy(out, &v, sizeof(v));
 }
 
 static inline uint32_t CRYPTO_load_u32_be(const void *in) {
   uint32_t v;
   OPENSSL_memcpy(&v, in, sizeof(v));
+#if !defined(RING_BIG_ENDIAN)
   return CRYPTO_bswap4(v);
+#else
+  return v;
+#endif
 }
 
 static inline void CRYPTO_store_u32_be(void *out, uint32_t v) {
+#if !defined(RING_BIG_ENDIAN)
   v = CRYPTO_bswap4(v);
+#endif
   OPENSSL_memcpy(out, &v, sizeof(v));
 }
 
 static inline uint64_t CRYPTO_load_u64_le(const void *in) {
   uint64_t v;
   OPENSSL_memcpy(&v, in, sizeof(v));
+#if defined(RING_BIG_ENDIAN)
+  return CRYPTO_bswap8(v);
+#else
   return v;
+#endif
 }
 
 static inline void CRYPTO_store_u64_le(void *out, uint64_t v) {
+#if defined(RING_BIG_ENDIAN)
+  v = CRYPTO_bswap8(v);
+#endif
   OPENSSL_memcpy(out, &v, sizeof(v));
 }
 
 static inline uint64_t CRYPTO_load_u64_be(const void *ptr) {
   uint64_t ret;
   OPENSSL_memcpy(&ret, ptr, sizeof(ret));
+#if !defined(RING_BIG_ENDIAN)
   return CRYPTO_bswap8(ret);
+#else
+  return ret;
+#endif
 }
 
 static inline void CRYPTO_store_u64_be(void *out, uint64_t v) {
+#if !defined(RING_BIG_ENDIAN)
   v = CRYPTO_bswap8(v);
-  OPENSSL_memcpy(out, &v, sizeof(v));
-}
-
-static inline crypto_word_t CRYPTO_load_word_le(const void *in) {
-  crypto_word_t v;
-  OPENSSL_memcpy(&v, in, sizeof(v));
-  return v;
-}
-
-static inline void CRYPTO_store_word_le(void *out, crypto_word_t v) {
+#endif
   OPENSSL_memcpy(out, &v, sizeof(v));
 }
 
--- a/mk/cargo.sh
+++ b/mk/cargo.sh
@@ -23,6 +23,7 @@ qemu_arm="qemu-arm -L /usr/arm-linux-gnueabihf"
 qemu_mipsel="qemu-mipsel -L /usr/mipsel-linux-gnu"
 qemu_powerpc64le="qemu-ppc64le -L /usr/powerpc64le-linux-gnu"
 qemu_riscv64="qemu-riscv64 -L /usr/riscv64-linux-gnu"
+qemu_s390x="qemu-s390x -L /usr/s390x-linux-gnu"
 
 # Avoid putting the Android tools in `$PATH` because there are tools in this
 # directory like `clang` that would conflict with the same-named tools that may
@@ -113,6 +114,16 @@ case $target in
     export AR_riscv64gc_unknown_linux_gnu=llvm-ar-$llvm_version
     export CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER=riscv64-linux-gnu-gcc
     export CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_RUNNER="$qemu_riscv64"
+    ;;
+  s390x-unknown-linux-gnu)
+    export CC_s390x_unknown_linux_gnu=clang-$llvm_version
+    export AR_s390x_unknown_linux_gnu=llvm-ar-$llvm_version
+    # XXX: Using -march=zEC12 to work around a z13 instruction bug in
+    # QEMU 8.0.2 and earlier that causes `test_constant_time` to fail
+    # (https://lists.gnu.org/archive/html/qemu-devel/2023-05/msg06965.html).
+    export CFLAGS_s390x_unknown_linux_gnu="--sysroot=/usr/s390x-linux-gnu -march=zEC12"
+    export CARGO_TARGET_S390X_UNKNOWN_LINUX_GNU_LINKER=s390x-linux-gnu-gcc
+    export CARGO_TARGET_S390X_UNKNOWN_LINUX_GNU_RUNNER="$qemu_s390x"
     ;;
   x86_64-unknown-linux-musl)
     export CC_x86_64_unknown_linux_musl=clang-$llvm_version
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -115,5 +115,9 @@ mod sealed {
     pub trait Sealed {}
 }
 
-// TODO: https://github.com/briansmith/ring/issues/1555.
-const _LITTLE_ENDIAN_ONLY: () = assert!(cfg!(target_endian = "little"));
+// XXX: 64-bit big endian is tested; 32-bit is not.
+// TODO: Add 32-bit big endian test coverage to CI.
+const _ENDIAN_TESTING: () = assert!(cfg!(any(
+    target_endian = "little",
+    target_pointer_width = "64"
+)));
--- a/include/ring-core/target.h
+++ b/include/ring-core/target.h
@@ -46,6 +46,9 @@
 #elif defined(__riscv) && __SIZEOF_POINTER__ == 8
 #define OPENSSL_64_BIT
 #define OPENSSL_RISCV64
+#elif defined(__s390x__)
+#define OPENSSL_64_BIT
+#define OPENSSL_S390X
 #elif defined(__wasm__)
 #define OPENSSL_32_BIT
 #else
--- a/mk/install-build-tools.sh
+++ b/mk/install-build-tools.sh
@@ -105,6 +105,14 @@ case $target in
     libc6-dev-riscv64-cross \
     qemu-user
   ;;
+--target=s390x-unknown-linux-gnu)
+  # Clang is needed for code coverage.
+  use_clang=1
+  install_packages \
+    qemu-user \
+    gcc-s390x-linux-gnu \
+    libc6-dev-s390x-cross
+  ;;
 --target=wasm32-unknown-unknown)
   cargo install wasm-bindgen-cli --bin wasm-bindgen-test-runner
   use_clang=1


uweigand:
Ping?  Any comments on this approach?

crypto/internal.h
```diff
@@ -254,19 +254,77 @@ static inline crypto_word constant_time_select_w(crypto_word mask,
 
 // Endianness conversions.
 
-#if defined(__GNUC__) && __GNUC__ >= 2
-static inline uint32_t CRYPTO_bswap4(uint32_t x) {
-  return __builtin_bswap32(x);
+static inline uint32_t CRYPTO_read_le32(const uint8_t *p) {
```
briansmith, line 257: I will do a merge from the BoringSSL sources to include these functions.

crypto/fipsmodule/ec/p256_shared.h
```diff
@@ -50,7 +50,14 @@ typedef unsigned char P256_SCALAR_BYTES[33];
 
 static inline void p256_scalar_bytes_from_limbs(
     P256_SCALAR_BYTES bytes_out, const BN_ULONG limbs[P256_LIMBS]) {
-  OPENSSL_memcpy(bytes_out, limbs, 32);
```
briansmith, line 53: I need to go through all the remaining uses of OPENSSL_memcpy and make sure they're OK.

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -156,3 +156,17 @@ int bn_from_montgomery_in_place(BN_ULONG r[], size_t num_r, BN_ULONG a[],
   }
   return 1;
 }
+
+#if !defined(OPENSSL_X86) && !defined(OPENSSL_X86_64) && \
+    !defined(OPENSSL_ARM) && !defined(OPENSSL_AARCH64)
+void bn_mul_mont(BN_ULONG *rp, const BN_ULONG *ap, const BN_ULONG *bp,
```
briansmith, line 162: This would need to have the prefixing stuff done to it in build.rs.

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -156,3 +156,17 @@ int bn_from_montgomery_in_place(BN_ULONG r[], size_t num_r, BN_ULONG a[],
...
+                 const BN_ULONG *np, const BN_ULONG *n0, size_t num) {
+  Limb tmp[2 * num];
```
briansmith, line 164: I agree this makes sense. This would be the first place we've used variable-length arrays in *ring* though, which makes me hesitate. I will circle back to this.

crypto/internal.h
uweigand: Note that the BoringSSL versions (CRYPTO_load_u32_le etc.) only work correctly on little-endian hosts.  (I guess this goes back to the BoringSSL team's decision to not support big-endian hosts at all.)

The versions I've provided in this PR work on any host, little-endian or big-endian.

crypto/fipsmodule/bn/montgomery.c
uweigand: As far as I can see, bn_mul_mont is already in SYMBOLS_TO_PREFIX in build.rs ?

crypto/fipsmodule/bn/montgomery.c
uweigand: I guess we could use a global maximum limit along the lines of MODULUS_MAX_LIMBS in src/arithmetic/bigint.rs.   I'd be happy to implement whatever change you prefer -- just let me know.

crypto/fipsmodule/bn/montgomery.c
briansmith: Thanks for pointing that out.

crypto/internal.h
briansmith: Yes, I know if I merge the BoringSSL versions then they'll need to be modified to support big endian. I also am very fond of the endian-neutral implementation style you chose. However, I remember that one compiler I tested (I think MSVC) didn't recognize those idioms as of a couple years ago. Also, it's easier to make use of the BoringSSL team's performance/benchmarking work and the AWS/Galois team's verification work if we keep the code the same for little-endian platforms. That's why I'm thinking to merge in the BoringSSL teams's functions and then we can modify them to support big-endian.

uweigand: Ok, I see.  That should work as well, thanks.

vimalk78:
hi @uweigand @briansmith, i am also interested in getting `ring` supported for s390x. Let me know if i can contribute in some way.
From this PR, I was able to build `ring` using `cross build --release --target s390x-unknown-linux-gnu`
btw, how do we get the above tests to run on the PR?

uweigand:
Hi @briansmith, let me try to make another attempt to get this going.

I've just updated the patch to use a different style of endian conversion, as you suggested.   This keeps the CRYPTO_bswap4 implementation from boringssl and adds the corresponding CRYPTO_bswap8 implementation from there as well.   The other endian-specific accessors are now all defined in term of CRYPTO_bswap4 / CRYPTO_bswap8 plus OPENSSL_memcpy (to handle any alignment issues if necessary).

Does this address your concerns?

If there's anything else I can do to help get this reviewed and merged, please let me know.

uweigand:
Fixed merge conflict against current mainline.   Any comments?

Xynnn007:
Hi @briansmith , can this pr be merged now?

uweigand:
Rebased and fixed merge conflicts against current mainline.

uweigand:
Added support for cross-building and testing as requested in https://github.com/briansmith/ring/issues/1555.
Split into three logically distinct commits.

uweigand:
PR updated after https://github.com/briansmith/ring/pull/1558 was merged.

ecnelises:
This works well and can be the foundation of more big endian targets (like AIX). Any blockers?

tnakaike:
I am expecting that this PR is merged because many components of [confidential containers](https://github.com/confidential-containers), which support s390x, are written in rust. For example, [Key Broker Service](https://github.com/confidential-containers/kbs) uses rustls, which depends on ring, to enable https.

angelnu:
As Ring is used by [PRISMA](https://www.prisma.io) we would depend on this patch to be able to use Prima on s390x. As we (Metaco and Ripple) offer multi-arch builds for our SW this would help us a lot. We are looking at moving more components to Rust so this crypto library being available in s390x would be very appreciated.

uweigand:
> @uweigand Could you (or whoever is working on this at IBM/RedHat) email me at [brian@briansmith.org](mailto:brian@briansmith.org) so we could coordinate how we could get this finished without it being blocked on me, and how s390x (and PPC?) support can be best maintained going forward. I have some ideas and I want to hear your ideas.

I've now reached out to you via email.

ecnelises:
Hi @uweigand could you please CC me? AIX stuff may be related, thanks.

crypto/fipsmodule/aes/aes_nohw.c
```diff
@@ -338,15 +338,15 @@ static inline void aes_nohw_compact_block(aes_word_t out[AES_NOHW_BLOCK_WORDS],
 #if defined(OPENSSL_SSE2)
   // No conversions needed.
 #elif defined(OPENSSL_64_BIT)
-  uint64_t a0 = aes_nohw_compact_word(out[0]);
```
briansmith, line 341: I noticed that you have added `CRYPTO_bswap8_le` to every invocation of `aes_nohw_compact_word`. Instead, you could move the `CRYPTO_bswap8_le` into `aes_nohw_compact_word` itself.

Further, if you look at `aes_nohw_compact_word`, you can see that `aes_nohw_compact_word` looks like this:

```
  // Numbering the 64/2 = 16 4-bit chunks, least to most significant, we swap
  // quartets of those chunks:
  //   0 1 2 3 | 4 5 6 7 | 8  9 10 11 | 12 13 14 15 =>
  //   0 2 1 3 | 4 6 5 7 | 8 10  9 11 | 12 14 13 15
  a = aes_nohw_delta_swap(a, UINT64_C(0x00f000f000f000f0), 4);
  // Swap quartets of 8-bit chunks (still numbering by 4-bit chunks):
  //   0 2 1 3 | 4 6 5 7 | 8 10  9 11 | 12 14 13 15 =>
  //   0 2 4 6 | 1 3 5 7 | 8 10 12 14 |  9 11 13 15
  a = aes_nohw_delta_swap(a, UINT64_C(0x0000ff000000ff00), 8);
  // Swap quartets of 16-bit chunks (still numbering by 4-bit chunks):
  //   0 2 4 6 | 1  3  5  7 | 8 10 12 14 | 9 11 13 15 =>
  //   0 2 4 6 | 8 10 12 14 | 1  3  5  7 | 9 11 13 15
  a = aes_nohw_delta_swap(a, UINT64_C(0x00000000ffff0000), 16);
```
So if you really wanted to, I think you might be able to change the inputs to the aes_nohw_delta_swap calls to optimize out the `CRYPTO_bswap8_le` call completely. If you want to save this change for the future, we can defer this change.

But at a minimum, let's move the byte swapping to `aes_nohw_compact_word` to make the diff from BoringSSL smaller.

crypto/fipsmodule/ec/p256_shared.h
```diff
@@ -51,7 +51,20 @@ typedef unsigned char P256_SCALAR_BYTES[33];
...
+  {
+#if BN_BITS2 == 64
+    BN_ULONG limb = CRYPTO_bswap8(limbs[i]);
+    OPENSSL_memcpy(bytes_out + i * 8, &limb, 8);
+#else
+    BN_ULONG limb = CRYPTO_bswap4(limbs[i]);
+    OPENSSL_memcpy(bytes_out + i * 4, &limb, 4);
+#endif
+  }
+#else
```
briansmith, line 65: In the recent merge from BoringSSL I brought in the following functions:

```
static inline crypto_word CRYPTO_load_word_le(const void *in) {
  crypto_word v;
  OPENSSL_memcpy(&v, in, sizeof(v));
  return v;
}

static inline void CRYPTO_store_word_le(void *out, crypto_word v) {
  OPENSSL_memcpy(out, &v, sizeof(v));
}
```
Please use these functions to eliminate the conditional compilation logic here.

crypto/internal.h
```diff
@@ -264,6 +264,10 @@ static inline crypto_word constant_time_select_w(crypto_word mask,
 static inline uint32_t CRYPTO_bswap4(uint32_t x) {
   return __builtin_bswap32(x);
 }
+
+static inline uint64_t CRYPTO_bswap8(uint64_t x) {
+  return __builtin_bswap64(x);
+}
```
briansmith, line 270: This was brought in by the recent BoringSSL merge.

crypto/internal.h
```diff
@@ -272,8 +276,44 @@ static inline uint32_t CRYPTO_bswap4(uint32_t x) {
...
+#else
+static inline uint32_t CRYPTO_bswap4(uint32_t x) {
+  x = (x >> 16) | (x << 16);
+  x = ((x & 0xff00ff00) >> 8) | ((x & 0x00ff00ff) << 8);
+  return x;
+}
+
+static inline uint64_t CRYPTO_bswap8(uint64_t x) {
+  return CRYPTO_bswap4(x >> 32) | (((uint64_t)CRYPTO_bswap4(x)) << 32);
+}
```
briansmith, line 292: I think these are all covered bythe recent BoringSSL merge too.

Note that `CRYPTO_bswap8` and `CRYPTO_bswap4` should probably only be used for implementing the functions mentioned below; they shouldn't be used outside of internal.h, AFAICT.

crypto/internal.h
```diff
@@ -272,8 +276,44 @@ static inline uint32_t CRYPTO_bswap4(uint32_t x) {
...
+#define CRYPTO_bswap4_le(x) CRYPTO_bswap4((x))
+#define CRYPTO_bswap8_le(x) CRYPTO_bswap8((x))
+#define CRYPTO_bswap4_be(x) (x)
+#define CRYPTO_bswap8_be(x) (x)
+#else
+#define CRYPTO_bswap4_le(x) (x)
+#define CRYPTO_bswap8_le(x) (x)
+#define CRYPTO_bswap4_be(x) CRYPTO_bswap4((x))
+#define CRYPTO_bswap8_be(x) CRYPTO_bswap8((x))
+#endif
```
briansmith, line 314: In general we don't use `#define` for functions; instead we use `static inline` functions. To help you, I brought in the BoringSSL definitions of the following functions in the recent BoringSSL merge:
```
static inline uint32_t CRYPTO_load_u32_le(const void *in);
static inline void CRYPTO_store_u32_le(void *out, uint32_t v);
static inline uint32_t CRYPTO_load_u32_be(const void *in);
static inline void CRYPTO_store_u32_be(void *out, uint32_t v);
static inline uint64_t CRYPTO_load_u64_le(const void *in);
static inline void CRYPTO_store_u64_le(void *out, uint64_t v);
static inline uint64_t CRYPTO_load_u64_be(const void *ptr);
static inline void CRYPTO_store_u64_be(void *out, uint64_t v);
static inline crypto_word_t CRYPTO_load_word_le(const void *in);
static inline void CRYPTO_store_word_le(void *out, crypto_word_t v);
static inline crypto_word_t CRYPTO_load_word_be(const void *in);
```
Please use these functions throughout the PR as they are more consistent with the Rust equivalents.

Basically I think your PR will mostly reduce to implementing the big-endian versions of these functions.

Note that some of these functions may be dead code in *ring* now and you may just remove the dead ones instead of porting them to big-endian, if you want.

crypto/poly1305/poly1305.c
```diff
@@ -29,14 +29,14 @@
...
   uint32_t r;
   OPENSSL_memcpy(&r, m, sizeof(r));
-  return r;
+  return CRYPTO_bswap4_le(r);
 }
 
 static void U32TO8_LE(uint8_t *m, uint32_t v) {
+  v = CRYPTO_bswap4_le(v);
   OPENSSL_memcpy(m, &v, sizeof(v));
 }
```
briansmith, line 41: The recent BoringSSL merge fixed this.

mk/cargo.sh
```diff
@@ -108,6 +109,13 @@ case $target in
       export CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_RUSTFLAGS="$rustflags_self_contained"
     fi
     ;;
+  s390x-unknown-linux-gnu)
+    export CC_s390x_unknown_linux_gnu=clang-$llvm_version
+    export AR_s390x_unknown_linux_gnu=llvm-ar-$llvm_version
+    export CFLAGS_s390x_unknown_linux_gnu="--sysroot=/usr/s390x-linux-gnu"
+    export CARGO_TARGET_S390X_UNKNOWN_LINUX_GNU_LINKER=s390x-linux-gnu-gcc
+    export CARGO_TARGET_S390X_UNKNOWN_LINUX_GNU_RUNNER="$qemu_s390x"
+    ;;
```
briansmith, line 118: Does clang work? Do you have a preferred toolchain, GCC vs clang? Generally I prefer clang because it is much lighter weight. If your team has an official preference for GCC then we can keep using GCC.

briansmith (changes requested):
I am sorry for the long delay here. I have now done enough of the BoringSSL merge where we can move forward with this. I am sorry so many changes are necessary but the suggested changes will reduce work in the long run.

briansmith:
I am planning to do a release very soon, probably later this week. If you can update your PR then I will include it in the release. Thanks for your help here.

briansmith:
> I am planning to do a release very soon, probably later this week. If you can update your PR then I will include it in the release. Thanks for your help here.

Otherwise, if we do the 0.17 release without these changes, we can do a 0.17.1 with them this gets merged.

crypto/fipsmodule/aes/aes_nohw.c
uweigand: Given your later comments, I've now changed this to use CRYPTO_load/store instead.  I think this makes the rest of the comments here obsolete - please let me know if you'd like me to do anything different here.

crypto/fipsmodule/aes/aes_nohw.c
```diff
@@ -392,6 +394,10 @@ static inline void aes_nohw_uncompact_block(
   b1 = aes_nohw_uncompact_word(b1);
   b2 = aes_nohw_uncompact_word(b2);
   b3 = aes_nohw_uncompact_word(b3);
+  b0 = CRYPTO_bswap4_le(b0);
+  b1 = CRYPTO_bswap4_le(b1);
+  b2 = CRYPTO_bswap4_le(b2);
+  b3 = CRYPTO_bswap4_le(b3);
```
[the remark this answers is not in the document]
uweigand: See above.

crypto/fipsmodule/aes/aes_nohw.c
```diff
@@ -920,11 +926,11 @@ void aes_nohw_ctr32_encrypt_blocks(const uint8_t *in, uint8_t *out,
     OPENSSL_memcpy(ivs.u8 + 16 * i, ivec, 16);
   }
 
-  uint32_t ctr = CRYPTO_bswap4(ivs.u32[3]);
+  uint32_t ctr = CRYPTO_bswap4_be(ivs.u32[3]);
   for (;;) {
     // Update counters.
     for (uint32_t i = 0; i < AES_NOHW_BATCH_SIZE; i++) {
-      ivs.u32[4 * i + 3] = CRYPTO_bswap4(ctr + i);
+      ivs.u32[4 * i + 3] = CRYPTO_bswap4_be(ctr + i);
```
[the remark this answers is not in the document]
uweigand: Fixed.

crypto/fipsmodule/ec/p256_shared.h
uweigand: On re-testing I noticed that all this code is never even used on s390x these days - the whole file is only compiled in on x86_64 and aarch64 targets.  Therefore I've just reverted this change now.

crypto/internal.h
uweigand: Thanks!

crypto/internal.h
uweigand: Agreed.

crypto/internal.h
uweigand: I've removed the load_word / store_word versions as they are unneded.  Everything else I've added big-endian versions for.

crypto/poly1305/poly1305.c
uweigand: Fixed.

uweigand:
I think I've addressed all comments above.  Let me know if you would like me to change anything else.

mk/cargo.sh
briansmith: I mean, is there a way to avoid using `s390x-linux-gnu-gcc` for linking and instead use rust-lld or clang as the linker?

crypto/fipsmodule/aes/aes_nohw.c
briansmith: Hi, I would rather we keep this code exactly as it is in BoringSSL as much as possible and instead add the `CRYPTO_bswap` call to the top of `aes_nohw_compact_word` along with a`"// TODO: merge byte swapping into delta swapping?` comment, and the same for `uncompact_word`.  This way we ensure we're not impeding the performance at all on the already-supported targets. I guess this is the only two places where we'll need to use `CRYPTO_bswap` outside internal.h, and hopefully it is temporary (NBD if not).

crypto/fipsmodule/aes/aes_nohw.c
```diff
@@ -392,10 +391,10 @@ static inline void aes_nohw_uncompact_block(
...
   b2 = aes_nohw_uncompact_word(b2);
   b3 = aes_nohw_uncompact_word(b3);
-  OPENSSL_memcpy(out, &b0, 4);
-  OPENSSL_memcpy(out + 4, &b1, 4);
-  OPENSSL_memcpy(out + 8, &b2, 4);
-  OPENSSL_memcpy(out + 12, &b3, 4);
+  CRYPTO_store_u32_le(out, b0);
+  CRYPTO_store_u32_le(out + 4, b1);
+  CRYPTO_store_u32_le(out + 8, b1);
+  CRYPTO_store_u32_le(out + 23, b1);
```
briansmith, line 397: Ditto here.

src/lib.rs
```diff
@@ -115,5 +115,3 @@ mod sealed {
     pub trait Sealed {}
 }
 
-// TODO: https://github.com/briansmith/ring/issues/1555.
-const _LITTLE_ENDIAN_ONLY: () = assert!(cfg!(target_endian = "little"));
```
briansmith, line 119: Please leave this in, but add:
```
// XXX: 64-bit big endian is tested; 32-bit is not. 
// TODO: Add 32-bit big endian test coverage to CI.
const _ENDIAN_TESTING: () = assert!(cfg!(any(target_endian = "little", target_pointer_width = "64")));
```

briansmith (changes requested):
Thank you! The thing I notice now is that the coverage reporting complains that the generic code isn't being picked up. This is because the coverage measurement is actually done in separate jobs in a separate test matrix.

In https://github.com/rust-lang/rust/pull/104304 you added profiler support for this target. So I think we can add this target to the `coverage` CI job's matrix. We already support `aarch64-unknown-linux-gnu` in CI through QEMU so I believe it should be easy to add this.

src/lib.rs
briansmith: Also, `cargo fmt`.

briansmith:
Note that you did all the work to enable s390x in CI but you didn't add s390x to the `test`, `test-features`, and `coverage` build matrices. We should do that so we have the right test coverage.

crypto/fipsmodule/aes/aes_nohw.c
uweigand: OK, changed to use `CRYPTO_bswap` now.  Note I don't quite see how this could be folded completely into the existing delta swapping calls, it seems to me we'd need at least one extra call.  In that case, just doing the byteswap separately is likely to be more efficient (as big-endian targets are likely to have an optimized byteswap instruction).

src/lib.rs
uweigand: Done.

mk/cargo.sh
uweigand: lld doesn't support the s390x target at this point.  It should be possible to use the clang driver to invoke the (system) linker, but that'll just do the same as the gcc driver.  Do you want me to make that change anyway?

briansmith: > lld doesn't support the s390x target at this point. It should be possible to use the clang driver to invoke the (system) linker, but that'll just do the same as the gcc driver. Do you want me to make that change anyway?

No, it's OK this way. Thanks!

briansmith:
I will fix the --no-default-features build failure.

briansmith:
Please rebase on top of main and resubmit.

briansmith:
Why did the earlier version of this have changes to `p256_shared.h` but the newest one doesn't?

uweigand:
> Please rebase on top of main and resubmit.

Done.

> Why did the earlier version of this have changes to `p256_shared.h` but the newest one doesn't?

Huh, I thought I added a comment but it seems to have disappeared.   I've removed that change as re-testing showed it is no longer necessary in the current code base.

The change was to fix an endian issue in `p256_scalar_bytes_from_limbs`, but that function is now only called from the `crypto/fipsmodule/ec/p256-nistz.c` file, which in turn is only even built at all on the (little-endian) AARCH64 and X86_64 targets.  So there is no need for this file to support big-endian platforms, and any code added there would be untested anyway.

briansmith:
Good news:
The refactoring to fix the --no-default-features build was merged into main.

Less good news:
There's a merge conflict in crypto/internal.h from the latest BoringSSL merge. Also, PR #1663 moved the stuff in base.h to target.h. I will merge PR #1663 ASAP so you can rebase this on top of main for the release. I am hoping to finish the BoringSSL merge today so there shouldn't be any new conflicts.

uweigand:
@briansmith: Rebased against current mainline to fix both merge conflicts.

uweigand:
@briansmith - this is unfortunate.  The recently added `test_constant_time` test now fails on s390x.  However, the test passes when run on a native machine, and it turns out this is actually a bug in qemu - and one that was already fixed here: https://lists.gnu.org/archive/html/qemu-devel/2023-05/msg06965.html

For some reason, clang chooses to use this particular z13 instruction (LOCFHR - which is generally quite rare) when implementing the `bssl_constant_time_test_main` routine.  

I'm wondering what the best way to proceed here would be.   I can imagine a number of options:
- Just disable this test on s390x (either generally, or only when running on qemu)
- Use a qemu that has the above fix - but that would mean building qemu yourself, or switching to Ubuntu 23.04 (if github actions already support that?), as older distro versions aren't recent enough
- Tell clang to not emit this instruction, e.g. by selecting an older target architecture (adding `-march=zEC12` to `CFLAGS_s390x_unknown_linux_gnu` works for me)

Any preferences?

briansmith:
> Just disable this test on s390x (either generally, or only when running on qemu)

Let's pick this choice. Note that this test isn't new. It is one of the oldest ones, which is why it is written in C!

uweigand:
> > Just disable this test on s390x (either generally, or only when running on qemu)
> 
> Let's pick this choice. Note that this test isn't new. It is one of the oldest ones, which is why it is written in C!

Huh, interesting.  Not sure what exactly triggered this to show up just now - the test seems to have passed before the recent boringssl merges.  Anyway, with things like that, some random changes in surrounding code may trigger different instruction selection choices.

I've updated the patch now to exclude this one test on s390x.

briansmith:
Thanks. Maybe IBM can send me one of these 390x boxes? Then my daughter could do her homework on it when I'm not using it for testing.

uweigand:
> Thanks. Maybe IBM can send me one of these 390x boxes? Then my daughter could do her homework on it when I'm not using it for testing. :)

Well, we're not usually sending out boxes, but if you're actually interested, you could in fact get free access to s390x machines for the purpose of open-source development and testing :-)

See here: https://www.ibm.com/community/z/open-source/virtual-machines-request/

uweigand:
Rebased once again after the riscv64 merge.

src/constant_time.rs
```diff
@@ -40,6 +40,7 @@ mod tests {
     use crate::{bssl, error};
 
     #[test]
+    #[cfg(not(target_arch = "s390x"))] // Triggers a qemu bug before 8.0.3
```
briansmith, line 43: OK, this is not what I was expecting. I thought we'd have an `#ifdef` around a line or two within `constant_time_test.cc`.

Maybe instead we should change ci.yml so that we have something like:
```
          - target: s390x-unknown-linux-gnu
            host_os: ubuntu-22.04
            # XXX: `constant_time_test` fails when run under QEMU 8.0.2 and earlier
            # (https://link that you shared with me in your earlier comment).
            cargo_options:  -- --skip constant_time_test
```

This way people would not ever run into the issue on bare metal testing and I could also play with it locally. WDYT?

.github/workflows/ci.yml
```diff
@@ -266,6 +267,9 @@ jobs:
           - target: x86_64-unknown-linux-gnu
             host_os: ubuntu-22.04
 
+          - target: s390x-unknown-linux-gnu
+            host_os: ubuntu-22.04
```
briansmith, line 271: In the analogous PPC64LE PR I just merged, the author noted that they kept the alphabetical order intact. Could you please do the same throughout this PR? It's an arbitrary rule but it avoids people needing to wonder where to put things.

briansmith (changes requested):
Your PR is the next one to merge after the requested changes are made. Sorry about the many conflicts and delays.

briansmith:
It seems like codecov doesn't understand s390x profiler output? Too bad, but we'll deal.

src/constant_time.rs
uweigand: I see, this is indeed nicer.  I've changed the PR accordingly.

.github/workflows/ci.yml
uweigand: Used alphabetic order everywhere now.

uweigand:
> It seems like codecov doesn't understand s390x profiler output? Too bad, but we'll deal.

Why do you say it doesn't understand the output?  I saw changes reported for most of the new lines added (except for those in unused inline functions), and also for the now-skipped test_constant_time test ...

uweigand:
Ah, the `--skip test_constant_time` unfortunately seems to break the full run:
```
     Running `qemu-s390x -L /usr/s390x-linux-gnu /home/uweigand/ring/target/s390x-unknown-linux-gnu/release/deps/aead-0c8df580230e6a2e --skip test_constant_time`
error: Found argument '--skip' which wasn't expected, or isn't valid in this context
```
It is passed to both tests and benchmarks, and benchmarks apparently do not expect the `--skip` argument.

I've now changed the patch to use `-march=zEC12` instead.   That seems the straightforward workaround (directly addresses the root cause that QEMU fails to correctly emulate a z13 instruction), only needs a change in one place, and also is used only when doing the QEMU-based cross testing.

What do you think?   If you have any other suggestion, please let me know.

briansmith:
Thanks!

uweigand:
Thanks for your support in getting this merged!

PR #1308, opened 2021-06-16
io: Move `io::der` unit tests to integration tests.

briansmith:
These are testing an API that is exposed publicly, so make them
integration tests.


briansmith:
Moving `positive_integer` might make less sense than originally thought. I'm looking at removing the uses of `positive_integer` from within *ring* and then moving it to webpki now.

complexspaces (approved):

briansmith:
I'm closing this for now because I think I'll end up removing `positive_integer()` which will change this PR substantially.

PR #1311, opened 2021-06-21
Add missing test file to Cargo's includes.

### LINKED ISSUE #1301: ed25519_verify_tests.txt missing from crates.io release
The newest crates.io releases (0.16.20 and 0.17.0-alpha.10) are missing the file tests/ed25519_verify_tests.txt, which is used in tests/ed25519_tests.rs.  That file presumably needs to be added to the include field in Cargo.toml/

jgalenson:
This should allow tests to pass on the crates.io release.

Fixes #1301.


briansmith:
Thanks!

PR #1314, opened 2021-06-23
RSA: Move `verification::Key` to `public::Key`.

briansmith:
Take a step towards having RSA keys support encryption/decryption
in addition to signing/verification.

No functional changes are intended.

Some documentation was moved around.

### DIFF
--- a/src/rsa/verification.rs
+++ b/src/rsa/verification.rs
@@ -14,74 +14,13 @@
 
 //! Verification of RSA signatures.
 
-use super::{parse_public_key, RsaParameters, N, PUBLIC_KEY_PUBLIC_MODULUS_MAX_LEN};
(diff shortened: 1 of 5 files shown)


src/rsa/verification.rs
```diff
@@ -14,75 +14,14 @@
...
 
-#[derive(Debug)]
-pub struct Key {
-    pub(super) n: bigint::Modulus<N>,
-    pub(super) e: bigint::PublicExponent,
-    pub(super) n_bits: bits::BitLength,
-}
-
-impl Key {
-    pub fn from_modulus_and_exponent(
```
briansmith, line 33: Note that when I review a change like this that purports to just move code around, I use a diff tool to explicitly compare the old and new file(s). So, I would open verification.rs as "old" and key.rs as "new" in my diff tool and compare them.

complexspaces (approved):

PR #1315, opened 2021-06-23
Add AEAD benchmarks

complexspaces:
This PR brings in the AEAD tests from the `crypto-bench` project to *ring*, and then adds a decryption benchmark to augment them.,

The benchmarks were also ported from the nightly only test framework to criterion so they can run on stable.

Here's a sample output of the benchmark suite and the naming convention used:
```
encrypt_tls12_finished_AES_128_GCM
                        time:   [251.70 us 262.92 us 272.48 us]
                        change: [-4.9424% +0.6600% +6.4524%] (p = 0.83 > 0.05)
                        No change in performance detected.
```

### DIFF
--- a/benches/aead.rs
+++ b/benches/aead.rs
@@ -1,0 +1,201 @@
+// Copyright 2015-2021 Brian Smith.
+//
+// Permission to use, copy, modify, and/or distribute this software for any
+// purpose with or without fee is hereby granted, provided that the above
+// copyright notice and this permission notice appear in all copies.
+//
+// THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHORS DISCLAIM ALL WARRANTIES
+// WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY
+// SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+// OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+#![allow(missing_docs)]
+
+use criterion::criterion_main;
+use ring::{aead, error};
+
+// All the AEADs we're testing use 96-bit nonces.
+pub const NONCE: [u8; 96 / 8] = [0u8; 96 / 8];
+
+// A TLS 1.2 finished message is always 12 bytes long.
+const TLS12_FINISHED_LEN: usize = 12;
+
+// A TLS 1.3 finished message is "[t]he size of the HMAC output for the
+// Hash used for the handshake," which is usually SHA-256.
+const TLS13_FINISHED_LEN: usize = 32;
+
+// In TLS 1.2, 13 bytes of additional data are used for AEAD cipher suites.
+const TLS12_AD: [u8; 13] = [
+    23, // Type: application_data
+    3, 3, // Version = TLS 1.2.
+    0x12, 0x34, // Length = 0x1234.
+    0, 0, 0, 0, 0, 0, 0, 1, // Record #1
+];
+
+// In TLS 1.3, 5 bytes of additional data are used for AEAD cihper suites.
+const TLS13_AD: [u8; 5] = [
+    0x17, // app data type
+    0x3,  // version
+    0x3,  // ..
+    0x0,  // Length
+    0x7,  // ..
+];
+
+struct NonceSequence(u64);
+impl NonceSequence {
+    const fn new() -> Self {
+        Self(0)
+    }
+}
+
+impl aead::NonceSequence for NonceSequence {
+    fn advance(&mut self) -> Result<aead::Nonce, error::Unspecified> {
+        let mut result = [0u8; aead::NONCE_LEN];
+        result[4..].copy_from_slice(&u64::to_be_bytes(self.0));
+        self.0 = self.0.checked_add(1).ok_or(error::Unspecified)?;
+        Ok(aead::Nonce::assume_unique_for_key(result))
+    }
+}
+
+macro_rules! function_bench_name {
+    ( $benchmark_name:ident, $algorithm:expr, $operation:ident) => {{
+        const FUNC_NAME_BASE: &str = concat!(
+            "aead_",
+            stringify!($operation),
+            "_",
+            stringify!($benchmark_name),
+            "_",
+            stringify!($algorithm)
+        );
+
+        FUNC_NAME_BASE.replace("&aead::", "")
+    }};
+}
+
+macro_rules! bench {
+    ( $benchmark_name:ident, $algorithm:expr, $chunk_len:expr, $ad:expr ) => {
+        pub(super) fn $benchmark_name(c: &mut criterion::Criterion) {
+            use super::{NonceSequence, NONCE};
+            use criterion::{black_box, BatchSize};
+            use ring::{
+                aead::{self, BoundKey},
+                rand::{SecureRandom, SystemRandom},
+            };
+
+            let rng = SystemRandom::new();
+
+            let mut key_bytes = vec![0u8; $algorithm.key_len()];
+            rng.fill(&mut key_bytes).unwrap();
+
+            {
+                let key = aead::UnboundKey::new($algorithm, &key_bytes).unwrap();
+                let mut key = aead::SealingKey::new(key, NonceSequence::new());
+
+                let mut in_out = black_box(vec![0u8; $chunk_len + $algorithm.tag_len()]);
+
+                c.bench_function(
+                    &function_bench_name!($benchmark_name, $algorithm, seal),
+                    |b| {
+                        b.iter(|| -> Result<(), ring::error::Unspecified> {
+                            let aad = aead::Aad::from(black_box($ad));
+                            let _tag = key.seal_in_place_separate_tag(aad, &mut in_out)?;
+                            Ok(())
+                        })
+                    },
+                );
+            }
+
+            {
+                let key =
+                    aead::LessSafeKey::new(aead::UnboundKey::new($algorithm, &key_bytes).unwrap());
+
+                let ciphertext = {
+                    let nonce = aead::Nonce::assume_unique_for_key(NONCE);
+                    let mut in_out = vec![0u8; $chunk_len + $algorithm.tag_len()];
+                    let aad = aead::Aad::from($ad);
+                    key.seal_in_place_append_tag(nonce, aad, &mut in_out)
+                        .unwrap();
+                    in_out
+                };
+
+                let num_batches = (std::cmp::max(1, 8192 / ciphertext.len()) * 10) as u64;
+
+                c.bench_function(
+                    &function_bench_name!($benchmark_name, $algorithm, open),
+                    |b| {
+                        b.iter_batched(
+                            || ciphertext.clone(),
+                            |mut ciphertext| -> Result<(), ring::error::Unspecified> {
+                                // Optimizes out
+                                let nonce = aead::Nonce::assume_unique_for_key(NONCE);
+
+                                let aad = aead::Aad::from(black_box($ad));
+                                let _result = key.open_in_place(nonce, aad, &mut ciphertext)?;
+
+                                Ok(())
+                            },
+                            BatchSize::NumBatches(num_batches),
+                        )
+                    },
+                );
+            }
+        }
+    };
+}
+
+macro_rules! benches {
+    ( $name:ident, $algorithm:expr ) => {
+        mod $name {
+            use criterion::criterion_group;
+
+            // A TLS 1.2 finished message.
+            bench!(
+                tls12_finished,
+                $algorithm,
+                super::TLS12_FINISHED_LEN,
+                super::TLS12_AD
+            );
+
+            // A TLS 1.3 finished message.
+            bench!(
+                tls13_finished,
+                $algorithm,
+                super::TLS13_FINISHED_LEN,
+                super::TLS13_AD
+            );
+
+            // For comparison with BoringSSL.
+            bench!(tls12_16, $algorithm, 16, super::TLS12_AD);
+
+            // ~1 packet of data in TLS.
+            bench!(tls12_1350, $algorithm, 1350, super::TLS12_AD);
+            bench!(tls13_1350, $algorithm, 1350, super::TLS13_AD);
+
+            // For comparison with BoringSSL.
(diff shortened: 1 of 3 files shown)


briansmith:
The `package` CI job is failing because the benchmark sources aren't listed in Cargo.toml.

I created PR #1316 to make it less tedious to manage this. You'll have to add "benches/*.rs" to Cargo.toml after that is merged.

briansmith:
PR #1316 was merged.

benches/aead.rs
```diff
@@ -0,0 +1,162 @@
...
+// In TLS 1.2, 13 bytes of additional data are used for AEAD cipher suites.
+const TLS12_AD: [u8; 13] = [
+    23, // Type: application_data
+    3, 3, // Version = TLS 1.2.
+    0x12, 0x34, // Length = 0x1234.
+    0, 0, 0, 0, 0, 0, 0, 1, // Record #1
+];
+
+// In TLS 1.3, no additional data is used for AEAD cipher suites.
+const TLS13_AD: [u8; 0] = [];
```
briansmith, line 25: At the time I wrote the benchmarks in crypto-bench, the TLS 1.3 AAD was empty. But now it is 5 bytes long in the final RFC. So let's change this to `[u8; 5]` and update the comment.

benches/aead.rs
```diff
@@ -0,0 +1,162 @@
...
+            const FUNC_NAME_BASE: &str =
+                concat!(stringify!($benchmark_name), "_", stringify!($algorithm));
+            let func_name_cleaned = FUNC_NAME_BASE.replace("&aead::", "");
+
+            {
+                let key = aead::UnboundKey::new($algorithm, &key_bytes).unwrap();
+                let mut key = aead::SealingKey::new(key, NonceSequence::new());
+
+                let mut in_out = black_box(vec![0u8; $chunk_len + $algorithm.tag_len()]);
+                let enc_func_name = format!("encrypt_{}", func_name_cleaned);
```
briansmith, line 67: It seems like we should create a function that encapsulates `FUNC_NAME_BASE`, `func_name_cleaned`, and the `format!`. I imagine for other benchmarks (digest, etc.) we'll need a similar thing.

benches/aead.rs
```diff
@@ -0,0 +1,162 @@
...
+                c.bench_function(&enc_func_name, |b| {
+                    b.iter(|| {
+                        let aad = black_box(aead::Aad::from($ad));
+                        key.seal_in_place_append_tag(aad, &mut in_out).unwrap();
```
briansmith, line 71: Isn't this going to grow `in_out` longer and longer? I think instead we should use `seal_in_place_separate_tag`.  Then we won't be benchmarking the performance of the memory management of `Vec`.

Maybe it doesn't matter, but I think it would be better to put the `black_box` for `in_out` within this closure like you did for `aad`. Or otherwise we could use a comment about why it is done differently.

We should avoid `unwrap()` since most callers won't call it and we shouldn't measure it.

Probably `let _foo = key.seal_in_place_separate_tag(...)?; Ok(())` or similar instead.

benches/aead.rs
```diff
@@ -0,0 +1,162 @@
...
+                    })
+                });
+            }
+
+            {
```
briansmith, line 76: This decryption benchmark should be defined in a separate function from the encryption benchmark.

benches/aead.rs
```diff
@@ -0,0 +1,162 @@
...
+                let key = aead::UnboundKey::new($algorithm, &key_bytes).unwrap();
+                let mut key = aead::SealingKey::new(key, NonceSequence::new());
+
+                let mut in_out = black_box(vec![0u8; $chunk_len + $algorithm.tag_len()]);
+                let enc_func_name = format!("encrypt_{}", func_name_cleaned);
+                c.bench_function(&enc_func_name, |b| {
+                    b.iter(|| {
+                        let aad = black_box(aead::Aad::from($ad));
```
briansmith, line 70: I think this should be `aead::Aad::from(block_box(&ad))`, no? The construction of the `Aad` should get optimized away in most common uses.

benches/aead.rs
```diff
@@ -0,0 +1,162 @@
...
+                        key.seal_in_place_append_tag(aad, &mut in_out).unwrap();
+                    })
+                });
+            }
+
+            {
+                let key =
+                    aead::LessSafeKey::new(aead::UnboundKey::new($algorithm, &key_bytes).unwrap());
+
+                let nonce = aead::Nonce::assume_unique_for_key(NONCE);
```
briansmith, line 80: This should be moved into the scope of `let ciphertext = { ... }` to make the scoping clearer.

benches/aead.rs
```diff
@@ -0,0 +1,162 @@
...
+                let dec_func_name = format!("decrypt_{}", func_name_cleaned);
+                c.bench_function(&dec_func_name, |b| {
+                    b.iter_batched(
+                        || ciphertext.clone(),
+                        |mut in_out| {
+                            // Optimizes out
+                            let nonce = aead::Nonce::assume_unique_for_key(NONCE);
+
+                            let aad = black_box(aead::Aad::from($ad));
+                            key.open_in_place(nonce, aad, &mut in_out).unwrap();
```
briansmith, line 99: Most/all the comments for the sealing benchmark apply to this opening benchmark too.

briansmith (changes requested):

benches/aead.rs
complexspaces: I've updated this to be 5 bytes and changed the comment to go along with it.

benches/aead.rs
complexspaces: I've pulled this out into a macro. Its not very useful as a function since it relies on the the `$benchmark_name` and `$algorithm`.

benches/aead.rs
complexspaces: Good points here thanks. To address `in_out` growing larger due to the tag, I've moved this benchmark to also use `iter_batched`, which lets you perform data setup. Every run now uses a fresh, pre-allocated vector to ensure the allocator isn't being measured. 

I've also tweaked the usage of `black_box` to be more consistent. Each benchmark has its ciphertext or plaintext setup as part of the benchmark setup and the AEAD AAD setup to obscure the data itself, not the `Aad::from` construction. 

Finally, the unwrapping was removed in favor of a changing the benchmark closure to return a result.

benches/aead.rs
complexspaces: Addressed as part of making all the `black_box` usages consistent 👍

benches/aead.rs
complexspaces: I've moved it inside.

benches/aead.rs
complexspaces: These should also be addressed since I modified both as part of reviewing the previous comments.

benches/aead.rs
complexspaces: I've kept the encryption and decryption benchmarks inside the same macro-generated function for a few reasons:
1. To avoid bloating the `criterion_group!()` call inside the `benches!` macro. By containing both the encryption and decrypt benchmarks under one function, the amount of manual work needed to keep the list up to date is small.
2. To avoid the boilerplate of needing a key generation function or duplicate code.
3. All benchmarks are ran in serial by the cargo benchmark runner, so there is no optimization that can be gained by splitting them up that's worthwhile considering point 1 and 2. The `bench_function` method from criterion handles separating these out instead.

briansmith:
Here's what the Critereon docs say:

> With that said, if the runtime of your function is small relative to the measurement overhead it will be difficult to take accurate measurements. In this situation, the best option is to use Bencher::iter which has next-to-zero measurement overhead.

I think that's likely the case here.

benches/aead.rs
```diff
@@ -0,0 +1,186 @@
...
+
+            {
+                let key = aead::UnboundKey::new($algorithm, &key_bytes).unwrap();
+                let mut key = aead::SealingKey::new(key, NonceSequence::new());
+
+                c.bench_function(
+                    &function_bench_name!($benchmark_name, $algorithm, encrypt),
+                    |b| {
+                        b.iter_batched(
+                            || black_box(vec![0u8; $chunk_len + $algorithm.tag_len()]),
```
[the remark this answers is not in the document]
complexspaces: I've moved this one back to `iter` and it now uses a single `Vec`, along with not adding the `tag` to the vector itself.

benches/aead.rs
```diff
@@ -0,0 +1,186 @@
...
+                            |mut in_out| -> Result<(), ring::error::Unspecified> {
+                                let aad = aead::Aad::from(black_box($ad));
+                                let _result = key.seal_in_place_append_tag(aad, &mut in_out)?;
```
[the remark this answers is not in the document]
complexspaces: This is done as well 👍

benches/aead.rs
```diff
@@ -0,0 +1,186 @@
...
+                    key.seal_in_place_append_tag(nonce, aad, &mut in_out)
+                        .unwrap();
+                    in_out
+                };
+
+                c.bench_function(
+                    &function_bench_name!($benchmark_name, $algorithm, decrypt),
+                    |b| {
+                        b.iter_batched(
+                            || black_box(ciphertext.clone()),
```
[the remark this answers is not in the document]
complexspaces: I've moved this to use `iter` and a ciphertext reset at the end instead of making a new allocation/deallocation every time.

Note that I've extracted the reset into its own function in order for Rust to see its just a `memcpy` and save a lot of instructions. For some reason, it didn't see the optimization on the `Vec`s themselves, or after coercing them to slices in the same function. You can try this out [here](https://rust.godbolt.org/z/Ke41hhrTo) if you'd like.

complexspaces:
After reworking how the Vec management is done, I agree. All of the benchmark's improved in both timing and in the number of outliers that were being reported.

benches/aead.rs
```diff
@@ -0,0 +1,194 @@
...
+                // Intentionally not inside `black_box` since with a random key, the data inside should not
+                // be optimizable. We also want `reset_vec` to be as fast as possible.
+                let mut ciphertext = original_ciphertext.clone();
+
+                // This is extracted so Rust can optimize this reset to a single memcpy.
+                fn reset_vec(original_ciphertext: &[u8], ciphertext: &mut [u8]) {
+                    for (b, c) in &mut ciphertext.iter_mut().zip(original_ciphertext.iter()) {
+                        *b = *c;
+                    }
+                }
```
briansmith, line 118: Couldn't we just use `ciphertext.copy_from_slice(original_ciphertext)`?

benches/aead.rs
```diff
@@ -0,0 +1,194 @@
...
+
+                c.bench_function(
+                    &function_bench_name!($benchmark_name, $algorithm, decrypt),
+                    |b| {
+                        b.iter(|| -> Result<(), ring::error::Unspecified> {
```
briansmith, line 123: Does this work?

```
iter_batched(|| { ciphertext.copy_from_slice(&original_ciphertext); &mut ciphertext }, |ciphertext| {
    ...
}, BatchSize::PerIteration);
```

This should subtract out the cost of the resetting the input vector from the measurements?

complexspaces: This doesn't compile unfortunately.

benches/aead.rs
complexspaces: Good point, thanks for pointing that out. I forgot the method existed.

briansmith:
OK, I ran these myself and I noticed two usability issues:

1. The benchmark names don't have `aead` in them, so if I do `cargo bench aead` nothing runs. I think every benchmark should have the module name if/when practical.
2. Along similar lines, I think having "seal" and "open" in the benchmark names instead of "encrypt" and "decrypt" is important, especially if/when we implement non-AEAD encryption/decryption.

Regarding whether we should use `iter` or iter_batched,` I found the benchmarks improved significantly by doing this:
```diff
b.iter_batched_ref(|| ciphertext.clone(),
-                        b.iter(|| -> Result<(), ring::error::Unspecified> {
+                        |ciphertext| -> Result<(), ring::error::Unspecified> {
                             // Optimizes out
                             let nonce = aead::Nonce::assume_unique_for_key(NONCE);

                             let aad = aead::Aad::from(black_box($ad));
-                            let _result = key.open_in_place(nonce, aad, &mut ciphertext)?;
-
-                            reset_vec(&original_ciphertext, &mut ciphertext);
+                            let _result = key.open_in_place(nonce, aad, &mut ciphertext[..])?;

                             Ok(())
-                        })
+                        },
+                        BatchSize::NumBatches(num_batches))
``

**Comment by briansmith** [comment]
I used:
```
                let num_batches = (cmp::max(1, 8192 / ciphertext.len()) * 10) as u64;
```

**Comment by briansmith** [comment]
Regarding the `iter` vs `iter_batched` for the opening benchmarks, I think my little experiment shows that the overhead of the `memcpy` ends up being significant.

**Comment by briansmith** [comment]
Here's my understanding of how `iter_batched_ref` version works, roughtly:

```
for i in 0..batch_size {
    inputs.push(f()); // <- The memcpy is now here, so it isn't counted, which is good.
}
measure(|| {
  for input in inputs {
    g(&mut input); // <- The memcpy was here, so it was counted, which was bad.
  }
})
```
If the inputs array is huge then the CPU data cache probably gets thrashed. If inputs stays small enough then the CPU cache becomes a non issue, and any memory overhead is smaller than the overhead saved from moving the `memcpy` out of the code being measured.

**Comment by complexspaces** [comment]
I've updated the benchmark names to follow the *ring*s naming conventions better. They now look like this:
```
Benchmarking aead_open_tls13_finished_AES_128_GCM: Warming up for 3.0000 s
```

**Comment by briansmith** [comment]
One thing we should consider: the AEAD functions will run almost exactly the same time whether the encryption succeeded or failed. The only difference, if everything is correct, is the construction of `Ok(plaintext)` vs `Err(error::Unspecified)` which will be very small. If the difference in timing between those two cases is significant then that would be weird and likely a bug. So we could just remove the `?` and avoid any setup and just accept that we'd be trying to open an invalid input all the time.

**Comment by briansmith** [comment]
> One thing we should consider: the AEAD functions will run almost exactly the same time whether the encryption succeeded or failed. The only difference, if everything is correct, is the construction of `Ok(plaintext)` vs `Err(error::Unspecified)` which will be very small. If the difference in timing between those two cases is significant then that would be weird and likely a bug. So we could just remove the `?` and avoid any setup and just accept that we'd be trying to open an invalid input all the time.

@complexspaces pointed out that this isn't currently true, because when we return `Err` we zero out the in/out buffer so that users don't accidentally use the partially-decrypted contents.

File: benches/aead.rs [hunk 13]
```diff
@@ -0,0 +1,188 @@
+#![allow(missing_docs)]
```
briansmith, line 1: Please copy over the license block from src/lib.rs to this file. That would what the license block that crypto-bench's aead.rs should have had.

briansmith (commented):
This looks good to me other than the need to add the license block to aead.rs.

Please squash all of these commits into a single commit. I don't mind having separate commits. Also add the contributing license text from https://github.com/briansmith/ring#contributing to the commit message.

benches/aead.rs
complexspaces: It has been added.

complexspaces:
I've squashed my commits and added the relevant licensing parts as requested.

briansmith (approved):

PR #1319, opened 2021-06-25
RSA: Create a new public key type.

briansmith:
Expose a set of public types with minimal functionality for RSA public key operations. We'll add new functionality on top of this.


briansmith:
@complexspaces You might have seen this PR before several weeks ago. As you've seen already, I've split it into multiple parts, some of which you've already reviewed. Some of the other parts will come later as PRs on top of this one.

complexspaces (approved):

PR #1325, opened 2021-07-06
Test benchmarks in CI/CD.

briansmith:


briansmith:
@complexspaces This seems to work.

complexspaces:
> @complexspaces This seems to work.

The [behavior in CI running in QEMU](https://github.com/briansmith/ring/pull/1325/checks?check_run_id=3003974560#step:8:1647) matches what I see running the same command locally, so I agree that this should do what we'd like.

complexspaces (approved):

PR #1327, opened 2021-07-13
Upgrade `spin` to the latest version.

briansmith:


briansmith:
I will merge this after looking at #1339.

PR #1329, opened 2021-07-19
Add open_in_place_separate_tag API to LessSafeKey

complexspaces:
This PR implements a new API for `LessSafeKey`, called `open_in_place_separate`. The intention of it is to mirror the API provided by [seal_in_place_separate_tag](https://docs.rs/ring/0.17.0-alpha.10/ring/aead/struct.LessSafeKey.html#method.seal_in_place_separate_tag), providing flexibility to users who may have their tags separate by default or as part of a higher level construction.

I chose not to implement this on `OpeningKey` since it seems like an API that doesn't match up cleanly with the higher level, misuse resistant properties of it.

### DIFF
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -22,7 +22,10 @@
 //! [`crypto.cipher.AEAD`]: https://golang.org/pkg/crypto/cipher/#AEAD
 
 use crate::{cpu, error, hkdf, polyfill};
-use core::ops::RangeFrom;
+use core::{
+    convert::{TryFrom, TryInto},
+    ops::RangeFrom,
+};
 
 pub use self::{
     aes_gcm::{AES_128_GCM, AES_256_GCM},
@@ -208,14 +211,31 @@ impl PartialEq for Algorithm {
 
 impl Eq for Algorithm {}
 
-/// An authentication tag.
+/// A possibly valid authentication tag.
 #[must_use]
 #[repr(C)]
+#[derive(Clone, Copy)]
 pub struct Tag([u8; TAG_LEN]);
 
 impl AsRef<[u8]> for Tag {
     fn as_ref(&self) -> &[u8] {
         self.0.as_ref()
+    }
+}
+
+impl TryFrom<&[u8]> for Tag {
+    type Error = error::Unspecified;
+
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
+        let raw_tag: [u8; TAG_LEN] = value.try_into().map_err(|_| error::Unspecified)?;
+        Ok(Self::from(raw_tag))
+    }
+}
+
+impl From<[u8; TAG_LEN]> for Tag {
+    #[inline]
+    fn from(value: [u8; TAG_LEN]) -> Self {
+        Self(value)
     }
 }
 
--- a/src/aead/less_safe_key.rs
+++ b/src/aead/less_safe_key.rs
@@ -14,7 +14,7 @@
 
 use super::{Aad, Algorithm, KeyInner, Nonce, Tag, UnboundKey, TAG_LEN};
 use crate::{constant_time, cpu, error, polyfill};
-use core::ops::RangeFrom;
+use core::{convert::TryInto, ops::RangeFrom};
 
 /// Immutable keys for use in situations where `OpeningKey`/`SealingKey` and
 /// `NonceSequence` cannot reasonably be used.
@@ -42,6 +42,24 @@ impl LessSafeKey {
             inner: (algorithm.init)(key_bytes, cpu_features)?,
             algorithm,
         })
+    }
+
+    /// Like [open_in_place](Self::open_in_place), except the authentication tag is
+    /// passed separately.
+    #[inline]
+    pub fn open_in_place_separate_tag<'in_out, A>(
+        &self,
+        nonce: Nonce,
+        aad: Aad<A>,
+        tag: Tag,
+        in_out: &'in_out mut [u8],
+        ciphertext: RangeFrom<usize>,
+    ) -> Result<&'in_out mut [u8], error::Unspecified>
+    where
+        A: AsRef<[u8]>,
+    {
+        let aad = Aad::from(aad.as_ref());
+        open_within_(self, nonce, aad, tag, in_out, ciphertext)
     }
 
     /// Like [`OpeningKey::open_in_place()`], except it accepts an arbitrary nonce.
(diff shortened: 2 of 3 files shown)


src/aead.rs
```diff
@@ -208,9 +211,10 @@ impl PartialEq for Algorithm {
 
 impl Eq for Algorithm {}
 
-/// An authentication tag.
+/// A possibly valid authentication tag.
 #[must_use]
 #[repr(C)]
+#[derive(Clone, Copy)]
 pub struct Tag([u8; TAG_LEN]);
```
briansmith, line 218: Use test `aead_test_aad_traits` as a model for covering the `Clone` and `Copy` implementations of `Tag`.

src/aead.rs
```diff
@@ -219,6 +223,20 @@ impl AsRef<[u8]> for Tag {
...
+impl TryFrom<&[u8]> for Tag {
+    type Error = error::Unspecified;
+
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
+        Ok(Self(value.try_into()?))
+    }
+}
+
+impl From<[u8; TAG_LEN]> for Tag {
+    fn from(value: [u8; TAG_LEN]) -> Self {
```
briansmith, line 235: `#[inline]`

src/aead/less_safe_key.rs
```diff
@@ -44,6 +44,26 @@ impl LessSafeKey {
...
+        &self,
+        nonce: Nonce,
+        aad: Aad<A>,
+        tag: Tag,
+        in_out: &'in_out mut [u8],
+    ) -> Result<&'in_out mut [u8], error::Unspecified>
+    where
+        A: AsRef<[u8]>,
+    {
+        let received_tag = tag;
```
briansmith, line 59: I think you don't need this line...we can just pass `tag` directly below.

src/aead/less_safe_key.rs
```diff
@@ -44,6 +44,26 @@ impl LessSafeKey {
...
+        let aad = Aad::from(aad.as_ref());
+
+        let ciphertext_len = in_out.len();
```
briansmith, line 62: I think you don't need to pass `ciphertext_len` as a parameter.

src/aead/less_safe_key.rs
```diff
@@ -140,20 +175,18 @@ fn open_within_<'in_out>(
     key: &LessSafeKey,
     nonce: Nonce,
     aad: Aad<&[u8]>,
+    received_tag: Tag,
     in_out: &'in_out mut [u8],
     src: RangeFrom<usize>,
+    ciphertext_len: usize,
```
briansmith, line 181: You don't need `ciphertext_len` here, as it can be computed as `in_out.get(src).ok_or(error::Unspecified).len()`, right?

src/aead/less_safe_key.rs
```diff
@@ -74,12 +94,27 @@ impl LessSafeKey {
...
+
+        let received_tag = (*received_tag).try_into()?;
+
         open_within_(
             self,
             nonce,
             Aad::from(aad.as_ref()),
+            received_tag,
             in_out,
             ciphertext_and_tag,
```
briansmith, line 116: Instead of passing `ciphertext_and_tag` here, pass in `in_out` shortened by the length of the tag.

src/aead/less_safe_key.rs
```diff
@@ -74,12 +94,27 @@ impl LessSafeKey {
...
+            ciphertext_len,
```
briansmith, line 117: You don't need this extra parameter if you take the above suggestion.

briansmith (changes requested):

src/aead/less_safe_key.rs
```diff
@@ -44,6 +44,26 @@ impl LessSafeKey {
...
     }
 
+    /// Like [open_in_place](Self::open_in_place), except it accepts an arbitrary authentication tag.
+    #[inline]
+    pub fn open_in_place_seperate_tag<'in_out, A>(
+        &self,
+        nonce: Nonce,
+        aad: Aad<A>,
+        tag: Tag,
+        in_out: &'in_out mut [u8],
```
briansmith, line 54: Why not keep the flexibility of shifting the output like the other `open_in_place` functions why adding a `src: RangeFrom<usize>` parameter? I don't see a reason to drop that functionality with the new API.

briansmith (commented):
> I chose not to implement this on `OpeningKey` since it seems like an API that doesn't match up cleanly with the higher level, misuse resistant properties of it.

I don't think this API is at odds with the intent of `OpeningKey` trying to minimize misuse, but also I think we could add it seperately.

src/aead/less_safe_key.rs
complexspaces: This was mostly an oversight on my part implementing it. I've added it back.

src/aead/less_safe_key.rs
complexspaces: 👍

src/aead/less_safe_key.rs
complexspaces: Thanks for the suggestion, I've made the change.

src/aead/less_safe_key.rs
complexspaces: That works as well, yes, but wasn't this way for the reason I stated above. However, I've made the change to remove the `ciphertext_len` parameter since it probably doesn't matter.

src/aead/less_safe_key.rs
complexspaces: I had originally passed it to avoid doing more work inside of `open_within_` to calculate something we already had, but I imagine that the optimizer will have an easy enough time eliding things that it shouldn't be an issue.

src/aead/less_safe_key.rs
complexspaces: That works for me, so I've made the change. This was mostly a stylistic choice.

src/aead.rs
complexspaces: I've added the attribute here.

src/aead.rs
complexspaces: Thanks for pointing those out, I've added the test coverage, along with some more cases for the `open_in_place_separate_tag` function itself.

briansmith (changes requested):
Sorry for the slow response. I thought I'd submitted this feedback before but apparently not. Just one change.

src/aead.rs
```diff
@@ -219,6 +223,21 @@ impl AsRef<[u8]> for Tag {
     }
 }
 
+impl TryFrom<&[u8]> for Tag {
+    type Error = error::Unspecified;
+
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
+        Ok(Self(value.try_into()?))
```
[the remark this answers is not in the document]
complexspaces: Good call, thank you. I've changed it to use `Self::from()` instead here.

briansmith:
FYI: The build seems to be faliing with the current changes.

briansmith (approved):

PR #1334, opened 2021-08-03
`kSecRandomDefault` is `NULL`.

### LINKED ISSUE #1333: ring::rand::generate is UB on macOS and iOS
In the following code we're taking the `kSecRandomDefault` symbol from Security Framework and interpreting it as `&'static SecRandomRef`.

https://github.com/briansmith/ring/blob/1bf024dacfeb38eef7400b7880f46af5620cab42/src/rand.rs#L374

https://github.com/briansmith/ring/blob/1bf024dacfeb38eef7400b7880f46af5620cab42/src/rand.rs#L390

According to https://developer.apple.com/documentation/security/ksecrandomdefault:

> **This constant is a synonym for NULL.**

Null is not a legal value for a value of type `&'static SecRandomRef`.

https://doc.rust-lang.org/reference/behavior-considered-undefined.html:

> Producing an invalid value, even in private fields and locals. "Producing" a value happens any time a value is assigned to or read from a place, passed to a function/primitive operation or returned from a function/primitive operation. The following values are invalid (at their respective type): 
>
> - A reference or Box<T> that is dangling, unaligned, or points to an invalid value.
>
> ...
>
> A reference/pointer is "dangling" if it is null or not all of the bytes it points to are part of the same allocation

briansmith:
Make it a `const` instead of referencing the value in the system library to save a relocation.

Fixes #1333.

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -382,20 +382,21 @@ mod darwin {
     // this when
     // https://github.com/rust-lang/rfcs/pull/1861#issuecomment-274613536 is
     // resolved.
+    //
+    // Defined in `SecRandom.h`
     #[repr(C)]
-    struct SecRandomRef([u8; 0]);
+    struct __SecRandom([u8; 0]);
+    type SecRandomRef = *const __SecRandom;
+
+    // https://developer.apple.com/documentation/security/ksecrandomdefault
+    // says "This constant is a synonym for NULL."
+    const kSecRandomDefault: SecRandomRef = core::ptr::null();
 
     #[link(name = "Security", kind = "framework")]
     extern "C" {
-        static kSecRandomDefault: &'static SecRandomRef;
-
         // For now `rnd` must be `kSecRandomDefault`.
         #[must_use]
-        fn SecRandomCopyBytes(
-            rnd: &'static SecRandomRef,
-            count: c::size_t,
-            bytes: *mut u8,
-        ) -> c::int;
+        fn SecRandomCopyBytes(rnd: SecRandomRef, count: c::size_t, bytes: *mut u8) -> c::int;
     }
 }
 


briansmith:
@complexspaces Please review this.

The ci/clippy failure is due to Rust 1.54 being released and is being fixed in PR #1335. The ci/audit failure is being fixed in PR #1336.

complexspaces (approved):

complexspaces:
This looks good to me since the reference is now gone.

briansmith:
I rebased this PR on top of #1335 and #1336 so all CI jobs should pass.

josephlr:
This implementation now matches what we have in `getrandom`, so it looks good to me.

src/rand.rs
```diff
@@ -385,17 +385,16 @@ mod darwin {
     #[repr(C)]
     struct SecRandomRef([u8; 0]);
 
+    // https://developer.apple.com/documentation/security/ksecrandomdefault
+    // says "This constant is a synonym for NULL."
+    const kSecRandomDefault: *const SecRandomRef = core::ptr::null();
```
dtolnay, line 390: I think the UB is fixed but the types here are misleading. SecRandomRef refers to a different thing in Security Framework than here.

```suggestion
    #[repr(C)]
    struct __SecRandom([u8; 0]);
    type SecRandomRef = *const __SecRandom;

    // https://developer.apple.com/documentation/security/ksecrandomdefault
    // says "This constant is a synonym for NULL."
    const kSecRandomDefault: SecRandomRef = core::ptr::null();
```

dtolnay (changes requested):

src/rand.rs
briansmith: Thanks! I made those changes plus also changed the declaration of of `SecRandomCopyBytes` accordingly.

Issue #1167, opened 2021-01-05
Windows arm64 support (aarch64-pc-windows-msvc)


Alovchin91:
Hi,

I'm trying to compile reqwest with rustls support on Windows arm64 and I'm getting the following error:

```
error: failed to run custom build command for `ring v0.16.19`

Caused by:
  process didn't exit successfully: `C:\Projects\reqwest\target\debug\build\ring-1879f9f353eaed0d\build-script-build` (exit code: 101)
  --- stdout
  OPT_LEVEL = Some("0")
  TARGET = Some("aarch64-pc-windows-msvc")
  HOST = Some("aarch64-pc-windows-msvc")
  CC_aarch64-pc-windows-msvc = None
  CC_aarch64_pc_windows_msvc = None
  HOST_CC = None
  CC = None
  CFLAGS_aarch64-pc-windows-msvc = None
  CFLAGS_aarch64_pc_windows_msvc = None
  HOST_CFLAGS = None
  CFLAGS = None
  CRATE_CC_NO_DEFAULTS = None
  CARGO_CFG_TARGET_FEATURE = None
  DEBUG = Some("true")
  aes_nohw.c
  C:\Users\alovchin\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.16.19\include\GFp/base.h(97): fatal error C1189: #error:  "Unknown target CPU"

  --- stderr
  running "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\VC\\Tools\\MSVC\\14.28.29333\\bin\\HostX86\\ARM64\\cl.exe" "-nologo" "-MD" "-Z7" "-Brepro" "-I" "include" "-W4" "/GS" "/Gy" "/EHsc" "/GR-" "/Zc:wchar_t" "/Zc:forScope" "/Zc:inline" "/Zc:rvalueCast" "/sdl" "/Wall" "/wd4127" "/wd4464" "/wd4514" "/wd4710" "/wd4711" "/wd4820" "/wd5045" "/Od" "/RTCsu" "-DNDEBUG" "-c" "/FoC:\\Projects\\reqwest\\target\\debug\\build\\ring-b786d3340d88e8a0\\out\\aes_nohw.obj" "crypto/fipsmodule/aes/aes_nohw.c"
  thread 'main' panicked at 'execution failed', C:\Users\alovchin\.cargo\registry\src\github.com-1ecc6299db9ec823\ring-0.16.19\build.rs:673:9
```

[...]

PR #1339, opened 2021-08-09
Windows ARM64 support

Alovchin91:
#1167

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,6 +149,9 @@ jobs:
           toolchain: stable
           profile: minimal
 
+      - run: echo "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\Llvm\x64\bin" >> $GITHUB_PATH
+        shell: bash
+
       - run: sh mk/package.sh
         shell: bash
 
@@ -167,6 +170,7 @@ jobs:
           - aarch64-apple-ios
           - aarch64-apple-darwin
           - aarch64-linux-android
+          - aarch64-pc-windows-msvc
           - aarch64-unknown-linux-gnu
           - aarch64-unknown-linux-musl
           - arm-unknown-linux-gnueabihf
@@ -232,6 +236,11 @@ jobs:
           - target: aarch64-linux-android
             host_os: ubuntu-18.04
             # TODO: https://github.com/briansmith/ring/issues/486
+            cargo_options: --no-run
+
+          - target: aarch64-pc-windows-msvc
+            host_os: windows-latest
+            # GitHub Actions doesn't have a way to run this target yet.
             cargo_options: --no-run
 
           - target: aarch64-unknown-linux-gnu
@@ -302,6 +311,12 @@ jobs:
         run: |
           sudo xcode-select -s /Applications/Xcode_12.4.app &&
           sudo rm -Rf /Library/Developer/CommandLineTools/SDKs/*
+
+      - if: ${{ matrix.target == 'aarch64-pc-windows-msvc' }}
+        run: |
+          echo "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\Llvm\x64\bin" >> $GITHUB_PATH
+          echo "RUSTFLAGS=-C target-feature=+crt-static" >> $GITHUB_ENV
+        shell: bash
 
       - if: ${{ !contains(matrix.host_os, 'windows') }}
         run: |
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -162,7 +162,7 @@ name = "ring"
 [dependencies]
 untrusted = { version = "0.9" }
 
-[target.'cfg(any(target_arch = "x86",target_arch = "x86_64", all(any(target_arch = "aarch64", target_arch = "arm"), any(target_os = "android", target_os = "fuchsia", target_os = "linux"))))'.dependencies]
+[target.'cfg(any(target_arch = "x86",target_arch = "x86_64", all(any(target_arch = "aarch64", target_arch = "arm"), any(target_os = "android", target_os = "fuchsia", target_os = "linux", target_os = "windows"))))'.dependencies]
 spin = { version = "0.9.2", default-features = false, features = ["once"] }
 
 [target.'cfg(any(target_os = "android", target_os = "linux"))'.dependencies]
@@ -176,7 +176,7 @@ once_cell = { version = "1.5.2", default-features = false, features=["std"] }
 web-sys = { version = "0.3.37", default-features = false, features = ["Crypto", "Window"] }
 
 [target.'cfg(target_os = "windows")'.dependencies]
-winapi = { version = "0.3.8", default-features = false, features = ["ntsecapi", "wtypesbase"] }
+winapi = { version = "0.3.9", default-features = false, features = ["ntsecapi", "wtypesbase", "processthreadsapi"] }
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 wasm-bindgen-test = { version = "0.3.18", default-features = false }
--- a/build.rs
+++ b/build.rs
@@ -228,6 +228,13 @@ const ASM_TARGETS: &[AsmTarget] = &[
         asm_extension: "asm",
         preassemble: true,
     },
+    AsmTarget {
+        oss: &[WINDOWS],
+        arch: "aarch64",
+        perlasm_format: "win64",
+        asm_extension: "S",
+        preassemble: true,
+    },
 ];
 
 struct AsmTarget {
@@ -378,7 +385,7 @@ fn pregenerate_asm_main() {
             let os = WINDOWS;
 
             if !std::mem::replace(&mut generated_prefix_headers, true) {
-                generate_prefix_symbols_nasm(&pregenerated, &ring_core_prefix()).unwrap();
+                generate_prefix_symbols(&pregenerated, &ring_core_prefix()).unwrap();
             }
             let srcs = asm_srcs(perlasm_src_dsts);
 
@@ -390,10 +397,8 @@ fn pregenerate_asm_main() {
                 force_warnings_into_errors: true,
             };
 
-            let compiler = cc::Build::new().get_compiler();
-
             for src in srcs {
-                compile(&src, &target, &compiler, &pregenerated);
+                compile(&src, &target, &pregenerated);
             }
         }
     }
@@ -441,7 +446,7 @@ fn build_c_code(
         out_dir
     };
 
-    generate_prefix_symbols(target, out_dir, ring_core_prefix).unwrap();
+    generate_prefix_symbols(out_dir, ring_core_prefix).unwrap();
 
     let asm_srcs = if let Some(asm_target) = asm_target {
         let perlasm_src_dsts = perlasm_src_dsts(asm_dir, asm_target);
@@ -500,13 +505,11 @@ fn build_library(
     srcs: &[PathBuf],
     additional_srcs: &[PathBuf],
 ) {
-    let compiler = cc::Build::default().get_compiler();
-
     // Compile all the (dirty) source files into object files.
     let objs = additional_srcs
         .iter()
         .chain(srcs.iter())
-        .map(|f| compile(f, target, &compiler, out_dir))
+        .map(|f| compile(f, target, out_dir))
         .collect::<Vec<_>>();
 
     // Rebuild the library if necessary.
@@ -545,14 +548,14 @@ fn build_library(
     println!("cargo:rustc-link-lib=static={}", lib_name);
 }
 
-fn compile(p: &Path, target: &Target, compiler: &cc::Tool, out_dir: &Path) -> String {
+fn compile(p: &Path, target: &Target, out_dir: &Path) -> String {
     let ext = p.extension().unwrap().to_str().unwrap();
     if ext == "o" {
         p.to_str().expect("Invalid path").into()
     } else {
         let out_path = obj_path(out_dir, p);
         let cmd = if target.os != WINDOWS || ext != "asm" {
-            cc(p, ext, target, compiler, &out_path, out_dir)
+            cc(p, ext, target, &out_path, out_dir)
         } else {
             nasm(p, &target.arch, &out_path, out_dir)
         };
@@ -571,27 +574,28 @@ fn obj_path(out_dir: &Path, src: &Path) -> PathBuf {
     out_path
 }
 
-fn cc(
-    file: &Path,
-    ext: &str,
-    target: &Target,
-    compiler: &cc::Tool,
-    out_path: &Path,
-    include_dir: &Path,
-) -> Command {
+fn cc(file: &Path, ext: &str, target: &Target, out_path: &Path, include_dir: &Path) -> Command {
     let mut c = cc::Build::new();
+
+    // FIXME: On Windows AArch64 we currently must use Clang to compile C code
+    if target.os == WINDOWS && target.arch == AARCH64 && !c.get_compiler().is_like_clang() {
+        let _ = c.compiler("clang");
+    }
+
+    let compiler = c.get_compiler();
+
     let _ = c.include("include");
     let _ = c.include(include_dir);
     match ext {
         "c" => {
-            for f in c_flags(compiler) {
+            for f in c_flags(&compiler) {
                 let _ = c.flag(f);
             }
         }
         "S" => (),
         e => panic!("Unsupported file extension: {:?}", e),
     };
-    for f in cpp_flags(compiler) {
+    for f in cpp_flags(&compiler) {
         let _ = c.flag(f);
     }
     if target.os != "none"
@@ -843,24 +847,18 @@ fn ring_core_prefix() -> String {
 
 /// Creates the necessary header file for symbol renaming and returns the path of the
 /// generated include directory.
-fn generate_prefix_symbols(
-    target: &Target,
-    out_dir: &Path,
-    prefix: &str,
-) -> Result<(), std::io::Error> {
+fn generate_prefix_symbols(out_dir: &Path, prefix: &str) -> Result<(), std::io::Error> {
     generate_prefix_symbols_header(out_dir, "prefix_symbols.h", '#', None, prefix)?;
 
-    if target.os == "windows" {
-        let _ = generate_prefix_symbols_nasm(out_dir, prefix)?;
-    } else {
-        generate_prefix_symbols_header(
-            out_dir,
-            "prefix_symbols_asm.h",
-            '#',
-            Some("#if defined(__APPLE__)"),
-            prefix,
-        )?;
-    }
+    generate_prefix_symbols_nasm(out_dir, prefix)?;
+
+    generate_prefix_symbols_header(
+        out_dir,
+        "prefix_symbols_asm.h",
+        '#',
+        Some("#if defined(__APPLE__)"),
+        prefix,
+    )?;
 
     Ok(())
 }
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -31,7 +31,12 @@ pub(crate) fn features() -> Features {
         target_arch = "x86_64",
         all(
             any(target_arch = "aarch64", target_arch = "arm"),
-            any(target_os = "android", target_os = "fuchsia", target_os = "linux")
+            any(
+                target_os = "android",
+                target_os = "fuchsia",
+                target_os = "linux",
+                target_os = "windows"
+            )
         )
     ))]
     {
@@ -49,7 +54,12 @@ pub(crate) fn features() -> Features {
 
             #[cfg(all(
                 any(target_arch = "aarch64", target_arch = "arm"),
-                any(target_os = "android", target_os = "fuchsia", target_os = "linux")
+                any(
+                    target_os = "android",
+                    target_os = "fuchsia",
+                    target_os = "linux",
+                    target_os = "windows"
+                )
             ))]
             {
                 arm::setup();
@@ -162,6 +172,27 @@ pub(crate) mod arm {
         }
     }
 
+    #[cfg(all(target_os = "windows", target_arch = "aarch64"))]
+    pub fn setup() {
+        // We do not need to check for the presence of NEON, as Armv8-A always has it
+        let mut features = NEON.mask;
+
+        let result = unsafe {
+            winapi::um::processthreadsapi::IsProcessorFeaturePresent(
+                winapi::um::winnt::PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE,
+            )
+        };
+
+        if result != 0 {
+            // These are all covered by one call in Windows
+            features |= AES.mask;
+            features |= PMULL.mask;
+            features |= SHA256.mask;
+        }
+
+        unsafe { OPENSSL_armcap_P = features };
+    }
+
     macro_rules! features {
         {
             $(
@@ -237,7 +268,12 @@ pub(crate) mod arm {
             }
 
             #[cfg(all(
-                any(target_os = "android", target_os = "fuchsia", target_os = "linux"),
+                any(
+                    target_os = "android",
+                    target_os = "fuchsia",
+                    target_os = "linux",
+                    target_os = "windows"
+                ),
                 any(target_arch = "arm", target_arch = "aarch64")
             ))]
             {
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -33,9 +33,12 @@ apply when building from crates.io:
   primitives (32- and 64- bit Intel, and 32- and 64-bit ARM), Perl must be
   installed and in `$PATH`.
 
-* For Windows targets, `target/tools/windows/nasm/nasm[.exe]` is used as the
-  assembler. The version to use and how to download it is documented in
-  [.github/workflows/ci.yml](.github/workflows/ci.yml).
+* For Windows targets except ARM64, `target/tools/windows/nasm/nasm[.exe]`
+  is used as the assembler. The version to use and how to download it is
+  documented in [.github/workflows/ci.yml](.github/workflows/ci.yml).
+
+* For Windows ARM64 target, Clang is used as the C compiler and the assembler.
+  See below "Building for Windows ARM64" section.
 
 Cross Compiling
 ---------------
@@ -57,10 +60,12 @@ that the current beta and nightly releases work.
 
 On Windows, *ring* supports the x86_64-pc-windows-msvc and i686-pc-windows-msvc
 targets best. These targets require the “Visual C++ Build Tools
-2015” package or Visual Studio 2015 Update 3 or later to be installed. Patches
-to get it working on other variants, including in particular Visual Studio 2017
-([#338]), Windows ARM platforms, Windows Universal Platform, Windows XP (the
-v140_xp toolchain; [#339]), and the -gnu targets ([#330]) are welcome.
+2015” package or Visual Studio 2015 Update 3 or later to be installed.
+*ring* now also supports the aarch64-pc-windows-msvc target. For the detailed
+instructions please see the next section.
+Patches to get it working on other variants, including in particular Visual Studio 2017
+([#338]), Windows Universal Platform, Windows XP (the v140_xp toolchain; [#339]),
+and the -gnu targets ([#330]) are welcome.
 
 For other platforms, GCC 4.6 or later and Clang 3.5 or later are currently
 supported best. The build script passes options to the C/C++ compiler that are
@@ -76,6 +81,34 @@ the wrapper automatically passes flags to the actual compiler to define the
 `__ANDROID_API__` macro. Otherwise, the macro `__ANDROID_API__` must be
 defined with a value of at least 21 on 64-bit targets or 18 on 32-bit targets;
 e.g. export `CFLAGS=-D__ANDROID_API__=21`.
+
+
+Building for Windows ARM64
+--------------------------
+
+Windows ARM64 target requires the “Visual C++ Build Tools 2019” package or
+Visual Studio 2019 or later to be installed. “Desktop development with C++”
+workflow should be installed, as well as
+“MSVC v142 - VS 2019 C++ ARM64 build tools” component.
+
+To build *ring* for Windows ARM64, you will need to install Clang as it is used
+as the C compiler and the assembler for that platform. You can either use
+the version of Clang installed by Visual Studio, a standalone version from
+llvm.org, or a mingw64 version of Clang, for example, from [llvm-mingw
+project](https://github.com/mstorsjo/llvm-mingw).
+
+If you're buiding *ring* on an ARM64 device like Surface Pro X, please note
+that llvm.org and llvm-mingw have native ARM64 versions of Clang available.
+Also, if you're building *ring* on an ARM64 device, you might want to use
+`aarch64-pc-windows-msvc` Rustup toolchain, which can be installed using
+`rustup toolchain add aarch64-pc-windows-msvc`.
+
+When building on an ARM64 device, due to a bug in the Visual Studio installer,
+if you're using `rustc` version < 1.55 you would need to run `cargo build` /
+`cargo test` commands from x86_arm64 Developer Command Prompt. You can use
+`C:\Program Files (x86)\Microsoft Visual Studio\2019\<edition>\VC\Auxiliary\Build\vcvarsx86_arm64.bat`
+batch script to configure the environment. If you use `rustc` 1.55 beta or newer,
+you can run `cargo` commands without configuring the dev environment beforehand.
 
 
 Additional Features that are Useful for Development
--- a/crypto/crypto.c
+++ b/crypto/crypto.c
@@ -15,6 +15,7 @@
 #include <ring-core/cpu.h>
 #include "internal.h"
 
+#if defined(OPENSSL_X86) || defined(OPENSSL_X86_64)
 // Our assembly does not use the GOT to reference symbols, which means
 // references to visible symbols will often require a TEXTREL. This is
 // undesirable, so all assembly-referenced symbols should be hidden. CPU
@@ -26,7 +27,6 @@
 #define HIDDEN __attribute__((visibility("hidden")))
 #endif
 
-#if defined(OPENSSL_X86) || defined(OPENSSL_X86_64)
 // This value must be explicitly initialised to zero in order to work around a
 // bug in libtool or the linker on OS X.
 //
--- a/crypto/fipsmodule/ec/ecp_nistz384.inl
+++ b/crypto/fipsmodule/ec/ecp_nistz384.inl
@@ -252,6 +252,6 @@ void nistz384_point_mul(P384_POINT *r, const BN_ULONG p_scalar[P384_LIMBS],
   add_precomputed_w5(r, wvalue, table);
 }
 
-#if defined(__GNUC__)
+#if defined(__GNUC__) || defined(__clang__)
 #pragma GCC diagnostic pop
 #endif
--- a/mk/package.sh
+++ b/mk/package.sh
@@ -10,5 +10,6 @@ if [[ $(git status --porcelain | wc -c) -ne 0 ]]; then
 fi
 
 cargo clean --target-dir=target/pregenerate_asm
-RING_PREGENERATE_ASM=1 cargo build --target-dir=target/pregenerate_asm
+RING_PREGENERATE_ASM=1 CC_AARCH64_PC_WINDOWS_MSVC=clang \
+  cargo build --target-dir=target/pregenerate_asm
 cargo package --allow-dirty


crypto/fipsmodule/ec/p256.c
```diff
@@ -35,6 +35,10 @@
 #pragma warning(disable: 4244)
 // 'initializing': conversion from 'size_t' to 'fiat_p256_limb_t'
 #pragma warning(disable: 4267)
+// 'function': incompatible types - from 'uint32_t *' to 'Limb *'
+// 'function': incompatible types - from 'fiat_p256_felem' to 'const Limb *'
+// etc.
+#pragma warning(disable: 4133)
```
Alovchin91, line 41: This is probably a bad idea, as I'm not sure if it's okay to use 32-bit math on ARM64 in this module

crypto/limbs/limbs.inl
```diff
@@ -35,9 +35,24 @@
...
+{ \
+  Limb tmp = (_b) + (_carry); \
+  *(_r) = (_a) + tmp; \
+  (_ret) = ((_a) > (*(_r))) | ((_b) > tmp); \
+} while (0)
+#define RING_CORE_SUBBORROW(_borrow, _a, _b, _r, _ret) do \
+{ \
+  Limb tmp = (_a) - (_borrow); \
+  *(_r) = tmp - (_b); \
+  (_ret) = ((_a) < tmp) | ((*(_r)) > tmp); \
```
Alovchin91, line 53: I did my best, you know the rest... Please validate this math

src/cpu.rs
```diff
@@ -162,6 +162,27 @@ pub(crate) mod arm {
...
+            winapi::um::processthreadsapi::IsProcessorFeaturePresent(
+                winapi::um::winnt::PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE
+            )
+        };
+
+        if result != 0 {
+            // These are all covered by one call in Windows
+            features |= AES.mask;
+            features |= PMULL.mask;
+            features |= SHA256.mask;
```
Alovchin91, line 180: Do I understand correctly that \*ring\* doesn't support SHA1? (see [BoringSSL patch](https://github.com/google/boringssl/commit/afd5dba756b6266fa99c11af6496b39d826769cd#diff-86205ea0f68c5c51fb60ca69d8da418f3f6bb855816ae047227ed69853c75eaaR36))

awakecoding:
@Alovchin91 I have improved the build.rs script to add dynamic detection of nasm.exe / clang.exe within the PATH + a few extra paths, which solves the annoying issue of having to copy nasm.exe inside the target folder every time. Next on my list is to figure out how the pregeneration works, I think one issue could be that both x64 and ARM object files use the same names currently:

`target\aarch64-pc-windows-msvc\debug\build\ring-bf8b835a0b88012d\out\aes_nohw.obj : fatal error LNK1112: module machine type 'ARM64' conflicts with target machine type 'x64'`

build.rs
```diff
@@ -287,6 +295,24 @@ fn read_env_var(name: &'static str) -> Result<String, std::env::VarError> {
     std::env::var(name)
 }
 
+// Find executable in PATH with optional extra search directories
+fn find_in_path(path: &Path, extra: Vec<&Path>) -> Option<PathBuf> {
+    #[cfg(target_os = "windows")]
+    let path = &path.with_extension("exe");
+
+    if let Some(found) = std::env::split_paths(&std::env::var_os("PATH").unwrap_or(OsString::new()))
```
Alovchin91, line 303: Do you think we could use [PathFindOnPath](https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-pathfindonpathw) Windows API here instead? 🤔

awakecoding: That's a Windows API, and I already have dreams of fixing build.rs for cross-compilation from Linux to Windows one day :)

Alovchin91:
@awakecoding Hmm, I don't think the issue is the file name but rather the lack of "target arch" parameter in clang.exe invocation.

build.rs
Alovchin91: Haha, makes sense 😅

awakecoding:
> @awakecoding Hmm, I don't think the issue is the file name but rather the lack of "target arch" parameter in clang.exe invocation.

I think you are correct, it doesn't hurt to explicitly specify it anyway, otherwise you rely on defaults (and that's never good).

awakecoding:
ok, I fixed it - I can build the tests now if I copy the "tests" folder inside the "target" folder to avoid the incorrect path issue

awakecoding:
For some odd reason, I don't seem to have the issue with compiling the test code anymore. I did a "git clean -fdx" and I didn't copy the tests directory, and it still builds. In other words, the latest code on your branch just builds correctly for me

Alovchin91:
> For some odd reason, I don't seem to have the issue with compiling the test code anymore. I did a "git clean -fdx" and I didn't copy the tests directory, and it still builds. In other words, the latest code on your branch just builds correctly for me

I think it's important to call both `cargo build` and `cargo test` either with a `--target` parameter, or without, but not mixing it. I think the reason is indeed the preassemble flag which likely builds assembly just once and then tries to reuse it. If you mix the calls with and without flag then, you get a difference of one folder there and that's why it likely fails.

awakecoding:
My next step is to add proper detection for the MSVC clang distribution (the one installed as an MSVC component) because I think this is what @briansmith would rather use. the llvm.org clang distribution has an installer that can add clang to the PATH, but the clang.exe installed via MSVC doesn't, so it'll need a bit more work to detect and find properly. The proper modern way might be to call vswhere.exe to find it, unless there's already something better available to do this detection.

image

Alovchin91:
@awakecoding I think you want to use `cc` for that:

https://docs.rs/cc/1.0.69/cc/windows_registry/fn.find_tool.html

BTW could you please also upgrade cc to 1.0.69? This version contains fixes to find MSVC tools on Windows ARM.

awakecoding:
@Alovchin91 sure, I absolutely hate vswhere.exe, so if cc has some helpers, I'm all for it!

awakecoding:
I'm trying to call cc::windows_registry::find_tool(&msvc_target, "clang.exe") with "x86", "x64" and "ARM64" as the msvc_target, but nothing appears to work. I'm not sure if I'm calling it right?

Alovchin91:
The name of the parameter is a bit confusing, but it means that it should be an `*-msvc` Rust target. So you need to just pass the target to it, e.g. `aarch64-pc-windows-msvc`.

awakecoding:
yeah, I tried all of those and even looked at the source code for CC, it just doesn't seem to want to find it. Even if I could get a proper result, it returns a Command, and there doesn't seem to be a stable feature right now to extract the path to the executable? That could be a problem

Alovchin91:
`find_tool` should return a `Tool` type that contains path to the executable: https://docs.rs/cc/1.0.69/cc/struct.Tool.html#method.path

awakecoding:
It appears to be broken, maybe you can have better luck than I do. I think I'll try a different method, I've been on it for an hour already: 

image

build.rs
```diff
@@ -365,7 +391,7 @@ fn pregenerate_asm_main() {
             let srcs = asm_srcs(perlasm_src_dsts);
             for src in srcs {
                 let obj_path = obj_path(&pregenerated, &src, MSVC_OBJ_EXT);
-                run_command(nasm(&src, asm_target.arch, &obj_path, &pregenerated));
+                run_command(win_asm(&src, asm_target.arch, &obj_path, &pregenerated));
```
briansmith, line 394: It seems like this should be made consistent with the code below that does this:

```
let cmd = if target.os != WINDOWS || (ext != "asm" && ext != "S") {
            cc(p, ext, target, warnings_are_errors, &out_path, out_dir)
        } else {
            win_asm(p, &target.arch, &out_path, out_dir)
        };
```

Probably factor out this conditonal into a function (named `asm` perhaps) and use it here and below.

In particular, it is confusing how `win_asm` is used for `.S` files here (IIUC) but not used for `.S` files below.

build.rs
```diff
@@ -642,10 +667,11 @@ fn cc(
     c
 }
 
-fn nasm(file: &Path, arch: &str, out_file: &Path, include_dir: &Path) -> Command {
+fn win_asm(file: &Path, arch: &str, out_file: &Path, include_dir: &Path) -> Command {
```
briansmith, line 670: If you take my suggestion above about creating the `asm` function, then you can undo the changes to this function.

awakecoding:
I can call cc::windows_registry::find_tool to find link.exe and other tools, but I think it is unfortunately limited to just the executable in the "primary" tools directory you get with the visual studio command prompt. maybe that's why it doesn't find clang.exe? The thing is that we can find clang.exe through the PATH in the visual studio command prompt.

build.rs
```diff
@@ -655,7 +681,23 @@ fn nasm(file: &Path, arch: &str, out_file: &Path, include_dir: &Path) -> Command
...
     )));
 
-    let mut c = Command::new("./target/tools/windows/nasm/nasm");
+    match arch {
+        "aarch64" => clang(out_file, include_dir, file),
+        _ => nasm(out_file, oformat, include_dir, file)
+    }
+}
+
+fn find_nasm() -> Option<PathBuf> {
```
briansmith, line 690: Especially since this PR doesn't actually use NASM, we don't need to change the way NASM is found in this PR. If we want to be more flexible with NASM then let's discuss that in another issue/PR.

Alovchin91:
I think you're right: https://github.com/alexcrichton/cc-rs/blob/4a6e8b7c7522392de0cda0325d4041f536745fe0/src/windows_registry.rs#L420

A good idea for extending the library then! 🙂

crypto/fipsmodule/ec/p256.c
briansmith: Yes, we shouldn't disable these warnings. When the code is correct, then there will be no warnings. These warnings help us ensure we're not accidentally using 32-bit integer types instead of 32-bit integer types or vice versa. If there are cases we need to do conversions then we should add explicit casts for those specific instances.

crypto/limbs/limbs.inl
```diff
@@ -35,9 +35,24 @@
 #endif
 typedef uint8_t Carry;
 #if LIMB_BITS == 64
+#if !defined(OPENSSL_AARCH64)
```
briansmith, line 38: I think we should delete this `#if` and instead change the above:

```
#if defined(_MSC_VER) && !defined(__clang__)
```
to something like:
```
#if defined(_MSC_VER) && !defined(__clang__) && (defined(OPENSSL_X86) || defined(OPENSSL_X86_64))
```

Alovchin91:
As a workaround you could ask it to find devenv.exe and given that it has a well-known path, build a new path based on it: https://github.com/alexcrichton/cc-rs/blob/4a6e8b7c7522392de0cda0325d4041f536745fe0/src/windows_registry.rs#L841

crypto/limbs/limbs.inl
briansmith: AFAICT, these macros are just doing what the `#else` branches below already do, so we can just undo this change if you take my suggestion above.

crypto/limbs/limbs.inl
```diff
@@ -60,6 +75,8 @@ static inline Carry limb_adc(Limb *r, Limb a, Limb b, Carry carry_in) {
   Carry ret;
 #if defined(RING_CORE_ADDCARRY_INTRINSIC)
   ret = RING_CORE_ADDCARRY_INTRINSIC(carry_in, a, b, r);
+#elif defined(RING_CORE_ADDCARRY)
+  RING_CORE_ADDCARRY(carry_in, a, b, r, ret);
 #else
```
briansmith, line 80: Why doesn't the existing `#else` case do the correct thing?

crypto/limbs/limbs.inl
```diff
@@ -74,6 +91,8 @@ static inline Carry limb_add(Limb *r, Limb a, Limb b) {
   Carry ret;
 #if defined(RING_CORE_ADDCARRY_INTRINSIC)
   ret = RING_CORE_ADDCARRY_INTRINSIC(0, a, b, r);
+#elif defined(RING_CORE_ADDCARRY)
+  RING_CORE_ADDCARRY(0, a, b, r, ret);
 #else
```
briansmith, line 96: Ditto.

crypto/limbs/limbs.inl
```diff
@@ -90,6 +109,8 @@ static inline Carry limb_sbb(Limb *r, Limb a, Limb b, Carry borrow_in) {
   Carry ret;
 #if defined(RING_CORE_SUBBORROW_INTRINSIC)
   ret = RING_CORE_SUBBORROW_INTRINSIC(borrow_in, a, b, r);
+#elif defined(RING_CORE_SUBBORROW)
+  RING_CORE_SUBBORROW(borrow_in, a, b, r, ret);
```
briansmith, line 113: ditto.

crypto/limbs/limbs.inl
```diff
@@ -104,6 +125,8 @@ static inline Carry limb_sub(Limb *r, Limb a, Limb b) {
   Carry ret;
 #if defined(RING_CORE_SUBBORROW_INTRINSIC)
   ret = RING_CORE_SUBBORROW_INTRINSIC(0, a, b, r);
+#elif defined(RING_CORE_SUBBORROW)
+  RING_CORE_SUBBORROW(0, a, b, r, ret);
```
briansmith, line 129: ditto.

src/cpu.rs
briansmith: Correct. *ring* currently doesn't use accelerated SHA-1. We will probably change that soon, as now there are commercial users that do need the SHA-1 to be fast. That's future work though.

crypto/fipsmodule/bn/internal.h
```diff
@@ -185,8 +185,13 @@ void bn_mul_mont(BN_ULONG *rp, const BN_ULONG *ap, const BN_ULONG *bp,
...
 static inline void bn_umult_lohi(BN_ULONG *low_out, BN_ULONG *high_out,
                                  BN_ULONG a, BN_ULONG b) {
-#if defined(OPENSSL_X86_64) && defined(_MSC_VER) && !defined(__clang__)
+#if defined(_MSC_VER) && !defined(__clang__)
+#if defined(OPENSSL_X86_64)
   *low_out = _umul128(a, b, high_out);
+#elif defined(OPENSSL_AARCH64)
+  *low_out = a * b;
+  *high_out = __umulh(a, b);
+#endif
```
briansmith, line 194: Doesn't the `#else` branch immediately below already do the right thing? I don't see the advantage of this change.

briansmith (changes requested):
I took a quick look at the patch and

briansmith:
Hi, I took a quick look at the PR and offered some suggested changes. I am impressed with your ability to get this mostly sorted out already on your own.

Could you please also extend the GitHub Actions configuration to include Windows AAarch64 as a build-but-don't-test target? This would make it clear how one gets `clang` and make it easy for me to test this locally, since I'd just need to run the same scripts as GitHub Actions uses.

briansmith:
> `find_tool` should return a `Tool` type that contains a path to the executable: https://docs.rs/cc/1.0.69/cc/struct.Tool.html#method.path

Again, I wouldn't worry about trying to search for `clang` (or `nasm`) right now. When I package *ring* on crates.io then the assembly will be packaged in `.obj` files in the packaged crate and the user won't need `clang` or `nasm`.

build.rs
```diff
@@ -287,6 +295,24 @@ fn read_env_var(name: &'static str) -> Result<String, std::env::VarError> {
     std::env::var(name)
 }
 
+// Find executable in PATH with optional extra search directories
+fn find_in_path(path: &Path, extra: Vec<&Path>) -> Option<PathBuf> {
```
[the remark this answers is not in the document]
awakecoding: I understand your point, so may I suggest to modify it such that specific directories are searched *before* checking the PATH instead? This would ensure that target/tools/windows/nasm/nasm.exe always gets picked up even if nasm.exe is also present in the PATH. Requiring a specific copy of a tool within a custom directory is particularly painful for everybody else. I have nasm installed in my CI environment for other dependencies that require it, I just make sure to make it available in my system PATH so they can pick it up.

crypto/fipsmodule/bn/internal.h
Alovchin91: TBH I copied it from here without looking 😬 https://github.com/microsoft/winrtc/blob/ef004bdb791282e11cfbe3312cda31f30f5afbeb/patches_for_WebRTC_org/m84/4001-Arm64-is-a-thing-and-has-intrinsic-to-mul-two-64bit-.patch#L24

The only difference seems to be that this branch uses an [`__umulh`](https://docs.microsoft.com/en-us/cpp/intrinsics/umulh?view=msvc-160) intrinsic. I think we can remove it indeed 😅

awakecoding:
> > `find_tool` should return a `Tool` type that contains a path to the executable: https://docs.rs/cc/1.0.69/cc/struct.Tool.html#method.path
> 
> Again, I wouldn't worry about trying to search for `clang` (or `nasm`) right now. When I package _ring_ on crates.io then the assembly will be packaged in `.obj` files in the packaged crate and the user won't need `clang` or `nasm`.

The concern is not limited to the way ring is packaged for crates.io, but how one can rebuild it entirely from source, including for the assembly files. You mentioned that you'd rather use the MSVC clang, and that's the only thing missing right now - properly detecting its install path so we can pick it up. If you don't care about detection then I guess you can manually add it the PATH in your build environment? We're just trying to make it simpler with minimal custom operations, such that all one needs to do is either install MSVC clang or llvm.org clang to rebuild everything.

crypto/limbs/limbs.inl
Alovchin91: It's not possible to use the `#else` branches below unfortunately because they use 128-bit `__uint128_t` type which is not available on MSVC 🤷

crypto/fipsmodule/ec/p256.c
Alovchin91: Okay, it looks like we have a problem then because of the `__uint128_t` type 😅

crypto/limbs/limbs.inl
briansmith: > It's not possible to use the `#else` branches below unfortunately because they use 128-bit `__uint128_t` type which is not available on MSVC 🤷

Thanks. I forgot about that. I will think about what to do. Probably it makes sense to change the `#else` branches so that they don't require `__uint128_t` and then are shared across MSVC and non-MSVC targets, if that doesn't negatively affect the performance.

Alovchin91: I'll leave this decision up to you if you don't mind 🙂

briansmith:
> The concern is not limited to the way ring is packaged for crates.io, but how one can rebuild it entirely from source, including for the assembly files. You mentioned that you'd rather use the MSVC clang, and that's the only thing missing right now - properly detecting its install path so we can pick it up. If you don't care about detection then I guess you can manually add it the PATH in your build environment? We're just trying to make it simpler with minimal custom operations, such that all one needs to do is either install MSVC clang or llvm.org clang to rebuild everything.

I understand. I would prefer to deal with it in a separate PR though. For a long time the crates.io distribution is the only one I "support." I am open to changing that but I'd rather deal with all the things that are involved in changing that policy separately.

src/cpu.rs
Alovchin91: Would you prefer to leave a TODO here so we don't forget to enable it later on?

awakecoding:
> > The concern is not limited to the way ring is packaged for crates.io, but how one can rebuild it entirely from source, including for the assembly files. You mentioned that you'd rather use the MSVC clang, and that's the only thing missing right now - properly detecting its install path so we can pick it up. If you don't care about detection then I guess you can manually add it the PATH in your build environment? We're just trying to make it simpler with minimal custom operations, such that all one needs to do is either install MSVC clang or llvm.org clang to rebuild everything.
> 
> I understand. I would prefer to deal with it in a separate PR though. For a long time the crates.io distribution is the only one I "support." I am open to changing that but I'd rather deal with all the things that are involved in changing that policy separately.

No problem, we can trim and remove some of the detection code and expect clang.exe to be already present in the PATH, and avoid touching nasm.exe to keep it all for a separate PR. Just to make sure you understand, we're not going to implement MSVC clang detection in this PR, mostly because it's quite painful and we've been on it for hours already, but also because it would bring even more complex detection code in the current PR. The most sensible way to detect it would be to implement calling vswhere.exe like in the screenshot below. I pointed to the x86 and x64 versions of the MSVC-distributed clang.exe:

image

crypto/fipsmodule/ec/p256.c
Alovchin91: @briansmith  What do you think about including a bigint library? For example boost's [cpp_int](https://www.boost.org/doc/libs/1_76_0/libs/multiprecision/doc/html/boost_multiprecision/tut/ints/cpp_int.html) or [calccrypto](https://github.com/calccrypto/uint128_t)? That could solve all the issues with math at once, and we could only include it on Windows ARM64.

awakecoding:
@briansmith @Alovchin91 I removed the dynamic nasm/clang detection from the current branch. From the build system point of view, it should be good for the current PR (excluding the update to the CI environment). I am done for now, let me know if you need anything, I think the rest has to do with the math and the lack of a uint128_t data type in MSVC for ARM64.

build.rs
briansmith: Disregard this. This was leftover feedback that was building on some other feedback I deleted before posting.

crypto/fipsmodule/bn/internal.h
briansmith: It seems like we need this because MSVC doesn't have `BN_ULLONG` for Aarch64 since that would be a 128-bit type. So no change needed here.

crypto/fipsmodule/ec/p256.c
briansmith: > Okay, it looks like we have a problem then because we don't have the __uint128_t type, and we also don't have a OPENSSL_USE_NISTZ256 implementation.

If I understand what you're saying correctly, there is another PR that adds the fallback `bn_mul_mont` implementation that would be needed for supporting MSVC to work.

HOWEVER, that code will be much slower than using the uint128_t-based arithmetic from Fiat Crypto that using clang would provide.

So, with that in mind, maybe we should first get this working with `TARGET_CC=clang` and then worry about supporting MSVC as the C compiler in a separate PR?


> @briansmith What do you think about including a bigint library? For example boost's cpp_int or calccrypto? That could solve all the issues with math at once, and we could only include it on Windows ARM64.

It would be very unlikely that we'd add an external dependency for the math.

crypto/limbs/limbs.inl
briansmith: There are three aspects of this:
1. Get something working. If we can rely on clang as the C compiler, then this is less much less work than what's in this PR.
2. Get it working with MSVC as the C compiler.
3. Unifying, if practical, the MSVC and non-MSVC code paths.

If we want to have, for expediency, a separate MSVC-only code path for now, then we "just" need to convert your `#define` macros into true inline functions, since we don't do that `#define` stuff except as a last resort.

src/cpu.rs
briansmith: No, because the other code paths also need to be modified if/when we do that.

briansmith:
> @briansmith @Alovchin91 I removed the dynamic nasm/clang detection from the current branch. From the build system point of view, it should be good for the current PR (excluding the update to the CI environment). I am done for now, let me know if you need anything, I think the rest has to do with the math and the lack of a uint128_t data type in MSVC for ARM64.

Thanks! Could you do one last thing? Could you please squash your commits into a single commit and add the following text to the commit message: "I agree to license my contributions to each file under the terms given at the top of each file I changed."? Then I can pick this up and finish it off.

crypto/fipsmodule/ec/p256.c
Alovchin91: One of the main goals for me is to make sure Rustup can compile with rustls-tls feature enabled. If we can somehow specify in build.rs that we want to use clang.exe as the C compiler, and that setting can be dragged into Rustup, I'm generally happy 🙂

Alovchin91: @briansmith Okay, before I change a ton of `#if defined(__GNUC__)` to `#if defined(__GNUC__) || defined(__clang__)` in this context:

```c
#if defined(__GNUC__) || defined(__clang__)
#pragma GCC diagnostic ignored "-Wsign-conversion"
#pragma GCC diagnostic ignored "-Wconversion"
#endif
```

-- is this an expected change? 🙃

briansmith: There is some code that does that in *ring*, because BoringSSL does that. But only code that we share with BoringSSL is supposed to do that, ideally.

For code not shared with BoringSSL, we should be using explicit casts and avoid disabling the warnings.

If you're going to hand this off to me, then go ahead and disable whatever warnings and then I'll fix that.

Alovchin91: I tried to be as conservative as I could when introducing those `#if ... defined(__clang__)` 🙂 It now builds on my SPX and all the tests run. I'm going to try cross-compiling everything from an x64 machine now, and if that succeeds, I think my job here is largely done 🙂 I'd prefer to leave the code itself to you since it likely has nothing to do with Windows ARM64 target but rather with the logic itself, and I don't trust myself there as I have zero knowledge of crypto 😬

crypto/limbs/limbs.inl
Alovchin91: I thought I'd leave the `crypto/fipsmodule/bn/internal.h` changes as is because they're present [upstream in BoringSSL](https://github.com/google/boringssl/blob/47c5f9d2f6c294622baf07c3373de752fd518b03/crypto/fipsmodule/bn/internal.h#L418), but if you'd prefer to revert that file, please let me know 🙂

Alovchin91:
@awakecoding Thanks a LOT for your invaluable help! Without your idea of compiling assembly using Clang we would've never gotten anywhere 🙏

briansmith:
@awakecoding I am not sure when you'll be done updating this or if now you've decided to keep working on it. If/when you are done, please squash the commits and add the contributor statement and mark this PR as ready to review.

awakecoding:
> @awakecoding I am not sure when you'll be done updating this or if now you've decided to keep working on it. If/when you are done, please squash the commits and add the contributor statement and mark this PR as ready to review.

My understanding is there are still a few changes you wanted to see done, but I would like to move one with just a cleaned up version of what is currently on the branch (MSVC + clang.exe to build the assembly). Was there anything else? Otherwise I can work on it some more tomorrow to get it ready for a clean merge, because I would like to get moving quickly on this - ring is the only thing blocking Windows ARM64 in all of my projects right now

Alovchin91:
Gimme 5 minutes and I'll squash everything and add a contributor statement. Do you mind me taking it over @awakecoding ?

briansmith:
> Was there anything else?

Anything you can do to help me do the build, push the built executables to my SPX, and run the tests, would be useful. E.g,:
* Update the documentation with Instructions (ideally a script, but anything) for installing the toolchain needed on a Windows host, assuming that the AAarch64-targetting tools and/or clang aren't installed, if they aren't installed by default.
* Whatever script you use for collecting the built test executables and running them on a SPX (presumably through SSH?).
* Updating .github/workflows/ci.yml to add matrix entries for Windows AAarch64 targets.

Alovchin91:
@briansmith I've squashed, added the contributor statement, and added @awakecoding as co-author. Hope all is correct now 🙂

awakecoding:
> @briansmith I've squashed, added the contributor statement, and added @awakecoding as co-author. Hope all is correct now 🙂
> 
> I suppose additional changes can be addressed in additional PRs? 🤔

Thanks a lot for squashing my commits, I won't be on on my computer until tomorrow. Feel free to finish cleaning it up for a merge, and we can then move on to creating separate PRs for additional improvements.

Feel free to create tickets to track the various planned improvements that have been discussed already and tag me on them, I can particularly be helpful for build-system related tasks.

@briansmith merging this PR will technically add initial support for Windows ARM64, even if we can improve it further. I would like to know your thoughts on what would be sufficient for a release, because in order to fix Windows ARM64 portability, other crates depending on Ring will need to be updated.

I ran cargo tree on my projects and Ring is mostly imported through webpki, so webpki would need a new release once ring is released. Otherwise I would have to backport and patch a few older ring releases to build for Windows ARM64.

awakecoding:
> > Was there anything else?
> 
> Anything you can do to help me do the build, push the built executables to my SPX, and run the tests, would be useful. E.g,:
> * Update the documentation with Instructions (ideally a script, but anything) for installing the toolchain needed on a Windows host, assuming that the AAarch64-targetting tools and/or clang aren't installed, if they aren't installed by default.
> * Whatever script you use for collecting the built test executables and running them on a SPX (presumably through SSH?).
> * Updating .github/workflows/ci.yml to add matrix entries for Windows AAarch64 targets.
> 

Updating the docs on how to install the additional tools would make sense in a separate PR with improvements for better detection of various toolchains (the stuff I removed, and the MSVC clang detection I haven't finished). That's probably the next thing I can get done after the current PR.

Updating the CI workflow to include Windows aarch64 targets is obviously quite an important one.

The last one is remote deployment and execution of the tests, I know it's something you need a hand with, but I would give it a try last on my list. My current top priority is simply getting Windows ARM64 portability fixed for everything depending on Ring in the Rust ecosystem, but I understand we need to get the testing in place.

briansmith:
The main issue is that I want to test this on my Surface Pro X before merging it and before the next release. It is fine if you don't want to do the extra steps involved in assisting with that--what is here is great. But, I try to test everything before I merge it, usually in CI, but in this case I'll have to do it manually.

Alovchin91:
@briansmith  Testing it on SPX would be as simple as running `cargo test` inside x86_arm64 developer command line (`vcvarsx86_arm64.bat`). Or if you use `rustc 1.55 beta` and a manually installed LLVM, you don't even need the developer command line. I'm not sure how the ci.yml would help in this case tbh, unless you want to directly use artifacts from its run?

Alovchin91:
@briansmith I have updated BUILDING.md to include instructions on how to build *ring* on Windows ARM64 (or, in fact, how to prepare the environment for building).

I've also added aarch64-pc-windows-msvc target to ci.yml, but that will still be cross-compilation. I've kept it in a separate commit in case you decide it's too early (you might probably want to setup a runner on your SPX instead).

briansmith:
Thanks. I read the instructions in the README and they look pretty reasonable for somebody building on a ARM64 host. I am building on an x86-64 host. I am hoping to avoid installing development tools onto my Surface Pro X so that it remains as much like an end-user device as possible. I uninstalled MS Visual Studio 2019 from my system and am in the installer. In the "Individual Components," tab what's the minimum I need to check to cross-compile to ARM64 from x86-64?

I'm going to try this and see what happens:
* MSVC v142 - VS 2019 C++ ARM build tools
* C++ Clang Compiler for Windows (12.0.0)
* MSVC v142 - VS 2019 C++ x64/x86 build tools (v14.29-16.10)
* Windows 10 SDK (10.0.16299.0)

Alovchin91:
Yeah, everything's correct except that you need ARM64 build tools (and you don't need ARM tools) 🙂 I'll update the instructions later today to include cross-compiling.

awakecoding:
The MSVC component I used to install the MSVC build tools in my Windows containers is Microsoft.VisualStudio.Component.VC.Tools.ARM64

briansmith:
I granted permission for the GitHub Actions jobs to run. Some of the jobs are failing, including the new test jobs. The `package` job is failing because, AFAICT, it just can't find clang; clang needs to be in the path. The new aarch64 jobs in the matrix are also failing for what looks to be similar reasons.

briansmith:
> Yeah, everything's correct except that you need ARM64 build tools (and you don't need ARM tools) 🙂 I'll update the instructions later today to include cross-compiling.

Thanks. I did switch things from ARM to ARM64. In order to get `vcvarsamd64_arm64.bat` it seems like we also need to install "C++ Universal Windows Platform support for v142 build tools (ARM64)." After I installed that, then from a command prompt I did:
```
"c:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliary\Build\vcvarsamd64_arm64.bat"
set PATH=%PATH%;c:\mozilla-build\msys\bin
cargo test --no-run --target=aarch64-pc-windows-msvc
```

The `set PATH` to add Perl to the path would be different for somebody else, especially somebody who didn't previously work for Mozilla.

Alovchin91:
> Thanks. I did switch things from ARM to ARM64. In order to get `vcvarsamd64_arm64.bat` it seems like we also need to install "C++ Universal Windows Platform support for v142 build tools (ARM64)."

Nope, you don't need that thankfully (it's HUGE). Here's the configuration I use on SPX that includes almost everything:

image

The only thing missing here is Windows SDK because I use the Insider SDK 22000 (as I'm on Windows 11).

I believe those `vsvars...` scripts are part of C++ Build Tools core features. So if you're installing Visual Studio, choosing "Desktop development with C++" workload should be enough.

Unless they broke something in 16.11... 😓

Alovchin91:
> The `package` job is failing because, AFAICT, it just can't find clang; clang needs to be in the path. The new aarch64 jobs in the matrix are also failing for what looks to be similar reasons.

Hmm, I'm explicitly adding C:\msys2\mingw64\bin to PATH as per https://github.com/actions/virtual-environments/issues/2257 as it seems to be the location where clang.exe can be found. Maybe it's msys64 though... 🤔 I'll try to run the pipeline on my fork to see if I can fix it.

awakecoding:
> > The `package` job is failing because, AFAICT, it just can't find clang; clang needs to be in the path. The new aarch64 jobs in the matrix are also failing for what looks to be similar reasons.
> 
> Hmm, I'm explicitly adding C:\msys2\mingw64\bin to PATH as per [actions/virtual-environments#2257](https://github.com/actions/virtual-environments/issues/2257) as it seems to be the location where clang.exe can be found. Maybe it's msys64 though... 🤔 I'll try to run the pipeline on my fork to see if I can fix it.
> 
> EDIT: Yeah, sorry, I'm an idiot, it's msys64 of course 🤦‍♂️

Is it building *inside* msys2? I had the unfortunate pleasure of being forced to use it to automate my libvpx builds recently, and I learned the hard way that you need to translate the paths you add to PATH inside the msys2 environment

awakecoding:
By the way, vcvarsand64_arm64.bat is just a wrapper for "vcvarsall.bat x64_arm64" so if the file is missing it could be worth a shot to call the entrypoint vcvarsall.bat with the x64_arm64 argument instead.

Alovchin91:
In fact, if you're cross-compiling @briansmith, you don't even need the vsvars batch script. It's needed only on Windows ARM64 and only on rustc < 1.55 due to a bug in Visual Studio Installer, but on x64 Windows rustc can find the right tools regardless.

awakecoding:
> In fact, if you're cross-compiling @briansmith, you don't even need the vsvars batch script. It's needed only on Windows ARM64 and only on rustc < 1.55 due to a bug in Visual Studio Installer, but on x64 Windows rustc can find the right tools regardless.

I do think he's cross-compiling here, this is definitely not being built *on* the Surface Pro X :)

Alovchin91:
Okay, I have updated the BUILDING.md to include all the info and fixed the msys64 path in ci.yml 🙂 Let's see if it builds now

briansmith:
> > Thanks. I did switch things from ARM to ARM64. In order to get vcvarsamd64_arm64.bat it seems like we also need to install "C++ Universal Windows Platform support for v142 build tools (ARM64)."
> 
> Nope, you don't need that thankfully (it's HUGE).

OK, when I have time I'll uninstall VS 2019 build tools and try a more minimal configuration and see if it still works.

> In fact, if you're cross-compiling @briansmith, you don't even need the vsvars batch script. It's needed only on Windows ARM64 and only on rustc < 1.55 due to a bug in Visual Studio Installer, but on x64 Windows rustc can find the right tools regardless.

Thanks. I verified that that is the case by starting a new command prompt. Now I just need to do this, in a new command prompt:

```
set PATH=%PATH%;c:\mozilla-build\msys\bin;c:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\Llvm\bin
cargo test --no-run --target=aarch64-pc-windows-msvc
```

briansmith:
Now I'm writing a script that will let me copy the test executable files through SSH from my desktop to my SPX and then run all the test executables remotely via SSH on the SPX and succeed/fail depending on whether all the tests succeed/fail.

Alovchin91:
Great, right at the moment when I was planning to test the CI workflow, GitHub Actions turned out to be broken 🤦‍♂️

image

I guess I'll go to sleep (it's 12am on this side of Atlantic 😛) and will continue with the CI workflow tomorrow.

crypto/fipsmodule/bn/internal.h
```diff
@@ -185,8 +185,13 @@ void bn_mul_mont(BN_ULONG *rp, const BN_ULONG *ap, const BN_ULONG *bp,
 
 static inline void bn_umult_lohi(BN_ULONG *low_out, BN_ULONG *high_out,
                                  BN_ULONG a, BN_ULONG b) {
-#if defined(OPENSSL_X86_64) && defined(_MSC_VER) && !defined(__clang__)
+#if defined(_MSC_VER) && !defined(__clang__)
```
Alovchin91, line 188: This set of conditions breaks compilation on x86 under MSVC 🤦‍♂️ I'll revert this file for now since we're using Clang for compiling under Windows ARM64 anyway.

.github/workflows/ci.yml
```diff
@@ -301,6 +307,10 @@ jobs:
           sudo xcode-select -s /Applications/Xcode_12.4.app &&
           sudo rm -Rf /Library/Developer/CommandLineTools/SDKs/*
 
+      # https://github.com/actions/virtual-environments/issues/2257
+      - if: ${{ matrix.target == 'aarch64-pc-windows-msvc' }}
+        run: echo "C:\msys64\mingw64\bin" >> $GITHUB_PATH
```
Alovchin91, line 312: Good news: there's Visual Studio Clang available on windows-latest!

Alovchin91:
@briansmith Would you prefer to use https://github.com/ilammy/msvc-dev-cmd GitHub Action to get developer command prompt, or to simply specify the full path to Clang which would be `C:\Program Files\Microsoft Visual Studio\2019\Enterprise\VC\Tools\Llvm\x64\bin`? (I'm talking about ci.yml here)

briansmith:
OK, I now have SSH server running on my Surface Pro X and I can `scp` files to it. To do this, I created a new non-admin user and then created ~/.ssh/authorized_keys as normal. I modified C:\ProgramData\ssh\sshd_config to disable password authentication and enable public key authentication. With this, I can SSH in and scp files to the device.

However, when I try to run one of the generated test executables on the device, through SSH, I get no output. I logged in locally as the non-admin user I created on the Surface Pro X and ran the program locally. I got a dialog box that says "The code execution cannot proceed because VCRUNTIME140.dll was not found. Reinstalling the program may fix this problem."

Is this expected? Are real-world applications expected to distribute VCRUNTIME140.dll? Or is something wrong with how this is being built?

I suspect people don't notice this if they're executing the build locally on the Surface Pro X since presumably the VS tools install that DLL.

Alovchin91:
Yeah, this is expected. VCRUNTIME140.dll is part of Visual C++ Redistributable package. Basically anything that's built with an *-msvc toolchain will have this dependency, regardless of the architecture. You can download and install it from [here](https://support.microsoft.com/en-us/topic/the-latest-supported-visual-c-downloads-2647da03-1eea-4433-9aff-95f26a218cc0).

Redistribution is a different topic 🙂 When you're redistributing the application, you have a choice. You can redistribute it as an installer dependency (so the setup program downloads and installs it). Or you can redistribute the DLL in your program's directory, this is permitted. Or, and that might be a better choice for your specific case, you can compile the program with [`-C target-feature=+crt-static`](https://rust-lang.github.io/rfcs/1721-crt-static.html), and then you don't have a dependency, you just link it statically.

If you want, I can add that to the CI tomorrow, maybe even for all Windows targets.

Alovchin91:
Okay, so it builds now: https://github.com/Alovchin91/test-ring-ci/actions/runs/1118176199

I'm going to wrap up for today, it's 2am here 🙈

briansmith:
> However, when I try to run one of the generated test executables on the device, through SSH, I get no output. I logged in locally as the non-admin user I created on the Surface Pro X and ran the program locally. I got a dialog box that says "The code execution cannot proceed because VCRUNTIME140.dll was not found. Reinstalling the program may fix this problem."
> 
> Is this expected? Are real-world applications expected to distribute VCRUNTIME140.dll? Or is something wrong with how this is being built?

It seems like this is indeed normal. So for now I'll have my script copy vcruntime140.dll to the device too, once I figure out which file in the SDK to copy.

Alovchin91:
> So for now I'll have my script copy vcruntime140.dll to the device too, once I figure out which file in the SDK to copy.

It would be in the `C:\Program Files (x86)\Microsoft Visual Studio\2019\<edition>\VC\Redist\MSVC\14.2x.xxxxx\arm64\Microsoft.VC142.CRT` folder.

awakecoding:
@Alovchin91 @briansmith I've got a MUCH easier fix for you guys: just set the RUSTFLAGS environment variable to "-C target-feature=+crt-static" before calling cargo, and magic will happen with regards to the MSVC runtime

awakecoding:
Of course you need to be careful about everything that manually builds C code and make sure the right flags are being passed to the compiler, a lot of build.rs scripts mess this up which is why I precompile them for my needs. Getting the static MSVC runtime everywhere is the story of my life: https://docs.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-160

briansmith:
> @briansmith Would you prefer to use https://github.com/ilammy/msvc-dev-cmd GitHub Action to get developer command prompt

I prefer to avoid third-party GitHub Actions for security reasons.

> or to simply specify the full path to Clang which would be `C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\Llvm\x64\bin`? (I'm talking about ci.yml here)

It isn't `Enterprise` on my machine :) On my machine it's "BuildTools".

I think we should just hard-code that path in the ci.yml for now.

The design of the build system and CI/CD for *ring* is optimized for making it easy to run locally what is run in CI/CD. That is why we have the relatively complicated mk/cargo.sh for non-Windows targets. Until now we'd not had to do anything non-trivial for Windows so Windows has been a bit different.

The other design aspect of the *ring* build system is that we don't require unusual tools to be installed. This is why we package generated files within the Cargo crate in release. I'll be making sure that the `package` step does the right thing for Windows targets here (aarch64 and x86 and x86-64) w.r.t. this before I do the release. Then people won't have to worry about having clang as an assembler or having to check the "clang" box during install. This will require that we do complete the follow-up work to get the C code compiling with MSVC for aarch64 before the release.

Alovchin91:
> It isn't `Enterprise` on my machine :) On my machine it's "BuildTools".

It's Enterprise on GitHub Actions :D

Okay, I got you, done.

awakecoding:
The cc crate has the static_crt option (which may be automatically mapped, not sure): https://docs.rs/cc/1.0.69/cc/struct.Build.html#method.static_crt

To check if you have correctly built with the static MSVC runtime you can use PPEE to check the import table, and make sure the vcruntime140.dll is missing: https://www.mzrst.com/

Alovchin91:
I've enabled static CRT (`-C target-feature=+crt-static`) on ARM64 build. Let me know if you want it on all Windows builds :)

Now I'm really going to sleep, I promise 🤣

Alovchin91:
@briansmith I've checked the pipeline, it should run without issue now. Please set it off when you're ready 🙂

crypto/curve25519/curve25519.c
```diff
@@ -31,7 +31,7 @@
 #pragma warning(disable: 4244)
 #endif
 
-#if defined(__GNUC__)
```
briansmith, line 34: Are these changes really needed? This code has been compiled by clang for a long time without complaint...

Alovchin91:
Damn, my bad, forgot to copy the last fix 🤦‍♂️  Tomorrow

crypto/curve25519/curve25519.c
Alovchin91: Yeah, so it turns out we're the first ones to compile this code with Clang on Windows 😅

Clang does its best to be compatible with other compilers. On Windows it does its best to be [compatible with MSVC](https://clang.llvm.org/docs/MSVCCompatibility.html), while on Unix it [builds the code in GNU mode](https://clang.llvm.org/compatibility.html#c) by default (whatever that means).

So in fact, if we replaced this line with `#if defined(__GNUC__) || defined(_MSC_VER)`, it would still compile! Also this is the exact reason why a few lines earlier there is `#if defined(_MSC_VER) && !defined(__clang__)`.

So in the end, the answer is "yes", we need these changes because Clang on Windows doesn't define __GNUC__, and then we cannot compile 🤷‍♂️

awakecoding:
Using a hardcoded path to add MSVC clang.exe is fine (don't change it) but for the record, here is the cleanest solution I could find after asking the question in the [vswhere GitHub repo](https://github.com/microsoft/vswhere/issues/242). This is something I could implement in a separate PR with improved detection of tools on the host.

```
$VSINSTALLDIR = $(vswhere.exe -latest -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath)
$VCINSTALLDIR = Join-Path $VSINSTALLDIR "VC"
$LLVM_ROOT = Join-Path $VCINSTALLDIR "Tools\Llvm\x64"
$Env:PATH += ";$LLVM_ROOT\bin"
```

image

awakecoding:
@briansmith I was thinking of backporting these patches to ring 0.16.20 such that I could use [patch] in my cargo.toml and fix existing dependencies before they are updated, but I couldn't find the match tag or commit on git. Is the one where the cargo.toml version number was bumped the right commit?

awakecoding:
Meh... forget a 0.16.20 backport, there's a lot of stuff that changed in the build system since then. I tried and failed, so instead I decided to make a fake 0.16.20 with the current branch, crossing fingers that the API didn't change too much for unsuspecting crates. If it weren't for webpki it might have worked :( it'll need an update after this PR, I'm getting a bunch of type mismatches:

image

If at the beginning of the week I had other Windows ARM64 blockers, I am now down to just Ring on my todo list, and it's a tough one because it's reused in many dependencies that will need to be updated as well. Are we close to a possible merge?

awakecoding:
@Alovchin91 I've been banging my head with weird winapi errors trying to build ring from this branch while imported from other projects, until I figured out that we didn't add the required winapi module in the features. If you're lucky I guess cargo can sometimes enable the corresponding feature due to another dependency, but it can lead to this:

```
error[E0433]: failed to resolve: could not find `processthreadsapi` in `um`
   --> C:\Users\mamoreau\.cargo\git\checkouts\ring-9b33fe7ed49b18d2\2f9ea83\src\cpu.rs:181:25
    |
181 |             winapi::um::processthreadsapi::IsProcessorFeaturePresent(
    |                         ^^^^^^^^^^^^^^^^^ could not find `processthreadsapi` in `um`
```

I pushed a commit on this branch to add the missing winapi feature + bumped it from 0.3.8 to 0.3.9

Alovchin91:
@awakecoding Thanks! 👍 I'll squash it with the previous commit if @briansmith is happy 🙂

Cargo.toml
```diff
@@ -176,7 +176,7 @@ once_cell = { version = "1.5.2", default-features = false, features=["std"] }
 web-sys = { version = "0.3.37", default-features = false, features = ["Crypto", "Window"] }
 
 [target.'cfg(target_os = "windows")'.dependencies]
-winapi = { version = "0.3.8", default-features = false, features = ["ntsecapi", "wtypesbase"] }
```
briansmith, line 179: Shouldn't we add this feature only for aarch64 targets? I think this would require adding `resolver = "2"` to Cargo.toml too?

awakecoding: Adding a single winapi module for all Windows targets is no big deal, what may happen is we'd forget to re-enable it if we use it on Windows x64. Is it even possible to tweak the Cargo.toml to limit just one such feature to Windows ARM64? As for the resolver=2, I do not know what you are talking about

briansmith:
I don't think this will be backportable to 0.16.x because, IIRC, there were BoringSSL changes picked up since then that are needed to support aarch64-pc-windows-msvc targets, which aren't in 0.16.x. I intend to release 0.17 very soon, along with new webpki. I will also help update Rustls. So hopefully 0.16.x is imminently obsolete.

I looked into BoringSSL's support for aarch64-pc-windows-msvc and they seem to only support this target when clang is used as the compiler--presumably both for C and assembly. It isn't clear whether or not they found ABI-related issues with trying to use MSVC + the aarch64 assembly code. I will ask them about it. But for now, it's prudent to assume that clang must be used as the C compiler and the assembler.

With that in mind, I have a series of PRs to support clang as the C compiler for -msvc targets: First, x86_64 and i686, and then aarch64 can be done on top of those changes. The first two PRs are #1343 and #1344. Once those two PRs are reviewed and merged, I will post the third one that I've verified enables clang-cl and clang to work for Windows for x86_64 and i686 by setting the `CC` environment variable prior to invoking `cargo`.

After that, I believe the changes to support specifically aarch64-pc-windows-msvc using clang as the compiler will be minimal, which is good, because we don't have any good way of testing this target in CI/CD yet.

Then I'll verify that the `package` mechanism of pre-assembly the assembly code for Windows works in the way we expect it to. I'll probably need to publish a pre-release of 0.17.0 to fully verify this.

As for supporting MSVC as the compiler for aarch64-pc-windows-msvc, that's highly desirable, but like I mentioned above, I need to talk to BoringSSL people to see if they know of any reason it would be problematic and/or what would need to be done to make that work.

Cargo.toml
briansmith: > Adding a single winapi module for all Windows targets is no big deal

I agree.

> Is it even possible to tweak the Cargo.toml to limit just one such feature to Windows ARM64? As for the resolver=2, I do not know what you are talking about

Yes, starting in Rust 1.151: https://doc.rust-lang.org/cargo/reference/features.html#feature-resolver-version-2

briansmith: I am fine to leave this the way you did it

Alovchin91:
It turns out #1343 introduced some conflicts with this story. Shall we maybe go ahead and merge it if everyone's happy before we continue working on build.rs? 🤔

Alovchin91:
I've rebased to resolve conflicts from #1343 and squashed the commits.

build.rs
```diff
@@ -552,6 +567,16 @@ fn cc(
     let is_musl = target.env.starts_with("musl");
 
     let mut c = cc::Build::new();
+
+    if target.os == WINDOWS && target.arch == AARCH64 {
+        let _ = c.compiler("clang-cl");
```
Alovchin91, line 572: @awakecoding @briansmith Gentlemen, please review this proposal. This would allow us to avoid Windows AArch64 branching in most cases. The additional flags to disable some warning are due to the fact that clang and clang-cl set different default warnings. BoringSSL sets `-W3` flag for clang-cl instead of `-Wall`, but in our case the `/Wall` flag is part of a static array that I didn't want to refactor, so I just disabled those warnings that prevent the code from compiling.

briansmith:
> It turns out #1343 introduced some conflicts with this story. Shall we maybe go ahead and merge it if everyone's happy before we continue working on build.rs? 🤔

My goal with the other PRs I have submitted (and will submit) is to reduce this PR to just the cpu.rs changes and the CI/CD changes, or close to that. The work I've been doing yesterday and today is to clarify build.rs in response to some of the parts of this PR that I found confusing or where I want to do things a different way. For example, my change to always use ".o" for the object file extension was done after having difficulty reviewing the logic regarding that in this PR. As another example, instead of having to create a separate `clang()` function in build.rs to force compilation with clang, we can use the existing `cc()` function so that Windows Aarch64 and other non-Windows assembly stuff is processed exactly the same way.

If you want to keep rebasing this PR on top of my other ones, go ahead. Or if you want to wait until the other changes are done, that could be less work.

build.rs
briansmith: > @awakecoding @briansmith Gentlemen, please review this proposal. This would allow us to avoid Windows AArch64 branching in most cases. The additional flags to disable some warnings are due to the fact that clang and clang-cl set different default warnings. BoringSSL sets `-W3` flag for clang-cl instead of `-Wall`, but in our case the `/Wall` flag is part of a static array that I don't want to refactor, so I just disabled those warnings that prevent the code from compiling. I expect @briansmith's work on supporting clang-cl for all platforms to work around this issue in a better way.

Indeed, I've already prepared a PR that enables compiling x86_64 and i686 Windows with clang and clang-cl, which deals with the warnings, which will obviate (I hope) the need for any such warning-related changes specific to AAarch64.

I am hoping that with my other changes, the user can just set `TARGET_CC=clang` or `TARGET_CC=clang-cl`, and that they'll have the freedom to choose between either clang or clang-cl or whatever other compiler they want (that supports uint128_t), when targeting AAarch64, without having to hard-code anything into build.rs, or where there's as little hard-coded into build.rs as possible.

awakecoding:
> > It turns out #1343 introduced some conflicts with this story. Shall we maybe go ahead and merge it if everyone's happy before we continue working on build.rs? 🤔
> 
> My goal with the other PRs I have submitted (and will submit) is to reduce this PR to just the cpu.rs changes and the CI/CD changes, or close to that. The work I've been doing yesterday and today is to clarify build.rs in response to some of the parts of this PR that I found confusing or where I want to do things a different way. For example, my change to always use ".o" for the object file extension was done after having difficulty reviewing the logic regarding that in this PR. As another example, instead of having to create a separate `clang()` function in build.rs to force compilation with clang, we can use the existing `cc()` function so that Windows Aarch64 and other non-Windows assembly stuff is processed exactly the same way.
> 
> If you want to keep rebasing this PR on top of my other ones, go ahead. Or if you want to wait until the other changes are done, that could be less work.

I was worried that cleanup in other PRs for specific things like the cc handling of clangs, and clang-related improvements, would ultimately create conflicts with the current PR. Can we keep it a minimum such that we can merge this PR sooner than later, and complete the improvements after the current PR has been merged? It would likely result in a lot less conflicts.

briansmith:
> I was worried that cleanup in other PRs for specific things like the cc handling of clangs, and clang-related improvements, would ultimately create conflicts with the current PR. Can we keep it a minimum such that we can merge this PR sooner than later, and complete the improvements after the current PR has been merged? It would likely result in a lot less conflicts.

There are parts of this PR that I don't want to maintain, and which are hard for me to verify the correctness of, including the `clang()` function separate from the `cc()` function for example.

Keep in mind that most people who contribute to *ring* are not interested in Windows AAarch64, so we have to optimize for those people, by minimizing the amount of Windows-AAarch64-specific logic.

In any case, I expect the work to be done today.

Alovchin91:
@briansmith  I'm sorry, I've made a comment about conflicts after I saw an email about #1343 but before I read your previous message 😅 Please go ahead, I totally support your argument that supporting Windows AArch64 should be a no-brainer compared to other architectures.

> As another example, instead of having to create a separate `clang()` function in build.rs to force compilation with clang, we can use the existing `cc()` function so that Windows Aarch64 and other non-Windows assembly stuff is processed exactly the same way.

I was hoping that would be possible, indeed 😄

@briansmith My only concern is that I hope that we could merge this PR before we enable compilation with MSVC, so there still will be some hardcoded `c.compiler("clang-cl")` code. Is that okay?

I'm also not sure if you can support both clang and clang-cl without branching between the two, since AFAIK clang-cl exists specifically to mimic MSVC's command line arguments.

Alovchin91:
_Discussion from #1346:_

> but whether it's in `$PATH` or not, or what that compiler is named (`clang` vs `clang-12`, for example) wouldn't be hard-coded in.

I'm afraid that would make usage of *ring* unnecessarily over-complicated on Windows AArch64. I'd probably still suggest to hardcode `clang-cl` for now, for better compatibility and for simpler usage.

At the same time, if you don't want to release a new *ring* version before it's compilable under MSVC on AArch64, and given that one still needs additional setup (Clang in the $PATH), maybe it doesn't really matter much. 🤔

I guess I don't have a strong opinion but am slightly inclined to hardcode `clang-cl` with a `FIXME` comment.

awakecoding:
> _Discussion from #1346:_
> 
> > but whether it's in `$PATH` or not, or what that compiler is named (`clang` vs `clang-12`, for example) wouldn't be hard-coded in.
> 
> I'm afraid that would make usage of _ring_ unnecessarily over-complicated on Windows AArch64. I'd probably still suggest to hardcode `clang-cl` for now, for better compatibility and for simpler usage.
> 
> At the same time, if you don't want to release a new _ring_ version before it's compilable under MSVC on AArch64, and given that one still needs additional setup (Clang in the $PATH), maybe it doesn't really matter much. 🤔
> 
> I guess I don't have a strong opinion but am slightly inclined to hardcode `clang-cl` with a `FIXME` comment.

I suggest we just make sure that it can be built for Windows ARM64, and then we could make further clang detection and flexibility improvements in a separate PR to handle clang vs clang-cl, and clang-12 vs clang, etc, if need be.

Alovchin91:
> and then we could make further clang detection and flexibility improvements in a separate PR to handle clang vs clang-cl, and clang-12 vs clang, etc, if need be.

@awakecoding To be honest, I don't think *ring* should know anything about where to find clang and which clang to use. I'd remove any knowledge of these specifics as soon as MSVC support is there. My suggestion to default to clang is just because I think it's easier to comprehend "clang-cl was not found" error message than something cryptic about int128_t 😅

> Maybe we can do something like "if the target arch is aarch64 and the env is msvc and no compiler has been set then default to whataever clang there is on the path." In the products I work on we will set TARGET_AARCH64_PC_WINDOWS_MSVC_CC or similar to a specific version of clang.

> I suggest the following course of action, which seems ideal to me:
PR 1. Merge this.
PR 2. Make a PR to get the build working when $TARGET_AARCH64_PC_WINDOWS_MSVC_CC points to VS2019's clang, either a full path or "clang" when it is in $PATH.
PR 3. Change the code to compile with VS2019's cl.exe (or clang or clang-cl).

@briansmith Sounds reasonable to me 🙂

build.rs
```diff
@@ -361,11 +368,20 @@ fn pregenerate_asm_main() {
...
             if !std::mem::replace(&mut generated_prefix_headers, true) {
                 generate_prefix_symbols_nasm(&pregenerated, &ring_core_prefix()).unwrap();
+                // We also need to generate prefix_symbols_asm.h for Windows ARM64
+                generate_prefix_symbols_header(
+                    &pregenerated,
+                    "prefix_symbols_asm.h",
+                    '#',
+                    None,
+                    &ring_core_prefix(),
+                )
```
briansmith, line 378: If I understand correctly, there are exactly two kinds of header that suffice for all targets: One for Nasm and one for all other cases. It would be good to DRY this up so that we don't have to repeat the constants "prefix_symbols_asm.h" and `#`, and so we don't need to use a different parameter for the `prefix_condition` parameter.

build.rs
```diff
@@ -524,10 +539,10 @@ fn compile(p: &Path, target: &Target, warnings_are_errors: bool, out_dir: &Path)
     } else {
         let mut out_path = out_dir.join(p.file_name().unwrap());
         assert!(out_path.set_extension(target.obj_ext));
-        let cmd = if target.os != WINDOWS || ext != "asm" {
+        let cmd = if target.os != WINDOWS || (ext != "asm" && ext != "S") {
             cc(p, ext, target, warnings_are_errors, &out_path, out_dir)
         } else {
-            nasm(p, &target.arch, &out_path, out_dir)
+            win_asm(p, &target.arch, &out_path, out_dir)
```
briansmith, line 545: I think we can now revert these changes.

build.rs
```diff
@@ -552,6 +567,16 @@ fn cc(
...
+
+    if target.os == WINDOWS && target.arch == AARCH64 {
+        let _ = c.compiler("clang-cl");
+        // clang-cl sets different default warnings than clang.
+        // See https://boringssl.googlesource.com/boringssl/+/refs/heads/3538/CMakeLists.txt#57
+        let _ = c.flag("-Wno-missing-prototypes");
+        let _ = c.flag("-Wno-unused-macros");
+        let _ = c.flag("-fmsc-version=1900");
+    }
+
```
briansmith, line 579: I think we can now revert these changes?

briansmith (changes requested):

Alovchin91:
I can't seem to build it with `clang-cl` anymore:

```
  running "clang-cl" "-nologo" "-MD" "-Z7" "-Brepro" "--target=aarch64-pc-windows-msvc" "-I" "include" "-I" "pregenerated" "-W4" "-fmsc-version=1900" "/GS" "/Gy" "/EHsc" "/GR-" "/Zc:wchar_t" "/Zc:forScope" "/Zc:inline" "/Zc:rvalueCast" "/sdl" "/Wall" "/wd4127" "/wd4464" "/wd4514" "/wd4710" "/wd4711" "/wd4820" "/wd5045" "/Od" "/RTCsu" "-DNDEBUG" "-WX" "-c" "/Fopregenerated\\aesv8-armx-win64.o" "pregenerated\\tmp\\aesv8-armx-win64.S"
  In file included from pregenerated\tmp\aesv8-armx-win64.S:14:
  include\ring-core/arm_arch.h(104,9): error: macro name is a reserved identifier [-Werror,-Wreserved-id-macro]
  #define __ARM_MAX_ARCH__ 8
          ^
  1 error generated.
```

Alovchin91:
@briansmith Please take a look at the changes I've made to `build.rs`. As mentioned earlier, I cannot compile the code with `clang-cl` now (for whatever reason) so maybe I'll try to figure out why later.

Alovchin91:
> I cannot compile the code with `clang-cl` now (for whatever reason) so maybe I'll try to figure out why later.

Hmm, forget about `clang-cl` I guess. The reason why I cannot compile with it now is because now it's also being used to compile assembly code which was not the case before. I'm not sure if we should worry about it or not.

Alovchin91:
Okay, so unless we care about the ability to compile assembly code with `clang-cl`, we should be good to go now @briansmith

awakecoding:
> Okay, so unless we care about the ability to compile assembly code with `clang-cl`, we should be good to go now @briansmith

I don't think we care about building the assembly code with clang-cl, mostly because we really intend to use clang for it, and I'm not even sure that clang-cl can compile the assembly code in question with the MSVC-style arguments.

Alovchin91:
> I don't think we care about building the assembly code with clang-cl, mostly because we really intend to use clang for it, and I'm not even sure that clang-cl can compile the assembly code in question with the MSVC-style arguments.

I'm not even sure it tries to compile it using a cl-driver when it sees a `.S` file 😅
In any case, if I assume pre-generated ASM, both `clang-cl` and `clang` now work for Windows AArch64 target 🙂

Alovchin91:
I must admit I'm pretty happy with the result! 🙂

briansmith:
> BTW could you please also upgrade cc to 1.0.69? This version contains fixes to find MSVC tools on Windows ARM.

Done in PR #1354.

briansmith:
Here's how I tested this on the device:

1. I copied
`C:\Program\ Files\apps-x86\Microsoft\ Visual\ Studio/2019\BuildTools\VC\Redist\MSVC\14.29.30133\arm64\Microsoft.VC142.CRT\vcruntime140.dll` into the working copy of the device via scp.

2. 
```
rm -Rf target/aarch64-pc-windows-msvc/release
cargo test --no-run --release --target=aarch64-pc-windows-msvc
find target/aarch64-pc-windows-msvc/release -name "*.exe" | xargs -i scp "{}" username@surfaceprox:
ssh username@surfaceprox
```

On the Surface Pro X:
```
for %i in (*.exe) do %i
```

I also tested the debug builds in an analogous way.

awakecoding:
> Here's how I tested this on the device:
> 
> 1. I copied
>    `C:\Program\ Files\apps-x86\Microsoft\ Visual\ Studio/2019\BuildTools\VC\Redist\MSVC\14.29.30133\arm64\Microsoft.VC142.CRT\vcruntime140.dll` into the working copy of the device via scp.
> 
> ```
> rm -Rf target/aarch64-pc-windows-msvc/release
> cargo test --no-run --release --target=aarch64-pc-windows-msvc
> find target/aarch64-pc-windows-msvc/release -name "*.exe" | xargs -i scp "{}" username@surfaceprox:
> ssh username@surfaceprox
> ```
> 
> On the Surface Pro X:
> 
> ```
> for %i in (*.exe) do %i
> ```
> 
> I also tested the debug builds in an analogous way.

Nice! If you want to get rid of the annoying vcruntime140.dll, you can use RUSTFLAGS='-C target-feature=+crt-static' as [documented here](https://rust-lang.github.io/rfcs/1721-crt-static.html#customizing-linkage-to-the-c-runtime).

briansmith:
@Alovchin91, @awakecoding Thanks for all your effort on this! I especially appreciate the effort you put in after the extra PR to make this easier to maintain (I hope) for people who don't have access to an aarch64-pc-windows-msvc device.

Alovchin91:
@briansmith Thanks for all your help along the way! 😊 That was a lot of fun gentlemen 😀

awakecoding:
@briansmith it's been a year to the day that these Windows ARM64 build fixes are been merged on the main branch, but no release has been done since then, and we're still unable to build some of our Rust dependencies for all platforms because of this. We're getting closer to a first possible Windows ARM64 build of Remote Desktop Manager, so it's becoming a more pressing issue that we get a working build.

I have tried backporting the fixes back on top of 0.16.20 (there is no release tag on GitHub, so finding the right commit is tricky) and force it to work for Windows ARM64, but that version of build.rs was already quite different from the version on top of which the fixes were made. I wasted a few hours on it, and I'm stick figuring out how the static library is built assuming a .a extension, and why it still wants to build a bunch of linux64 assembly files.

I tried using the version of ring right after this pull request got merged, and edited Cargo.toml to make a fake 0.16.20 release out of it, such that it would get picked up by other crates like webpki, but then I hit "expected struct `Reader`, found struct `untrusted::reader::Reader`" issues. The latest ring release uses untrusted 0.9.0, while webpki is still on 0.7.1, and trying to change it to 0.9.0 causes a whole lot of other issues that I'm not familiar with.

What I need is a version of ring that can be built for Windows ARM64 *and* webpki, and I guess this means updating webpki to an newer version of the untrusted crate matching the latest version of ring. What would be the simplest way to get this unblocked? Is a ring 0.16.20 backport the only option, or is a new release of compatible crates coming soon?

tr3ysmith:
@briansmith What can we do to get a new release out for the main branch? Rust is basically unusable on Windows ARM I you're relying on major libraries like Reqwest, Rustls, OAuth, etc.

Issue #1234, opened 2021-03-24
Remove `spin` dependency in favor of `once_cell::race::OnceBool` for no_std environments.


briansmith:
Switching to `once_cell::race` would allow us to eliminate `spin` as a dependency. We don't need or really want a spin lock anyway,; the race semantics seem better.

PR #1340, opened 2021-08-10
Remove spin dependency in favor of once_cell::race::OnceBool for no_s…

chrismiceli:
…td environments.

remove spin dependency
use race from existing dependency

### DIFF
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -35,9 +35,9 @@ pub(crate) fn features() -> Features {
         )
     ))]
     {
-        use once_cell::race::OnceBool;
+        use once_cell::sync::OnceCell;
 
-        static INIT: OnceBool = OnceBool::new();
+        static INIT: OnceCell<bool> = OnceCell::new();
         let _ = INIT.get_or_init(|| {
             #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
             {
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -35,8 +35,10 @@ pub(crate) fn features() -> Features {
         )
     ))]
     {
-        static INIT: spin::Once<()> = spin::Once::new();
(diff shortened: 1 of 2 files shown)


chrismiceli:
FYI @briansmith, this is for https://github.com/briansmith/ring/issues/1234.

chrismiceli:
> 
> 
> @chrismiceli Thanks. Did you review the once_cell::race implementation to ensure it implements a _full_ barrier that would protect the implicit initialization of `OPENSSL_ia32cap_P` and/or `OPENSSL_armcap_P` in addition to the bool value?

No, but now I read the documentation more clearly and switched to sync::OnceCell as it is described as such:
https://github.com/matklad/once_cell/blob/master/src/lib.rs#L946
whereas the original race::OnceCell explicitly says the init function can be called multiple times:
https://github.com/matklad/once_cell/blob/master/src/race.rs#L51

src/cpu.rs
```diff
@@ -35,8 +35,10 @@ pub(crate) fn features() -> Features {
         )
     ))]
     {
-        static INIT: spin::Once<()> = spin::Once::new();
-        let () = INIT.call_once(|| {
+        use once_cell::sync::OnceCell;
+
+        static INIT: OnceCell<bool> = OnceCell::new();
+        let _ = INIT.get_or_init(|| {
```
briansmith, line 41: This will not work for a `no_std` environment, though, right?

I think the right thing to do is use `std::sync::Once` or `OnceCell` for environments that have `libstd` but something else for `no_std` environments.

briansmith (changes requested):

chrismiceli:
There is more history here than I was aware of when I started. After reading the issues and previous commits related to this, I think I'm in over my head and will close this PR for the moment.

briansmith:
@chrismiceli Thanks for looking into this!

PR #1343, opened 2021-08-13
build.rs: Use cc-rs's built-in logic for warnings-as-errors.

briansmith:
Take a step towards supporting Clang as the compiler for -msvc targets.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -614,12 +614,7 @@ fn cc(
     }
 
     if warnings_are_errors {
-        let flag = if &target.env != "msvc" {
-            "-Werror"
-        } else {
-            "/WX"
-        };
-        let _ = c.flag(flag);
+        c.warnings_into_errors(true);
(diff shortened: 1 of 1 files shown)


briansmith:
@awakecoding @Alovchin91 Please review this. This is one of many changes I'm making to officially support Clang for -msvc targets.

awakecoding:
It looks good to me, I guess cc automatically detects the compiler and sets the correct flags.   Is there a way to tell cc to use clang/MSVC and not the regular MSVC compiler? And if so, does it use clang-cl (same flags as cl.exe) or clang.exe (same flags as clang on all platforms)? Did you have a way to check which compiler was picked up by cc in the end?

briansmith:
The codecov job's failure is caused by there being less covered code, not by there be any new uncovered code, so it's benign (good even).

PR #1344, opened 2021-08-13
build.rs: Always use ".o" for object files.

briansmith:
Take a step towards supporting clang as the C compiler for -msvc
targets.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -273,7 +273,6 @@ const WINDOWS: &str = "windows";
 
 const MSVC: &str = "msvc";
 const MSVC_OBJ_OPT: &str = "/Fo";
-const MSVC_OBJ_EXT: &str = "obj";
 
 /// Read an environment variable and tell Cargo that we depend on it.
 ///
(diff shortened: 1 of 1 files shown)


briansmith:
@awakecoding @Alovchin91 Please review this. This is one of many changes I'm making to officially support Clang for -msvc targets. This eliminates the confusion about which object file extension is to be used for -msvc targets; it doesn't need to vary based on the compiler.

build.rs
```diff
@@ -535,9 +525,12 @@ fn compile(p: &Path, target: &Target, warnings_are_errors: bool, out_dir: &Path)
...
 }
 
-fn obj_path(out_dir: &Path, src: &Path, obj_ext: &str) -> PathBuf {
+fn obj_path(out_dir: &Path, src: &Path) -> PathBuf {
     let mut out_path = out_dir.join(src.file_name().unwrap());
-    assert!(out_path.set_extension(obj_ext));
+    // To eliminate unnecessary conditional logic, use ".o" as the extension,
+    // even when the compiler (e.g. MSVC) would normally use something else
+    // (e.g. ".obj"). cc-rs seems to do the same.
+    assert!(out_path.set_extension("o"));
```
Alovchin91, line 533: I hope this is fine to do. When I was trying to compile *ring* for Windows AArch64, clang refused to accept any extension other than `.S` for assembly files 🤷‍♂️ Hopefully MSVC is not so opinionated.

briansmith: > I hope this is fine to do. When I was trying to compile _ring_ for Windows AArch64, clang refused to accept any extension other than `.S` for assembly files 🤷‍♂️ Hopefully MSVC is not so opinionated.

In GCC/clang land, ".S" means "use the preprocessor, ".s" means "don't use the preprocessor, so that's kind of a different issue.

PR #1346, opened 2021-08-13
build.rs: Refactor preassembly in `package` process in preparation for aarch64-pc-windows-msvc.

briansmith:


briansmith:
@awakecoding @Alovchin91 Please review this. This generalizes the logic in the `package` process so that it can use the C compiler to assemble code instead of assuming that `nasm` should be used. for aarch64-pc-windows-msvc, you'd still need to add the `AsmTarget` entry and do some minimal changes, but I think this will allow you to get rid of `clang()` in your PR.

build.rs
```diff
@@ -348,13 +348,28 @@ fn pregenerate_asm_main() {
...
             if !std::mem::replace(&mut generated_prefix_headers, true) {
                 generate_prefix_symbols_nasm(&pregenerated, &ring_core_prefix()).unwrap();
             }
             let srcs = asm_srcs(perlasm_src_dsts);
+
+            let target = Target {
+                arch: asm_target.arch.to_owned(),
+                os: os.to_owned(),
+                env: env.to_string(),
+                is_git: true,
```
Alovchin91, line 365: Is it always `is_git` because preassembly is supposed to always run from a GitHub Action only?

build.rs
```diff
@@ -348,13 +348,28 @@ fn pregenerate_asm_main() {
...
+                is_debug: false,
```
Alovchin91, line 366: Same question basically 🙂

awakecoding (approved):
LGTM

build.rs
```diff
@@ -348,13 +348,28 @@ fn pregenerate_asm_main() {
...
+            };
+
+            let compiler = cc::Build::new().get_compiler();
```
Alovchin91, line 369: Windows AArch64 (for now) will always be compiled with clang. Would we need to hardcode it here as well or, because preassembly is only expected to run by package.sh, we will need to set environment variables there?

I personally would prefer to keep that hardcoded compiler in a single place somewhere, so maybe we could introduce a new function to create compiler in one place? (Maybe this is something for our PR actually)

briansmith:
Good questions. I just submitted PR #1347 to replace `Target::is_git` and to document why we do this weird stuff. I've verified that it is easy to rebase this PR on top of that one, so once that's merged, I'll update this PR with the clarified version.

build.rs
briansmith: > Windows AArch64 (for now) will always be compiled with clang. Would we need to hardcode it here as well or, because preassembly is only expected to run by package.sh, we will need to set environment variables there?
>
> I personally would prefer to keep that hardcoded compiler in a single place somewhere, so maybe we could introduce a new function to create compiler in one place? (Maybe this is something for our PR actually)

I think here we would add a check that `compiler.is_like_clang()`, and then change `mk/package.sh` to set `TARGET_CC` to the path to VS2019's clang.

Alovchin91: My concern in this case is that we'll have to put a check for Windows AArch64 in 2 places – package.sh as you've mentioned and build.rs, because the users of *ring* will still have to compile C code with Clang.

Alovchin91: On the other hand, it might be temporary only for C code, while for assembly it'll probably stay permanent 🤔

briansmith: > My concern in this case is that we'll have to put a check for Windows AArch64 in 2 places – package.sh as you've mentioned and build.rs, because the users of ring will still have to compile C code with Clang. Since it's a (hopefully) temporary thing, I'd suggest to have this knowledge in 1 place.

It would be nice to have this in one place but OTOH the process for `package` is very different than the normal build. For `package` ideally the inputs will become completely fixed: a specific version of clang from a specific download, similar to how `nasm` is done.

For the build, at least when building from crates.io, ideally we don't force any specific compiler on the user. It seems temporarily the build will only work if the compiler is clang or clang-cl, but whether it's in `$PATH` or not, or what that compiler is named (`clang` vs `clang-12`, for example) wouldn't be hard-coded in.

briansmith:
> Good questions. I just submitted PR #1347 to replace `Target::is_git` and to document why we do this weird stuff. I've verified that it is easy to rebase this PR on top of that one, so once that's merged, I'll update this PR with the clarified version.

I merged PR #1347 and rebased this PR on top of it.

build.rs
briansmith: Now `is_git` has been replaced here by `force_warnings_into_errors` which I think is clearly the right thing to do. LMK if you disagree.

build.rs
briansmith: Let's move the remaining discussion of aarch64-specific stuff into PR #1339.

build.rs
briansmith: Regarding `is_debug`, we do want to have the preassembled stuff be a release build as we only really want to support debug building (`NDEBUG` and the like) during development, as `assert()` and similar things can introduce unwanted side channels.

briansmith (commented):
Thanks for the review. I think the construction of `Target` is now much clearer thanks to the changes prompted by your questions.

build.rs
Alovchin91: > but whether it's in `$PATH` or not, or what that compiler is named (`clang` vs `clang-12`, for example) wouldn't be hard-coded in.

I'm afraid that would make usage of *ring* unnecessarily over-complicated on Windows AArch64. I'd probably still suggest to hardcode `clang-cl` for now, for better compatibility and for simpler usage.

At the same time, if you don't want to release a new *ring* version before it's compilable under MSVC on AArch64, and given that you still need additional setup (Clang in the PATH), maybe it doesn't really matter much. 🤔

I guess I don't have a strong opinion but slightly inclined to hardcode `clang-cl` with a `FIXME` comment.

build.rs
```diff
@@ -370,13 +370,28 @@ fn pregenerate_asm_main() {
         perlasm(&perlasm_src_dsts, asm_target);
 
         if asm_target.preassemble {
+            // Preassembly is currently only done for Windows targets.
+            assert_eq!(&asm_target.oss, &[WINDOWS]);
+            let os = WINDOWS;
+            let env = "msvc";
```
Alovchin91, line 376: Is it not done for Windows *-gnu targets though? 🤔

build.rs
Alovchin91: Sorry, again didn't notice the last comment 🤦‍♂️  Will move the discussion to #1339

briansmith: > I'm afraid that would make usage of ring unnecessarily over-complicated on Windows AArch64. I'd probably still suggest to hardcode clang-cl for now, for better compatibility and for simpler usage.

Maybe we can do something like "if the target arch is aarch64 and the env is msvc and no compiler has been set then default to whataever `clang` there is on the path." In the products I work on we will set `TARGET_AARCH64_PC_WINDOWS_MSVC_CC` or similar to a specific version of clang.

> At the same time, if you don't want to release a new ring version before it's compilable under MSVC on AArch64, and given that one still needs additional setup (Clang in the $PATH), maybe it doesn't really matter much. 🤔

I suggest the following course of action, which seems ideal to me:
PR 1. Merge this.
PR 2. Make a PR to get the build working when `$TARGET_AARCH64_PC_WINDOWS_MSVC_CC` points to VS2019's `clang`, either a full path or "clang" when it is in `$PATH`.
PR 3. Change the code to compile with VS2019's cl.exe (or clang or clang-cl).

Note that defaulting to "clang" in "$PATH" creates some problems in some environments where people have a different `clang` in `$PATH`. In some projects I work on, this can happen because they need `clang` to target wasm32.

Alovchin91: Responded in #1339 😅

build.rs
briansmith: @Alovchin91 Another excellent question! I submitted PR #1348 to make it clearer. I will then rebase this on top of #1348 if that makes sense. Then there won't be any "msvc" here and inspecting the rest of the code will make it clear that the preassembled object files are used by all windows targets. See this existing code in build.rs that determines whether to use the preassembled code:

```
if use_pregenerated && target.os == WINDOWS {
```

briansmith: Also, at least to me it makes sense to use the same preassembled object files for -gnu targets since ultimately the ABI is the same, at least for code that doesn't use the C standard library, which the assembly code doesn't.

Alovchin91: 😅  Much clearer now indeed, thanks!

> Also, at least to me it makes sense to use the same preassembled object files for -gnu targets since ultimately the ABI is the same, at least for code that doesn't use the C standard library, which the assembly code doesn't.

I'm not super-professional here so the only thing I know is that GCC object files are usually not advised to be mixed with MSVC, but yet again — I'm dumb in this kind of questions 😬

briansmith:
I've now rebased this on top of #1348.

Alovchin91 (dismissed):

briansmith:
Not sure what's going on with the aarch64-apple-darwin nightly build in CI that is failing. Possibly a regression in the Rust Nightly toolchain?

briansmith:
> Not sure what's going on with the aarch64-apple-darwin nightly build in CI that is failing. Possibly a regression in the Rust Nightly toolchain?

I rebased this on top of PR #1349 to allow CI to work.

PR #1347, opened 2021-08-13
build.rs: Refactor away `Target::is_git` to clarify logic.

briansmith:
Localize the use of the "is git" check to a single place and use more
descriptive variables for the various effects that this check has.

As a nice side effect, the loosly-typed `warnings_are_errors` parameters
that were threaded through functions are now gone.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -308,19 +308,41 @@ fn ring_build_rs_main() {
 
     let is_git = std::fs::metadata(".git").is_ok();
 
-    // Published builds are always release builds.
+    // Published builds are always built in release mode.
     let is_debug = is_git && env::var("DEBUG").unwrap() != "false";
+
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -364,11 +386,23 @@ struct Target {
     arch: String,
     os: String,
     env: String,
-    is_git: bool,
+
+    // Is this a debug build? This affects whether assertions might be enabled
```
complexspaces, line 390: Nit: This (and the one below) could be made into a doc comment so IDE hover actions show the text without needing to jump to the field definition.

complexspaces (approved):

build.rs
briansmith: Thanks. i made that change!

PR #1349, opened 2021-08-15
CI: Use nightly-2021-08-13 to work around regression in latest Rust nightly.

briansmith:


briansmith:
I verified that nightly-2021-08-14 seems to break the build. It's not clear if this is a bug in Rustc or a bug in *ring* or both. For now, work around it by using nightly-2021-08-13 for testing.

briansmith:
The codecov coverage reduction seems like a false positive, maybe due to rounding.

PR #1358, opened 2021-08-18
Merge BoringSSL d4f877e: Reference the newer ChaCha20-Poly1305 RFC.

briansmith:


briansmith:
The tricky parts of this PR were:
1. Git doesn't recognize that `aead_chacha20_poly1305_tests.txt` is derived from the corresponding BoringSSL file.
2. There are references to the old RFC in *ring* that aren't in BoringSSL and vice-versa.
3. Grepping for "7539" shows a lot of false positives because that sequence of digits appears in many test vectors for other algorithms.

complexspaces (approved):

PR #1368, opened 2021-08-26
bigint: Remove some unnecessary assertions in `elem_exp_vartime_()`.

briansmith:


src/arithmetic/bigint.rs
```diff
@@ -611,11 +613,22 @@ impl<M> One<M, RR> {
...
         // `elem_squared` by replacing some of the early squarings with shifts.
         // TODO: Benchmark shift vs. squaring performance to determine the
-        // optimal value of `lg_base`.
-        let lg_base = 2usize; // Shifts vs. squaring trade-off.
-        debug_assert_eq!(lg_base.count_ones(), 1); // Must 2**n for n >= 0.
-        let shifts = r - bit + lg_base;
-        let exponent = (r / lg_base) as u64;
+        // optimal value of `LG_BASE`.
+        const LG_BASE: usize = 2; // Shifts vs. squaring trade-off.
+        debug_assert_eq!(LG_BASE.count_ones(), 1); // Must be 2**n for n >= 0.
```
complexspaces, line 618: Since *ring* has a higher MSRV, this could be replaced with a compile-time check akin to how [static-assertions](https://github.com/nvzqz/static-assertions-rs) allows values to be checked at compile time. This would entirely remove this runtime overhead for a known-good value.

briansmith: > Since _ring_ has a higher MSRV, this could be replaced with a compile-time check akin to how [static-assertions](https://github.com/nvzqz/static-assertions-rs) allows values to be checked at compile time. This would entirely remove this runtime overhead for a known-good value.

Note that this is a debug assertion so it doesn't apply to release builds, and also the optimizer should be able to easily optimize it away. I do agree with the idea of replacing these checks with static assertions. I filed #1371 for this.

complexspaces (approved):

PR #1385, opened 2021-09-21
Factor out `rsa::public::Modulus`

briansmith:


src/rsa/public/modulus.rs
```diff
@@ -0,0 +1,69 @@
...
+pub struct Modulus {
+    value: bigint::Modulus<N>,
+    bits: bits::BitLength,
+}
+
+impl core::fmt::Debug for Modulus {
+    fn fmt(&self, fmt: &mut ::core::fmt::Formatter) -> Result<(), ::core::fmt::Error> {
+        self.value.fmt(fmt)
+    }
+}
```
briansmith, line 13: I'm not too worried about the coverage failure here. 5 or so PRs from now, we can format the debug output the way we want, but we can't do that now. At that future point in time, we can add a test for this.

src/rsa/public/key.rs
```diff
@@ -40,30 +39,11 @@ impl Key {
         // documentation for RSA keys.
 
         // Step 3 / Step c for `n` (out of order).
```
complexspaces, line 41: This comment appears to have been moved into `Modulus::from_be_bytes`. Should it be removed here too?

src/rsa/public/modulus.rs
```diff
@@ -0,0 +1,69 @@
...
+        let bits_rounded_up =
+            bits::BitLength::from_usize_bytes(bits.as_usize_bytes_rounded_up()).unwrap(); // TODO: safe?
+        if bits_rounded_up < min_bits {
+            return Err(error::KeyRejected::too_small());
+        }
+        if bits > max_bits {
+            return Err(error::KeyRejected::too_large());
+        }
+
+        // TODO: what about this?
```
complexspaces, line 53: Perhaps this could be checked inside `Key::from_modulus_and_exponent`?

briansmith: I removed the comment here since it is still in `Key`. I do think I need to follow up and clarify (probably remove) that comment after I investigate the history of it.

complexspaces: That sounds fine since its technically out of scope for this MR either way.

complexspaces (approved):

PR #1389, opened 2021-09-23
RSA: Correct misleading documentation about keypair limitations.

briansmith:
Originally *ring* did require `p > q` but that limitation was removed,
so remove the comment about that limitation. Document the maximum size
of the public exponent next to where the minimum value is documented.


briansmith:
@complexspaces PTAL.

PR #1391, opened 2021-09-23
Internal: Replace some manual big-endian encoding w/ standard library use.

briansmith:
Refactor `limb::big_endian_from_limbs` to use an approach based on
iterators. We will be then be able to use the new `limb::be_bytes`
to implement `rsa::public::Exponent::be_bytes()` and
`rsa::public::Modulus::be_bytes()` and eventually other similar functions.
We want those functions to return `ExactSizeIterator`s.

This is also part of an ongoing process to eliminate replace all the
big-endian/little-endian encoding logic in *ring* w/ use of core APIs.

### DIFF
--- a/src/limb.rs
+++ b/src/limb.rs
@@ -18,7 +18,7 @@
 //! Limbs ordered least-significant-limb to most-significant-limb. The bits
 //! limbs use the native endianness.
 
-use crate::{c, error};
+use crate::{c, error, polyfill::ArrayFlatMap};
 
 #[cfg(feature = "alloc")]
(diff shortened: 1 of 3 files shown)


briansmith:
@complexspaces PTAL. You might remember that I had implemented a function for stripping leading zeros that allocated and returned the result into a `Box<[u8]>`. This is one of the stepping stones towards doing the same thing in a non-allocating way, in conjunction with an upcoming `StripLeadingZeroes` iterator.

complexspaces (approved):

PR #1392, opened 2021-09-24
RSA: Replace `RsaSubjectPublicKey` with `rsa::public::Key`.

briansmith:
Retain the `AsRef<[u8]>` implementation until we have a better design for
getting the public key bytes from the key pair.


briansmith:
Going to break this up into smaller parts.

PR #1393, opened 2021-09-24
RSA: Replace `RsaSubjectPublicKey::{modulus,exponent}` with `rsa::Public::{Modulus,Exponent}::be_bytes()`.

briansmith:
This is a step towards removing the heap-allocated and usually-unnecessary `public_key: RsaSubjectPublicKey` field. The new API allows the caller to better control how it stores/allocates the component values.

This also removes a couple of infallible `unwrap()`s.

This is a step towards removing `io::Positive` from the public API.

This is a breaking API change.


briansmith:
@complexspaces PTAL.

src/rsa/public/exponent.rs
```diff
@@ -77,6 +78,19 @@ impl Exponent {
 
         Ok(Self(value))
     }
+
+    /// The big-endian encoding of the exponent.
+    ///
+    /// There are no leading zeros.
```
complexspaces, line 84: Nit: This may be better worded as `All leading zeros have been stripped from the value.`

src/rsa/public/modulus.rs
```diff
@@ -48,6 +48,14 @@ impl Modulus {
         Ok(Self { value, bits })
     }
 
+    /// The big-endian encoding of the modulus.
+    ///
+    /// There are no leading zeros.
```
complexspaces, line 53: Ditto here, too.

src/rsa/public/exponent.rs
briansmith: I thought about that too, but there's no reason to imply that there were any leading zeros in the first place, as that's an implementation detail.

complexspaces: I see. If that's an implementation detail then the current comment is fine.

complexspaces (approved):

src/rsa/public/exponent.rs
```diff
@@ -77,6 +78,18 @@ impl Exponent {
...
+    pub fn be_bytes(&self) -> impl ExactSizeIterator<Item = u8> + Clone + '_ {
```
briansmith, line 85: FYI, before merging this, I removed `#[inline]` from here. It doesn't make to inline this now that it is so complicated.

complexspaces: Thanks for the heads up. I generally don't pay too much attention to the `#[inline]` attribute because rustc and LLVM are already extremely aggressive about inlining.

PR #1395, opened 2021-09-25
RSA: Replace `RsaSubjectPublicKey` with `rsa::public::Key`.

briansmith:


briansmith:
@complexspaces PTAL.

complexspaces (approved):

briansmith:
@complexspaces In CI, when checked out on Windows, "*.txt" files seem to be given Windows newlines, and then the test would fail with the way this PR originally modified it, because the test was expecting a Unix newline. That's a bug in its own right that should be dealt with; maybe this is what you were hitting before when you were having difficulties with newlines? However, the change that triggered it was not critical to this PR, so I just removed the newline from the `.txt` file that the earlier version of this PR added.

PR #1396, opened 2021-09-27
Derive `Clone` and `Copy` more instead of explicitly implementing them

briansmith:


briansmith:
@complexspaces PTAL. This is based on your comment in another PR suggesting we could derive these more.

src/agreement.rs
```diff
@@ -209,7 +196,7 @@ where
     }
 }
 
-impl<B: AsRef<[u8]>> UnparsedPublicKey<B> {
```
complexspaces, line 212: Will removing the `AsRef<[u8]>` bound cause any usability problems? In `src/signature.rs` a bound was added on `verify` to make up for it but nothing was added here.

briansmith: `agreement::agree_ephemeral` already has an `AsRef<[u8]>` bound so nothing needed to be added.

I think if this change were problematic, the tests would fail to compile. Maybe I'm overlooking something?

src/agreement.rs
```diff
@@ -178,25 +178,12 @@ impl PublicKey {
 }
 
 /// An unparsed, possibly malformed, public key for key agreement.
-pub struct UnparsedPublicKey<B: AsRef<[u8]>> {
+#[derive(Clone, Copy)]
+pub struct UnparsedPublicKey<B> {
```
briansmith, line 182: Regarding the code coverage for `UnparsedPublicKey`, I guess the explicit implementations were also untested, so I'm going to call that out of scope.

src/agreement.rs
complexspaces: Ah ok, I missed that it already had that bound present. I think everything is fine then in that case.

complexspaces (approved):

src/endian.rs
```diff
@@ -25,21 +25,10 @@ pub trait FromByteArray<T> {
 
 macro_rules! define_endian {
     ($endian:ident) => {
+        #[derive(Clone, Copy)]
```
briansmith, line 28: Ditto. I guess we'll need a coverage test for this, which wasn't being tested either way. (It's not tested because it's also `Copy` and the `Copy` implementation is used instead of `Clone`, which we don't need, except that `Copy` forces us to implement it.)

src/signature.rs
```diff
@@ -352,26 +352,13 @@ pub trait VerificationAlgorithm: core::fmt::Debug + Sync + sealed::Sealed {
 }
 
 /// An unparsed, possibly malformed, public key for signature verification.
-pub struct UnparsedPublicKey<B: AsRef<[u8]>> {
+#[derive(Clone, Copy)]
+pub struct UnparsedPublicKey<B> {
```
briansmith, line 356: Ditto.

src/agreement.rs
complexspaces: Would a follow-up issue be best to cover adding these?

briansmith: If it were anything other than "missing code coverage," probably. Right now we can use the code coverage reports to see what code isn't hit by code coverage yet. Currently about 7%, although that's an over-estimate because there's some testing done not in CI that isn't counted towards the measurement yet.

PR #1397, opened 2021-09-27
RSA internals: Move digest instead of passing a reference when doing padding.

briansmith:
The digest is never used after encoding, so move it instead of referencing it.
This is more correct since for signing (and soon encryption) the padded value
is only supposed to be used once.


briansmith:
@complexspaces PTAL.

complexspaces (approved):

PR #1398, opened 2021-09-27
RSA/bigint: Thread proof of CPU feature detection through bigint.

briansmith:
The tests of `bigint` were not doing CPU feature detection themselves.
Thus they were depending on some other tests that run before them to do
it, or else they were not making use of all the CPU optimizations
possible, and thus not testing all the interesting code paths.

Also, as we are expanding the functionality of the RSA module, it has
become more difficult to track where CPU feature detection has been done
and where it needs to be done. Move the proof that the CPU feature
detection has been done down into the callers of the `bn_` functions
that need CPU feature detection to have been done.

This will also be helpful if/when we expand the use of the `bigint`
module beyond RSA.

### DIFF
--- a/src/arithmetic/bigint.rs
+++ b/src/arithmetic/bigint.rs
@@ -38,7 +38,7 @@
 
 use crate::{
     arithmetic::montgomery::*,
-    bits, bssl, c, error,
+    bits, bssl, c, cpu, error,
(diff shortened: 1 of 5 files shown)


briansmith:
@complexspaces PTAL.

complexspaces (approved):

PR #1399, opened 2021-09-28
RSA: Remove `From<Exponent> for NonZeroU64` implementation.

briansmith:
We shouldn't promise to external users that the exponent will always
fit in a 64-bit integer as in the future we may need to allow larger
exponents.


briansmith:
@complexspaces PTAL.

complexspaces (approved):

PR #1400, opened 2021-09-28
Refactor RSA PSS MGF processing to enable future use by OAEP.

briansmith:
Previously, `mgf1()` wrote the mask to the buffer, and then we XOR'd
the data onto the mask. Now, `mgf1()` XOR's the mask onto the data
that is already in the `out` buffer.

### DIFF
--- a/src/rsa/padding.rs
+++ b/src/rsa/padding.rs
@@ -229,10 +229,6 @@ pub struct PSS {
 
 impl crate::sealed::Sealed for PSS {}
 
-// Maximum supported length of the salt in bytes.
-// In practice, this is constrained by the maximum digest length.
-const MAX_SALT_LEN: usize = digest::MAX_OUTPUT_LEN;
-
 impl Padding for PSS {
     fn digest_alg(&self) -> &'static digest::Algorithm {
         self.digest_alg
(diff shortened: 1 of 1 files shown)


briansmith:
@complexspaces PTAL.

src/rsa/padding.rs
```diff
@@ -269,42 +265,36 @@ impl RsaEncoding for PSS {
...
-        // Step 4.
-        let mut salt = [0u8; MAX_SALT_LEN];
-        let salt = &mut salt[..metrics.s_len];
-        rng.fill(salt)?;
-
-        // Step 5 and 6.
-        let h_hash = pss_digest(self.digest_alg, m_hash, salt);
+        let (db, digest_terminator) = em.split_at_mut(metrics.db_len);
+        let h;
+        {
```
complexspaces, line 270: Nit: `h` could be assigned to the output of this block to potentially clarify at what point its bound when scrolling the code.

briansmith: Good point. I actually just eliminated the scoping block as it's no longer important with the refactored code.

complexspaces (approved):

src/rsa/padding.rs
complexspaces: That works too. I think its much easier to follow whats happening now 👍

PR #1401, opened 2021-09-28
RSA: Split RSA public key arithmetic from signature verification logic.

briansmith:
`Exponent::exponentiate` will eventually be used for both signature
verification and encrypting using a public key.


briansmith:
@complexspaces PTAL.

src/rsa/public/key.rs
```diff
@@ -87,6 +98,55 @@ impl Key {
     pub fn e(&self) -> Exponent {
         self.e
     }
+
+    /// Calculates base**e (mod n), filling the first part of `out_buffer` with
+    /// the result.
+    ///
+    /// This is constant-time with respect to the value in `input` (only).
```
complexspaces, line 105: Should this mention `base` instead as well?
```suggestion
    /// This is constant-time with respect to the value in `base` (only).
```

complexspaces (approved):

PR #1403, opened 2021-09-28
RSA: Remove misleading documentation about side channels protection.

briansmith:
When *ring* first started, BoringSSL and OpenSSL upstream were both
using an implementation of constant-time-ish exponentiation that took
shortcuts that made it clearly not constant-time. Long ago, that code
was replaced here and in BoringSSL (and probably OpenSSL upstream), so
this comment is no longer correct.


briansmith:
@complexspaces PTAL.

complexspaces (approved):

PR #1405, opened 2021-09-28
RSA: Simplify API

briansmith:
See the individual commit messages for details. It will likely be easiest to review them commit-by-commit.

Basically, this simplifies the `rsa` module's API to something that better aligns with what we can do generically (using traits) for all the asymmetric algorithm implementations. We don't need to expose `public::Modulus` and `public::Exponent` since the only thing the user could do is measure them and get their encoded components; the latter is what the `Components` types are designed to do.


briansmith:
@complexspaces PTAL. You're best off reviewing it commit-by-commit, especially because of the renaming of rsa/signing.rs to rsa/keypair.rs, which replaced a previsouly-much-different version of that file.

complexspaces (approved):

PR #1406, opened 2021-09-30
Add UEFI target support

xiaoyuxlu:
####  build.rs:   
ABI of rust UEFI target [x86_64-unknown-uefi](https://github.com/rust-lang/rust/blob/master/compiler/rustc_target/src/spec/x86_64_unknown_uefi.rs#L5) is like windows. So add an AsmTarget like windows and it also need preassemble.

Currently must use Clang to compile c code for rust UEFI target.

#### rand.rs: 
Add [rdrand](https://crates.io/crates/rdrand) crate to generate random number. It can only be built when the target_os is `uefi`.


build.rs
```diff
@@ -235,6 +235,13 @@ const ASM_TARGETS: &[AsmTarget] = &[
         asm_extension: "S",
         preassemble: true,
     },
+    AsmTarget {
+        oss: &[UEFI],
```
briansmith, line 239: Does it not work correctly if we add UEFI to the `AsmTarget` that has `oss: &[WINDOWS],
        arch: "x86_64"`?

build.rs
```diff
@@ -602,6 +612,7 @@ fn cc(file: &Path, ext: &str, target: &Target, include_dir: &Path, out_file: &Pa
         && target.os != "redox"
         && target.os != "windows"
         && target.arch != "wasm32"
+        && target.os != "uefi"
```
briansmith, line 615: It was a mistake here that we use `"windows"` instead of `WINDOWS`, so I understand why you copied that, but we should just use `UEFI` here.

build.rs
```diff
@@ -460,7 +470,7 @@ fn build_c_code(
 
         // For Windows we also pregenerate the object files for non-Git builds so
         // the user doesn't need to install the assembler.
-        if use_pregenerated && target.os == WINDOWS {
+        if use_pregenerated && (target.os == WINDOWS || target.os == UEFI) {
```
briansmith, line 473: I suggest we add a:

```
fn supports_preassembly(arch: &str, os: &str) -> bool {
    ASM_TARGETS.iter().find(|asm_target| {
         asm_target.preassemble && asm_target.arch == arch && asm_target.oss.contains(os)
   }).is_some()
)
```

This will fix a bug in the existing logic wherein we don't screen out aarch64 Windows, for which we just added support, where we don't preassemble.

build.rs
```diff
@@ -642,6 +653,13 @@ fn cc(file: &Path, ext: &str, target: &Target, include_dir: &Path, out_file: &Pa
         }
     }
 
+    // UEFI is a baremental freestanding environment without stdlib.
+    if target.os == "uefi" && target.arch == "x86_64" {
```
briansmith, line 657: Is `target.arch == "x86_64"` a necessary condition? I guess we'd need these flags for any UEFI target, regardless of architecture.

I suggest:
```
let freestanding = target.os = UEFI;
if freestanding {
    let _ = c.flat("-ffreestanding");
}
```
Then, de-dupe this with the existing code block that adds `-nostdlibinc` and `RING_CORE_NOSTDLIBINC` by adding `freestanding || <existing conditions>`.

src/rand.rs
```diff
@@ -419,3 +422,14 @@ mod fuchsia {
         fn zx_cprng_draw(buffer: *mut u8, length: usize);
     }
 }
+
+#[cfg(any(target_os = "uefi"))]
+mod uefi {
+    use crate::error;
+    use rdrand::RdRand;
```
briansmith, line 429: Hi. Presently our policy is to not depend on third-party crates for any crypto primitives. Let's just use the `core::arch::x86_64` intrinsics directly, instead of depending on that crate.

At the top of this file is a long comment with a table that indicates the source of randomness for the target. Please add a comment about UEFI to that table.

briansmith (changes requested):
Thanks for this.

In .github/ci.yml where have some blocks like this:

```
          - target: aarch64-apple-darwin
            host_os: macos-latest
            # GitHub Actions doesn't have a way to run this target yet.
            cargo_options: --no-run
```
Please add an analogous one for the x86_64 UEFI target.

Similarly there is:
```
        target:
          - aarch64-apple-ios
          - ...
          - x86_64-unknown-linux-gnu
```
Add the x86_64 UEFI target to that.

Also, could you point me to a way that we could emulate the UEFI environment well enough to run tests? It would be ideal if we could get this emulated environment working in GitHub Actions, but if this is a large amount of work, we can go with the build-only approach for now.

briansmith:
Since we need to force clang for this target, mk/install-build-tools.sh can be modified here to add this target:
```
--target=i686-unknown-linux-musl|--target=x86_64-unknown-linux-musl)
  use_clang=1
```

Please make the analogous change in mk/cargo.sh.

This should make CI work and will also make it easier for anybody, especially me, to build locally.

briansmith:
Does CPUID work reliabily (as expected) in the UEFI environment?

briansmith:
Also, are there any limitations or special considerations on CPU features (AVX2, AES-NI, etc.) for the EUFI environment?

src/rand.rs
briansmith: > Hi. Presently our policy is to not depend on third-party crates for any crypto primitives. Let's just use the core::arch::x86_64 intrinsics directly, instead of depending on that crate.

Scratch this. BoringSSL upstream has rdrand-x86_64.pl that exposes a `CRYPTO_rdrand` function that we can use instead. I also need to update cpu-intel.c to disable RDRAND on AMD CPUs due to bugs in some of them, and then I'll update cpu.rs so that you can detect whether RDRAND is available.

xiaoyuxlu:
A few days ago was China National Day. Sorry for the late reply. 
I appreciate you for watching the pull request carefully and give a lot of comments.

I will continue to work on it and make this patch better. 

> 
> 
> Thanks for this.
> 
> In .github/ci.yml where have some blocks like this:
> 
> ```
>           - target: aarch64-apple-darwin
>             host_os: macos-latest
>             # GitHub Actions doesn't have a way to run this target yet.
>             cargo_options: --no-run
> ```
> 
> Please add an analogous one for the x86_64 UEFI target.
> 
> Similarly there is:
> 
> ```
>         target:
>           - aarch64-apple-ios
>           - ...
>           - x86_64-unknown-linux-gnu
> ```
> 
> Add the x86_64 UEFI target to that.
> 
> Also, could you point me to a way that we could emulate the UEFI environment well enough to run tests? It would be ideal if we could get this emulated environment working in GitHub Actions, but if this is a large amount of work, we can go with the build-only approach for now.

Currently, I don't have a good way to run tests. I will try to find a way.  I think we can build-only for now.  

> 
> 
> Does CPUID work reliabily (as expected) in the UEFI environment?

I think so. 

> 
> 
> Also, are there any limitations or special considerations on CPU features (AVX2, AES-NI, etc.) for the EUFI environment?

UEFI environment always run in long-mode, in x86_64 it always run in 64bit. 
I think there is no special things.

jyao1:
@briansmith, would you please review this patch to see if there is any other comment?

tiziano88:
This patch would be useful to us too as part of https://github.com/project-oak/oak , is there anything in particular that is blocking it? Can we help resolve it? Thanks @xiaoyuxlu and @briansmith for the work so far

jyao1:
This has been there for half a year.

@briansmith, would you please review this patch to see if there is any other comment?

dnovick:
Just wondering what the status of this pull request is. Are there any blocking issues remaining? Thanks.

.github/workflows/ci.yml
```diff
@@ -453,3 +453,55 @@ jobs:
...
+      - if: ${{ contains(matrix.host_os, 'windows') }}
+        run: ./mk/install-build-tools.ps1
+
+      - uses: briansmith/toolchain@v1
+        with:
+          toolchain: nightly
+          override: true
+          components: rust-src
+
+      # Currently, there is no 'rust-std' for target 'x86_64-unknown-uefi'.
```
briansmith, line 504: Is this still the case? Or can we simplify all this now?

Cargo.toml
```diff
@@ -82,6 +82,7 @@ include = [
     "crypto/fipsmodule/modes/asm/ghash-x86.pl",
     "crypto/fipsmodule/modes/asm/ghash-x86_64.pl",
     "crypto/fipsmodule/modes/asm/ghashv8-armx.pl",
+    "crypto/fipsmodule/rand/asm/rdrand-x86_64.pl",
```
briansmith, line 85: This can go away since we'd use `getrandom`.

build.rs
```diff
@@ -64,6 +64,7 @@ const RING_SRCS: &[(&[&str], &str)] = &[
     (&[X86_64], "crypto/fipsmodule/ec/asm/p256-x86_64-asm.pl"),
     (&[X86_64], "crypto/fipsmodule/modes/asm/aesni-gcm-x86_64.pl"),
     (&[X86_64], "crypto/fipsmodule/modes/asm/ghash-x86_64.pl"),
+    (&[X86_64], "crypto/fipsmodule/rand/asm/rdrand-x86_64.pl"),
```
briansmith, line 67: Ditto.

build.rs
```diff
@@ -222,7 +223,7 @@ const ASM_TARGETS: &[AsmTarget] = &[
         preassemble: true,
     },
     AsmTarget {
-        oss: &[WINDOWS],
+        oss: &[WINDOWS, UEFI],
```
briansmith, line 226: This deserves a comment about why UEFI uses Windows ASM stuff unlike everything else.

build.rs
```diff
@@ -381,8 +384,9 @@ fn pregenerate_asm_main() {
         perlasm(&perlasm_src_dsts, asm_target);
 
         if asm_target.preassemble {
-            // Preassembly is currently only done for Windows targets.
-            assert_eq!(&asm_target.oss, &[WINDOWS]);
+            // Preassembly is currently done for Windows and UEFI targets.
```
briansmith, line 387: let's just remove the comment since there's now a `preassemble` field that makes this clear, and it duplicates the assertion below.

build.rs
```diff
@@ -460,7 +464,7 @@ fn build_c_code(
 
         // For Windows we also pregenerate the object files for non-Git builds so
         // the user doesn't need to install the assembler.
-        if use_pregenerated && target.os == WINDOWS {
+        if use_pregenerated && supports_preassembly(&target.arch, &target.os) {
```
briansmith, line 467: Change this to look at the `preassemble` field.

build.rs
```diff
@@ -554,7 +558,7 @@ fn compile(p: &Path, target: &Target, include_dir: &Path, out_dir: &Path) -> Str
         p.to_str().expect("Invalid path").into()
     } else {
         let out_file = obj_path(out_dir, p);
-        let cmd = if target.os != WINDOWS || ext != "asm" {
+        let cmd = if !supports_preassembly(&target.arch, &target.os) || ext != "asm" {
```
briansmith, line 561: Ditto.

build.rs
```diff
@@ -627,17 +632,25 @@ fn cc(file: &Path, ext: &str, target: &Target, include_dir: &Path, out_file: &Pa
         }
     }
 
+    // UEFI is a baremental freestanding environment without stdlib.
+    let freestanding = target.os == UEFI;
```
briansmith, line 636: Is this still true? I heard there is a libstd now.

build.rs
```diff
@@ -680,7 +693,10 @@ fn nasm(file: &Path, arch: &str, include_dir: &Path, out_file: &Path) -> Command
         std::path::MAIN_SEPARATOR,
     )));
 
-    let mut c = Command::new("./target/tools/windows/nasm/nasm");
+    let mut c = Command::new(&get_command(
+        "NASM_EXECUTABLE",
+        "./target/tools/windows/nasm/nasm",
+    ));
```
briansmith, line 699: Please move this to a separate PR.

briansmith (changes requested):
Instead of all the rdrand stuff, please add a `less-safe-getrandom-rdrand` non-default feature to *ring*. Then change rand.rs so to add UEFI to the lost of supported operating systems/environments if and only if that feature is also enabled.

Presumably then SGX and similar would do likewise.

briansmith:
This is a very old PR. If you were to rebase it onto today's main, most of it would go away as explained above. If/when this is ready, I will release a 17.x update that adds UEFI support.

xiaoyuxlu:
> This is a very old PR. If you were to rebase it onto today's main, most of it would go away as explained above. If/when this is ready, I will release a 17.x update that adds UEFI support.

You're right.
Will do. Thank you very much.

hugusmaximus:
Hi there! I'm not a developers nor I am too familiarized with some specific OS internals and faced a problem trying to compile "tiny_http" crate with "rustls" support on a aarch64 platform and Hermit (https://github.com/hermit-os/hermit-rs/tree/main) unikernel. After reading this thread I think I may be facing the same problem as other people trying to compile on other platforms... This is my error:

```
error[E0425]: cannot find function `fill_impl` in this scope
   --> /home/hugo/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ring-0.16.20/src/rand.rs:160:9
    |
160 |         fill_impl(dest)
    |         ^^^^^^^^^
    |
help: you might have meant to call the method
    |
160 |         self.fill_impl(dest)
    |         +++++

For more information about this error, try `rustc --explain E0425`.
error: could not compile `ring` (lib) due to 1 previous error
```

@briansmith will the release 17.x of ring solve my problem too? Any hint on how I need to proceed to solve that? Does it require intervention of the developer of Hermit OS?

Best,

PR #1407, opened 2021-10-04
digest: Remove public fields

briansmith:


briansmith:
@complexspaces PTAL.

complexspaces (dismissed):

src/digest.rs
```diff
@@ -306,6 +295,29 @@ impl Eq for Algorithm {}
...
+    /// The size of the chaining value of the digest function, in bytes.
+    ///
+    /// For non-truncated algorithms (SHA-1, SHA-256, SHA-512), this is equal
+    /// to [`Self::output_len()`]. For truncated algorithms (e.g. SHA-384,
+    /// SHA-512/256), this is equal to the length before truncation. This is
+    /// mostly helpful for determining the size of an HMAC key that is
+    /// appropriate for the digest algorithm.
+    pub fn chaining_len(&self) -> usize {
+        self.chaining_len
+    }
```
briansmith, line 313: I noticed when writing this PR that `chaining_len` was and is untested. Since that isn't new in this PR, I'm going to defer it for later.

src/hkdf.rs
```diff
@@ -47,7 +47,7 @@ pub static HKDF_SHA512: Algorithm = Algorithm(hmac::HMAC_SHA512);
 
 impl KeyType for Algorithm {
     fn len(&self) -> usize {
-        self.0.digest_algorithm().output_len
+        self.0.digest_algorithm().output_len()
```
briansmith, line 50: Ditto: We should test that one can derive a `Salt` using HKDF.

src/hmac.rs
```diff
@@ -256,7 +258,7 @@ impl Key {
 
 impl hkdf::KeyType for Algorithm {
     fn len(&self) -> usize {
-        self.digest_algorithm().output_len
+        self.digest_algorithm().output_len()
```
briansmith, line 261: Ditto for HMAC.

PR #1408, opened 2021-10-04
signature: Fix doc test when `--feature=std` is used.

briansmith:


briansmith:
@complexspaces PTAL. This fixes breakage from #1405.

complexspaces (approved):

PR #1410, opened 2021-10-05
RSA: Refactor `rsa::KeyPair::sign()` in preparation for adding RSA encryption

briansmith:
Refactor the implementation of `rsa::KeyPair::sign()` so that the arithmetic can be reused for RSA encryption. Clarify the implementation design decisions.

Also clarify the API, especially w.r.t. failure conditions.

### DIFF
--- a/src/rsa/keypair.rs
+++ b/src/rsa/keypair.rs
@@ -569,13 +569,37 @@ impl KeyPair {
         msg: &[u8],
         signature: &mut [u8],
     ) -> Result<(), error::Unspecified> {
-        let mod_bits = self.public.n().len_bits();
-        if signature.len() != mod_bits.as_usize_bytes_rounded_up() {
+        if signature.len() != self.public().modulus_len() {
             return Err(error::Unspecified);
         }
 
         let m_hash = digest::digest(padding_alg.digest_alg(), msg);
-        padding_alg.encode(m_hash, signature, mod_bits, rng)?;
+
+        // Use the output buffer as the scratch space for the signature to
+        // reduce the required stack space.
+        padding_alg.encode(m_hash, signature, self.public().n().len_bits(), rng)?;
+
+        // RFC 8017 Section 5.1.2: RSADP, using the Chinese Remainder Theorem
+        // with Garner's algorithm.
+
+        // Steps 1 and 2.
+        let m = self.private_exponentiate(signature)?;
+
+        // Step 3.
+        m.fill_be_bytes(signature);
+
+        Ok(())
+    }
+
+    /// Returns base**d (mod n).
+    ///
+    /// This does not return or write any intermediate results into any buffers
+    /// that are provided by the caller so that no intermediate state will be
+    /// leaked that would endanger the private key.
+    ///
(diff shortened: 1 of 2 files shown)


src/rsa/keypair.rs
```diff
@@ -563,21 +569,48 @@ impl KeyPair {
...
+        // reduce the required stack space.
+        padding_alg.encode(m_hash, signature, self.public().n().len_bits(), rng)?;
+
+        // RFC 8017 Section 5.1.2: RSADP, using the Chinese Remainder Theorem
+        // with Garner's algorithm.
+
+        // Steps 1 and 2.
+        let m = self.private_exponentiate(signature)?;
+
+        // Step 3.
```
complexspaces, line 588: Should this still keep the reference to "Manger's Attack revisited"?

tests/rsa_tests.rs
```diff
@@ -142,24 +142,19 @@ fn test_signature_rsa_pkcs1_sign_output_buffer_len() {
...
-    assert!(key_pair
-        .sign(&signature::RSA_PKCS1_SHA256, &rng, MESSAGE, &mut signature)
-        .is_ok());
-
-    // The output buffer is one byte too long.
-    signature.push(0);
-    assert!(key_pair
-        .sign(&signature::RSA_PKCS1_SHA256, &rng, MESSAGE, &mut signature)
-        .is_err());
+    // Wen the output buffer is not exactly the right length, `sign()` returns
```
complexspaces, line 145: Typo: 
```suggestion
    // When the output buffer is not exactly the right length, `sign()` returns
```

tests/rsa_tests.rs
```diff
@@ -142,24 +142,19 @@ fn test_signature_rsa_pkcs1_sign_output_buffer_len() {
...
+    // an error (and does not panic or invoke UB). if `sign` doesn't check that
```
complexspaces, line 146: Based on this comment, shouldn't there be some tests checking the error conditions like before?

briansmith: Sorry, I don't understand what you mean. Note that the test is testing that the condition "the buffer is the right length" is equal to (`assert_eq!`) the condition "the result is Ok(_).".

src/rsa/keypair.rs
briansmith: I mentioned why I removed the reference in the third commit message.

tests/rsa_tests.rs
briansmith: Thanks. I fixed it and pushed a new version of this set of commits.

src/rsa/keypair.rs
complexspaces: Ah ok, thank you for the clarification.

tests/rsa_tests.rs
complexspaces: Oh, this is my bad. I re-read the test and noticed that the left-hand side of the `assert_eq!()` is dependent on `len`'s value. I missed that on my first read-through and combined with the comment, was tripped up.

complexspaces (approved):

briansmith:
I had to re-push this because rustfmt.

PR #1412, opened 2021-10-06
RSA internals: Split `rsa::padding` into submodules

briansmith:
Prepare for adding OAEP by organizing the submodule in a way that separates what's shared from what's not shared.


briansmith:
This should help with the review:
* `git diff main:src/rsa/padding.rs src/rsa/padding/pkcs1.rs`
* `git diff main:src/rsa/padding.rs src/rsa/padding/pss.rs`

PR #1423, opened 2021-11-12
fix dead link for AEAD

name1e5s:
The [original link to AEAD Paper](http://www-cse.ucsd.edu/~mihir/papers/oem.html) is dead.
According to https://cseweb.ucsd.edu//~mihir/pubs.html we can find the new link is [here](https://eprint.iacr.org/2000/025.pdf).

### DIFF
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -18,7 +18,7 @@
 //! generic composition paradigm][AEAD] for an introduction to the concept of
 //! AEADs.
 //!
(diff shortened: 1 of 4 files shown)


ShadowJonathan:
Thanks, i was wondering why the link wasn't working.

briansmith (approved):

briansmith:
Thank you!

PR #1428, opened 2021-11-18
all: convert uses of static to const

### LINKED ISSUE #1427: Change statics to const, where possible
In rustls-ffi, we are hoping to export some ciphersuites (`rustls::SupportedCipherSuite` structs) to C code, by putting them in a const array (see https://github.com/rustls/rustls-ffi/pull/165). One obstacle we are running into is that each `rustls::SupportedCipherSuite` has several statics inside of it, and you can't declare a const that points to statics:

```rust
pub static TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256: SupportedCipherSuite =
    SupportedCipherSuite::Tls12(&Tls12CipherSuite {
        common: CipherSuiteCommon {
            suite: CipherSuite::TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
            bulk: BulkAlgorithm::Chacha20Poly1305,
            aead_algorithm: &ring::aead::CHACHA20_POLY1305,
        },
        kx: KeyExchangeAlgorithm::ECDHE,
        sign: TLS12_ECDSA_SCHEMES,
        fixed_iv_len: 12,
        explicit_nonce_len: 0,
        aead_alg: &ChaCha20Poly1305,
        hmac_algorithm: ring::hmac::HMAC_SHA256,
    });
```

I am not a Rust expert, but from my understanding, the Rust compiler has gotten smarter about its ability to evaluate consts, which may make the conversion from static to const possible now, especially if your compatibility guarantee is only for the latest stable version of Rust. 

If there is otherwise no difference, would it be possible to change some of the `static` values in ring to `const` values instead?

kevinburke:
It may not have been possible to do this when this code was written,
but it should be possible now.

Fixes #1427.

### DIFF
--- a/src/aead/aes_gcm.rs
+++ b/src/aead/aes_gcm.rs
@@ -21,7 +21,7 @@ use crate::{aead, cpu, error, polyfill};
 use core::ops::RangeFrom;
 
 /// AES-128 in GCM mode with 128-bit tags and 96 bit nonces.
-pub static AES_128_GCM: aead::Algorithm = aead::Algorithm {
+pub const AES_128_GCM: aead::Algorithm = aead::Algorithm {
     key_len: 16,
     init: init_128,
     seal: aes_gcm_seal,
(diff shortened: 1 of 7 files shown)


kevinburke:
Github Actions run here: https://github.com/kevinburke/ring/actions/runs/1478224095

Issue #728, opened 2018-11-26
Provide an easier-to-use rand API


briansmith:
Consider adding a function like this to `ring::rand`.

```rust
pub fn value<T: Default + AsMut<[u8]>>(rng: &SecureRandom) -> Result<T, error::Unspecified> {
    let mut r: T = Default::default();
    rng.fill(r.as_mut())?;
    Ok(r)
}
```

Then we could initialize fields like this:
```rust
struct Key {
   value: [u8; KEY_LENGTH],
}

impl Key {
    fn new(rng: &rand::SecureRandom) -> Result<Self, error::Unspecified> {
        Self { value: rand::value(rng) }
    }
}
```

Ideally `rand::value()` would work for arrays of all primitive integer types (usize, u128, u64, u32, u16, u8) and vector types. Thus we'd need to replace the `AsMut<[u8]>` bound with a more general one that admits those types.

sconybeare:
I decided to take a stab at this. I haven't tried to make anything work for vectors, but I've implemented this with a trait bound that allows for the construction of primitive integer types, as well as fixed sized arrays thereof. The lengths of those arrays are arbitrarily limited to 32 elements because we don't have support for generic compile-time integers, and because I had to pick a number.

briansmith:
In the interim, until we get this all sorted out, I created a new function:

```rust
pub fn generate<T: RandomlyConstructable>(
    rng: &SecureRandom,
) -> Result<Random<T>, error::Unspecified> { ... }
```

I implemented `RandomlyConstructable` for `u8` arrays of lengths 4, 8, 16, 32, 48, and 64, since those are the lengths I know to be useful for *ring* users.

Notably, I was able to do so without using any `unsafe`. I think that's something we should strive for in a more generic implementation.

> The lengths of those arrays are arbitrarily limited to 32 elements because we don't have support for generic compile-time integers, and because I had to pick a number.

`[u8; 64]` and `[u8; 48]` are commonly needed. In my implementation I avoid using the standard library's `AsRef`/`AsMut` implementations for this reason.

PR #1431, opened 2021-12-07
Provide an easier-to-use rand API

Cassy343:
I saw that issue #728 hit a dead end a while ago, and I figured that with the somewhat recent stabilization of const generics that it could be picked up again. I extend random construction to arrays of all integer types, however I do use unsafe code to accomplish this. If it is desirable to stick with strictly safe code, then that is definitely possible, however it will likely be more inefficient.

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -72,6 +72,7 @@ where
 
 pub(crate) mod sealed {
     use crate::error;
+    use core::{mem, slice};
 
     pub trait SecureRandom: core::fmt::Debug {
         /// Fills `dest` with random bytes.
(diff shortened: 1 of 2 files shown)


src/rand.rs
```diff
@@ -83,21 +84,64 @@ pub(crate) mod sealed {
...
+        }
+
+        #[inline]
+        fn as_mut_bytes(&mut self) -> &mut [u8] {
+            // Since all values in this expression are constant, the compiler should optimize this
+            // assertion out.
+            assert!(
+                N == 0 || (isize::MAX as usize) / N >= mem::size_of::<T>(),
+                "Array too large to be converted to slice of u8"
+            );
```
briansmith, line 124: Is this assertion necessary? I thought it was already UB to have `self` be larger than isize bytes?

briansmith: We should have tests for the case where N == 0 and where `core::mem::size_of<T>` is zero.

briansmith: When I want to be sure the compiler can evaluate an expression at compile time, I do:
```
const NO_OVERFLOW: bool = N == 0 || (isize::MAX as usize) / N >= mem::size_of::<T>();
assert!(NO_OVERFLOW);
```
I'd rather not have the string in the assertion. I'm working to eliminate all such strings.

src/rand.rs
```diff
@@ -83,21 +84,64 @@ pub(crate) mod sealed {
...
+            // Safety:
+            // - `data` is valid (aligned, non-null, non-dangling) for reads and writes of size
+            //   `len` bytes
+            //   - The entire memory range of this slice is in a single allocated object because
+            //     it came from another mutable slice
+            //   - `data` is not dangling and properly aligned for all types because, again, it was
+            //     derived from another mutable slice
+            // - Because T is POD, and because `len` is calculated based on the size of T,
+            //   `data` points to `len` initialized u8s
+            // - We co-opt the lifetime of `self`, so no borrow checker rules are broken
```
briansmith, line 138: I'm not sure about this:
>             // - We co-opt the lifetime of `self`, so no borrow checker rules are broken

It's hard for me to remember the exact circumstances, but I recall trying to do something similar and finding that I was running into a conflict with the mutable noalias optimization or similar.

This, in conjunction with wanting to avoid the need to initially zero all the bytes to start with, made me want to rewrite this whole module in terms of `MaybeUninit` and pointers.

WDYT?

briansmith (changes requested):
The idea seems good. I think we need to have some tests added. In particular, expand the test coverage to all the types you added, and also arrays of length 0.

src/rand.rs
Cassy343: > Is this assertion necessary? I thought it was already UB to have `self` be larger than isize bytes?

Ope I somehow managed to not read the `* mem::size_of<T>` part in the docs. I'll take that out - it makes matters simpler anyway.

src/rand.rs
Cassy343: So I had not heard of the mutable noalias optimization before, and quick google searches of it suggest that it's actually [buggy in the LLVM](https://stackoverflow.com/a/57259339/5619133). In a world where there is no mis-compilation, the function signature should essentially be equivalent to `fn as_mut_bytes<'a>(&'a mut self) -> &'a mut [u8]`, and since `'a` will be inferred by the usage of the return value, it will extend the borrow of `self` accordingly. I can ask some other rustaceans I know about it to double check though.

When reading through the module I also thought about what it would look like to switch it over to using raw pointers. I think that's a viable option (and skips the pain of creating an initially zeroed-out array), but I didn't want to get waist deep with changing `SecureRandom` on my first contribution.

Cassy343:
Is there any interest in moving the issue this PR addresses forward or should I close this PR?

briansmith:
Thanks for the PR. See https://github.com/rust-random/getrandom/pull/293 and the discussion in the related `getrandom` issue, where I discuss some of the API considerations and propose an alternative to defining a `Pod`/zerocopy/bytemuch-style API. I'm going to close this in favor of moving in that direction.

PR #1434, opened 2021-12-10
constant_time: Rewrite (mostly) in Rust

briansmith:


ShadowJonathan (approved):
Godbolt output of this looks good; https://godbolt.org/z/fv3EbWPnT

Though, take in mind that I've removed the `CONSTANT_TIME_value_barrier_w` call there for simplicity, but it seems that is also good enough.

I believe that `CONSTANT_TIME_value_barrier_w` can be replaced with an `asm!` call in the future, if it ever needs be.

Though I believe with some further research into LLVM and Rust's compilation attributes, we can see if this is possible in pure and safe rust.

---

In the future, we should look into build-time checking of the ASM generates by these functions to really verify if they're constant-time, so that at least if the compiler optimises it away despite all protections, the build will fail, preventing the worst vulnerabilities to surface.

Or we could optionally just make this all-`asm!` for these kinds of functions.

PR #1435, opened 2021-12-10
cpu: Split `cpu.rs` submodules into their own files.

briansmith:
No functional change.


briansmith:
@complexspaces  @ShadowJonathan PTAL. This is a prerequisite for replacing crypto/cpu-intel.c with Rust code.

ShadowJonathan (commented):
I'm not sure what i can contribute to this at the moment, i havent gotten as far as to know the ins and outs of every bit of the library (yet), but this LGTM, a simple move of platform-specific code into separate files.

complexspaces (approved):
This looks fine to me as well :thumbsup:

briansmith:
Not to get too meta, but did anybody actually check if the code is the same between the old and the new? It is (at least I intend it to be) a simple move and rustfmt.

ShadowJonathan:
I did compare some sections, but I admit, I didn't closely study if there were any (subtle) changes, I was just interested in getting the gist of the change :sweat_smile:

complexspaces:
I used `git difftool` to look between the two files on each revision. The first one lost `#[cfg(attr)]` blocks and a lot of whitespace changes, and removed the ARM code. The second commit was a clean split of the ARM code into its own file and removal of more `#[cfg()` blocks in the module.

Issue #1419, opened 2021-10-24
Build for riscv64gc-unknown-linux-gnu failed.


xfbs:
Steps to reproduce. I did this on latested Debian, use a docker container if you don't have it.

```
apt update
apt install gcc-riscv64-linux-gnu g++-riscv64-linux-gnu 
export CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER=/usr/bin/riscv64-linux-gnu-gcc
cargo build --target riscv64gc-unknown-linux-gnu
cargo build --target riscv64gc-unknown-linux-gnu --tests
```

I'm able to fix some simple compilation errors in the beginning by adding RISCV to the architectures:

```
diff --git a/include/ring-core/base.h b/include/ring-core/base.h
index f1a027d1a..92c9a4372 100644
--- a/include/ring-core/base.h
+++ b/include/ring-core/base.h
@@ -91,6 +91,10 @@
 #define OPENSSL_MIPS64
 #elif defined(__wasm__)
 #define OPENSSL_32_BIT
+#elif __riscv && __riscv_xlen == 64
+#define OPENSSL_64_BIT
+#elif __riscv && __riscv_xlen == 32
+#define OPENSSL_32_BIT
 #else
 // Note BoringSSL only supports standard 32-bit and 64-bit two's-complement,
 // little-endian architectures. Functions will not produce the correct answer
```

This is all I've gotten to so far.

kirana1996:
@xfbs,
any update.

briansmith:
We have Rust or C implementations for almost every algorithm, so most things work on every architecture. Unfortunately, there are one or two cases where we don't have an architecture-neutral implementation yet.

In the case of RISC-V, because it doesn't have a carry flag at all, we may also need to implement some bignum algorithms differently, since the current C implementations are optimized for having a carry flag presently.

xfbs:
Anything I can do to help? I'm not very familiar with the codebase, but I think if you point me in the right direction I can possibly figure this out. I have RISC-V hardware that we can use for testing things (but obviously cross-compilation + qemu-user-static works, too). Am hoping to be able to help get ring working because I'd like to be able to build some stuff for RISC-V that depend on it!

kirana1996:
@xfbs 

     While trying to run  a sample ring application, I am facing the error as below,

error: failed to run custom build command for `ring v0.16.20`

Caused by:
[...]

PR #1436, opened 2021-12-11
Hacks taken from @amock to make RISC-V build.

xfbs:
As stated in #1419, trying to get RISC-V to build. This uses generic C code for montomery multiplication taken from @amock. But it builds and tests successfully like this.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -30,6 +30,8 @@ const X86: &str = "x86";
 const X86_64: &str = "x86_64";
 const AARCH64: &str = "aarch64";
 const ARM: &str = "arm";
+const RISCV64: &str = "riscv64";
+const RISCV32: &str = "riscv32";
 
 #[rustfmt::skip]
 const RING_SRCS: &[(&[&str], &str)] = &[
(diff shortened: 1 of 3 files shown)


briansmith:
Thanks for the PR! Could you rebase this on top of #1558? I think we just need the base.h changes now.

briansmith:
Also, could you add the changes to the scripts in mk/ to get `cargo.sh test` working for some RISC-V targets in QEMU? This will enable us to more easily work on adding RISC-V-specific code paths in the future.

olof-nord:
This PR allows me to build ring locally, thanks!
But with this branch several years old, it is 442 commits behind :)

It would be fantastic to get this working on riscv - for example to run lemmy on riscv hardware.

I understand the generic approach suggested there, I will keep my eyes open for updates!
https://github.com/briansmith/ring/issues/1455

briansmith:
Thanks. For now I'm planning to go with the simpler PR #1627. If you want, we can revisit the extended changes you have here, afterward.

PR #1437, opened 2021-12-14
cpu: Move some of the CPUID logic to Rust.

briansmith:
Separate the generic CPUID logic from the OpenSSL-specific logic for constructing `OPENSSL_ia32cap_P`. Rewrite the generic logic in Rust.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -926,7 +926,7 @@ fn prefix_all_symbols(pp: char, prefix_prefix: &str, prefix: &str) -> String {
         "LIMBS_window5_unsplit_window",
         "LIMB_shr",
         "OPENSSL_armcap_P",
-        "OPENSSL_cpuid_setup",
+        "OPENSSL_ia32cap_init",
         "OPENSSL_ia32cap_P",
         "OPENSSL_memcmp",
         "aes_hw_ctr32_encrypt_blocks",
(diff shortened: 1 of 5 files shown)


briansmith:
@ShadowJonathan PTAL. This demonstrates some of the non-technical difficulties (license) with the process of replacing the remaining C code with Rust.

ShadowJonathan:
> This demonstrates some of the non-technical difficulties (license) with the process of replacing the remaining C code with Rust.

I'm not seeing those difficulties in action in this PR necessarily, could you explain them to me?

briansmith:
Closing this. The original idea was to separate the "obvious" and non-OpenSSL/BoringSSL-specific bits from the OpenSSL/BoringSSL-specific bits. But on its own it doesn't really accomplish much. Also testing the logic of `is_intel` and similar is tricky when we don't all reliably have access to Intel (or AMD) CPUs.

I think the right way to do this is to have a safe abstraction for the CPUID instruction. But then using our safe abstraction would differ from what BoringSSL does because it unconditionally queries CPUID leaf 0 and leaf 1. So then we have to choose between the safe abstraction or doing exactly what BoringSSL does. In order to get away from doing exactly what BoringSSL does we need to have more (all?) of the CPU feature detection logic in Rust and none of it in assembly. Then we'll know enough about exactly what the assembly code is doing to determine how to deal with situations where we can't query leaf 0 and/or 1.

PR #1440, opened 2021-12-15
signature: Enable Ed25519 support for wasm32 targets.

briansmith:
Ed25519 was disabled for WebAssembly due to some unrelated issues with
getting the X25519 code working in WebAssembly. Temporarily remove the
`agreement` API when targetting WebAssembly to work around those issues
in a way that lets us enabled Ed25519.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -33,6 +33,7 @@ const ARM: &str = "arm";
 
 #[rustfmt::skip]
 const RING_SRCS: &[(&[&str], &str)] = &[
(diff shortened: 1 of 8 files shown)


ivanschuetz:
Any updates? We're using this in an SDK https://github.com/manuelmauro/algonaut/blob/main/algonaut_transaction/src/account.rs#L35 and it would be great if it didn't break WASM compatibility..

PR #1456, opened 2022-02-04
Add support for EC private keys

farcaller:
This adds support for [EC PRIVATE KEY](https://datatracker.ietf.org/doc/html/rfc5915), i.e. non pkcs8-wrapped keys.


ctz (approved):

farcaller:
Tentative ping.

cpu:
@briansmith Would you be open to this feature if it were rebased on main? We have [a bit of a hack upstream](https://github.com/rustls/rustls/pull/998) in rustls patching over this and it might be nice to drop it one day.

briansmith:
@cpu Yes, I would. Would you do RSA and EdDSA too?

farcaller:
I don’t mind adding the others too. Let me see if I remember how it works, was a while since I touched this code.

cpu:
Great thank you. Happy to support as appropriate.

farcaller:
@briansmith do you know any good producers of RSA keys? I thought ssh ones would fit but they are actually in a different format.

briansmith:
> @briansmith do you know any good producers of RSA keys? I thought ssh ones would fit but they are actually in a different format.

Does this work? https://gist.github.com/briansmith/2ee42439923d8e65a266994d0f70180b#extracting-an-rsa-public-key-from-the-private-key-without-the-subjectpublickeyinfo-metadata

farcaller:
Sorry, I was out of free cycles the previous week to follow up on this.

@briansmith do you mind those two being decoupled, given how rebase was pretty smooth? I think the ec keys can go forward straight on and unblock rustls, and I'll work on the RSA keys in a separate PR then.

briansmith:
@farcaller I looked more into how Rustls is using the existing APIs. I think we should "just" refactor the *ring* PKCS#8 API so that it works more like what Rustls needs, to avoid Rustls needing to jump through hoops. Please see https://github.com/briansmith/ring/issues/1889.

PR #1465, opened 2022-02-21
Use LLVM 14 in CI/CD for coverage.

briansmith:


briansmith:
I investigated Codecov's complaints about decreased coverage, but I didn't see anything significant.

PR #1466, opened 2022-02-21
Test both variants of Ed25519 PKCS#8 parser with all test case inputs

briansmith:


briansmith:
The test code coverage report shows a notably improvement, though it shows things aren't perfect either.

PR #1489, opened 2022-04-26
Added feature to allow ring to work under cosmwasm.

rhomber:
As per my comment on the related issue, not sure if you'd like to merge this (or provide something similar). But this allows me to include ring as it produces no wasm exports.


janik-cloudflare:
Related: #1486

briansmith:
Does cosmwasm have any API analogous to [wasi::random_get](https://docs.rs/wasi/0.9.0+wasi-snapshot-preview1/wasi/fn.random_get.html)? If so, could you add `cosmwasm` support to the `getrandom` crate? Then we could support this target very easily.

briansmith:
What is the Rust target that is used for cosmwasm? wasm32-unknown-unknown? wasm32-wasi?

PR #1493, opened 2022-05-11
Add `sign_digest_less_safe` to rsa `KeyPair`

reduf:
The goal of this MR is to allow a user to sign a digest without having to have all the message at disposition. I tried to follow the advises given in [MR-915](https://github.com/briansmith/ring/pull/915), for instance the name of the function.

In addition, I'm not convince the testing code I added is the best or most elegant way to do it, but I didn't want to add custom stuff to support testing. Arguably, it's a little awkward to specify the digest algorithm twice. You might have some recommendation on that.

### DIFF
--- a/src/rsa/keypair.rs
+++ b/src/rsa/keypair.rs
@@ -575,9 +575,31 @@ impl KeyPair {
 
         let m_hash = digest::digest(padding_alg.digest_alg(), msg);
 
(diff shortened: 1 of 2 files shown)


reduf:
Closed since no reaction and don't want to maintain the branch.

PR #1495, opened 2022-05-14
Add `Digest::from_bytes`

reduf:
`Digest::from_bytes` can be used to create a `Digest` structure from an external source, without any knowledge of the digested message.

This can be particularly useful in combination with MR #1493 to sign a digest on behalf of a user authenticated through other means without exposing the private key of the user.

### DIFF
--- a/src/digest.rs
+++ b/src/digest.rs
@@ -239,6 +239,28 @@ pub struct Digest {
 }
 
 impl Digest {
+    /// Create a `Digest` from an algorithm and an already computed digest.
+    ///
+    /// The `data` needs to match the expected output length of the `algorithm`.
+    pub fn from_bytes(
+        algorithm: &'static Algorithm,
+        data: &[u8]
+    ) -> Result<Digest, crate::error::Unspecified> {
+        if algorithm.output_len != data.len() {
+            return Err(crate::error::Unspecified);
+        }
+
+        const SIZE: usize = 512 / 8 / core::mem::size_of::<BigEndian<u64>>();
+        let mut value = Output {
+            as64: [BigEndian::<u64>::from(0); SIZE]
+        };
+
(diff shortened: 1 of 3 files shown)


src/digest.rs
```diff
@@ -239,6 +239,28 @@ pub struct Digest {
...
+    /// The `data` needs to match the expected output length of the `algorithm`.
+    pub fn from_bytes(
+        algorithm: &'static Algorithm,
+        data: &[u8]
+    ) -> Result<Digest, crate::error::Unspecified> {
+        if algorithm.output_len != data.len() {
+            return Err(crate::error::Unspecified);
+        }
+
+        const SIZE: usize = 512 / 8 / core::mem::size_of::<BigEndian<u64>>();
```
reduf, line 253: That's a bit dodgy and there is certainly a much better way. Do you have a recommendation?

briansmith:
I'm not excited about this idea. The idea of `Digest` is that when you have a `Digest`, you are certain that the digest was computed from ourselves. That means that we know (could have known) the value that was digested. If you just have a bunch of bytes that happen to be the right length for the algorithm you are interested in, you don't know what value was digested. 

What is the use case you're trying to address?

reduf:
In my case, it's impractical to have the original message, because it can be 100s of MBs, on a different machine. Furthermore, I can't send the private key to the machine with the data, for security reasons.

The goal is to split the digesting of this message and the signing of the message, which can be done with this PR in combination with #1493.

reduf:
I double checked and it was indeed an incorrect assumption, because the digest received doesn't need to be encrypted, but instead the digest of a "document" containing this digest. This mean that we still compute the digest ourselves.

Anyway, thanks for your time and have a nice day!

Issue #1455, opened 2022-02-04
C (or Rust) fallback for all functions that work for all little-endian targets


ellenhp:
Related: #1419

Just want to start a conversation around portability to embedded architectures.

Despite having a lot of assembly, ring actually pretty close to being portable. All that's preventing it from building on architectures like RISC-V and xtensa is a Rust or pure-C implementation of what appears to be a montgomery multiplier. There's an open PR (#1436) to add one (written in C) but I can't figure out where the code actually came from, and there are no tests so I don't really trust it.

I'm pretty motivated to get crypto stuff working for an ESP32 project of mine, so let me know if I can help make this happen. I'd be up for trying to write a C or Rust fallback for this function, or writing tests for the existing one, trying to analyze it with KLEE to prove its memory safety and constant-time properties, etc, whatever's needed to meet code standards.

It would be amazing to have the entire rust ecosystem of crypto libraries working on the ESP32 especially if I don't have to maintain a patch like I am right now. :)

ellenhp:
Also related to #1297 it seems!

briansmith:
I am generalizing this issue to be about automatically supporting every little-endian architecture with a generic fallback. Big-endian architectures are tracked separately in #1555; the work in supporting them is a superset of the work in supporting the little-endian architectures.

Here's what we need to do, AFAICT:
* Have a C or Rust fallback for each of the few functions that are currently only implemented in Rust. In the case of `bn_mul_mont` we may just want to replace the callers with callers that call the Rust equivalent, or implement `bn_mul_mont` in terms of that Rust code.
* Replace the allowlist of target architectures in `include/ring-core/base.h` with a more general mechanism; see below.
* Run tests in CI (probably using qemu like we do for 32-bit ARM and Aarch64) for an architecture that doesn't have any assembly language implementations in use. I suggest wasm32-wasi.

Regarding the allowlist in `include/ring-core/base.h`: It looks like this:
```
#if defined(__x86_64) || defined(_M_AMD64) || defined(_M_X64)
#define OPENSSL_64_BIT
#define OPENSSL_X86_64
#elif defined(__x86) || defined(__i386) || defined(__i386__) || defined(_M_IX86)
#define OPENSSL_32_BIT
#define OPENSSL_X86
#elif defined(__AARCH64EL__) || defined(_M_ARM64)
#define OPENSSL_64_BIT
#define OPENSSL_AARCH64
#elif defined(__ARMEL__) || defined(_M_ARM)
[...]

PR #1500, opened 2022-06-02
Adding RISCV from #1419

fede2cr:
Copy paste from @xfbs, to fix bug #1419.

Built on Fedora-riscv64, and run the tests. It is working fine as a dependency for sccache, and compiled without any errors at all.


briansmith:
I don't think this PR is enough to actually get RISC-V working because it is missing a solution to the `bn_mul_mont` problem. I outlined a more general solution in issue #1455 that should cover RISC-V, and I'm going to close this issue so we can focus on that more general approach.

PR #1503, opened 2022-06-27
Expose RSA_PKCS1_SHA1 for RSA signing

kjvalencik:
`RSA_PKCS1_SHA1_FOR_LEGACY_USE_ONLY` is currently _intentionally_ not available for RSA signing (only verification). However, it may be necessary when interacting with services that only support RSA SHA1 (e.g., Azure DevOps). This PR proposes exposing it publicly.

There were no existing SHA1 test vectors because they were removed in 2015 are not present in FIPS 186-4. I updated `convert_nist_rsa_test_vectors.py` to be able to accept an algorithm filter and generated a new test vector file with the SHA1 test vectors from FIPS 186-2.


briansmith:
@kjvalencik Do you still need this? According to https://stackoverflow.com/a/60417115 Azure DevOps now does support better algorithms.

kjvalencik:
Yes, this is still needed. Even though Azure DevOps now supports `rsa-sha2-256` and `rsa-sha2-512` there are still many use cases that do not. For example, older switches and routers.

nealwon:
Is this PR still active?

We strongly need this feature when interaction with some old systems.

kjvalencik:
@nealwon I still need it and have been using a fork for it. Hopefully it can be eventually merged.

PR #1459, opened 2022-02-05
Add support for ESP32 and RISC-V

ellenhp:
Related to #1436 and #1297 

Fixes #1455
Will likely address #1419 but I don't know for sure.

A few points:
* The code for bn_mult_mont is from #1297
* I know you expressed [reluctance to accepting variable-length arrays](https://github.com/briansmith/ring/pull/1297/files#r657301788) in the past, so let me know if you want that changed to something else.
* Embedded targets are usually compiled with size-optimization so the compiler is very, very reluctant to inline things like the value barrier code in `crypto/internal.h`. Since we treat warnings as errors I needed to add a way to disable the `-Winline` flag.
* I'm adding architecture support here, and also adding support for the ESP-IDF RTOS. Architecture support seems fairly uncontroversial if the code meets standards, but I understand if you're reluctant to add support for a new OS target for maintenance burden reasons. Just let me know and I can yank that part and maintain it myself as a patch. I'd obviously prefer it we could upstream it though.
* This has been lightly tested through rustls on both xtensa based ESP32 micros and RISC-V based ESP32's, but we haven't been able to run a full test suite on them on account of not having a full general purpose OS, so no `cargo test`. Let me know if this is a blocker and I'll figure out how to at least make the tests work on RISC-V. I think linux on RISC-V would work in qemu or something.

PR #1506, opened 2022-07-27
Support Xtensa, RISC-V and ESP-IDF

MabezDev:
This is a continuation of the upstreaming effort from https://github.com/briansmith/ring/pull/1459.

We would really like some feedback on this PR, we wish to use `ring` to secure millions of Espressif devices. If could let us know what reservations you have about this PR, I would be more than happy to address them.

Quoting from the original PR for comments about the PR itself:

A few points:

- The code for bn_mult_mont is from https://github.com/briansmith/ring/pull/1297
- I know you expressed [reluctance to accepting variable-length arrays](https://github.com/briansmith/ring/pull/1297/files#r657301788) in the past, so let me know if you want that changed to something else.
- Embedded targets are usually compiled with size-optimization so the compiler is very, very reluctant to inline things like the value barrier code in crypto/internal.h. Since we treat warnings as errors I needed to add a way to disable the -Winline flag.
- I'm adding architecture support here, and also adding support for the ESP-IDF RTOS. Architecture support seems fairly uncontroversial if the code meets standards, but I understand if you're reluctant to add support for a new OS target for maintenance burden reasons. Just let me know and I can yank that part and maintain it myself as a patch. I'd obviously prefer it if we could upstream it though.
- This has been lightly tested through rustls on both xtensa based ESP32 micros and RISC-V based ESP32's, but we haven't been able to run a full test suite on them on account of not having a full general purpose OS, so no cargo test. Let me know if this is a blocker and I'll figure out how to at least make the tests work on RISC-V. I think linux on RISC-V would work in qemu or something.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -167,7 +167,7 @@ untrusted = { version = "0.9" }
 [target.'cfg(any(target_arch = "x86",target_arch = "x86_64", all(any(target_arch = "aarch64", target_arch = "arm"), any(target_os = "android", target_os = "fuchsia", target_os = "linux", target_os = "windows"))))'.dependencies]
 spin = { version = "0.9.2", default-features = false, features = ["once"] }
 
-[target.'cfg(any(target_os = "android", target_os = "linux"))'.dependencies]
+[target.'cfg(any(target_os = "android", target_os = "linux", target_os = "espidf"))'.dependencies]
 libc = { version = "0.2.100", default-features = false }
 once_cell = { version = "1.8.0", default-features = false, features=["std"], optional = true }
 
@@ -201,6 +201,7 @@ slow_tests = []
 std = ["alloc"]
 test_logging = []
 wasm32_unknown_unknown_js = ["web-sys"]
+size_optimized = []
 
 # XXX: debug = false because of https://github.com/rust-lang/rust/issues/34122
 
--- a/build.rs
+++ b/build.rs
@@ -41,11 +41,11 @@ const RING_SRCS: &[(&[&str], &str)] = &[
     (&[], "crypto/mem.c"),
     (&[], "crypto/poly1305/poly1305.c"),
 
-    (&[AARCH64, ARM, X86_64, X86], "crypto/crypto.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/ecp_nistz.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/gfp_p256.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/gfp_p384.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/p256.c"),
+    (&[], "crypto/crypto.c"),
+    (&[], "crypto/fipsmodule/ec/ecp_nistz.c"),
+    (&[], "crypto/fipsmodule/ec/gfp_p256.c"),
+    (&[], "crypto/fipsmodule/ec/gfp_p384.c"),
+    (&[], "crypto/fipsmodule/ec/p256.c"),
 
     (&[X86_64, X86], "crypto/cpu-intel.c"),
 
@@ -126,6 +126,7 @@ fn cpp_flags(compiler: &cc::Tool) -> &'static [&'static str] {
             "-Wenum-compare",
             "-Wfloat-equal",
             "-Wformat=2",
+            #[cfg(not(feature = "size_optimized"))]
             "-Winline",
             "-Winvalid-pch",
             "-Wmissing-field-initializers",
--- a/crypto/fipsmodule/bn/montgomery.c
+++ b/crypto/fipsmodule/bn/montgomery.c
@@ -156,3 +156,17 @@ int bn_from_montgomery_in_place(BN_ULONG r[], size_t num_r, BN_ULONG a[],
   }
   return 1;
 }
+
+#if !defined(OPENSSL_X86) && !defined(OPENSSL_X86_64) && \
+    !defined(OPENSSL_ARM) && !defined(OPENSSL_AARCH64)
+void bn_mul_mont(BN_ULONG *rp, const BN_ULONG *ap, const BN_ULONG *bp,
+                 const BN_ULONG *np, const BN_ULONG *n0, size_t num) {
+  Limb tmp[2 * num];
+  for (size_t i = 0; i < num; i++)
+    tmp[i] = 0;
+  for (size_t i = 0; i < num; i++)
+    tmp[num + i] = limbs_mul_add_limb(tmp + i, ap, bp[i], num);
+
+  bn_from_montgomery_in_place(rp, num, tmp, 2 * num, np, num, n0);
+}
+#endif
--- a/include/ring-core/base.h
+++ b/include/ring-core/base.h
@@ -91,6 +91,12 @@
 #define OPENSSL_MIPS64
 #elif defined(__wasm__)
 #define OPENSSL_32_BIT
+#elif defined(__xtensa__)
+#define OPENSSL_32_BIT
+#elif defined(__riscv) && __riscv_xlen == 64
+#define OPENSSL_64_BIT
+#elif defined(__riscv) && __riscv_xlen == 32
+#define OPENSSL_32_BIT
 #else
 // Note BoringSSL only supports standard 32-bit and 64-bit two's-complement,
 // little-endian architectures. Functions will not produce the correct answer
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -169,6 +169,7 @@ impl crate::sealed::Sealed for SystemRandom {}
         not(feature = "dev_urandom_fallback")
     ),
     target_arch = "wasm32",
+    target_os = "espidf",
     windows
 ))]
 use self::sysrand::fill as fill_impl;
@@ -223,6 +224,21 @@ mod sysrand_chunk {
                 // will cause the caller to try again.
                 return Ok(0);
             }
+            return Err(error::Unspecified);
+        }
+        Ok(r as usize)
+    }
+}
+
+#[cfg(target_os = "espidf")]
+mod sysrand_chunk {
+    use crate::{c, error};
+
+    #[inline]
+    pub fn chunk(dest: &mut [u8]) -> Result<usize, error::Unspecified> {
+        let chunk_len: c::size_t = dest.len();
+        let r = unsafe { libc::getrandom(dest.as_mut_ptr() as *mut libc::c_void, chunk_len, 0) };
+        if r < 0 {
             return Err(error::Unspecified);
         }
         Ok(r as usize)
@@ -286,6 +302,7 @@ mod sysrand_chunk {
     target_os = "android",
     target_os = "linux",
     target_arch = "wasm32",
+    target_os = "espidf",
     windows
 ))]
 mod sysrand {


bunnie:
This patch would also be extremely enabling for Xous on Precursor! We're trying to add secure messaging to our RISC-V mobile platform and getting `rustls` to compile past the `ring` dependency is a major stumbling block. Something like this would allow us to move ahead on application development for our end users.

Thanks for opening the PR @MabezDev, I really appreciate it!

Janrupf:
We are also currently using a patched version of ring for Espressif based platforms - would really love to see this getting upstreamed!

My thoughts regarding the dynamically sized arrays:
I don't really know where the size comes from or how large it can get - but I think if the maximum is reasonably small it would make sense to keep this as a static stack array in combination with a check. The comment above the function in the header seems to suggest that the function has no way of signalling an error, so I'm not sure how a size check would look in this context.

MabezDev:
@briansmith would you mind taking a look at this PR when you get a few moments? It would be really great to get some feedback :).

Darkspirit:
Just have seen https://twitter.com/feistyduck/status/1577644279740387328:
This PR is linked in this article: https://www.crowdsupply.com/sutajio-kosagi/precursor/updates/bringing-up-tls-on-precursor
They have made a temporary pure-Rust fork of ring with c2rust: https://github.com/betrusted-io/ring-xous/tree/0.16.20-cleanup/src/c2rust

juztamau5:
I can confirm that this patch fixed building ring for RISC-V for me. I also had to apply this patch:

```patch
--- a/build.rs
+++ b/build.rs
@@ -236,6 +236,13 @@ const ASM_TARGETS: &[AsmTarget] = &[
         asm_extension: "S",
         preassemble: true,
     },
+    AsmTarget {
+        oss: LINUX_ABI,
+        arch: "riscv64",
+        perlasm_format: "linux64",
+        asm_extension: "S",
+        preassemble: false,
+    },
 ];
 
 struct AsmTarget {
```

I was also building for a non-standard RISC-V arch and had to do:

```bash
export CRATE_CC_NO_DEFAULTS=1
```

before calling cargo build.

src/rand.rs
```diff
@@ -229,6 +230,21 @@ mod sysrand_chunk {
     }
 }
 
+#[cfg(target_os = "espidf")]
```
briansmith, line 233: This part should be updated to just add `target_os = "espidf"` to the allowlist in the new version of rand.rs, after reviewing that the `getrandom` crate implements this functionality for this operating system correctly. Please do this change in a separate PR.

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -156,3 +156,17 @@ int bn_from_montgomery_in_place(BN_ULONG r[], size_t num_r, BN_ULONG a[],
   }
   return 1;
 }
+
+#if !defined(OPENSSL_X86) && !defined(OPENSSL_X86_64) && \
+    !defined(OPENSSL_ARM) && !defined(OPENSSL_AARCH64)
+void bn_mul_mont(BN_ULONG *rp, const BN_ULONG *ap, const BN_ULONG *bp,
```
briansmith, line 162: Let's solve this in a PR that isn't specific to any OS/architecture. I will contact the original author of this code to get it submitted so there's no IPR concerns.

build.rs
```diff
@@ -41,11 +41,11 @@ const RING_SRCS: &[(&[&str], &str)] = &[
     (&[], "crypto/mem.c"),
     (&[], "crypto/poly1305/poly1305.c"),
 
-    (&[AARCH64, ARM, X86_64, X86], "crypto/crypto.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/ecp_nistz.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/gfp_p256.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/gfp_p384.c"),
-    (&[AARCH64, ARM, X86_64, X86], "crypto/fipsmodule/ec/p256.c"),
+    (&[], "crypto/crypto.c"),
```
briansmith, line 44: I don't think a change for crypto.c is needed.

Cargo.toml
```diff
@@ -167,7 +167,7 @@ untrusted = { version = "0.9" }
 [target.'cfg(any(target_arch = "x86",target_arch = "x86_64", all(any(target_arch = "aarch64", target_arch = "arm"), any(target_os = "android", target_os = "fuchsia", target_os = "linux", target_os = "windows"))))'.dependencies]
 spin = { version = "0.9.2", default-features = false, features = ["once"] }
 
-[target.'cfg(any(target_os = "android", target_os = "linux"))'.dependencies]
+[target.'cfg(any(target_os = "android", target_os = "linux", target_os = "espidf"))'.dependencies]
```
briansmith, line 170: I don't think this change will be needed once you update the `rand.rs` changes.

Cargo.toml
```diff
@@ -201,6 +201,7 @@ slow_tests = []
 std = ["alloc"]
 test_logging = []
 wasm32_unknown_unknown_js = ["web-sys"]
+size_optimized = []
```
briansmith, line 204: I'd rather not control this with a feature flag. Perhaps you can work around this with `CFLAGS` in your build environment for now?

include/ring-core/base.h
```diff
@@ -91,6 +91,12 @@
 #define OPENSSL_MIPS64
 #elif defined(__wasm__)
 #define OPENSSL_32_BIT
+#elif defined(__xtensa__)
+#define OPENSSL_32_BIT
+#elif defined(__riscv) && __riscv_xlen == 64
+#define OPENSSL_64_BIT
+#elif defined(__riscv) && __riscv_xlen == 32
+#define OPENSSL_32_BIT
```
briansmith, line 99: Please reconsider this in light of the approach I suggest in #1455.

briansmith (changes requested):
Thanks for the PR. Please see the inline comments. I'd like to separate the PR that adds support for the OS from the PR that adds the fallback implementation, and I'd like us to implement the approach outlined in #1455 if at all practical.

We should explore some alternate solution to the '-Winline` issue as I understand why it is problematic for you, but also I do find the warnings to be very helpful. Perhaps we should just remove `-Winline` for now.

briansmith:
@MabezDev Thanks again for the PR. Because my comments are tantamount to "redo the whole thing and split it up into multiple PRs," and because you aren't the original author of the `bn_mult_mont` code,  I don't think this PR makes sense any more. I'm going to close it so we can concentrate on resolving #1455 and then rebasing the remaining work on top of that solution.

reinismu:
@MabezDev Been wanting to try use rustls in my ESP32S3 project. First had to port these changes to version `0.16.20` (https://github.com/reinismu/ring/commit/d82e61ddb27c0485c5a405e0d97144864f1f417b). It compiled fine, but at linker stage it fails with:

```
xtensa-esp32s3-elf/bin/ld: /home/reinis/projects/hub/target/xtensa-esp32s3-espidf/release/deps/libring-01be472a3cba49f7.rlib: error adding symbols: file format not recognized
```

<details>
  <summary>objdump</summary>
  
  ```
  lib.rmeta:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000010:
HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.0.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.1.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.10.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.11.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.12.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.13.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.14.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.15.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.2.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.3.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.4.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.5.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.6.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.7.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.8.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


ring-01be472a3cba49f7.ring.54f8b82c-cgu.9.rcgu.o:     file format elf32-little
architecture: UNKNOWN!, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x00000000


aes_nohw.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


montgomery.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


montgomery_inv.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


limbs.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


mem.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


poly1305.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


crypto.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


curve25519.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


ecp_nistz.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


ecp_nistz256.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


gfp_p256.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


gfp_p384.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000


constant_time_test.o:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000011:
HAS_RELOC, HAS_SYMS
start address 0x0000000000000000
  ```
  
</details>

Here I notice `architecture: i386:x86-64` could that be the cause? If that is the case how do I fix it? :thinking:

PR #1513, opened 2022-09-11
Support powerpc64-ibm-aix target

ecnelises:
Add support for both 64-bit PowerPC and AIX OS. The Perl scripts are from upstream OpenSSL. Keep this in draft status until it's been official Rust target.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ include = [
     "crypto/cipher_extra/test/aes_256_gcm_siv_tests.txt",
     "crypto/constant_time_test.c",
     "crypto/cpu-intel.c",
+    "crypto/cpu-ppc.c",
     "crypto/crypto.c",
     "crypto/curve25519/asm/x25519-asm-arm.S",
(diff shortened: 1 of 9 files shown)


briansmith:
I think we should de-conflate two issues: Big-endian PowerPC support and AIX support.

Regarding supporting any Big-endian target architectures at all, first I want us to resolve #1555 so that we have a general fallback implementation that works for all big-endian target architectures. Once #1555 is resolved then all PowerPC targets should "just work" as long as the operating system is allowlisted (see next item).

Regarding support for the AIX operating system, this should just be a matter of submitting a PR that updates the allowlist in rand.rs to include "aix" after verifying that the `getrandom` crate does use the best operating system RNG API API on AIX.

After those two things are done, then we can (re)consider whether we want to use the PowerPC optimized assembly implementations from BoringSSL.

briansmith:
Thanks for the PR. I'm going to close this since ultimately it will need to be broken up into multiple PRs to separately address each of the aspects mentioned in the preceding comment.

PR #1517, opened 2022-09-22
Add loongarch64 support

zhaixiaojuan:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html

### DIFF
--- a/crypto/fipsmodule/bn/asm/loongarch-mont.pl
+++ b/crypto/fipsmodule/bn/asm/loongarch-mont.pl
@@ -1,0 +1,339 @@
+#! /usr/bin/env perl
+# Copyright 2010-2018 The OpenSSL Project Authors. All Rights Reserved.
+#
+# Licensed under the OpenSSL license (the "License").  You may not use
+# this file except in compliance with the License.  You can obtain a copy
+# in the file LICENSE in the source distribution or at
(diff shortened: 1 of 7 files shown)


xen0n:
Please wait before the port is even merged into Rust upstream: rust-lang/rust#96971.

briansmith:
Thanks for the PR. Please factor out the changes to base.h and the scripts in `mk/` into a separate PR so that this target will start working after PR #1558. It is unlikely I will be able to accept the assembly code in this PR any time soon because I don't have time to learn how to review the assembly language for this. I suggest instead we work together on improving the performance of the Rust/C implementation of `bn_mul_mont` so that we don't need the assembly code.

zhaixiaojuan:
I've factored out the changes to base.h and the scripts in mk/ into a separate PR:
https://github.com/briansmith/ring/pull/1565
@briansmith @xen0n

PR #1518, opened 2022-09-23
Fix Ed25519 PKCS#8 v2 generation to make it parse with `der` >=0.5

### LINKED ISSUE #1464: Ed25519 `pkcs8` documents do not have RFC-valid public key part
I discovered this while working with `pkcs8` 0.8;

In the key template, `ring` prefixes the public key bit with `A1 23 03 21` (Context Specific `[1]`, then `BITSTRING`), while the [RFC](https://datatracker.ietf.org/doc/html/rfc8410#section-10.3) works with "Context Specific `[1]`" *only*, the parser is then supposed to infer the following data is a `BITSTRING`, without prefix.

---

Furthermore, together with https://github.com/briansmith/ring/issues/1299 and https://github.com/briansmith/ring/issues/834, i strongly recommend switching to a pkcs8-parser library, instead of manually wrapping and generating keys like this.

NoelTautges:
Currently, the Ed25519 PKCS#8 v2 template encodes the public key bitstring in its constructed form, something not allowed by the DER spec, which causes an error when attempting to parse it with `der` >=0.5 thanks to its stricter parsing rules. More details in [my comment on relevant issue #1464](https://github.com/briansmith/ring/issues/1464#issuecomment-1255412858).

This PR instead encodes the bitstring in its primitive form, which makes parsing in `der` >=0.5 work, but breaks parsing in `der` <0.5 because [`der` didn't have support for that until 0.5](https://github.com/RustCrypto/formats/commit/923727c6629d0dc6dde4bca5e99c7f0df9ec5158#diff-c2abe9f690e21bd07d196a983b13fcfd82dc9a7272d071060f908e08f6a5273dL172) (the relevant tag number is 0x81).

And I'd like to second at least exploring the option of outsourcing DER/PKCS#8 encoding to another crate. It may be a loss for compile times, but it may be a benefit overall.

Resolves #1464.

`pkcs8` 0.9 PoC that no longer panics with this fix:

```rust
use base64::display::Base64Display;
use pkcs8::der::Decode;
use pkcs8::PrivateKeyInfo;
use ring::rand::SystemRandom;
use ring::signature::Ed25519KeyPair;

fn main() {
    let system_random = SystemRandom::new();
    let key_pair_doc = Ed25519KeyPair::generate_pkcs8(&system_random).unwrap();
    let key_pair_serialized = key_pair_doc.as_ref();
    let wrapper = Base64Display::with_config(key_pair_serialized, base64::STANDARD);
    println!("{}", wrapper);
    PrivateKeyInfo::from_der(key_pair_serialized).unwrap();
}
```


chanced:
Any chance this will land at some point? *ring* not being compatible with the `pkcs8` is unfortunate.

By the way, to the authors of *ring*, thank you for the crate.

bkstein:
We'd also appreciate a fix. For the time being, it is not possible to create valid ed25519 key pairs with _rcgen_, too. _rcgen_ also relies on _ring_ for key generation. See my [closed PR in RustCrypto/formats#948](https://github.com/RustCrypto/formats/issues/948), which was placed in the wrong project.

bkstein:
I have cherry-picked the relevant commits from @NoelTautges ' branch. I can confirm, that generating and parsing `KeyPair`s works with the new template using rcgen, too:
```rust
fn main() {
    let key_pair = rcgen::KeyPair::generate(&rcgen::PKCS_ED25519).unwrap();
    let key_pair_der = key_pair.serialize_der();
    let _key_pair_pkcs8 = pkcs8::PrivateKeyInfo::from_der(key_pair_der.as_slice()).unwrap();
}
```
This is similar to the PoC mentioned in @NoelTautges ' above, but it also tests DER serialization.

briansmith:
Thank you! I have included your PR's commit as-is into #1680, where I also added a regression test. Sorry for the long delay and I appreciate your help. I'm closing this one since it doesn't have a test.

PR #1520, opened 2022-10-04
OpenBSD: Use getentropy(2)

phy1729:
Unconditionally use `getentropy(2)` on OpenBSD. The syscall was introduced in OpenBSD 5.6 released about 8 years ago. Since only the two latest releases are supported, the last release to not have getentropy has been unsupported for about 7.5 years.

`cargo test` and `cargo test --no-default-features` pass on OpenBSD-current.

Related #797.

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -134,6 +134,8 @@ impl<T> RandomlyConstructable for T where T: self::sealed::RandomlyConstructable
 ///
 /// On macOS and iOS, `fill()` is implemented using `SecRandomCopyBytes`.
 ///
+/// On OpenBSD `fill()` is implemented using the [`getentropy`] syscall.
+///
 /// On wasm32-unknown-unknown when the "wasm32_unknown_unknown_js" feature is
 /// enabled, `fill()` is implemented using `window.crypto.getRandomValues()`.
 /// It must be used in a context where the global object is a `Window`; i.e. it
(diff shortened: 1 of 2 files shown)


src/rand.rs
```diff
@@ -229,6 +232,23 @@ mod sysrand_chunk {
...
+#[cfg(target_os = "openbsd")]
+mod sysrand_chunk {
+    use core::cmp::min;
+    use crate::{c, error};
+
+    #[inline]
+    pub fn chunk(dest: &mut [u8]) -> Result<usize, error::Unspecified> {
+        let chunk_len: c::size_t = min(dest.len(), 256);
+        // See `SYS_getentropy` in #include <sys/syscall.h>.
+        let r = unsafe { libc::syscall(7, dest.as_mut_ptr(), chunk_len) };
```
josephlr, line 244: Instead of manually using `syscall`, why not just use [`libc::getentropy`](https://rust-lang.github.io/libc/x86_64-unknown-openbsd/doc/libc/fn.getentropy.html)

josephlr (approved):
I think this implementation makes sense, given that [this is what the `getrandom` crate does](https://github.com/rust-random/getrandom/blob/master/src/openbsd.rs).

src/rand.rs
phy1729: Fixed up to use libc::getentropy instead.

briansmith:
See the draft PR https://github.com/briansmith/ring/pull/1531 where I propose we delegate the randomness stuff to the getrandom crate. I'm going to close this in favor of that; I'll reopen this if we end up not merging that for some reason.

PR #1524, opened 2022-10-20
CI: Use Ubuntu 20.04 in GitHub Actions.

briansmith:
"The Ubuntu 18.04 Actions runner image is being deprecated and will be removed by 12/1/22" according to the GitHub Changelog blog.


briansmith:
One problem here is that Firefox and `geckodriver` aren't in the Ubuntu 22.04 images. In this PR I've just disabled the testing in Firefox. We could move that testing to Windows (like `getrandom`did, IIRC) but for now I've just commented out the Firefox testing so that Chrome is the only browser tested.

PR #1531, opened 2022-10-25
Use the `getrandom` crate to implement `ring::rand`.

briansmith:


oliverdunk:
Silly question, per #1508 have we tested to see if this works in a worker in WASM? My experience has been that most libraries don’t handle that case.

If not I’ll give this a try!

cemeyer (dismissed):

gaojiaqi7:
Can we add a feature flag for `getrandom/rdrand` so that it can support more `x86/x86_64` targets?

PR #1534, opened 2022-10-26
Add rdrand feature to enable getrandom/rdrand feature.

xiaoyuxlu:
The default features of getrandom does not support some targets, such as x86_64-unknown-uefi x86_64-unknown-none.

This patch enable getrandom/rdrand feature for those targets.


xiaoyuxlu:
https://github.com/briansmith/ring/blob/2bdcdb233fe74a93b392be76a58275e508539d12/src/rand.rs#L133

PR #1540, opened 2022-10-29
Use `cfg!(target_feature)` for static ARM/AARCH64 feature detection.

briansmith:
Our MSRV supports `if` in const expressions. Implement a workaround for buggy detection on older versions of Rust.

### DIFF
--- a/src/cpu/arm.rs
+++ b/src/cpu/arm.rs
@@ -116,7 +116,8 @@ pub fn setup() {
 #[cfg(all(target_os = "windows", target_arch = "aarch64"))]
 pub fn setup() {
     // We do not need to check for the presence of NEON, as Armv8-A always has it
-    let mut features = NEON.mask;
+    const _ASSERT_NEON_DETECTED: () = assert!((ARMCAP_STATIC & NEON.mask) == NEON.mask);
+    let mut features = ARMCAP_STATIC;
 
     let result = unsafe {
         windows_sys::Win32::System::Threading::IsProcessorFeaturePresent(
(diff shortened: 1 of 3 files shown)


briansmith:
My conclusion is that there's not much benefit to doing this until we decide to use Rust 1.61 or later as the MSRV. The main goal I had here was to switch to using the dynamic [std](https://doc.rust-lang.org/stable/std/index.html)::[arch](https://doc.rust-lang.org/stable/std/arch/index.html)::[is_aarch64_feature_detected](https://doc.rust-lang.org/stable/std/arch/macro.is_aarch64_feature_detected.html#) to avoid needing a bunch of OS-specific code for supporting BSD and other operating systems.

src/cpu/arm.rs
```diff
@@ -266,10 +245,28 @@ prefixed_export! {
     static mut OPENSSL_armcap_P: u32 = ARMCAP_STATIC;
 }
 
-const _APPLE_TARGETS_HAVE_ALL_FEATURES: () = assert!(
-    (ARMCAP_STATIC == _ALL_FEATURES_MASK)
-        || !cfg!(all(target_arch = "aarch64", target_vendor = "apple"))
-);
+// MSRV: Enforce 1.61.0 on some aarch64-* targets (aarch64-apple-*, in particular) prior to. Earlier
+// versions of Rust before did not report the AAarch64 CPU features correctly for these targets.
+// Cargo.toml specifies `rust-version` but versions before Rust 1.56 don't know about it.
```
briansmith, line 250: I verified that `cargo test +1.60.0 --target=aarch64-apple-darwin` fails due to this, and `cargo test +1.61.0 --target=aarch64-apple-darwin` passes.

PR #1546, opened 2022-11-01
Make it compile for wasm32-unknown-unknown

Niederb:
Currently ring does not compile for `wasm32-unknown-unknown` without the feature `wasm32_unknown_unknown_js`:
Compiling with `cargo build --release  --target wasm32-unknown-unknown` leads to the error
```
error: the wasm32-unknown-unknown target is not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support
   --> /home/tniederberger/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.8/src/lib.rs:263:9
    |
263 | /         compile_error!("the wasm32-unknown-unknown target is not supported by \
264 | |                         default, you may need to enable the \"js\" feature. \
265 | |                         For more information see: \
266 | |                         https://docs.rs/getrandom/#webassembly-support");
```
This PR fixes the compilation, though I don't know if there is a more elegant way that does not duplicate the `getrandom` dependency.


briansmith:
In this code in rand.rs:
```
    all(
        feature = "wasm32_unknown_unknown_js",
        target_arch = "wasm32",
        target_os = "unknown",
    ),
```
Remove the `feature = "wasm32_unknown_unknown"` condition. Then we need to add documentation saying that if you can't use the `wasm32_unknown_unknown_js` feature because you're building for a non-browser environment, then you need to add 
`getrandom { version = "...", features = ["custom"] }` as a dependency of your project. (I don't want to add a `wasm32_unknown_unknown_custom` feature to *ring*.)

Niederb:
@josephlr Yes, I want to use _ring_ in a non JS-Environment. I only want to use it for verification though so I assume I don't need the rand functionality. 
As far as I can see there is currently no option to disable `rand` and as written above `wasm32-unknown-unknown` currently does not compile. A `verification-only` mode that would not need a source for randomness would be ideal for me.

PR #1556, opened 2022-11-10
WIP: Add support for {mipsel,riscv64gc}-unknown-linux-gnu

briansmith:


briansmith:
This is based on @LinusU's PR #1120.

These two things succeed:
```
mk/cargo.sh test --target=riscv64gc-unknown-linux-gnu --test=aead_tests
mk/cargo.sh test --target=mipsel-unknown-linux-gnu --test=aead_tests
```
But, tests that require `bn_mul_mont` fail because we don't have a fallback; that is what PR #1455 is about.

Also, code coverage measurement doesn't work because the compiler wasn't built with the profiler runtime support for either target:
```
$ RING_COVERAGE=1 mk/cargo.sh test --target=riscv64gc-unknown-linux-gnu --test=aead_tests
error[E0463]: can't find crate for `profiler_builtins`
  |
  = note: the compiler may have been built without the profiler runtime
$ RING_COVERAGE=1 mk/cargo.sh test --target=mipsel-unknown-linux-gnu --test=aead_tests
error[E0463]: can't find crate for `profiler_builtins`
  |
  = note: the compiler may have been built without the profiler runtime
```

briansmith:
Closing this in favor of #1558.

PR #1558, opened 2022-11-11
bigint: Provide a fallback implementation for `bn_mul_mont`.

briansmith:
Add an implementation of `bn_mul_mont`. Intentionally break the build on big-endian targets so they don't accidentally build/run with incorrect logic. See the individual commit messages for more details. Add mipsel-unknown-linux-gnu to the GitHub Actions test matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,6 +180,7 @@ jobs:
           - i686-pc-windows-msvc
           - i686-unknown-linux-gnu
           - i686-unknown-linux-musl
+          - mipsel-unknown-linux-gnu
           - x86_64-pc-windows-gnu
           - x86_64-pc-windows-msvc
           - x86_64-apple-darwin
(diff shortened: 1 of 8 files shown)


briansmith:
Note that one commit in this PR is by @LinusU from PR #1120.

amjoseph-nixpkgs:
Do you know that the machine code this produces executes in [constant time](http://crypto.stanford.edu/~dabo/abstracts/ssl-timing.html)?

Can you be sure that future improvements to rustc will maintain those constant time properties?

This isn't something that can be tested for or "fuzzed" away.  Throwing random test vectors at a routine will give you back white noise timing plots.  The sophisticated attacks involve thinking about what kinds of inputs would create carries in certain bit positions, and possibly even using feedback to make those decisions.  The necessary countermeasure is to be correct by construction for anything handling raw key material.  Chacha20 makes this easy, and it can be done portably for AES if you accept a performance hit.  But RSA is a minefield.

PR #1561, opened 2022-11-18
NFC: Polyfill const_option feature.

briansmith:


briansmith:
This doesn't work in Rust 1.57 because `T: Copy` isn't supported in `const fn`, so this should probably be deferred until we have a better reason for increasing the MSRV.

briansmith:
@joshlf Just stumbled across this old PR, which I've updated on top of your changes. This is how I normally try to polyfill language/library features.

briansmith:
Will rebase this on main after MSRV is updated to 1.61, which is planned to happen in PR #1540.

Issue #1043, opened 2020-08-07
wasm32-wasi doesn't build


ruslan-belinskyy:
When building with:

```
rustup toolchain add 1.40.0
rustup target add wasm32-wasi
cargo wasi build
```
Getting:
```
error[E0432]: unresolved import `super::sysrand_chunk`
   --> /Users/210038/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.16.15/src/rand.rs:304:16
    |
304 |     use super::sysrand_chunk::chunk;
    |                ^^^^^^^^^^^^^ could not find `sysrand_chunk` in `super`


```

Cargo.toml
```
[package]
name = "sumo-collect"
version = "0.1.0"
authors = ["ruslan-belinskyy"]
edition = "2018"


[dependencies]
serde = "1.0.114"
serde_json = "1.0"
jsonwebtoken = "7.2.0"
```

Geal:
I have the same issue (along with the wasm32-unknown-unknown target requiring web-sys). For the context, I'd need ring to work in wasm in a server side setting, so that I can use rustls there. I can provide help in testing solutions for this or even try to fix it in a few days

Geal:
related PR: https://github.com/briansmith/ring/pull/900

briansmith:
I'm happy to get WASI working. I think it is really simple to do now that `wasm32-unknown-unknown` is working.

What I'm missing is a recipe for testing it. I would love help getting the CI/CD running with a good, minimal-dependency, WASI runtime.

ruslan-belinskyy:
We're using Fastly Lucet: https://www.fastly.com/blog/announcing-lucet-fastly-native-webassembly-compiler-runtime
But i didn't try to run it locally, because it didn't compile.

briansmith:
I think I'm fine using Lucet or whatever. What I'm hoping for is a small shell script that downloads the necessary thing to run the tests under wasi and then runs the test under wasi, so that I can copy/paste that into the *ring* CI/CD scripts. Even better would be a PR that modifies the CI/CD scripts to get it working.

briansmith:
I had to switch from Travis CI to GitHub Actions.

Please check out how the new GitHub Actions stuff in `.github/workflows/ci.yml` does it, particularly the `test-wasm32` build matrix.

However, I think we can run the tests:

[...]

PR #1568, opened 2022-12-16
Support for wasm32-wasi

JanKaul:
This PR adds support for the wasm32-wasi target.

To get the CI working, i had to use the dtolnay/rust-toolchain@stable action. The toolchain only seems to work when it is installed with rustup, see a similar issue here: https://bugs.archlinux.org/task/75269.

Currently the clippy and audit CI runs are failing. I didn't include them in this PR because I thought it would make it more difficult to see the actual changes. If I should include the clippy and audit changes, please let me know.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -497,11 +497,9 @@ jobs:
 
       - run: mk/install-build-tools.sh --target=${{ matrix.target }} ${{ matrix.features }}
 
-      - uses: briansmith/actions-rs-toolchain@v1
+      - uses: dtolnay/rust-toolchain@stable
         with:
-          override: true
           target: ${{ matrix.target }}
-          toolchain: ${{ matrix.rust_channel }}
 
       - name: Install wasi-sdk
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -507,7 +507,9 @@ jobs:
           wget -O ./wasi-sdk-linux.tar.gz $WASI_URL
           mkdir wasi-sdk
           tar -C ./wasi-sdk -xf wasi-sdk-linux.tar.gz --strip-components=1
-          echo "WASI_SDK_PATH=`pwd`/wasi-sdk" >> $GITHUB_ENV
+          export WASI_SDK_PATH=`pwd`/wasi-sdk
+          echo "WASI_SDK_PATH=${WASI_SDK_PATH}" >> $GITHUB_ENV
+          echo "CC=${WASI_SDK_PATH}/bin/clang --sysroot=${WASI_SDK_PATH}/share/wasi-sysroot" >> $GITHUB_ENV
       
       - name: Install precompiled wasmtime
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -497,9 +497,11 @@ jobs:
 
       - run: mk/install-build-tools.sh --target=${{ matrix.target }} ${{ matrix.features }}
 
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: briansmith/actions-rs-toolchain@v1
         with:
+          override: true
           target: ${{ matrix.target }}
+          toolchain: ${{ matrix.rust_channel }}
 
       - name: Install wasi-sdk
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -513,10 +513,10 @@ jobs:
       
       - name: Install precompiled wasmtime
         run: |
-          export WASMTIME_URL=$(curl -s https://api.github.com/repos/bytecodealliance/wasmtime/releases/latest | jq -r '.assets[] | select(.name | contains("x86_64-linux.tar.xz")) | .browser_download_url')
-          wget -O /tmp/binaries.tar.xz $WASMTIME_URL
-          tar -C /tmp -xf /tmp/binaries.tar.xz --strip-components=1
-          mv /tmp/wasmtime ~/.cargo/bin
+          VERSION=v3.0.0
+          URL=https://github.com/bytecodealliance/wasmtime/releases/download/${VERSION}/wasmtime-${VERSION}-x86_64-linux.tar.xz
+          wget -O - $URL | tar -xJ --strip-components=1 -C ~/.cargo/bin
+          wasmtime --version
 
       - run: |
           mk/cargo.sh test -vv --target=${{ matrix.target }} ${{ matrix.mode }}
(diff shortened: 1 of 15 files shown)


xxchan:
just kindly fwd: @dynamite-bud said this doesn't work https://github.com/briansmith/ring/issues/1043#issuecomment-1643928508
            
```bash
WASI_SDK_PATH=/(...)/wasi/wasi-sdk cargo build --target wasm32-wasi                                                              1.71.0-nightly  2.90  4.88G 14:24:35  100%

   Compiling ring v0.17.0-not-released-yet (/(...)/ring)
error: no rules expected the token `unsafe`
  --> src/arithmetic/bigint/bn_mul_mont_fallback.rs:26:5
   |
26 |     unsafe fn bn_mul_mont(
   |     ^^^^^^ no rules expected this token in macro call
   |
  ::: src/prefixed.rs:1:1
   |
1  | macro_rules! prefixed_extern {
   | ---------------------------- when calling this macro
   |
note: while trying to match `fn`
  --> src/prefixed.rs:6:22
   |
6  |             $vis:vis fn $name:ident ( $( $arg_pat:ident : $arg_ty:ty ),* $(,)? )
   |                      ^^

warning: unused macro definition: `prefixed_export`
  --> src/prefixed.rs:44:14
   |
44 | macro_rules! prefixed_export {
   |              ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_macros)]` on by default

warning: unused imports: `Limb`, `MODULUS_MAX_LIMBS`, `N0`, `limbs_from_mont_in_place`, `limbs_mul`
  --> src/arithmetic/bigint/bn_mul_mont_fallback.rs:22:13
   |
22 | use super::{limbs_from_mont_in_place, limbs_mul, Limb, MODULUS_MAX_LIMBS, N0};
   |             ^^^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^  ^^^^  ^^^^^^^^^^^^^^^^^  ^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `crate::c`
  --> src/arithmetic/bigint/bn_mul_mont_fallback.rs:23:5
   |
23 | use crate::c;
   |     ^^^^^^^^

warning: `ring` (lib) generated 3 warnings
error: could not compile `ring` (lib) due to previous error; 3 warnings emitted
```

JanKaul:
It works with 
```
export WASI_SDK_PATH=/opt/wasi-sdk CC="${WASI_SDK_PATH}/bin/clang --sysroot=${WASI_SDK_PATH}/share/wasi-sysroot"
```

And the wasi tests in the CI also succeed.

tests/digest_tests.rs
```diff
@@ -47,7 +47,7 @@ fn digest_misc() {
 }
 
 // wasm_bindgen doesn't build this correctly.
-#[cfg(not(target_arch = "wsam32"))]
+#[cfg(not(target_arch = "wasm32"))]
```
briansmith, line 50: Thanks for pointing out the typo. I fixed this the right way in PR #1647, so you can remove this change.

build.rs
```diff
@@ -494,6 +494,19 @@ fn build_c_code(
...
 
+fn cc_builder() -> cc::Build {
+    let mut c = cc::Build::new();
+    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
+    if target_os == "wasi" {
+        let wasi_sdk_path =
+            &std::env::var("WASI_SDK_PATH").expect("missing environment variable: WASI_SDK_PATH");
+        c.compiler(format!("{}/bin/clang", wasi_sdk_path).as_str());
+        c.archiver(format!("{}/bin/llvm-ar", wasi_sdk_path).as_str());
+        c.flag(format!("--sysroot={}/share/wasi-sysroot", wasi_sdk_path).as_str());
```
briansmith, line 505: I feel like these changes belong in cc-rs. We traditionally avoid doing this kind of discovery. I suggest we remove this change from this PR.

.github/workflows/ci.yml
```diff
@@ -464,6 +464,69 @@ jobs:
       # TODO: Do this check on Windows too.
       - run: mk/check-symbol-prefixes.sh --target=${{ matrix.target }}
 
+  # The wasm32-wasi targets have a different set of feature sets.
```
briansmith, line 467: I do not understand this comment about feature sets, and thus I don't understand why the wasi target needs its own build matrix.

.github/workflows/ci.yml
```diff
@@ -464,6 +464,69 @@ jobs:
...
+
+      - run: mk/install-build-tools.sh --target=${{ matrix.target }} ${{ matrix.features }}
+
+      - uses: briansmith/actions-rs-toolchain@v1
+        with:
+          override: true
+          target: ${{ matrix.target }}
+          toolchain: ${{ matrix.rust_channel }}
+
+      - name: Install wasi-sdk
```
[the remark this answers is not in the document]
JanKaul: When I install the wasi-sdk with the `install-build-tools.sh` script, where would you place the install files? On linux you would typically place them in `/opt/wasi-sdk` but I'm not sure it I can assume that.

tests/digest_tests.rs
JanKaul: Okay, I will, Thanks for the review.

jedisct1:
You should consider supporting `cargo zigbuild` instead. This is less painful than `wasi-sdk`, and the same command can be used to cross-compile to other targets.

Support for WASI using `cargo zigbuild` was done in https://github.com/jedisct1/ring-wasi

Without supporting pre-compiled files, it doesn't require any changes to the build scripts, nor messing with environment variables.

.github/workflows/ci.yml
```diff
@@ -464,6 +464,69 @@ jobs:
...
+        run: |
+          export WASI_URL=$(curl -s https://api.github.com/repos/WebAssembly/wasi-sdk/releases/latest | jq -r '.assets[] | select(.name | contains("linux.tar.gz")) | .browser_download_url')
+          wget -O ./wasi-sdk-linux.tar.gz $WASI_URL
+          mkdir wasi-sdk
```
briansmith, line 510: Actually, why do we need the WASI SDK? If you look at the existing wasm32 targets, we manage to build them with standard clang, and I don't see anything WASI-specific here besides the rand.rs, which doesn't add any new C stuff. Maybe we can get rid of this SDK dependency by treating the `-wasi` target just like the `-unknown-unknown` target?

briansmith:
@JanKaul Thanks for your PR. Please see #1745 where I incorporated these changes and added the CI integration. Closing this in favor of that PR.

JanKaul:
Thanks a lot for taking this on. I wasn't entirely sure how to do this.

PR #1580, opened 2023-01-28
Fix `build.rs`: Use `env::var_os` to retrieve `Path`

NobodyXu:
- Fix retrieveing `OUT_DIR` in `build.rs`
 - Fix getting `CARGO_MANIFEST_DIR`: Use `env::var_os`
 - Modify `run_command_with_args` to take `OsStr` instead of `String` for args
 - Call `env::var_os` in `read_env_var`
 - Change return type to `PathBuf` instead of `String` for fn `compile`

Signed-off-by: Jiahao XU <Jiahao_XU@outlook.com>

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -338,7 +338,7 @@ fn ring_build_rs_main() {
         is_debug,
         force_warnings_into_errors,
     };
-    let pregenerated = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap()).join(PREGENERATED);
+    let pregenerated = PathBuf::from(env::var_os("CARGO_MANIFEST_DIR").unwrap()).join(PREGENERATED);
 
     build_c_code(
         &target,
--- a/build.rs
+++ b/build.rs
@@ -642,11 +642,7 @@ fn cc(b: &cc::Build, file: &Path, ext: &str, out_file: &Path) -> Command {
     let mut c = cc.to_command();
     let _ = c
         .arg("-c")
-        .arg(format!(
-            "{}{}",
-            obj_opt,
-            out_file.to_str().expect("Invalid path")
-        ))
+        .arg(format!("{}{}", obj_opt, out_file.display()))
         .arg(file);
     c
 }
--- a/build.rs
+++ b/build.rs
@@ -639,11 +639,11 @@ fn cc(b: &cc::Build, file: &Path, ext: &str, out_file: &Path) -> Command {
 
     let cc = b.get_compiler();
     let obj_opt = if cc.is_like_msvc() { "/Fo" } else { "-o" };
+    let mut arg = OsString::from(obj_opt);
+    arg.push(out_file);
+
     let mut c = cc.to_command();
-    let _ = c
-        .arg("-c")
-        .arg(format!("{}{}", obj_opt, out_file.display()))
-        .arg(file);
+    let _ = c.arg("-c").arg(arg).arg(file);
     c
 }
 
--- a/build.rs
+++ b/build.rs
@@ -529,10 +529,16 @@ fn build_library(
     println!("cargo:rustc-link-lib=static={}", lib_name);
 }
 
-fn compile(b: &cc::Build, p: &Path, target: &Target, include_dir: &Path, out_dir: &Path) -> String {
+fn compile(
+    b: &cc::Build,
+    p: &Path,
+    target: &Target,
+    include_dir: &Path,
+    out_dir: &Path,
+) -> PathBuf {
     let ext = p.extension().unwrap().to_str().unwrap();
     if ext == "o" {
-        p.to_str().expect("Invalid path").into()
+        p.into()
     } else {
         let out_file = obj_path(out_dir, p);
         let cmd = if target.os != WINDOWS || ext != "asm" {
@@ -542,7 +548,7 @@ fn compile(b: &cc::Build, p: &Path, target: &Target, include_dir: &Path, out_dir
         };
 
         run_command(cmd);
-        out_file.to_str().expect("Invalid path").into()
+        out_file
     }
 }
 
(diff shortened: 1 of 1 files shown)


briansmith:
Is the entire purpose of this PR to support non-Unicode(-compatible) paths? Is there any other goal with this PR?

briansmith:
I agree with the idea that we should be using `OsString`s for paths. Do you have a realistic example of a configuration that works with this PR but doesn't work without this PR, so I can understand the real-world implications better? Thanks in advance!

NobodyXu:
> Is the entire purpose of this PR to support non-Unicode(-compatible) paths? Is there any other goal with this PR?

Yes



> I agree with the idea that we should be using `OsString`s for paths. Do you have a realistic example of a configuration that works with this PR but doesn't work without this PR, so I can understand the real-world implications better? Thanks in advance!

Linux filename can be anything that is `\0` terminated and various unix program accepts anything that is ascii-compatible, that is, encoded in bytes and can use all ascii characters.

So the user can use arbitary non-unicode characters in their home dir or whatever dir, `cargo new --lib .` a project, then add this project as a dependency and tries to `cargo b` it.

Without non-unicode support, this will fail since the `CARGO_MANIFEST_DIR` is likely to be a absolute path.
It is also possible that `OUR_DIR` is a absolute path and even if it's not, the user could change the location and name of the `target` dir.

So it's possible for these environment variables to contain non-unicode and I'd like ring's `build.rs` to properly handle it.

briansmith:
> Linux filename can be anything that is \0 terminated and various unix program accepts anything that is ascii-compatible, that is, encoded in bytes and can use all ascii characters.

That is true, but if somebody is taking full advantage of that flexibility, it's really their own fault.

> Without non-unicode support, this will fail since the CARGO_MANIFEST_DIR is likely to be a absolute path.

I understand. I'm simply asking could you provide a plausible scenerio where `CARGO_MANIFEST_DIR` contains non-Unicode-compatible characters? For example, "If a user with username "X" tries to build *ring* when their system character set is Y then the build fails." I.e. something I could test out myself without feeling silly.

NobodyXu:
> I understand. I'm simply asking could you provide a plausible scenerio where CARGO_MANIFEST_DIR contains non-Unicode-compatible characters? For example, "If a user with username "X" tries to build ring when their system character set is Y then the build fails." I.e. something I could test out myself without feeling silly.

On linux, I am able to create a non-unicode dir by running:

```bash
mkdir $(printf '\xFF')
```

And then cd into it by running:

```bash
cd $(printf '\xFF')
```

I can create a user with non-unicode home:

```bash
adduser a --home /home/"$(printf 'a\xFF')"
```

and I can cd into its home using:

```bash
cd /home/a\377/
```

briansmith:
@NobodyXu Is this an issue that's causing trouble on a real system, or just a theoretical issue? `0xFF` in a path is a theoretical issue. This will help me prioritize this.

NobodyXu:
> @NobodyXu Is this an issue that's causing trouble on a real system, or just a theoretical issue? `0xFF` in a path is a theoretical issue. This will help me prioritize this.

Unfortunately I don't have any real world example for this, so right now it just stays as a theoretical issue, but I won't be surprised if somebody out there is having non-unicode path in their system, hence I think it's better to support any `Path`.

build.rs
```diff
@@ -289,18 +290,18 @@ const WINDOWS: &str = "windows";
...
     println!("cargo:rerun-if-env-changed={}", name);
-    std::env::var(name)
+    std::env::var_os(name)
 }
 
 fn main() {
     const RING_PREGENERATE_ASM: &str = "RING_PREGENERATE_ASM";
     match read_env_var(RING_PREGENERATE_ASM).as_deref() {
-        Ok("1") => {
+        Some(val) if val == OsStr::new("1") => {
```
briansmith, line 301: You can do `Some(val) if val == "1"`; see #1698.

build.rs
```diff
@@ -692,9 +689,10 @@ fn nasm(file: &Path, arch: &str, include_dir: &Path, out_file: &Path) -> Command
     c
 }
 
-fn run_command_with_args<S>(command_name: S, args: &[String])
+fn run_command_with_args<S, Arg>(command_name: S, args: impl IntoIterator<Item = Arg>)
```
briansmith, line 692: We should just get rid of the generics here; see #1698.

briansmith (changes requested):
PTAL at #1698. Your changes here are more thorough #1698. I think it would be great if you could review #1698 and then rebase this on that. If you do I will review it and merge it right away.

NobodyXu:
@briansmith I've rebased this PR against main, please review this.

build.rs
```diff
@@ -529,10 +529,16 @@ fn build_library(
...
 }
 
-fn compile(b: &cc::Build, p: &Path, target: &Target, include_dir: &Path, out_dir: &Path) -> String {
+fn compile(
+    b: &cc::Build,
+    p: &Path,
+    target: &Target,
+    include_dir: &Path,
+    out_dir: &Path,
+) -> PathBuf {
```
briansmith, line 538: Nice!

build.rs
```diff
@@ -633,15 +639,11 @@ fn cc(b: &cc::Build, file: &Path, ext: &str, out_file: &Path) -> Command {
...
     let mut c = cc.to_command();
-    let _ = c
-        .arg("-c")
-        .arg(format!(
-            "{}{}",
-            obj_opt,
-            out_file.to_str().expect("Invalid path")
-        ))
-        .arg(file);
+    let _ = c.arg("-c").arg(arg).arg(file);
```
briansmith, line 646: Also very nice.

briansmith (approved):
Thank you for doing this. If you have other things you want to clean up in build.rs I would be happy to work with you on them. PR #1699 is another one you may be interested in.

briansmith (approved):

briansmith:
Thank you!

PR #1586, opened 2023-02-15
CI: Fix `audit` job.

briansmith:
See individual commit messages for details.

PR #1587, opened 2023-02-15
build.rs: Address new clippy warning.

briansmith:


briansmith:
There are several things in the CI configuration that need to be updated before all jobs will pass. PR #1586 addresses the `audit` job and (an)other PR(s) will address the remaining issues. Since this change "clearly" only materially affects (fixes) the clippy job I'm going to merge this even though not all checks passed.

PR #1591, opened 2023-02-16
Enable `build.rs` to build each source file in parallel

NobodyXu:
This PR enables `build.rs` to run `perl` and `nasm` and `cc` in parallel to speedup compilation.
Parallel execution of `perl` and `nasm` is done manually while parallel execution of `cc` is done using `cc::Build`.

This PR also builds src and test in parallel.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -183,7 +183,7 @@ wasm-bindgen-test = { version = "0.3.26", default-features = false }
 libc = { version = "0.2.100", default-features = false }
 
 [build-dependencies]
-cc = { version = "1.0.69", default-features = false }
+cc = { version = "1.0.69", default-features = false, features = ["parallel"] }
 
 [dev-dependencies]
 criterion = { version = "0.4", default-features = false }
--- a/build.rs
+++ b/build.rs
@@ -530,14 +530,8 @@ fn build_library(
         let _ = c.flag(f);
     }
 
-    match target.os.as_str() {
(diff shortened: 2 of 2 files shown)


NobodyXu:
On my powerful M1, this reduces compilation time from 4.3s => 4.2s, but on CI and other env with a much weaker CPU and much less ram, the reduction can be more significant.

I also submit a PR to `cc` to speedup parallel compilation, would also help `ring` once it is merged: https://github.com/rust-lang/cc-rs/pull/780

briansmith:
Let's split this up into multiple PRs.

> parallel execution of `cc` is done using `cc::Build`.

PR #1699 will make it so that on non-Windows targets, `cc::Build` can build all .c and all .S files in a library in parallel, if we enable its parallel build feature.

>  `perl` 

Let's do this last. Most users of *ring* will be building it from the crates.io release that packages the output of the PerlAsm so that perl doesn't need to be run at all.

> `nasm` 

Again, most users will be using the crates.io release where the `nasm` output has been prepackaged into the crate, so this isn't as urgent as the rest.

> This PR also builds src and test in parallel.

I think issue #1705 describes a better solution for this.

WDYT?

NobodyXu:
> > This PR also builds src and test in parallel.
> 
> I think issue #1705 describes a better solution for this.
> 
> WDYT?

I think #1705 is indeed a better solution.
Closing this PR.

PR #1596, opened 2023-03-07
Let cc-rs choose the correct optimization flags for MSVC (`/O2` instead of `/Ox`)

andoalon:
`/Ox` is widely misunderstood to be the flag to enable maximum optimization with MSVC, partly because it used to be called "Full optimzation" in the Visual Studio GUI (it has been changed and now it's called "Optimizations" or "Enable Most Speed Optimizations" which describes it better). The flag to enable maximum optimizaitions however is actually `/O2` ("Maximize Speed"), which was and still is the default flag for release builds in Visual Studio.

https://learn.microsoft.com/en-us/cpp/build/reference/ox-full-optimization?view=msvc-170
https://stackoverflow.com/questions/5063334/what-is-the-difference-between-the-ox-and-o2-compiler-options

If the intention was indeed to have `/O2` minus the flags `/Ox` is missing I would suggest adding a comment

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -613,14 +613,9 @@ fn cc(file: &Path, ext: &str, target: &Target, include_dir: &Path, out_file: &Pa
         let _ = c.define("NDEBUG", None);
     }
 
-    if compiler.is_like_msvc() {
-        if std::env::var("OPT_LEVEL").unwrap() == "0" {
-            let _ = c.flag("/Od"); // Disable optimization for debug builds.
-                                   // run-time checking: (s)tack frame, (u)ninitialized variables
-            let _ = c.flag("/RTCsu");
-        } else {
-            let _ = c.flag("/Ox"); // Enable full optimization.
-        }
+    if compiler.is_like_msvc() && std::env::var("OPT_LEVEL").unwrap() == "0" {
+        // run-time checking: (s)tack frame, (u)ninitialized variables
+        let _ = c.flag("/RTCsu");
     }
 
     // Allow cross-compiling without a target sysroot for these targets.


build.rs
```diff
@@ -619,7 +619,10 @@ fn cc(file: &Path, ext: &str, target: &Target, include_dir: &Path, out_file: &Pa
                                    // run-time checking: (s)tack frame, (u)ninitialized variables
             let _ = c.flag("/RTCsu");
         } else {
-            let _ = c.flag("/Ox"); // Enable full optimization.
```
briansmith, line 622: Do you agree that cc-rs does the right thing automatically, so that we can just remove both this line and the analogous "/Od" line above? I would prefer to do that, if cc-rs is automatically doing the right thing. (In general, I want to remove as much logic from build.rs that duplicates that cc-rs does as possible; see recent PRs that change build.rs to this effect.)

briansmith (changes requested):

build.rs
andoalon: I haven't previously used cc-rs so I assumed manually specifying optimization flags was somehow necessary 😅  
Well, cc-rs can determine the correct flag for each compiler: https://github.com/rust-lang/cc-rs/blob/2447a2ba5f455c00b1563193e125b60eebbd8ebe/src/lib.rs#L1748-L1749

I quickly checked the docs and there's [this function](https://docs.rs/cc/latest/cc/struct.Build.html#method.opt_level) that lets you manually specify the optimization level, but it also says that the `OPT_LEVEL` environment variable is used to determine the correct value otherwise (which is what is done manually here).

About `"/Od"`, it is the default so no need to specify it: https://learn.microsoft.com/en-us/cpp/build/reference/od-disable-debug?view=msvc-170

It's not clear to me whether `"/RTCsu"` is on by default:
https://learn.microsoft.com/en-us/cpp/build/reference/rtc-run-time-error-checks?view=msvc-170

The Visual Studio IDE does enable it for debug builds but that might not be true when using the compiler directly. We might need to read a little more, or just keep it to be sure.

tl;dr  
Yes, I totally agree, no optimization flags should be manually set unless there's a good reason for doing it. The `"/RTCsu"` situation is not 100% clear to me

andoalon: I took another look, but I couldn't conclusively find anything indicating that `/RTCsu` is enabled by default. Therefore I left that part but removed the rest. If you think you don't need `/RTCsu`, then the whole if statement can be removed, otherwise I would say the change is ready to be merged

briansmith (approved):

build.rs
briansmith: @andoalon I would be happy to have the /RTCsu removed.

briansmith:
I also investigated this and found the following: We are presently (before this PR) using `/Ox` for release builds instead of `/O2`. 
According to what I read the difference between `/Ox` and `/O2` is `/Gs /GF /Gy`. Our build.rs unconditionally enables `/Gy` which is function-level linking. `/GF` is irrelevant to us because we don't deal with strings in C. Not sue about `'/Gs` but we unconditionally enable `/GS` presently.

I suggest, in addition to your changes, we should also remove '/GS' and '/RTCsu` and just go with the defaults for this. We should keep on unconditionally enabling `/Gy`' function-level linking.

briansmith:
Thanks!

PR #1604, opened 2023-04-19
Add i686-apple-darwin target to asm matcher in 0.16 branch

melvyn2:
The 0.17 branch & pre-releases have fixed this somewhere in their overhaul, but haven't been released yet. Until the 0.17 series is released, this patch is useful for projects relying on stable _ring_ versions.

Tested in a macOS 10.14 VM.


briansmith:
Thanks. I am surprised anybody needs a 32-bit apple-darwin target any more. What's the use case?

melvyn2:
Writing modules that are loaded by software out of my control and hasn't been updated but is still used, basically. I don't use ring directly in these modules but important dependencies do.

briansmith:
Hi, I am not planning to do a 0.16 release so I'm going to close this. Thanks for your help though!

PR #1606, opened 2023-04-27
Build script fixing

ssrlive:
Build failed under windows, fix a warning and a path search.


briansmith:
We intentionally avoid trying to "detect" whether an executable exists in build.rs.

We already do this:
`run_command_with_args(&get_command("PERL_EXECUTABLE", "perl"), &args);`

I would be open to supporting an analogous `NASM_EXECUTABLE` that works the same way.

PR #1624, opened 2023-08-09
Automatically install build tools on Windows without pre-compiled assembly

george-cosma:
`ring` right now expects the developer to manually use the `install-build-tools.ps1` script when using this library. This isn't that feasible for when you want to include `ring` as a git dependency rather than from crates.io. Specifically, we wanted to use the latest version from main, but couldn't compile on Windows because `nasm` was missing.

This PR makes it so whenever an assembly file is compiled, it checks for the presence of the compiler or otherwise runs the install script.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -676,7 +676,19 @@ fn nasm(file: &Path, arch: &str, include_dir: &Path, out_file: &Path) -> Command
         std::path::MAIN_SEPARATOR,
     )));
 
-    let mut c = Command::new("./target/tools/windows/nasm/nasm");
+    let nasm_path = "./target/tools/windows/nasm/nasm";
+    if !Path::new(nasm_path).exists() {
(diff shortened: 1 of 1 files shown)


briansmith:
@george-cosma build.rs scripts shouldn't be doing network I/O. in the short term, my plan is to get the next release out ASAP so fewer people need this. In the long term, I intend to get rid of the nasm dependency and the precompilation steps completely. I will be proposing a solution to that soon.

 I'm going to close this because I don't see any way to get to an agreement about adding a downloading step to *ring*.

Besides, it isn't just nasm that is missing; Perl is also required when building from Git. That is something I also intend to fix long-term.

briansmith:
Also, thanks for your contribution, and sorry it wasn't accepted.

george-cosma:
It's alright, and I understand your reasoning.
Amazing project btw 👍

PR #1627, opened 2023-08-19
Add support and CI workflow for riscv64

leso-kn:
As there's been few recent activity on related PRs, this is #1574 rebased on latest `main` with:
* git commit-attribution to involved developers and
* sanetized commit messages to match the naming used by ring

Tested and working on Alpine Linux (riscv64 with musl libc).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,6 +180,7 @@ jobs:
           - i686-pc-windows-msvc
           - i686-unknown-linux-gnu
           - i686-unknown-linux-musl
+          - riscv64gc-unknown-linux-gnu
           - x86_64-pc-windows-gnu
           - x86_64-pc-windows-msvc
           - x86_64-apple-darwin
@@ -244,6 +245,9 @@ jobs:
             host_os: ubuntu-22.04
 
           - target: i686-unknown-linux-musl
+            host_os: ubuntu-22.04
+
+          - target: riscv64gc-unknown-linux-gnu
             host_os: ubuntu-22.04
 
           - target: x86_64-pc-windows-gnu
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -488,6 +488,7 @@ jobs:
         target:
           - aarch64-unknown-linux-gnu
           - i686-unknown-linux-gnu
+          - riscv64gc-unknown-linux-gnu
           - x86_64-unknown-linux-musl
 
         mode:
@@ -509,6 +510,9 @@ jobs:
           # https://github.com/rust-lang/rust/issues/79556 and
           # https://github.com/rust-lang/rust/issues/79555 are fixed.
           - target: i686-unknown-linux-gnu
+            host_os: ubuntu-22.04
+
+          - target: riscv64gc-unknown-linux-gnu
             host_os: ubuntu-22.04
 
           - target: x86_64-unknown-linux-musl
--- a/mk/cargo.sh
+++ b/mk/cargo.sh
@@ -21,6 +21,7 @@ rustflags_self_contained="-Clink-self-contained=yes -Clinker=rust-lld"
 qemu_aarch64="qemu-aarch64 -L /usr/aarch64-linux-gnu"
 qemu_arm="qemu-arm -L /usr/arm-linux-gnueabihf"
 qemu_mipsel="qemu-mipsel -L /usr/mipsel-linux-gnu"
+qemu_riscv64="qemu-riscv64 -L /usr/riscv64-linux-gnu"
 
 # Avoid putting the Android tools in `$PATH` because there are tools in this
 # directory like `clang` that would conflict with the same-named tools that may
@@ -98,6 +99,12 @@ case $target in
     export AR_mipsel_unknown_linux_gnu=mipsel-linux-gnu-gcc-ar
     export CARGO_TARGET_MIPSEL_UNKNOWN_LINUX_GNU_LINKER=mipsel-linux-gnu-gcc
     export CARGO_TARGET_MIPSEL_UNKNOWN_LINUX_GNU_RUNNER="$qemu_mipsel"
+    ;;
+  riscv64gc-unknown-linux-gnu)
+    export CC_riscv64gc_unknown_linux_gnu=clang-$llvm_version
+    export AR_riscv64gc_unknown_linux_gnu=llvm-ar-$llvm_version
+    export CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER=clang-$llvm_version
+    export CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_RUNNER="$qemu_riscv64"
     ;;
   x86_64-unknown-linux-musl)
     export CC_x86_64_unknown_linux_musl=clang-$llvm_version
--- a/mk/cargo.sh
+++ b/mk/cargo.sh
@@ -103,7 +103,7 @@ case $target in
   riscv64gc-unknown-linux-gnu)
     export CC_riscv64gc_unknown_linux_gnu=clang-$llvm_version
     export AR_riscv64gc_unknown_linux_gnu=llvm-ar-$llvm_version
-    export CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER=clang-$llvm_version
+    export CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER=riscv64-linux-gnu-gcc
     export CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_RUNNER="$qemu_riscv64"
     ;;
   x86_64-unknown-linux-musl)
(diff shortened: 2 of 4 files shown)


build.rs
```diff
@@ -45,7 +47,7 @@ const RING_SRCS: &[(&[&str], &str)] = &[
     (&[], "crypto/mem.c"),
     (&[], "crypto/poly1305/poly1305.c"),
 
-    (&[AARCH64, ARM, X86_64, X86], "crypto/crypto.c"),
+    (&[AARCH64, ARM, X86_64, X86, RISCV64, RISCV32], "crypto/crypto.c"),
```
briansmith, line 50: Why do we need to build crypto.c for this arch? I expect we don't need to, since crypto.c is just for architectures for which we have arch-specific non-Rust code.

include/ring-core/base.h
```diff
@@ -89,6 +89,14 @@
 #elif defined(__MIPSEL__) && defined(__LP64__)
 #define OPENSSL_64_BIT
 #define OPENSSL_MIPS64
+#elif defined(__riscv) && __riscv_xlen == 64
+#define OPENSSL_64_BIT
+#define OPENSSL_RISCV64
+#define OPENSSL_NO_ASM
```
briansmith, line 95: BoringSSL doesn't define OPENSSL_NO_ASM here. Is it really needed?

include/ring-core/base.h
```diff
@@ -89,6 +89,14 @@
...
+#elif defined(__riscv) && __riscv_xlen == 32
+#define OPENSSL_32_BIT
+#define OPENSSL_RISCV32
+#define OPENSSL_NO_ASM
```
briansmith, line 99: Ditto.

mk/cargo.sh
```diff
@@ -99,6 +100,12 @@ case $target in
     export CARGO_TARGET_MIPSEL_UNKNOWN_LINUX_GNU_LINKER=mipsel-linux-gnu-gcc
     export CARGO_TARGET_MIPSEL_UNKNOWN_LINUX_GNU_RUNNER="$qemu_mipsel"
     ;;
+  riscv64gc-unknown-linux-gnu)
+    export CC_riscv64gc_unknown_linux_gnu=riscv64-linux-gnu-gcc
```
briansmith, line 104: Why are we using GCC here instead of clang? I generally prefer to use clang unless we have a good reason not to.

src/aead/aes.rs
```diff
@@ -266,7 +266,7 @@ impl Key {
                 ctr32_encrypt_blocks!(vpaes_ctr32_encrypt_blocks, in_out, src, &self.inner, ctr)
             }
 
-            #[cfg(any(target_arch = "x86"))]
```
briansmith, line 269: Please rebase on top of the latest main. to get rid of the need for this change.

briansmith (commented):
Thank you for this. I appreciate the thoroughness of this PR. A few questions.

briansmith:
There is a merge conflict in ci.yml that needs to be resolved before I can submit this to Github Actions. Please rebase this onto main.

include/ring-core/base.h
briansmith: Is there a real-world need for the 32-bit RISC-V? The PR adds the 64-bit stuff to the GitHub Actions matrix but not the 32-bit. 

I noticed BoringSSL doesn't define OPENSSL_RISCV32.

build.rs
leso-kn: Indeed! Fixed in 71dc9e9

include/ring-core/base.h
leso-kn: Right, this is deprecated. Removed in 71dc9e9

include/ring-core/base.h
leso-kn: Same as above, removed in 71dc9e9

mk/cargo.sh
leso-kn: Changed to clang in 71dc9e9

src/aead/aes.rs
leso-kn: Rebasing done (71dc9e9 and 04282ec). See what changed [here](https://github.com/briansmith/ring/compare/05e5107560048a683dd2521b61c9273eb005ead5..04282ec737556361fdd724ac08b490df1b5c84ad).

leso-kn:
Thanks for the thorough review :) Applied the pointed out suggestions and rebased onto main!

mk/install-build-tools.sh
```diff
@@ -91,6 +91,12 @@ case $target in
     libc6-dev-mipsel-cross \
     qemu-user
   ;;
+--target=riscv64gc-unknown-linux-gnu)
+  install_packages \
+    gcc-riscv64-linux-gnu \
+    libc6-dev-riscv64-cross \
```
briansmith, line 97: Does this need to be changed with the switch to clang? Look at the other `use_clang` targets.

briansmith:
Have you really tested this on riscv32? I'm wondering if we should just drop that part.

mk/install-build-tools.sh
leso-kn: Yes, indeed, I did check if `gcc-riscv64-linux-gnu` and `libc6-dev-riscv64-cross` were still applicable but missed the `use_clang` setting. Adding that

leso-kn:
I personally only tested riscv64. The commit contains both as it's based on the work of @amock and @xfbs, while the second commit is based on @light4's implementation, which was targeted for riscv64 only.

I believe we will have tests through the CI workflow for riscv64, do you prefer that we add an entry in the CI config for riscv32 as well or rather to remove that part from the first commit?

briansmith:
> I believe we will have tests through the CI workflow for riscv64, do you prefer that we add an entry in the CI config for riscv32 as well or rather to remove that part from the first commit?

I'd rather we just remove it for now. We can add it back if/when we have somebody who is really needing it and who can help test it.

mk/install-build-tools.sh
leso-kn: Added in f2ad112

include/ring-core/base.h
```diff
@@ -89,6 +89,9 @@
 #elif defined(__MIPSEL__) && defined(__LP64__)
 #define OPENSSL_64_BIT
 #define OPENSSL_MIPS64
+#elif defined(__riscv) && __riscv_xlen == 64
```
briansmith, line 92: BoringSSL uses `__SIZEOF_POINTER__`. I don't think that using the pointer size is as relevant for *ring* as it is for BoringSSL so I am open to using `__riscv_xlen` instead. Any thoughts on this?
```
#elif defined(__riscv) && __SIZEOF_POINTER__ == 8
#define OPENSSL_64_BIT
#define OPENSSL_RISCV64
```

briansmith:
@leso-kn You probably didn't notice but the CI failed. It seems like the linking configuration isn't correct. 

I would stil llike to use the clang-based build system, but if you can't get it working, I'm OK with reverting back to what you had before with the GCC-based build system.

briansmith:
> @leso-kn You probably didn't notice but the CI failed. It seems like the linking configuration isn't correct.
> 
> I would stil llike to use the clang-based build system, but if you can't get it working, I'm OK with reverting back to what you had before with the GCC-based build system.

It seems like you just need to change the linking so that it continues to be done by the GCC. See #1297 and #1057 where we had to do the same thing.

briansmith:
PR #1663 is moving the stuff in base.h to target.h. I will merge PR #1663 ASAP so you can rebase this on top of main for the release. I am hoping to finish the BoringSSL merge today so there shouldn't be any new conflicts.

briansmith:
I posted PR #1669 which rebases your changes on top of the current main branch.

briansmith:
Thank you for your help with this!

leso-kn:
Hey :) Sorry for the late reply, it was a quite busy week, TL is on vacation and the customer requested a lot of things at work.

Thanks for rebasing it! I'll do it here as well just for completedness, but feel free to merge either request – probably yours as the pipeline is already in progress I see.

leso-kn:
Rebasing done, with commit signing and original committer dates in case you prefer that. But feel free to merge either of the PRs :)

leso-kn:
@briansmith Rebasing done, with commit signing and original committer dates in case you prefer that. But feel free to merge either of the PRs :)

leso-kn:
To see the difference between the PRs see [here](https://github.com/briansmith/ring/pull/1627/commits) and [here](https://github.com/briansmith/ring/pull/1669/commits).

briansmith:
> @briansmith Rebasing done, with commit signing and original committer dates in case you prefer that. But feel free to merge either of the PRs :)

OK, If you prefer this way, I will let CI run your PR and then merge yours!

.github/workflows/ci.yml
```diff
@@ -507,6 +512,9 @@ jobs:
           - target: i686-unknown-linux-gnu
             host_os: ubuntu-22.04
 
+          - target: riscv64gc-unknown-linux-gnu
+            host_os: ubuntu-22.04
```
briansmith, line 516: The previous CI run passed except `coverage`. `coverage` failed because this was missing. I added it and now `coverage` is passing.

briansmith:
Thank you for all your assistance with this!

Issue #1567, opened 2022-12-08
build panic on ppc


frank6600:
As title, build whill panic on ppc,
Option::unwrap() on a None value, build.rs:358:10

he32:
This is `ring`s way of telling you "not implemented for this platform".
I have the same itch, as bulk pkgsrc builder for NetBSD/powerpc, and I see an increasing number of rust packages fail to build due to the vendored inclusion of this module.
The list of supported platforms can be found around line 216 in `build.rs` in the `ASM_TARGETS` table.

briansmith:
Closing this as a duplicate of an earlier PPC issue.

PR #1629, opened 2023-08-21
Add support for building on Little Endian PPC64

amock:
This should fix https://github.com/briansmith/ring/issues/1567


ecnelises:
There is https://github.com/briansmith/ring/pull/1057 (and maybe other) for enabling ppc64le, and more work at https://github.com/IBM/ring/commits/ppc-0.16.20. These (including patches for other archs) are pending on review.

runlevel5 (approved):
LGTM 👍

build.rs
```diff
@@ -45,7 +46,7 @@ const RING_SRCS: &[(&[&str], &str)] = &[
     (&[], "crypto/mem.c"),
     (&[], "crypto/poly1305/poly1305.c"),
 
-    (&[AARCH64, ARM, X86_64, X86], "crypto/crypto.c"),
```
briansmith, line 48: This isn't needed.

include/ring-core/base.h
```diff
@@ -89,6 +89,9 @@
 #elif defined(__MIPSEL__) && defined(__LP64__)
 #define OPENSSL_64_BIT
 #define OPENSSL_MIPS64
+#elif defined(__PPC64__) && defined(__LITTLE_ENDIAN__)
+#define OPENSSL_64_BIT
+#define OPENSSL_NO_ASM
```
briansmith, line 94: I hope that `OPENSSL_NO_ASM` isn't needed because we don't define it for any other targets.

build.rs
```diff
@@ -30,6 +30,7 @@ const X86: &str = "x86";
 const X86_64: &str = "x86_64";
 const AARCH64: &str = "aarch64";
 const ARM: &str = "arm";
+const PPC64LE: &str = "ppc64le";
```
briansmith, line 33: This is also not needed.

PR #1631, opened 2023-09-01
Implement NIST P521 signature verification

vkrasnov:
Need a way to work on the 32bit part, as 521 is kinda awkward in this regard, because number of limbs in 32 bit mode is not double that of 521 mode.


briansmith:
Vlad, my understanding is that you have it working for 64-bit but not 32-bit targets. is that right? Happy to have a chat with you about how to resolve the 32-bit issues.

vkrasnov:
Yes, this is correct

briansmith:
Vlad, I just created a new branch `b/generate-curve` that contains the test case generator I used for P-256 and P-384 many years ago. This will help you generate the analogous tests for P-521.

I am still looking for the program I used to generate the tops of `ops/p{256,384}.rs`. I will share it with you when I find it.

briansmith:
> I am still looking for the program I used to generate the tops of ops/p{256,384}.rs. I will share it with you when I find it.

It seems like I do have that program, but it isn't going to help us. I will review your PR #1702 and come back to this one.

vkrasnov:
I copied the ec tests from boringssl, the ops tests could perhaps help understand why 32bit code fails, but my understanding that they are provided in montromery representation, which is different between 64 and 32 bits

briansmith:
> I copied the ec tests from boringssl, the ops tests could perhaps help understand why 32bit code fails, but my understanding that they are provided in montromery representation, which is different between 64 and 32 bits

I will generate the ops tests for P-521 so that they work for 32-bit and 64-bit targets and share them in a PR. This should let us identify the failure for 32-bit targets. I think we need to also do merge 1758 and rebase this on top of that, as I'm not confortable with the type punning we're doing, especially when BoringSSL is no longer doing it.

src/ec/suite_b/ops.rs
```diff
@@ -985,6 +982,14 @@ mod tests {
         );
     }
 
+    #[test]
+    fn p521_point_mul_base_test() {
+        point_mul_base_tests(
+            &p521::PRIVATE_KEY_OPS,
+            test_file!("ops/p521_point_mul_base_tests.txt"),
```
briansmith, line 989: Is the problem possilby that the expected result is a Point, which is Montgomery-encoded, and you generated p521_point_mul_base_tests.txt for 64-bit and not for 32-bit?

vkrasnov: I changed it to have two separate files, although we may want to either change the file format, or alternatively use non-montgomery results, like boringssl does and convert the result instead.

vkrasnov: Seems to pass now

crypto/fipsmodule/ec/ecp_nistz.inl
```diff
@@ -0,0 +1,449 @@
...
+static inline void copy_conditional(Elem r, const Elem a,
+                                                const Limb condition) {
+  for (size_t i = 0; i < FE_LIMBS; ++i) {
+    r[i] = constant_time_select_w(condition, a[i], r[i]);
+  }
+}
+
+static inline void elem_add(Elem r, const Elem a, const Elem b) {
+  LIMBS_add_mod(r, a, b, Q, FE_LIMBS);
+}
```
briansmith, line 71: It's annoying, but it's better for us to keep the stuff that was from gfp_p384.c (that I wrote) in a separate file from the stuff that was originally from Intel so that when we move things to Rust, we can keep track of the provinence correctly to make sure the license stuff is done correctly. So please split gfp_p384.c into `gfp.inl` and `gfp_p384.c`. (In the Rust equivalent to this, I will probably not end up deriving the code from Intel-sourced code, and I don't want to muddy things.)

Making this change will also (I hope) make it so GitHub will show my the actual diffs between the original files and the new files, which would make this easier to review.

crypto/fipsmodule/ec/ecp_nistz.inl
```diff
@@ -0,0 +1,449 @@
...
+  elem_mul_mont(S2, S2, in2_y); /* S2 = Y2*Z1^3 */
+  elem_sub(R, S2, S1);          /* R = S2 - S1 */
+
+  elem_mul_mont(U1, in1_x, Z2sqr); /* U1 = X1*Z2^2 */
+  elem_mul_mont(U2, in2_x, Z1sqr); /* U2 = X2*Z1^2 */
+  elem_sub(H, U2, U1);             /* H = U2 - U1 */
+
+  BN_ULONG is_exceptional = is_equal(U1, U2) & ~in1infty & ~in2infty;
+  if (is_exceptional) {
+    if (is_equal(S1, S2)) {
```
briansmith, line 257: Let's split this PR into two PRs: One PR with the first commit that makes the P-384 code generic, and one that adds the P-521 logic.

I expect to see that with the first PR, where this is only used by P-384, this is unreached code. But the code coverage report for the second (P-521) PR should show that this is actually reached by the tests.

In th P-521 PR, please add a comment along the lines of the one added in https://github.com/openssl/openssl/pull/4424/files.

briansmith (changes requested):

crypto/fipsmodule/ec/gfp_p384.c
```diff
@@ -63,120 +65,12 @@ static const BN_ULONG ONE[P384_LIMBS] = {
...
-  elem_add(r, doubled, a);
-}
-
-static inline void elem_sqr_mont(Elem r, const Elem a) {
-  /* XXX: Inefficient. TODO: Add a dedicated squaring routine. */
-  elem_mul_mont(r, a, a);
-}
+#include "ecp_nistz.inl"
 
 void p384_elem_sub(Elem r, const Elem a, const Elem b) {
```
briansmith, line 75: It seems like all, or at least most, of these `extern` functions' definitions are boilerplate that can be moved to what would be gfp.h like the point multiplication/addition/doubling functions.

briansmith (changes requested):
Some more suggestions.

briansmith:
Regarding the missing tests: I am working on changing the point arithmetic tests so that the test data files are NOT Montgomery-encoded. See PR #1769 and PR #1770.

Regarding generating tests for p521_elem_*, it looks like the Montgomery encoding issue only affects the multiplication tests. I guess I will do something similar to what I am doing in #1770 for those.

briansmith:
Now that PR #1756 is merged, you can run `python3 mk/generate_curves.py` which will generate target/curves/p{256,384,521}.rs. When you update this PR, could you merge in the changes from what's generated into target/curves (using two-way merge) and/or change mk/generate_curves.py as needed? This will make it easier to review this as well. Simiarly, if you have code that generates gfp_521.c, it would be awesome to have that added to mk/generate_curves.py.

yacinehmito:
@briansmith @vkrasnov Is there a way I can help getting this merged? (see https://github.com/denoland/deno/issues/21169).

I can take action on the review, but I am wary of doing this without asking first.

tgross35:
If it helps for reference, rustcrypto recently added the same feature for P521 https://github.com/RustCrypto/elliptic-curves/pull/953 https://github.com/RustCrypto/elliptic-curves/pull/956

vkrasnov:
Converting to draft until #1881 and #1883  are merged

briansmith:
The clippy failure is due to the newest Rust toolchain having a stricter clippy. If you rebase on top of main then you will get the fix, which was in https://github.com/briansmith/ring/pull/1888.

yacinehmito:
How can I help move this forward?

rami3l:
This could be interesting for the Rustup project as well. We've switched from `ring` to `aws-lc-rs` for our `rustls` backend mostly for P521 support, but the latter's building pipeline is longer and is prone to platform-specific cmake errors... It would definitely be nice if we had more options to choose from.

PR #1565, opened 2022-11-30
Add loongarch64 support

zhaixiaojuan:

PR #1632, opened 2023-09-03
Add support for LoongArch64

heiher:
Follow up https://github.com/briansmith/ring/pull/1565

### DIFF
--- a/include/ring-core/target.h
+++ b/include/ring-core/target.h
@@ -34,6 +34,9 @@
 #elif defined(__ARMEL__) || defined(_M_ARM)
 #define OPENSSL_32_BIT
 #define OPENSSL_ARM
+#elif defined(__loongarch_lp64)
+#define OPENSSL_64_BIT
+#define OPENSSL_LOONGARCH64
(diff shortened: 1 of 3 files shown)


briansmith (approved):

briansmith (changes requested):
@heiher The stuff in `base.h` is now to be in `target.h`. Note that the `clippy` job failed; try `mk/clippy.sh`. Please rebase on the main branch and resubmit.

briansmith:
Is there no QEMU support for this target? Could you please add the rest of the stuff like the other targets have for QEMU to cargo.sh and install-build-tools.sh? This makes it very easy for us to all test this target manually, even if we don't add it to CI.

briansmith:
I couldn't find the toolchain support in Ubuntu 22.04. Please help Ubuntu get your cross-compilation and QEMU support into Ubuntu if you haven't already. Thanks!

heiher:
You're right. Fortunately, Debian/LoongArch port is already in progress. Thanks!

PR #1635, opened 2023-09-03
CI: Use LLVM 16 tools.

briansmith:
Rust now uses LLVM 16 and writes object files that LLVM 15's `nm` cannot fully understand.


briansmith:
Some CI jobs are failing due to unrelated issues. These will be resolved in separate PRs.

PR #1636, opened 2023-09-03
CI: Increase MSRV to 1.60.

briansmith:


briansmith:
Some CI jobs fail, but this makes progress towards addressing other failures. The rest of the failures will be addressed in future PRs.

PR #1637, opened 2023-09-03
CI: (Temporarily?) remove mipsel-unknown-linux-gnu from build matrix.

briansmith:
The CI jobs for this target are failing with this error:
```
error: component 'rust-std' for target 'mipsel-unknown-linux-gnu' is
unavailable for download for channel 'nightly'
```

Remove the target while we investigate.


briansmith:
The test-features jobs are failing because of an MSRV issue that will be fixed in #1638.

PR #1638, opened 2023-09-03
CI: Use 1.60.0 as MSRV for "features" test jobs too.

briansmith:


briansmith:
With this change, everything in CI is passing again.

PR #1652, opened 2023-09-26
NFC: Move low-level Montgomery arithmetic out of `bigint`.

briansmith:
When the `alloc` feature is disabled, on lesser-used targets we don't build `bigint` but we still need some of the Montgomery arithmetic.

```
git diff \
   HEAD^1:src/arithmetic/bigint/bn_mul_mont_fallback.rs \
   src/arithmetic/montgomery.rs
```

```
git diff \
   HEAD^1:src/arithmetic/bigint.rs \
   src/arithmetic/montgomery.rs
```


briansmith:
@uweigand @erickt Could you please try this PR to see if it fixes the `--no-default-features` build problem for you? The two `git diff` commands in the summary can be used (probably `git difftool` is better) to verify that the actual code that was moved is not changed.

briansmith:
I just pushed a new version to fix the `#[cfg(...)]` logic.

src/arithmetic/montgomery.rs
```diff
@@ -86,3 +86,159 @@ impl ProductEncoding for (RR, Unencoded) {
...
+        use super::BIGINT_MODULUS_MAX_LIMBS;
+
+        // The mutable pointer `r` may alias `a` and/or `b`, so the lifetimes of
+        // any slices for `a` or `b` must not overlap with the lifetime of any
+        // mutable for `r`.
+
+        // Nothing aliases `n`
+        let n = unsafe { core::slice::from_raw_parts(n, num_limbs) };
+
+        let mut tmp = [0; 2 * MODULUS_MAX_LIMBS];
```
erichte-ibm, line 120: I think this should be `BIGINT_MODULUS_MAX_LIMBS`, with this tweak it successfully builds and passes tests with `--no-default-features`.

briansmith:
@erickt Sorry, it was a typo. I meant to ping erichte-ibm.

uweigand:
> @uweigand @erickt Could you please try this PR to see if it fixes the `--no-default-features` build problem for you? 

Yes, it does.  Thanks!

PR #1656, opened 2023-09-28
Merge BoringSSL through 7b9b9baa95449d49019f7ce45b94963f8763005f

briansmith:


briansmith:
I compared the changed files with those of 7b9b9baa95449d49019f7ce45b94963f8763005f and verified that they differ by only what's expected: removed code, and using "ring-core/" instead of "openssl/" as the include prefix.

PR #1673, opened 2023-10-01
NFC: CPU: Clarify safety invariants around `OPENSSL_armcap_P`.

briansmith:


briansmith:
@joshlf PTAL.

briansmith:
Once we're done with this, I will do the same for the Intel side.

briansmith:
Another change I need to make depends on this, so I'm going to merge it as-is. If you comment on any of the SAFETY comment stuff, I'll update them in a future PR based on your comments.

joshlf:
> @joshlf PTAL.

I don't know enough about `OPENSSL_armcap_P` to be able to give very precise feedback, but in broad strokes this looks like exactly the kind of reasoning that we try to encode in zerocopy 👍

PR #1675, opened 2023-10-01
Expand 32-bit ARM test coverage

briansmith:
See individual commit messages for details.


briansmith:
Sadly this needs work:
* We need to compile with clang in order to get the coverage information for the C code, but clang cannot build the sha256 and x25519 assembly for arm-unknown-linux-gnueabi. We need to improve the conditional compilation for this target in the assembly and/or change our build system.
* armv7-unknown-linux-gnueabihf doesn't have profiler_builtins; see https://github.com/rust-lang/rust/issues/79640.

PR #1676, opened 2023-10-01
Add 32-bit big-endian PowerPC support

briansmith:


briansmith:
@erichte-ibm PTAL.

briansmith:
```
error[E0463]: can't find crate for `profiler_builtins`
  |
  = note: the compiler may have been built without the profiler runtime
```
We need the change to rust-lang/rust analogous to https://github.com/rust-lang/rust/pull/105389 for this target and probably for the 64-bit big-endian version too.

briansmith:
@runlevel5 Perhaps you know who should look at this?

runlevel5:
> @runlevel5 Perhaps you know who should look at this?

@classilla I am wondering if you could help Brian review this Pull Request? Many thanks in advance

pkubaj:
This commit probably introduces misdetection of powerpc64 (big-endian). The reason is that `__powerpc__` is defined on both 64-bits and 32-bits. Since you check for `__powerpc__` before checking for `__powerpc64__` it will go first. The solution would be to explicitly check `#if defined(__powerpc__) && !defined(__powerpc64__)` or simply move the `__powerpc__` check below `__powerpc64__`.

pkubaj:
Also, this line:
```
#elif (defined(__PPC64__) || defined(__powerpc64__)) && \
      (defined(_LITTLE_ENDIAN) || defined(_BIG_ENDIAN))
```

It's not necessary to check endianness here, there were no middle-endian PPC and I doubt ring even supports middle-endian :)

classilla:
What @pkubaj said. Please note that I have relatively few 32-bit systems now running anything other than old versions of Mac OS or Mac OS X.

PR #1677, opened 2023-10-02
Add support for 64-bit big-endian PowerPC.

briansmith:
Not added to CI. We need to find a better solution for supporting PowerPC (and s390x and other less-common archs) in CI.


briansmith:
@runlevel5 @ecnelises @q66 @erichte-ibm PTAL.

PR #1688, opened 2023-10-03
Enable FE25519_ADX Curve25519 implementation on x86-64.

briansmith:
When merging BoringSSL 43f88915f97d1f7df489b4b706afdfa556ba2050 I brought in the source files but I did not hook them up since some non-trivial build system and Rust source changes were necessary. Make those changes so we can use the new implementation.


crypto/curve25519/curve25519.c
```diff
@@ -777,10 +778,9 @@ static void table_select(ge_precomp *t, const int pos, const signed char b) {
 //
 // Preconditions:
 //   a[31] <= 127
-void x25519_ge_scalarmult_base(ge_p3 *h, const uint8_t a[32]) {
+void x25519_ge_scalarmult_base(ge_p3 *h, const uint8_t a[32], int use_adx) {
```
briansmith, line 781: Our goal is to move all runtime processor feature detection to Rust. We've already removed all the C functions for runtime detection.

crypto/curve25519/internal.h
```diff
@@ -28,6 +28,27 @@ void x25519_NEON(uint8_t out[32], const uint8_t scalar[32],
                      const uint8_t point[32]);
 #endif
 
+#if !defined(OPENSSL_NO_ASM) && !defined(OPENSSL_SMALL) && \
```
briansmith, line 31: When I merged 43f88915f97d1f7df489b4b706afdfa556ba2050 I should have merged this, but I had wrongly thought it was going to be unneeded, so I didn't. Do it now.

crypto/internal.h
```diff
@@ -277,6 +277,34 @@ static inline crypto_word_t constant_time_select_w(crypto_word_t mask,
   return (mask & a) | (~mask & b);
 }
 
+// constant_time_select_8 acts like |constant_time_select| but operates on
```
briansmith, line 280: Until this, these constant-time utilities were dead code and so they were removed or not merged in earlier merges. Now we need them, so bring them from the current BoringSSL (we've already merged all changes that BoringSSL has published up to today).

src/cpu/intel.rs
```diff
@@ -41,6 +41,21 @@ impl Feature {
     }
 }
 
+pub(crate) const ADX: Feature = Feature {
```
briansmith, line 44: You can constants are the same as those used in the BoringSSL functions that were previously being called in curve25519.c.

briansmith:
@vkrasnov Do you have time to review this? Right now this optimized code is not being used and we have some evidence that 0.17.0 has a performance regression from 0.16.20 without this.

vkrasnov:
I will take a look, just need to dust off my Intel laptop

briansmith:
The tests are failing on x86_64-apple-darwin in GitHub Actions with:
```
   process didn't exit successfully: `/Users/runner/work/ring/ring/target/x86_64-apple-darwin/release/deps/ring-5c9f1d78ad85243d` (signal: 4, SIGILL: illegal instruction)
```
We need to look into the CPU feature detection and see if there are additional CPU features being used by this code, or if there is some bug in this patch regarding feature detection.

vkrasnov:
So it doesn't look like there is a benchmark for x25519 in ring? But I ran my own benchmarks on an i7-108450H which is an iteration of skylake essentially.

I do see a regression from 0.16 to 0.17 for the public key op, from 72.5 Kops/sec to around but the shared key is actually faster going from 27.5 Kops/s to 30 Kops/s. This specific branch isn't really faster on either however. It looks like a better optimization target would be to reimplement the public key op.

briansmith:
> So it doesn't look like there is a benchmark for x25519 in ring? But I ran my own benchmarks on an i7-108450H which is an iteration of skylake essentially.
> 
> I do see a regression from 0.16 to 0.17 for the public key op, from 72.5 Kops/s to around 26 Kops/s but the shared key is actually faster going from 27.5 Kops/s to 30 Kops/s. This specific branch isn't really faster on either however. It looks like a better optimization target would be to reimplement the public key op.

Hmm....that doesn't seem to match what the BoringSSL people got in their benchmarking in https://boringssl.googlesource.com/boringssl/+/9d4f833eec8205e7ad257fb7e7cb321270d3e3cb:
```
Did 75000 Ed25519 key generation operations in 1007110us (74470.5 ops/sec) [+26.9%]
Did 72000 Ed25519 signing operations in 1011133us (71207.2 ops/sec) [+25.5%]
Did 78000 Curve25519 base-point multiplication operations in 1006737us (77478.0 ops/sec) [+27.5%]
```
I will investigate further.

briansmith:
@vkrasnov All of this code, except for the Rust integration bits, is from BoringSSL. Previously in this PR I had only added the base point multiplication optimization. I've now updated the PR to add the variable point optimization too. But, it is still failing with "illegal instruction" errors that I didn't anticipate in GitHub Actions; usually on macOS but not always. I need to dig out an old PC to run Intel SDE to debug further.

briansmith:
From https://github.com/briansmith/ring/actions/runs/6412019109/job/17408596518#step:9:5:
```
$ sysctl -a | grep machdep.cpu.features
machdep.cpu.features: FPU VME DE PSE TSC MSR PAE MCE CX8 APIC SEP MTRR PGE MCA CMOV PAT PSE36 CLFSH MMX FXSR SSE SSE2 SS HTT SSE3 PCLMULQDQ MON VMX SSSE3 CX16 SSE4.1 SSE4.2 x2APIC POPCNT AES VMM PCID XSAVE OSXSAVE TSCTMR AVX1.0 RDRAND F16C
```
So the CPU doesn't have the features.

I put in some logging of what `OPENSSL_cpuid_setup()` calculates (all hex):
```
5f8bfbff, ffba222b, 283, 0
```
0x283 in binary is 0b1010000011 which clearly doesn't have the any of the bits set.

Further I put:
```
assert!(!cfg!(target_vendor = "apple") || !has_fe25519_adx(cpu));
```
But the test still fails on x86-64 Apple with:
```
process didn't exit successfully: `/Users/runner/work/ring/ring/target/x86_64-apple-darwin/debug/deps/ring-1365db4137ff7907 --nocapture` (signal: 4, SIGILL: illegal instruction)
```

And further it's not just the Curve25519 tests are failing. `SHA1_FOR_LEGACY_USE_ONLY`, `SHA256`, `SHA384`, and `SHA512` tests failed too, with a panic.

It's starting to look like the inclusion of these functions marked as target BMI2 + ADX in the C code is causing codegen to go wrong.

briansmith:
OK, I got this working. I simply had forgotten to add the `use_adx` parameter on the Rust side of `x25519_public_from_private_generic_masked`. Note for future reference that an aarch64-apple-darwin system can be used to simulate an x86_64-apple-darwin system (`cargo test --target=x86_64-apple-darwin`) as its CPU feature support in emulation seems quite similar.

briansmith:
I added benchmarks for X25519 in PR #1694. They cover the underlying math for Ed25519 as well. They show that the base point multiplication is about 25% faster on my AMD 7950x running Fedora.

briansmith:
BTW, if you benchmark this, I recommend doing so on a modern box, not an old one like Skylake. The ADX assembly was scheduled by a superoptimizing compiler for some particular, I believe newer, CPUs.

vkrasnov:
Well with bssl speed I am getting:

```
Did 34000 Curve25519 base-point multiplication operations in 1012096us (33593.7 ops/sec)
Did 30000 Curve25519 arbitrary point multiplication operations in 1017157us (29494.0 ops/sec)
```

Comet-lake is hardly old though, modern Intels aren't really far off in terms of performance, I will try an AMD server too

vkrasnov:
Ran on an `AMD EPYC 7713`, there is some improvement for public key only: 28,024.19 ops/sec vs. 21,482.63 ops/sec
 for reference bssl speed gets 27237.0.

briansmith:
> Ran on an AMD EPYC 7713, there is some improvement for public key only: 28,024.19 ops/sec vs. 21,482.63 ops/sec
for reference bssl speed gets 27237.0.

I think you mean to say that the base point multiplication got faster but the arbitrary multiplication stayed about the same performance? That's not too different than what I found.

Do you think this PR is good to merge?

build.rs
```diff
@@ -472,6 +475,19 @@ fn build_c_code(
     let core_srcs = sources_for_arch(&target.arch)
         .into_iter()
         .filter(|p| !is_perlasm(p))
+        .filter(|p| {
+            if let Some(extension) = p.extension() {
+                // We don't (and can't) use any .S on Windows since MSVC and NASM can't assemble
+                // them.
+                if extension == "S"
+                    && (target.arch == X86_64 || target.arch == X86)
```
vkrasnov, line 483: Can rustc build it perhaps? I don't currently have a win machine, but what happen if you include it as a glob asm?

ctz:
FWIW this improves ed25519 signing by 10% for me compared to 0.17, that is on Skylake i7-7820X, using different benchmarks we have down in rustls.

> BTW, if you benchmark this, I recommend doing so on a modern box, not an old one like Skylake.

Will try my AMD box this weekend.

build.rs
briansmith: If clang is available then we could use it. We could also preassemble `.S` using clang for Windows, like we do for `.asm` for nasm source files.

BoringSSL doesn't use this (nor nistz256) on Windows so IDK if there are ABI issues. I think there may be some concerns that it doesn't work with unwinding but I'm not sure that's a big deal for us. More investigation is warranted. We do enable nistz256 on Windows in *ring* so it wouldn't be unprecedented.

briansmith:
The following were done on AMD 7950x using Fedora 39 preview and the system-supplied compilers. (Note that `x25519_generate_key` is probably dominated by `getrandom` syscall cost.)

Here is *ring* 0.17 using Clang 17 (`clang version 17.0.1 (Fedora 17.0.1-1.fc39`):
```
$ CC=clang cargo bench x25519
x25519_generate_key       time:   [230.00 ns 230.01 ns 230.03 ns]
x25519_compute_public_key time:   [7.9842 µs 7.9852 µs 7.9862 µs]
x25519_agree_ephemeral    time:   [20.414 µs 20.435 µs 20.458 µs]
```

Here is *ring* 0.16 using the same Clang 17:
```
$ CC=clang cargo bench x25519
x25519_generate_key       time:   [238.65 ns 238.66 ns 238.67 ns]
x25519_compute_public_key time:   [11.251 µs 11.253 µs 11.254 µs]
x25519_agree_ephemeral    time:   [28.737 µs 28.752 µs 28.766 µs]
```

Here is *ring* 0.17 using GCC 13 (`cc (GCC) 13.2.1 20230918 (Red Hat 13.2.1-3)`):
```
$ CC= cargo bench x25519
x25519_generate_key       time:   [241.54 ns 241.55 ns 241.55 ns]
x25519_compute_public_key time:   [20.361 µs 20.371 µs 20.379 µs]
x25519_agree_ephemeral    time:   [25.587 µs 25.591 µs 25.596 µs]
```

Here is *ring* 0.16 using the same GCC 13:
```
x25519_generate_key       time:   [238.92 ns 238.92 ns 238.92 ns]
x25519_compute_public_key time:   [10.781 µs 10.783 µs 10.785 µs]
x25519_agree_ephemeral    time:   [27.973 µs 27.985 µs 27.995 µs]
```

The fastest combination is *ring* 0.17 with Clang 17.

briansmith:
Here are the BoringSSL measurements on the same system:
```
$ git rev-parse HEAD
b975f12a2fa3f313b98b0dc40a9bc9e8c0667d08

$ clang --version
clang version 17.0.1 (Fedora 17.0.1-1.fc39)

$ CC=clang CXX=clang++ cmake -DCMAKE_BUILD_TYPE=Release -B build-release-clang
$ build-release-clang/tool/bssl speed -filter 25519
Did 119974 Ed25519 key generation operations in 1000576us (119,904.9 ops/sec)
Did 115000 Ed25519 signing operations in 1000041us (114,995.3 ops/sec)
Did 31000 Ed25519 verify operations in 1027620us (30,166.8 ops/sec)
Did 126000 Curve25519 base-point multiplication operations in 1006698us (125,161.7 ops/sec)
Did 49000 Curve25519 arbitrary point multiplication operations in 1007039us (48,657.5 ops/sec)

$ cc --version
cc (GCC) 13.2.1 20230918 (Red Hat 13.2.1-3)

$ CC= CXX= cmake -DCMAKE_BUILD_TYPE=Release -B build-release-gcc
$ build-release-gcc/tool/bssl speed -filter 25519
Did 49000 Ed25519 key generation operations in 1002435us (48,881.0 ops/sec)
Did 49000 Ed25519 signing operations in 1009440us (48,541.8 ops/sec)
Did 35000 Ed25519 verify operations in 1008613us (34,701.1 ops/sec)
Did 51000 Curve25519 base-point multiplication operations in 1015004us (50,246.1 ops/sec)
Did 41000 Curve25519 arbitrary point multiplication operations in 1021011us (40,156.3 ops/sec)
```

cpu:
I did a rough bisect of *ring* using the Crtierion benchmark and I believe it was https://github.com/briansmith/ring/commit/7b59320e3e92f2c21e677b267b0431d8a59f15ef ([boringssl upstream](https://boringssl-review.googlesource.com/c/boringssl/+/60107)) that introduces the gcc performance regression w/ `x25519_compute_public_key`.

```
[nix-shell:~/Code/Rust/ring/bench]$ git rev-parse HEAD
7b59320e3e92f2c21e677b267b0431d8a59f15ef
[nix-shell:~/Code/Rust/ring/bench]$ CC= cargo bench x25519
<snipped>
x25519_generate_key     time:   [220.56 ns 221.69 ns 222.88 ns]
<snipped>
x25519_compute_public_key
                        time:   [28.470 µs 28.639 µs 28.859 µs]
<snipped>                        
x25519_agree_ephemeral  time:   [25.590 µs 25.624 µs 25.663 µs]
<snipped>

[nix-shell:~/Code/Rust/ring/bench]$ git checkout 2d8fbe09e9f7622df7c7cc0116781a5d65613f9f
Previous HEAD position was 7b59320e3 Merge BoringSSL 'd605df5': Use packed representation for large Curve25519 table
HEAD is now at 2d8fbe09e Import currently-unused utilities in crypto/internal.h
[nix-shell:~/Code/Rust/ring/bench]$ CC= cargo bench x25519
<snipped>
x25519_generate_key     time:   [221.39 ns 222.39 ns 223.52 ns]
<snipped>
x25519_compute_public_key
                        time:   [10.700 µs 10.736 µs 10.785 µs]
<snipped>
x25519_agree_ephemeral  time:   [25.506 µs 25.818 µs 26.290 µs]
<snipped>
```

Benchmarking done w/ gcc 11.3.0 on a 12th Gen Intel(R) Core(TM) i7-1280P.

cpu:
I went down a bit of a rabbit hole getting `boringssl` upstream building w/ `gcc` on my system[^gcc] to try and reproduce the `bssl` results. Sharing those results here but I think bringing these findings upstream to the boringssl folks is a good next step.

Testing with https://github.com/google/boringssl/commit/d605df5b6f8462c1f3005da82d718ec067f46b70:
```
[daniel@blanc:~/Code/C/boringssl-clean]$ git checkout d605df5b6f8462c1f3005da82d718ec067f46b70
A	flake.lock
A	flake.nix
Previous HEAD position was be0fdf7fd Constant-time test that X25519 has a single path.
HEAD is now at d605df5b6 Use packed representation for large Curve25519 table
[daniel@blanc:~/Code/C/boringssl-clean]$ CFLAGS="-Wno-error=stringop-overflow" CC= CXX= cmake -DCMAKE_BUILD_TYPE=Release -B build-release-gcc
<snipped>
[daniel@blanc:~/Code/C/boringssl-clean]$ make -C build-release-gcc
<snipped>
[daniel@blanc:~/Code/C/boringssl-clean]$ build-release-gcc/tool/bssl speed -filter 25519
Did 33744 Ed25519 key generation operations in 1006475us (33526.9 ops/sec)
Did 34000 Ed25519 signing operations in 1011973us (33597.7 ops/sec)
Did 32000 Ed25519 verify operations in 1032193us (31002.0 ops/sec)
Did 36000 Curve25519 base-point multiplication operations in 1021745us (35233.8 ops/sec)
Did 39000 Curve25519 arbitrary point multiplication operations in 1020887us (38202.1 ops/sec)
```

Testing with https://github.com/google/boringssl/commit/4a0393fcf37d7dbd090a5bb2293601a9ec7605da, the parent commit to https://github.com/google/boringssl/commit/d605df5b6f8462c1f3005da82d718ec067f46b70:

```
[daniel@blanc:~/Code/C/boringssl-clean]$ git checkout 4a0393fcf37d7dbd090a5bb2293601a9ec7605da
A	flake.lock
A	flake.nix
Previous HEAD position was d605df5b6 Use packed representation for large Curve25519 table
HEAD is now at 4a0393fcf Add SHA256 and SHA512 bindings to bssl-crypto
[daniel@blanc:~/Code/C/boringssl-clean]$ rm -rf build-release-gcc/
[daniel@blanc:~/Code/C/boringssl-clean]$ CFLAGS="-Wno-error=stringop-overflow" CC= CXX= cmake -DCMAKE_BUILD_TYPE=Release -B build-release-gcc
<snipped>
[daniel@blanc:~/Code/C/boringssl-clean]$ make -C build-release-gcc
<snipped>
[daniel@blanc:~/Code/C/boringssl-clean]$ build-release-gcc/tool/bssl speed -filter 25519
Did 90900 Ed25519 key generation operations in 1006408us (90321.2 ops/sec)
Did 94000 Ed25519 signing operations in 1002192us (93794.4 ops/sec)
Did 33000 Ed25519 verify operations in 1029750us (32046.6 ops/sec)
Did 103000 Curve25519 base-point multiplication operations in 1005442us (102442.5 ops/sec)
Did 39000 Curve25519 arbitrary point multiplication operations in 1010017us (38613.2 ops/sec)
```

Benchmarking done w/ gcc 13.2.0 (_I switched from gcc 11.3 while debugging the `stringop-overflow` issue_) on a 12th Gen Intel(R) Core(TM) i7-1280P.

[^gcc]: I was seeing a compile error with `gcc` (but not `clang`) building `crypto/asn1/a_bitstr.c`. The error was of the form "/nix/store/y8wfrgk7br5rfz4221lfb9v8w3n0cnyd-glibc-2.37-8-dev/include/bits/string_fortified.h:29:10: error: ‘__builtin_memcpy’ specified bound between 18446744071562067968 and 18446744073709551615 exceeds maximum object size 9223372036854775807 [-Werror=stringop-overflow=]". I was able to get the build to proceed by adding `-Wno-error=stringop-overflow` to my `CFLAGS`. Based on a [similar addition](https://github.com/NixOS/nixpkgs/blob/f99e5f03cc0aa231ab5950a15ed02afec45ed51a/pkgs/development/libraries/boringssl/default.nix#L34-L37) to the NixPkgs derivation I suspect it's a quirk of gcc >12...

cpu:
> I think bringing these findings upstream to the boringssl folks is a good next step.

https://bugs.chromium.org/p/boringssl/issues/detail?id=655

briansmith:
> I think bringing these findings upstream to the boringssl folks is a good next step.

I had already let them know about it so they could decide what they want to do about it, if anything. There are a lot of factors to consider and they are doing really good work to improve performance that hasn't landed yet. I'd rather not distract them from their forward progress. It would be better to file a bug against GCC to get GCC to improve.

cpu:
> I had already let them know about it so they could decide what they want to do about it, if anything. 

Understood. For what it's worth that kind of context would be useful to capture in the issue tracker somewhere so we don't duplicate effort or bother the upstream across multiple channels. I would have held off from reporting the bug if I knew you were discussing it already in another venue.

>  I'd rather not distract them from their forward progress. 

From my perspective if your goal is to have Rustls expedite the upgrade to ring 0.17 I don't think it's a distraction to pursue a fix with BoringSSL concurrent to opening a bug with GCC.

In either case I've also opened a bug with the GCC maintainers: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111774

PR #1693, opened 2023-10-04
[WIP] Use zerocopy byteorder module to replace endianness stuff

joshlf:
@briansmith Here's a prototype of replacing endianness support with zerocopy. It's mergeable as-is, but there are two gaps to mention which are worth calling out:
- There's still a dependency on the `byteorder` crate
- The byteorder types have no alignment, which may pessimize codegen

Both of these are on our roadmap to fix.


src/aead/aes.rs
```diff
@@ -356,7 +355,7 @@ impl From<Counter> for Iv {
 
 impl Iv {
     pub(super) fn as_bytes_less_safe(&self) -> &[u8; 16] {
-        self.0.as_byte_array()
+        zerocopy::transmute_ref!(&self.0)
```
briansmith, line 358: Is there not a convenient way to do it without using macros? Not the end of the world, for sure.

briansmith (commented):
In general this is looking really good. It would be good for us to talk through the timing side channel stuff to see what we can do about that.

src/aead/aes.rs
joshlf: The short answer is no; we rely on the fact that `core::mem::transmute` uses compiler magic to ensure that the types' sizes are equal (we also use the same trick to ensure that alignment of the source type is at least as large as the alignment of the destination type). 

The long answer is that it's _technically_ possible, but it has a lot of sharp edges; here are some links if you're curious:
* https://github.com/google/zerocopy/pull/183#discussion_r1208678157
* https://github.com/rust-lang/unsafe-code-guidelines/issues/409
* https://github.com/rust-lang/rust/issues/112090

At some point in the future, we may be able to support this as a function without needing to rely on the const eval hacks. In particular, that'll come once we are able to encode more facts about types in the type system itself. This is ongoing work, but it'll be a while before we're at a place where we can replace `transmute_ref!`.

joshlf:
> In general this is looking really good. It would be good for us to talk through the timing side channel stuff to see what we can do about that.

Sounds good. Can you say more about which operations you're thinking about wrt timing side channels? I see that there's an impl of `BitXorAssign` in [endian.rs](https://github.com/briansmith/ring/pull/1693/files#diff-3caa8a83069b768f998288fb48544b018d16246ae51bdeb28fb7de44ea873e34) and there are byte order swapping operations, but there are also methods to expose those types as e.g. byte arrays; I presume this means that some operations happen on the byte array representation, which the code in `endian.rs` (and similarly in zerocopy) doesn't have any control over.

briansmith:
> Sounds good. Can you say more about which operations you're thinking about wrt timing side channels? I see that there's an impl of BitXorAssign in endian.rs and there are byte order swapping operations,

Byte order swapping, XOR, and eventually we'll need to write a constant-time comparison of arrays. I think right the constant-time comparison is done using the byte array representation but we're hoping we'll be able to make use of SIMD for the comparison in the future, probably by using inline assembly.

briansmith:
By "constant-time comparison," we mean "a == b" in constant time for an array of u8/BEU32/etc./values.

src/aead/aes.rs
CryZe: bytemuck can do this without macros. Unfortunately bytemuck doesn't have byteorder types built-in. I wanted to PR them in at some point. So maybe soon.

joshlf: Which bytemuck API are you referring to?

joshlf: Ah looks like you're probably referring to [`cast_ref`](https://docs.rs/bytemuck/latest/bytemuck/fn.cast_ref.html)? Looks like that's not exactly the same API - if the sizes don't match or the source alignment is less than the destination alignment, `cast_ref` panics at runtime, while `transmute_ref!` fails to compile.

CryZe: Yeah I guess that's true.

joshlf:
Makes sense. My gut is that it wouldn't make sense for zerocopy make any promises about constant time-ness because the compiler can always undermine those, and it'd make more sense to do any operations that need to be guaranteed constant time on the byte representation. (Obviously all of the standard caveats about constant time execution technically being impossible without formal language support, the compiler can do whatever it wants, etc etc).

joshlf:
What would it take to gain confidence in performance and/or constant time-ness of this PR? I still need to write the changes to get rid of the byteorder crate dependency, but it'd be good if I can just do that after we've confirmed that this PR even makes sense from a performance and constant time-ness perspective.

src/aead/aes.rs
briansmith, line 358: Once we import in the benchmarks for `ring::{digest,hmac,pbkdf2}` then we can evaluate the performance of PR #1731. I expect that even if PR #1731 as written has performance issues, we will be able to tweak it so it doesn't. This will almost definitely eliminate this code completely.

src/digest.rs
```diff
@@ -114,7 +114,8 @@ impl BlockContext {
 
         Digest {
             algorithm: self.algorithm,
-            value: (self.algorithm.format_output)(self.state),
+            // SAFETY: TODO
```
briansmith, line 117: PR #1731 will (I hope) address this as well, by making the code 100% safe as long as we are happy with the safety guarantees for the code that writes to the `self.state` union.

Which functions within `digest` need to be marked `unsafe` and which can be safe with `unsafe` blocks is something that could use discussion as it seems we may disagree, which means I might be misunderstanding something.

PR #1695, opened 2023-10-05
Fix powerpc64 detection

pkubaj:
Both powerpc64 and powerpc define `__powerpc__`. Also remove endianness check for powerpc64* since it's not necessary.


briansmith:
Thank you!

PR #1697, opened 2023-10-05
Clarify build.rs C compiler and linker configuration.

briansmith:
See each individual commit message for details.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -511,7 +511,6 @@ fn build_library(
 
     match target.os.as_str() {
         "macos" => {
-            let _ = c.flag("-fPIC");
             let _ = c.flag("-Wl,-dead_strip");
         }
         _ => {
--- a/build.rs
+++ b/build.rs
@@ -387,8 +387,9 @@ fn pregenerate_asm_main() {
                 force_warnings_into_errors: true,
             };
 
+            let b = new_build(&target, &pregenerated_tmp);
             for src in srcs {
-                compile(&src, &target, &pregenerated_tmp, &pregenerated);
+                compile(&b, &src, &target, &pregenerated_tmp, &pregenerated);
             }
         }
     }
@@ -483,6 +484,12 @@ fn build_c_code(
     );
 }
 
+fn new_build(target: &Target, include_dir: &Path) -> cc::Build {
+    let mut b = cc::Build::new();
+    configure_cc(&mut b, target, include_dir);
+    b
+}
+
 fn build_library(
     target: &Target,
     out_dir: &Path,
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -485,37 +484,31 @@ fn build_c_code(
...
     additional_srcs: &[PathBuf],
 ) {
+    let mut c = new_build(&target, out_dir);
+
     // Compile all the (dirty) source files into object files.
     let objs = additional_srcs
         .iter()
         .chain(srcs.iter())
-        .map(|f| compile(f, target, out_dir, out_dir))
+        .map(|f| compile(&c, f, target, out_dir, out_dir))
```
briansmith, line 506: Now all the calls to `compile()` share a single build configuration that is constructed once per library, instead of once per source file.

build.rs
```diff
@@ -485,37 +484,31 @@ fn build_c_code(
...
         .collect::<Vec<_>>();
 
     // Rebuild the library if necessary.
     let lib_path = PathBuf::from(out_dir).join(format!("lib{}.a", lib_name));
 
-    let mut c = cc::Build::new();
-
-    for f in LD_FLAGS {
```
briansmith, line 507: LD_FLAGS was empty so this wasn't doing anything.

build.rs
```diff
@@ -485,37 +484,31 @@ fn build_c_code(
...
-        let _ = c.flag(f);
-    }
-    match target.os.as_str() {
-        "macos" => {
-            let _ = c.flag("-fPIC");
-            let _ = c.flag("-Wl,-dead_strip");
-        }
-        _ => {
-            let _ = c.flag("-Wl,--gc-sections");
```
briansmith, line 516: Move to `configure_cc`.

build.rs
```diff
@@ -591,8 +578,11 @@ fn cc(file: &Path, ext: &str, target: &Target, include_dir: &Path, out_file: &Pa
     if target.os.as_str() == "macos" {
         // ``-gfull`` is required for Darwin's |-dead_strip|.
         let _ = c.flag("-gfull");
+        let _ = c.flag("-fPIC");
+        let _ = c.flag("-Wl,-dead_strip");
     } else if !compiler.is_like_msvc() {
         let _ = c.flag("-g3");
+        let _ = c.flag("-Wl,--gc-sections");
```
briansmith, line 585: `-Wl,--gc-sections` is no longer given to MSVC, but otherwise the effect is equivalent to before.

build.rs
```diff
@@ -573,10 +564,6 @@ fn cc(file: &Path, ext: &str, target: &Target, include_dir: &Path, out_file: &Pa
 
     let _ = c.include("include");
     let _ = c.include(include_dir);
-    match ext {
-        "c" | "S" => (),
-        e => panic!("Unsupported file extension: {:?}", e),
-    };
```
briansmith, line 579: Moved to the top of the new `cc()`.

PR #1698, opened 2023-10-06
build.rs: Use `OsStr`/`OsString` more to handle non-Unicode paths.

briansmith:


briansmith:
@NobodyXu I saw your PR #1580. I think if we simplify some of the stuff in build.rs then your PR will become simpler. Please review this PR and see if it makes sense. Then we could rebase #1580 on top if it?

NobodyXu (approved):

NobodyXu:
Thanks, I reviewed it and it looks good to me.

Would rebase my PR after this lands.

PR #1699, opened 2023-10-06
build.rs: Use cc-rs in a more canonical way in most cases.

briansmith:
See individual commit comments. The intent here is to help people understand what's required to adapt build.rs to an alternative build system by making its use of cc-rs more idiomatic.


briansmith:
@NobodyXu WDYT about this, in light of https://github.com/briansmith/ring/pull/1591#issuecomment-1751210063? Do you think this is good to check in?

NobodyXu (approved):
LGTM

PR #1702, opened 2023-10-06
Replace limbs initializer with const from_hex

vkrasnov:
The current macro initializer is not uniform across architectures, and doesn't work when the number of 64bit limbs is not exactly half that of 32bit limbs, (i.e. 521 bit).
This change replaces all limbs! initializers with a new const function with hex string initializers, which is more portable and more ergonomic.

### DIFF
--- a/src/ec/suite_b/ops/p384.rs
+++ b/src/ec/suite_b/ops/p384.rs
@@ -12,54 +12,24 @@
 // OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
+use crate::arithmetic::limbs_from_hex;
+
 use super::{
     elem::{binary_op, binary_op_assign},
     elem_sqr_mul, elem_sqr_mul_acc, Modulus, *,
 };
-use core::marker::PhantomData;
-
-macro_rules! p384_limbs {
-    [$($limb:expr),+] => {
-        limbs![$($limb),+]
-    };
-}
 
 pub static COMMON_OPS: CommonOps = CommonOps {
     num_limbs: 384 / LIMB_BITS,
 
     q: Modulus {
-        p: p384_limbs![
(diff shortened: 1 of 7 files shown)


src/ec/suite_b/ops/p384.rs
```diff
@@ -12,54 +12,24 @@
...
-            0xe3374bee, 0x1f022094, 0xb62b21f4, 0x604fbff9, 0xcd08114b
-        ],
-        m: PhantomData,
-        encoding: PhantomData, // Unreduced
-    },
-
+    n: Elem::from_hex("ffffffffffffffffffffffffffffffffffffffffffffffffc7634d81f4372ddf581a0db248b0a77aecec196accc52973"),
+    a: Elem::from_hex("fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffbfffffffc0000000000000003fffffffc"),
+    b: Elem::from_hex("cd08114b604fbff9b62b21f41f022094e3374bee94938ae277f2209b1920022ef729add87a4c32ec081188719d412dcc")
+,
```
briansmith, line 32: This comma shouldn't be on its own line; it should be at the end of the line above. I will fix this after I merge this.

src/ec/suite_b/ops/p384.rs
```diff
@@ -332,10 +272,7 @@ unsafe extern "C" fn p384_elem_sqr_mont(
     p384_elem_mul_mont(r, a, a);
 }
 
-const N_RR_LIMBS: [Limb; MAX_LIMBS] = p384_limbs![
-    0x19b409a9, 0x2d319b24, 0xdf1aa419, 0xff3d81e5, 0xfcb82947, 0xbc3e483a, 0x4aab1cc5, 0xd40d4917,
-    0x28266895, 0x3fb05b7a, 0x2b39bf21, 0x0c84ee01
-];
+const N_RR_HEX: &str = "0c84ee012b39bf213fb05b7a28266895d40d49174aab1cc5bc3e483afcb82947ff3d81e5df1aa4192d319b2419b409a9";
```
briansmith, line 275: This doesn't need to have the leading zero, right? I will fix this after I merge it.

briansmith (approved):

PR #1708, opened 2023-10-06
0.17.2

briansmith:
This is 0.17.1 with one addition to Cargo.toml to add a header file that wasn't listed.


briansmith:
I went ahead and yanked 0.17.1 after 0.17.2 was published.

PR #1716, opened 2023-10-09
build.rs: Remove `-fno-strict-aliasing` from the C compiler configuration.

briansmith:
Enable further compiler optimizations.


briansmith:
@vkrasnov @ctz I was looking into the build flags *ring* uses vs. what BoringSSL uses and this is the main optimization-related one I see that could affect performance. I didn't see this affecting performance using clang-16 though.

PR #1717, opened 2023-10-09
Curve25519: Conservatively force noinline on ADX code paths.

briansmith:
See https://github.com/rust-lang/rust/issues/116573.


RalfJung:
These functions don't seem to use any SIMD types in their signature so I don't think they are affected by https://github.com/rust-lang/rust/issues/116573. But of course there might be other bugs lurking in target-feature handling...

PR #1718, opened 2023-10-09
Use `CRYPTO_memcmp` name the same as BoringSSL.

briansmith:
*ring* defined a function named `OPENSSL_memcmp` that did what `CRYPTO_memcmp` does in BoringSSL, and BoringSSL has a different function called `OPENSSL_memcmp`. *ring* doesn't need `OPENSSL_memcmp` so sync the `CRYPTO_memcmp` stuff with BoringSSL.

This eliminates unnecessary differences from BoringSSL.

### DIFF
--- a/crypto/mem.c
+++ b/crypto/mem.c
@@ -57,10 +57,11 @@
 #include <ring-core/mem.h>
 #include "internal.h"
 
-int OPENSSL_memcmp(const void *av, const void *bv, size_t len) {
+int CRYPTO_memcmp(const void *in_a, const void *in_b, size_t len) {
+  const aliasing_uint8_t *a = in_a;
+  const aliasing_uint8_t *b = in_b;
   uint8_t x = 0;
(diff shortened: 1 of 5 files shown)


crypto/mem.c
```diff
@@ -57,10 +57,11 @@
...
 #include "internal.h"
 
-int OPENSSL_memcmp(const void *av, const void *bv, size_t len) {
+int CRYPTO_memcmp(const void *in_a, const void *in_b, size_t len) {
+  const aliasing_uint8_t *a = in_a;
+  const aliasing_uint8_t *b = in_b;
   uint8_t x = 0;
-  const aliasing_uint8_t *a = av;
-  const aliasing_uint8_t *b = bv;
+
```
briansmith, line 64: This is just making the function structure the same as BoringSSL's version. The only difference is the use of `aliasing_uint8_t` instead of `uint8_t`.

PR #1729, opened 2023-10-11
AEAD: Remove all use of `ring::endian` and u64 from `Block`.

briansmith:
In particular, eliminate the use of `ArrayEncoding::as_byte_array` as we work towards removing that function because it uses `unsafe`.

 Where `Block` is used in parameters to C/assembly code, I verified that the C/assembly code uses `uint8_t *` as the function type (meaning `uint8_t[16]`) in the BoringSSL headers. Until recently the stuff in GCM was using `uint64_t` or a union containing `uint64_t`, which is why block was previously defined in terms of `BigEndian<u64>`.


briansmith:
@joshlf This removes the `BitXorAssign` implementation from `ring::endian`.

joshlf (approved):
Cool!

PR #1731, opened 2023-10-11
Remove ArrayEncoding usage from Digest and delete ArrayEncoding.

briansmith:


briansmith:
I still need to benchmark this. Will add benchmarks for `ring::digest` in another PR.

briansmith:
@joshlf This removes the last `unsafe` from `ring::endian`.

src/digest.rs
```diff
@@ -475,17 +466,30 @@ pub const MAX_OUTPUT_LEN: usize = 512 / 8;
...
 fn sha512_format_output(input: State) -> Output {
-    let input = unsafe { &input.as64 };
-    Output {
-        as64: input.map(BigEndian::from),
-    }
+    let input = unsafe { input.as64 };
+    format_output::<_, _, { core::mem::size_of::<u64>() }>(input, u64::to_be_bytes)
+}
+
+#[inline]
```
joshlf, line 478: I could be wrong about this, but my understanding is that inside of a crate (as opposed to between crates), `#[inline]` has little to no effect.

joshlf (approved):
Nice!

briansmith:
This has been blocked on adding (back) the benchmarks for `ring::digest` to ensure this doesn't negatively affect performance. Those are in #1809 now.

briansmith:
The codecov "project" coverage % "failure" is expected since this reduces the amount of covered code.

PR #1732, opened 2023-10-12
Take some steps toward eliminating the last `unsafe` in `ring::polyfill`

briansmith:
Replace a few uses of `ChunksFixed` with a new `ArraySplitMap`. `ArraySplitMap` avoids the need to construct an intermediate array; we would hope the optimizer would handle that, but at least in the past it did not always do so.

The approach here isn't very scalable since the source code of each implementation of `ArraySplitMap`  grows twice as fast as the size of the input array. Maybe we can find a cleverer solution, at least for the larger ones.

### DIFF
--- a/src/polyfill/chunks_fixed.rs
+++ b/src/polyfill/chunks_fixed.rs
@@ -27,7 +27,5 @@ macro_rules! define_chunks_fixed {
 }
 
 // Sorted by the first value, then the second value.
-define_chunks_fixed!(16, 8);
-define_chunks_fixed!(32, 4);
 define_chunks_fixed!(64, 32);
 define_chunks_fixed!(80, 20);
--- a/src/polyfill/chunks_fixed.rs
+++ b/src/polyfill/chunks_fixed.rs
@@ -27,7 +27,6 @@ macro_rules! define_chunks_fixed {
 }
 
 // Sorted by the first value, then the second value.
-define_chunks_fixed!(12, 4);
 define_chunks_fixed!(16, 4);
 define_chunks_fixed!(16, 8);
 define_chunks_fixed!(32, 4);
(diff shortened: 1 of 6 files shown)


briansmith:
@joshlf PTAL. This is mostly to start a discussion. Definitely I am curious about whether there's a simpler way of doing this. I was hoping the array_* standard library facilities would be available by now, but it seems unlikely that they will become stable soon.

joshlf:
> @joshlf PTAL. This is mostly to start a discussion. Definitely I am curious about whether there's a simpler way of doing this. I was hoping the array_* standard library facilities would be available by now, but it seems unlikely that they will become stable soon.

I don't have any strong opinions here, but the approach seems reasonable. Anything that removes `unsafe`s is good in my book 🙂

src/polyfill/chunks_fixed.rs
```diff
@@ -27,9 +27,5 @@ macro_rules! define_chunks_fixed {
 }
 
 // Sorted by the first value, then the second value.
-define_chunks_fixed!(12, 4);
-define_chunks_fixed!(16, 4);
-define_chunks_fixed!(16, 8);
-define_chunks_fixed!(32, 4);
```
briansmith, line 33: For the small arrays that are converted here (the ones for which lines are removed above) this approach seems to optimize well. For the other two larger array sizes (mentioned below), I will post a separate PR that uses different techniques.

briansmith:
This change ends up being +91 −33 (+58 net) LoC. The future changes to eliminate `ChunksFixed` comletely are +13 -48 (-35 net) LoC. So it looks like the project to eliminate `ChunksFixed`, for the purpose of eliminating the one `unsafe` line, costs about +23 net LoC. 

I believe all the changes here don't worsen the code, and slightly improve the callers. The remaining changes to remove `ChunksFixed` are more debatable. So I'm going to merge this for now and we'll discuss the issues in that PR.

PR #1736, opened 2023-10-13
Remove `ChunksFixed` which contains one line of `unsafe` code.

briansmith:
See the individual commit messages.

chunks_fixed.rs didn't have a license header when it should have. That means that the original LoC total is 14 LoC lower than it should have been. Thus this change is  +13 -48 -14 (net -49) LoC, which combined with PR #1732's net +58 means that this project cost 11 LoC using the most naive metric of complexity we can get.

Unlike in PR #1732, here we're using potentially-panicking array slicing or equivalent (e.g. `copy_from_slice`). So we're trading static guarantee of not panicking for static guarantee of safety. In this code it is obvious that no panic will be occur so it's not a giant deal. We do the same elsewhere, unfortunately.

I don't feel super great about that. However, I think it does help us evaluate using zerocopy or other dependencies (ideally an improved standard library and/or language) to eliminate use of potentially-panicking APIs in addition to eliminating use of `unsafe` and/or making our `unsafe` use safer.


briansmith:
@joshlf PTAL.

src/polyfill/chunks_fixed.rs
```diff
@@ -1,31 +0,0 @@
...
-    ( $unchuncked_len:expr, $chunk_len:expr ) => {
-        define_chunks_fixed!($unchuncked_len, $chunk_len, $unchuncked_len / $chunk_len);
-    };
-
-    ( $unchuncked_len:expr, $chunk_len:expr, $chunked_len:expr ) => {
-        impl<'a, T> ChunksFixed<'a, &'a [[T; $chunk_len]; $chunked_len]>
-            for &'a [T; $unchuncked_len]
-        {
-            #[inline(always)]
-            fn chunks_fixed(self) -> &'a [[T; $chunk_len]; $chunked_len] {
```
joshlf, line 20: Since you've been asking about other places zerocopy might help: this method should be able to be replaced by [`transmute_ref!`](https://docs.rs/zerocopy/latest/zerocopy/macro.transmute_ref.html), at least in contexts where `T` and `Self` are concrete types.

joshlf (approved):

PR #1737, opened 2023-10-13
Bump dependency requirements to what's been tested.

briansmith:
Older versions may work but we've not been testing those old versions.


briansmith:
The MSRV CI jobs verified that this doesn't affect MSRV.

PR #1738, opened 2023-10-13
Remove `target-dir` workaround for Clippy cache issues.

briansmith:
For a long while this hasn't been needed. See
https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-152

### DIFF
--- a/mk/clippy.sh
+++ b/mk/clippy.sh
@@ -19,7 +19,6 @@ IFS=$'\n\t'
 
 export NULL=""
 cargo clippy \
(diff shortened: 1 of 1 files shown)


briansmith:
@djc Do I understand this correctly? (This is part of a project to move most of this logic from mk/clippy.sh to clippy.toml.)

djc:
I don't recall ever having used clippy's `--target-dir` option, so pretty sure this isn't needed!

PR #1743, opened 2023-10-14
CI: Run tests for `armv7-unknown-linux-musleabihf`.

briansmith:


briansmith:
Not sure why codecov is saying coverage decreased here since this did not affect coverage. Probably it is just being slow to sync with GitHub.

PR #1749, opened 2023-10-15
Add support for Hermit OS.

tema3210:
the impl in getrand: https://github.com/rust-random/getrandom/blob/master/src/hermit.rs


briansmith:
Sure. But see https://github.com/rust-random/getrandom/issues/377. I approved the CI run and will merge as-is once it passes.

PR #1754, opened 2023-10-16
add an opt-in less-safe-getrandom-custom feature

### LINKED ISSUE #1734: provide a mechanism to mark a custom `getrandom` implementation as being `SecureRandom`
Context: I'm working on adding no-std support to rustls and as part of that I'm building a small no-std demo that uses libc (instead of libstd) to perform networking. I'm using the "OS agnostic" `x86_64-unknown-none` target, which lacks a std implementation, to ensure libstd doesn't get sneakily used by a dependency but the demo will run on x86_64 Linux (for simplicity of testing it on CI but the demo should run on any (RT)OS that has a POSIX layer but lacks a libstd port).

I'm trying to use *ring* as the cryptographic backend of rustls in no-std mode and I'm able to build rustls with the *ring* Cargo feature enabled and the "std" feature disabled (WIP PR: rustls/rustls#1502) for the `x86_64-unknown-linux-gnu` target but the build fails for the `x86_64-unknown-none` target. I have tracked down the issue to this `SecureRandom` implementation:

https://github.com/briansmith/ring/blob/2e8363b433fa3b3962c877d9ed2e9145612f3160/src/rand.rs#L127-L149

The `x86_64-unknown-none` target has its "target_os" set to "none" so it doesn't match any of those conditionals. That means the `SystemRandom` type cannot be used with API like [`RsaKeyPair::sign`](https://github.com/rustls/rustls/blob/ac10422a898335230f64104b169461a50e9d61db/rustls/src/crypto/ring/sign.rs#L148) and [`EphemeralPrivateKey::generate`](https://github.com/rustls/rustls/blob/ac10422a898335230f64104b169461a50e9d61db/rustls/src/crypto/ring/kx.rs#L27) , which rustls uses when configured to use *ring* as its default crypto provider.

Would it be possible for *ring* to provide a mechanism to mark a custom `SystemRandom` as being `SecureRandom`? As of *ring* v0.17.0, one can enable `getrandom`'s "custom" feature to provide a custom `getrandom` implementation which is what `SystemRandom` uses. My demo will run on Linux so my custom `getrandom` implementation is the same as the `x86_64-unknown-linux-gnu` implementation so it's also a "secure" `getrandom` implementation.

Perhaps a Cargo feature can be used as the mechanism? Similar to rustls' "dangerous" Cargo feature. Let's say "secure-custom-getrandom":

``` toml
# *ring*'s Cargo.toml
[features]
secure-custom-getrandom = ["getrandom/custom"]
```

``` rust
// ring/src/rand.rs
#[cfg(any(
    feature = "secure-custom-getrandom",
    target_os = "android",
    // ..
))]
impl sealed::SecureRandom for SystemRandom { /* .. */ }
```

I think adding the `feature = ` at the `any` level is fine because `getrandom`'s documentation states that custom implementations [only have an effect on unsupported targets](https://docs.rs/getrandom/0.2.10/getrandom/macro.register_custom_getrandom.html#registering-a-custom-getrandom-implementation) so you cannot, for example, override the Linux `getrandom` implementation.

japaric:
This Cargo feature treats a user-provided `getrandom` implementation as a secure random number generator (`SecureRandom`). The feature only has effect on targets not supported by `getrandom`.

closes #1734

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -197,6 +197,7 @@ cc = { version = "1.0.83", default-features = false }
 default = ["alloc", "dev_urandom_fallback"]
 alloc = []
 dev_urandom_fallback = []
+less-safe-getrandom-custom = ["getrandom/custom"]
 slow_tests = []
 std = ["alloc"]
 unstable-testing-arm-no-hw = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -197,7 +197,7 @@ cc = { version = "1.0.83", default-features = false }
 default = ["alloc", "dev_urandom_fallback"]
 alloc = []
 dev_urandom_fallback = []
-less-safe-getrandom-custom = ["getrandom/custom"]
+less-safe-getrandom-custom-or-rdrand = []
 slow_tests = []
 std = ["alloc"]
 unstable-testing-arm-no-hw = []
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,6 +22,14 @@
 //!     <th>Description
 //! <tr><td><code>alloc (default)</code>
 //!     <td>Enable features that require use of the heap, RSA in particular.
+//! <tr><td><code>less-safe-getrandom-custom</code>
+//!     <td>Treat a user-provided ("custom") <code>getrandom</code>
+//!         implementation as a secure random number generator (see
+//!         <code>SecureRandom</code>). Only has effect on targets
+//!         <strong>not</strong> supported by <code>getrandom</code>.
+//!         See <a href="https://docs.rs/getrandom/0.2.10/getrandom/macro.register_custom_getrandom.html">
+//!             <code>register_custom_getrandom</code>
+//!         </a> for details.
 //! <tr><td><code>std</code>
 //!     <td>Enable features that use libstd, in particular
 //!         <code>std::error::Error</code> integration. Implies `alloc`.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,14 +22,16 @@
 //!     <th>Description
 //! <tr><td><code>alloc (default)</code>
 //!     <td>Enable features that require use of the heap, RSA in particular.
-//! <tr><td><code>less-safe-getrandom-custom</code>
-//!     <td>Treat a user-provided ("custom") <code>getrandom</code>
-//!         implementation as a secure random number generator (see
-//!         <code>SecureRandom</code>). Only has effect on targets
-//!         <strong>not</strong> supported by <code>getrandom</code>.
-//!         See <a href="https://docs.rs/getrandom/0.2.10/getrandom/macro.register_custom_getrandom.html">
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -197,6 +197,7 @@ cc = { version = "1.0.83", default-features = false }
 default = ["alloc", "dev_urandom_fallback"]
 alloc = []
 dev_urandom_fallback = []
+less-safe-getrandom-custom = ["getrandom/custom"]
```
briansmith, line 200: I don't think that *ring* needs to activate the `getrandom/custom` feature. The project will have to actually define the custom implementation in some crate, and *that* crate can activate the feature. Or they could decide to implement the `rdrand` feature instead.

Note that `getrandom`'s "custom" feature also doesn't work in the expected way on targets that have RDRAND; see https://github.com/rust-random/getrandom/issues/346.

Accordingly, let's rename this to `less-safe-getrandom-custom-or-rdrand` and use `[]` as the dependency.

src/lib.rs
```diff
@@ -22,6 +22,14 @@
 //!     <th>Description
 //! <tr><td><code>alloc (default)</code>
 //!     <td>Enable features that require use of the heap, RSA in particular.
+//! <tr><td><code>less-safe-getrandom-custom</code>
```
briansmith, line 25: Please update this too.

src/lib.rs
```diff
@@ -22,6 +22,14 @@
...
+//!     <td>Treat a user-provided ("custom") <code>getrandom</code>
+//!         implementation as a secure random number generator (see
+//!         <code>SecureRandom</code>). Only has effect on targets
+//!         <strong>not</strong> supported by <code>getrandom</code>.
```
briansmith, line 29: This is misleading because `getrandom` does "support" these targets in some way and also the user will get to choose whether to use `rdrand` and/or a custom implementation depending on the target.

src/rand.rs
```diff
@@ -125,6 +125,10 @@ impl crate::sealed::Sealed for SystemRandom {}
 // system's) CSPRNG. Avoid using it on targets where it uses the `rdrand`
 // implementation.
 #[cfg(any(
+    // NOTE `getrandom`'s (v0.2.10) docs state that a custom implementation will
+    // NOT override the implementation of supported targets, like the ones
+    // listed below.
+    feature = "less-safe-getrandom-custom",
```
briansmith, line 131: I (we?) don't want people to this new "less-safe" feature as a shortcut to avoiding this allowlist, so this should be `all(feature = "less-safe-....", target_os = "none")`.

briansmith (changes requested):
Thanks for doing this.

japaric:
thanks for the review. I have addressed your comments in a new commit. let me know if I should squash the commits.

japaric:
is there anything else that should be done in this PR?

PR #1756, opened 2023-10-17
Add simplified elliptic curve code generator.

briansmith:
Add a tool to generate and/or verify NIST curve boilerplate.

Manually sync the formatting of the existing code with the generator.

### DIFF
--- a/mk/generate_curves.py
+++ b/mk/generate_curves.py
@@ -16,8 +16,6 @@
 
 from textwrap import wrap
 
-limb_bits = 32
-
 curve_template = """
 // Copyright 2016-2023 Brian Smith.
 //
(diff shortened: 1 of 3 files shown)


vkrasnov:
Yes, I generated the params with python as well, they are the same, not sure why my code doesn't pass the 32bit tests, since code is reused from p384. I was unable to get a 32bit dev enviroment unfortunately.

briansmith:
I updated this PR substantially so that it can generate everything in in p384.rs except the addition chains.

I also extended the P-521 support so that it does the right thing for the 32-bit vs. 64-bit Montgomery stuff; it does the `#[cfg(...)]` stuff slightly differently than your P-521 PR does it. See the second commit in this PR. I verified the numbers match your numbers for both 64-bit and 32-bit.

> Yes, I generated the params with python as well, 

I didn't extend the code generator to generate gfp_p{384,521}.c. If you have Python code that can generate the constants in the C code, then could you please submit it as a change to this Python script?

PR #1757, opened 2023-10-17
ECDSA verification: Refactor a*g + b*p multiplication

briansmith:
Only the nistz256 implementation should be doing this multiplication in the naïve way. Even then, we should be using the optimized variable-time implementation from BoringSSL, so bring it in and use it.

After this PR, we can more easily replace `twin_mul_inefficient` with a proper twin multiplication implementation that works for any curve.

See the individual commit messages to see the changes made to the imported BoringSSL code; the second commit imports it unmodified, and the third commit makes the needed changes.


briansmith:
@vkrasnov PTAL.

PR #1758, opened 2023-10-18
EC: Use only array types when crossing Rust<->C boundary

briansmith:
Don't assume that a struct that contains only arrays has no padding, and don't assume we can cast between a `P{256,384}_POINT` and a `Limb[3][P{256,384}_LIMBS]` array.


briansmith:
@vkrasnov I don't know that this presently solves any problem but it is definitely safer to do it this way. I wrote this when trying to figure out what's wrong with your P-521 code on a 32-bit machine.

PR #1759, opened 2023-10-18
P-256 ECDSA: Use BoringSSL's WNAF implementation

briansmith:
For P-256 ECDSA verification, when not using nistz256, use the WNAF-based multiplication. This is faster.

This will serve as a model for the Rust WNAF-based multiplication for P-384 and P-521.


briansmith:
This is blocked on #1757 and #1758.

briansmith:
PR #1764 proposes a generalized solution to this. We should compare the performance of that with the performance of this.

briansmith:
> PR https://github.com/briansmith/ring/pull/1764 proposes a generalized solution to this. We should compare the performance of that with the performance of this.

Accordingly, blocked on #1772.

PR #1764, opened 2023-10-19
Use WNAF-based multiplication for ECDSA verification

briansmith:
Implement a generic ECDSA WNAF-based twin-multiplication function.


briansmith:
Closing this in favor of #1768, which strips out the `fallback` point addition and doubling functions.

PR #1769, opened 2023-10-24
EC Suite B: Remove reudndant "point_mul" tests.

briansmith:
When I generated these test vectors, I gave all of them the same point: the generator of the curve. Consequently these input files are 100% redundant with the `point_mul_base.txt` input files. So just remove them and use the `point_mul_base.txt` files instead.


briansmith:
@vkrasnov This is the first step towards checking in the (new) generator for the curve arithmetic test generator.

PR #1770, opened 2023-10-25
EC: Avoid Montgomery encoding in point arithmetic test data files.

briansmith:
Add a new scalar base point multiplication test case generator that where the points are *not* Montgomery-encoded. This way we don't need to generate different test data files when the Montgomery encoding for a curve isn't the same for 32-bit and 64-bit targets (P-521).

This version of the generator produces the test cases for all the scalars that the current P-256 and P-384 tests generate, in the same format; the only exception is that the point is not Montgomery-encoded.


briansmith:
Remaining work:

- [ ] Change the p{256,384}_point_{double,sum} tests so that they don't use Montgomery encoding for the input or output points. For the testing of edge cases of point doubling and point addition (handling of points at infinity in inputs and outputs), Unfortunately Go's API (even the deprecated one these tests use) doesn't expose Jacobian points so we probably can't write these generators in Go.
- [ ] Change the test scaffolding for the (base) point multiplication tests so that it Montgomery-encodes the expected result from the input file before comparing to the actual computed result. Then regenerate the .txt files by running the generators so that the input points are no longer Montgomery-encoded.

briansmith:
I'm going to merge the test case generator as-is in the interest of helping others contribute towards solving the bigger problem.

Issue #1772, opened 2023-10-26
Add ECDSA verification benchmarks


briansmith:
We need benchmarks that verify ECDSA signatures in order to make progress on various improvements to ECDSA and to support the project of moving from C to Rust.

These benchmarks should use a fixed ECDSA public key loaded from a checked-in file and a fixed (set of) signature(s). Don't generate a new ECDSA keypair as part of the benchmark scaffolding, and don't generate new signatures in the scaffolding, because the ECDSA verification code is variable-time with respect to the bit pattern in the signature (the values of r and s), and we want the benchmarks to provide reproducible/comparable timing.

briansmith:
PR #1773 provides a good example of how to do this.

PR #1776, opened 2023-10-27
Benchmarks: add ECDSA benchmarks

vkrasnov:


bench/ecdsa.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+}
+
+fn verify(c: &mut Criterion) {
+    for (alg_name, sign_alg, verify_alg) in ALGORITHMS {
+        c.bench_function(&format!("ecdsa_{alg_name}_verify"), |b| {
+            let rng = rand::SystemRandom::new();
+            let pkcs8_bytes = EcdsaKeyPair::generate_pkcs8(sign_alg, &rng).unwrap();
+            let key_pair = EcdsaKeyPair::from_pkcs8(sign_alg, pkcs8_bytes.as_ref(), &rng).unwrap();
+            let signature = key_pair.sign(&rng, &[]).unwrap();
```
briansmith, line 55: We're going to run into issue described in https://github.com/briansmith/ring/issues/1772 where the performance of each run is going to vary based on the randomized values of `r` and `s` in this signature.

bench/ecdsa.rs
```diff
@@ -0,0 +1,70 @@
...
+    (
+        "p384",
+        &signature::ECDSA_P384_SHA384_FIXED_SIGNING,
+        &signature::ECDSA_P384_SHA384_FIXED,
+    ),
+];
+
+fn sign(c: &mut Criterion) {
+    for (alg_name, alg, _) in ALGORITHMS {
+        c.bench_function(&format!("ecdsa_{alg_name}_sign"), |b| {
```
briansmith, line 37: We're going to end up with the "fixed" and non-fixed (ASN.1) benchmarks having the same benchmark name.

I suggest we just remove the `_FIXED_` variants from the bemchmarking to resolve this. I don't think we need to worry about the difference in performance between the two ways of formatting the signature and also the performance difference between the two is probably impossible to measure in a meaningful way.

bench/ecdsa.rs
```diff
@@ -0,0 +1,70 @@
...
+            b.iter(|| {
+                key_pair.sign(&rng, black_box(&[])).unwrap();
+            })
+        });
+    }
+}
+
+fn verify(c: &mut Criterion) {
+    for (alg_name, sign_alg, verify_alg) in ALGORITHMS {
+        c.bench_function(&format!("ecdsa_{alg_name}_verify"), |b| {
```
briansmith, line 51: ditto here.

briansmith (changes requested):

bench/ecdsa.rs
vkrasnov: Not sure what do you mean, the name is whatever string I put there in the ALGORITHMS array. I also think padding performance is not critical, so I only put the FIXED variants.

briansmith: I don't remember what it looked like before but it is fine the way it is now.

PR #1780, opened 2023-10-30
Extend elliptic curve code generator to generate C code boilerplate.

briansmith:
Generate some of the C boilerplate, particularly the large constants. The output is written into target/curves/, and can be merged into the actual code in crypto/fipsmodule/ec/ using a two-way merge tool; this is the same as the Rust code generation.

Changes to gfp_p{256,384}.c are due to differences in the generator's output:

* The generator doesn't generate trailing commas in arrays.
* The generator consistently avoids adding leading zeros to hex constants, and consistently format values less than 10 in decimal; the exiting code used a mix of styles.
* The generator wraps arrays consistently; the existing code used a mix of wrapping styles.
* The generator does not nest constants in the functions that need them. This was changed to support future refactorings.
* In order to support P-521, we avoid `TOBN` as it won't work for 32-bit targets for P-521 since there are an odd number of limbs in that case.

### DIFF
--- a/crypto/fipsmodule/ec/gfp_p256.c
+++ b/crypto/fipsmodule/ec/gfp_p256.c
@@ -24,10 +24,12 @@ typedef Limb Scalar[P256_LIMBS];
 #include "../bn/internal.h"
 
 static const BN_ULONG N[P256_LIMBS] = {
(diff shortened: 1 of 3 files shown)


briansmith:
@vkrasnov PTAL.

PR #1794, opened 2023-11-03
generate asm symbols for `target_os = "none"` targets

### LINKED ISSUE #1793: *ring* fails to build for `target_os = "none"` targets
I'm tried compiling a no-std executable that depends on *ring* to the `x86_64-unknown-none` target and got a linker error listing several "undefined references" like `ring_core_0_17_5_ChaCha20_ctr32` and `ring_core_0_17_5_bn_mul_mont`.

I have tracked down the problem to these being symbols in the (pregenerated?) asm group. The build script does not produce the assembly files because `none` is not in [the `LINUX_ABI` list](https://github.com/briansmith/ring/blob/8ed48604d0e76d27858b314b661adc78e835f123/build.rs#L249-L263). Adding `"none"` to this list fixes the linker error and produces a working binary. I have also tested that my program works (+) with the `aarch64-unknown-none` target. I'll submit a PR with that change.

(+) I'll eventually push the source code of my program to [this draft rustls PR](https://github.com/rustls/rustls/pull/1534)

japaric:
the built-in rustc targets `{x86,x86_64,arm,aarch64}-none` all use the ELF format

fixes #1793 

more context can be found in issue #1793 . I have tested this with the `x86_64-unknown-none` and `aarch64-unknown-none` targets and run the binary on Linux. I plan to test `thumbv7em-none-eabihf` in the future.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -260,6 +260,7 @@ const LINUX_ABI: &[&str] = &[
     "linux",
     "redox",
     "solaris",
+    "none",
 ];
 
 /// Operating systems that have the same ABI as macOS on every architecture


briansmith:
> the built-in rustc targets {x86,x86_64,arm,aarch64}-none all use the ELF format

As I'm sure you know alredy, the issue isn't ELF vs. non-ELF but rather ABI (and calling convention) compatibility. See my comment in the issue #1793.

PR #1805, opened 2023-11-13
getauxval isn't defined on uclibc, so disable dynamic feature detection

skrap:
This is my attempt at the suggested alternative approach @briansmith suggested in https://github.com/briansmith/ring/pull/1800

### DIFF
--- a/src/cpu/arm.rs
+++ b/src/cpu/arm.rs
@@ -17,9 +17,18 @@
     allow(dead_code)
 )]
 
+// uclibc: When linked statically, uclibc doesn't provide getauxval.
+// When linked dynamically, recent versions do provide it, but we
+// want to support older versions too. Assume that if uclibc is being
+// used, this is an embedded target where the user cares a lot about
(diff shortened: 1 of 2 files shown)


src/cpu/arm.rs
```diff
@@ -19,7 +19,8 @@
 
 #[cfg(all(
```
briansmith, line 20: Please add a comment above this along the following lines (with any typos fixed):
> uclibc: When linked statically, uclibc doesn't provide `getauxval`. When linked dynamically, recent versions do provide it, but we want to support older versions too. Assume that if uclibc is being used, this is an embedded target where the user cares a lot about minimizing code size and also that they know in advance exactly what target features are supported, so rely only on static feature detection.

skrap: Done!

briansmith (approved):

skrap:
Drat I forgot to run rustfmt after adding the comment.  Fixed.  Sorry!

skrap:
Thank you!

PR #1825, opened 2023-11-26
Remove libc dependency.

briansmith:
Use `core::ffi` instead of `libc`.


Cargo.toml
```diff
@@ -177,18 +177,12 @@ untrusted = { version = "0.9" }
 [target.'cfg(any(target_arch = "x86",target_arch = "x86_64", all(any(target_arch = "aarch64", target_arch = "arm"), any(target_os = "android", target_os = "fuchsia", target_os = "linux", target_os = "windows"))))'.dependencies]
 spin = { version = "0.9.8", default-features = false, features = ["once"] }
 
-[target.'cfg(all(any(target_os = "android", target_os = "linux"), any(target_arch = "aarch64", target_arch = "arm")))'.dependencies]
-libc = { version = "0.2.148", default-features = false }
-
```
briansmith, line 182: `getauxval` is provided by libc so if we're using `getauxval` we might as well keep using the `libc` crate so that it will handle the linking. However, if we keep using libc on these targets, we should limit the `cfg` to where we actually use libc::getauxval.

briansmith:
Replaced by PR #1829.

PR #1842, opened 2023-12-01
cpu arm: Fix static feature detection initialization.

briansmith:
Commit f932b941bd1f59782cb3db8f7cd7b8b2c9842ee9 was incomplete and wrong. On targets where we do any static or dynamic feature detection and where we have these global variables, we need to unconditionally write the detected features to the global variable so that assembly can see them. Since we do static feature detection regardless of operating system, the initialization of the global most be done without any conditions on the operating system.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -174,7 +174,7 @@ name = "ring"
 getrandom = { version = "0.2.10" }
 untrusted = { version = "0.9" }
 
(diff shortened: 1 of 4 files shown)


briansmith:
I expanded the CI testing in #1843 and verified that such testing would have prevented this bug.

PR #1851, opened 2023-12-05
montgomery: Do runtime limb lengths during Montgomery multiplication.

briansmith:
Our expectation is that these checks will never trigger errors because the prerequisites are always checked. However, the checks don't cost anything significant and this clarifies the memory safety.


briansmith:
Rebased on main after #1856 was merged.

briansmith:
Superceded by another PR that did the same.

PR #1864, opened 2023-12-14
AEAD: Refactor to make adding out-of-place open/seal APIs easier.

briansmith:


src/inout.rs
```diff
@@ -0,0 +1,134 @@
...
+
+use crate::{error, polyfill};
+use core::{convert::TryInto, marker::PhantomData, mem::MaybeUninit, ops::RangeFrom};
+
+pub struct InOut<'i, 'o, T> {
+    input: *const T,
+    output: *mut MaybeUninit<T>,
+    len: usize,
+    _input: PhantomData<&'i [T]>,
+    _output: PhantomData<&'o mut [T]>,
```
briansmith, line 11: This is the main problem with the sketched implementation. We end up using these pointers and then trying to emulate overlapping mutable and slices with them. Since Rust is designed to not support a mutable slice overlapping with anything, the whole premise is really questionable. So I will reimplement in terms of actual slices so that the borrow checker can help ensure the safety.

PR #1868, opened 2023-12-20
Simplified `SystemRandom` with const constructor and `rand` (crate) API impl.

elfeiin:
Quality of life PR to make rng easier to use. Anyone who understands `rand` can follow their examples to use `SystemRandom`.

### Considered changes not done in this PR (yet)
Remove superfluous `new` constructor
Add more docs
Back `ring::rand` with crate `rand` under feature flag

If you merge I will be like this: :>

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -109,13 +109,13 @@ impl<T> RandomlyConstructable for T where T: self::sealed::RandomlyConstructable
 /// `fill()` once at a non-latency-sensitive time to minimize latency for
 /// future calls.
 #[derive(Clone, Debug)]
-pub struct SystemRandom(());
+pub struct SystemRandom;
 
 impl SystemRandom {
     /// Constructs a new `SystemRandom`.
     #[inline(always)]
     pub fn new() -> Self {
-        Self(())
+        Self
     }
 }
 
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -18,6 +18,10 @@
 //! parameter to make it clear which functions are non-deterministic.
 
 use crate::error;
+#[cfg(feature = "rand-api")]
+use core::num::NonZeroU32;
+#[cfg(feature = "rand-api")]
+use rand::RngCore;
 
 /// A secure random number generator.
 pub trait SecureRandom: sealed::SecureRandom {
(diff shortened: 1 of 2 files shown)


src/rand.rs
```diff
@@ -109,13 +113,43 @@ impl<T> RandomlyConstructable for T where T: self::sealed::RandomlyConstructable
...
 /// future calls.
 #[derive(Clone, Debug)]
-pub struct SystemRandom(());
+pub struct SystemRandom;
 
 impl SystemRandom {
     /// Constructs a new `SystemRandom`.
     #[inline(always)]
-    pub fn new() -> Self {
-        Self(())
```
briansmith, line 118: It is intentional that we reserve the right to add fields to `SystemRandom` and we reserve the right for `SystemRandom` to not be `const`-constructable. We will need to wait until the FIPS project is further along before we can revisit this.

src/rand.rs
```diff
@@ -109,13 +113,43 @@ impl<T> RandomlyConstructable for T where T: self::sealed::RandomlyConstructable
...
+    pub const fn new() -> Self {
+        Self
+    }
+}
+
+#[cfg(feature = "rand-api")]
+impl RngCore for SystemRandom {
```
briansmith, line 127: It is intentional that we do not implement `rand` traits. If somebody is implementing cryptography they shouldn't use `rand`. If they aren't implementing cryptography then they probably shouldn't be using `ring::rand`. If they are implementing something that maybe is cryptography and maybe not then we'll leave it up to them as to how to solve that.

briansmith (commented):
I expect to make changes to this API as part of the FIPS project for *ring*, and then after that we can revisit the API to see how we can make it more convenient. I filed #1875 to help people understand some of the design intent and rough plans.

src/rand.rs
elfeiin: Understood. Thank you.

src/rand.rs
elfeiin: I see- I thought there might be a reason. Thank you for your time.

briansmith:
I'm going to close this. In part, I doubt we'd add a new dependency on `rand` just to implement the `rand` traits. But also because these two changes would be better done as separate PRs if/when we do want to consider them.

PR #1880, opened 2024-01-04
Add riscv32 support

blue-freedom-technologies:
> I am happy to accept a PR that adds riscv32 support. I suggest we start with riscv32gc-unknown-linux-gnu since we can just copy-paste-edit the changes from PR https://github.com/briansmith/ring/pull/1627, at least for for target.h, mk/cargo.sh, mk/install-build-tools.sh.

### DIFF
--- a/include/ring-core/target.h
+++ b/include/ring-core/target.h
@@ -52,7 +52,6 @@
 #define OPENSSL_RISCV64
 #elif defined(__riscv) && __SIZEOF_POINTER__ == 4
 #define OPENSSL_32_BIT
-#define OPENSSL_RISCV32
 #elif defined(__s390x__)
 #define OPENSSL_64_BIT
 #define OPENSSL_S390X
--- a/include/ring-core/target.h
+++ b/include/ring-core/target.h
@@ -50,6 +50,9 @@
 #elif defined(__riscv) && __SIZEOF_POINTER__ == 8
 #define OPENSSL_64_BIT
 #define OPENSSL_RISCV64
+#elif defined(__riscv) && __SIZEOF_POINTER__ == 4
+#define OPENSSL_32_BIT
+#define OPENSSL_RISCV32
 #elif defined(__s390x__)
 #define OPENSSL_64_BIT
 #define OPENSSL_S390X
--- a/mk/install-build-tools.sh
+++ b/mk/install-build-tools.sh
@@ -128,6 +128,12 @@ case $target in
     libc6-dev-riscv64-cross \
     qemu-user
   ;;
+--target=riscv32gc-unknown-linux-gnu)
+  use_clang=1
+  install_packages \
+    #https://github.com/riscv-collab/riscv-gnu-toolchain#installation-newliblinux-multilib
+    qemu-user
+  ;;
 --target=s390x-unknown-linux-gnu)
   # Clang is needed for code coverage.
   use_clang=1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,6 +153,7 @@ jobs:
           - i686-unknown-linux-musl
           - powerpc-unknown-linux-gnu
           - riscv64gc-unknown-linux-gnu
+          - riscv32gc-unknown-linux-gnu
           - wasm32-wasi
           - x86_64-pc-windows-gnu
           - x86_64-pc-windows-msvc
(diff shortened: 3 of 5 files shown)


blue-freedom-technologies:
Hi @briansmith,

There are two points here that we need your advise:
-[OPENSSL_32_BIT](https://github.com/blue-freedom-technologies/ring/blob/riscv32_support/include/ring-core/target.h#L54)
-[newliblinux-multilib](https://github.com/blue-freedom-technologies/ring/blob/riscv32_support/mk/install-build-tools.sh#L134)

Thank you for your time.

Cheers,

include/ring-core/target.h
```diff
@@ -50,6 +50,9 @@
 #elif defined(__riscv) && __SIZEOF_POINTER__ == 8
 #define OPENSSL_64_BIT
 #define OPENSSL_RISCV64
+#elif defined(__riscv) && __SIZEOF_POINTER__ == 4
+#define OPENSSL_32_BIT
+#define OPENSSL_RISCV32
```
briansmith, line 55: This is what BoringSSL has:
```
#elif defined(__riscv) && __SIZEOF_POINTER__ == 8
#define OPENSSL_64_BIT
#define OPENSSL_RISCV64
#elif defined(__riscv) && __SIZEOF_POINTER__ == 4
#define OPENSSL_32_BIT
```
In particular, they don't define `OPENSSL_RISCV32`. So I suggest we remove that. My plan is to remove all of the `OPENSSL_{arch}` defines for architectures where we do not have architecture-specific C/asm code, including `OPENSSL_RISCV64` above, in the near future.

mk/install-build-tools.sh
```diff
@@ -128,6 +128,12 @@ case $target in
     libc6-dev-riscv64-cross \
     qemu-user
   ;;
+--target=riscv32gc-unknown-linux-gnu)
+  use_clang=1
+  install_packages \
+    #https://github.com/riscv-collab/riscv-gnu-toolchain#installation-newliblinux-multilib
```
briansmith, line 134: I see that this target is quite difficult to support. I read "If you need riscv32-linux, then yes you will need to build it." from https://github.com/riscv-collab/riscv-gnu-toolchain/issues/808.

It seems like my niave and underinformed guess that riscv32gc-unknown-linux-gnu would be the easiest riscv32 target to support was wrong.

Is riscv32gc-unknown-linux-gnu the actual target you are trying to support, or were you just trying to follow my suggestion? If you don't need riscv32gc-unknown-linux-gnu but instead need a different target, I think we should focus on the one that's actually useful.

include/ring-core/target.h
blue-freedom-technologies: Agreed, I have remove the "OPENSSL_RISCV32".

mk/install-build-tools.sh
blue-freedom-technologies: Thank you for taking the time for this analysis.

Ok, don't forget you ask for it not me.... don't kill the messenger. 

Now the bad news: 
our target [riscv32imac-esp-espidf](https://doc.rust-lang.org/nightly/rustc/platform-support/esp-idf.html) is a [Tier 3](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-3). Meaning target is supported but not built, distributed, or guaranteed to work. 

So, for our target to work we need to have the [ESP-IDF Toolchain](https://docs.espressif.com/projects/esp-idf/en/v5.1.2/esp32/get-started/linux-macos-setup.html#standard-toolchain-setup-for-linux-and-macos) and then have a config.toml thing similar to this :

```rust
[build]
target = "riscv32imac-esp-espidf"

[target.riscv32imac-esp-espidf]
linker = "/tools/riscv32-esp-elf/esp-12.2.0_20230208/riscv32-esp-elf/bin/ld"

[unstable]
build-std = ["std", "panic_abort"]

[env]
MCU="esp32c6"
ESP_IDF_VERSION = "v5.1.1"
TARGET_CC = "/tools/riscv32-esp-elf/esp-12.2.0_20230208/riscv32-esp-elf/bin/riscv32-esp-elf-cc"
```

For the ring crate build.

As example let's run the test case ed25519_tests:

```bash
cargo build --test ed25519_tests
```

Result...

image


So, I am open to suggestions on your side....when you have the time of course.  :frowning_face:

blue-freedom-technologies:
On a more happy note I have added the [target_os = "espidf"](https://github.com/blue-freedom-technologies/ring/commit/41d60763ef99254746e4df7f95809a02e99321de#diff-ad78f8e30b4e145564968b663360cb796a75ff17cd7eb6b9373ed7a853d38b91).

Thank you for your time addressing all of these troubles.

Cheers,

PR #1882, opened 2024-01-04
B/generate curve

vkrasnov:


briansmith:
Vlad, I have create branch [boringssl-generate_tests](https://github.com/briansmith/ring/tree/boringssl-generate_tests) which is BoringSSL (not *ring*) with these changes applied. I split out the clang-format and a few other changes into separate commits. The main difference is a small refactoring so that the long multiline comments don't get trashed by the clang-format. The other change was to remove the .patch file, since I think it will be annoying to maintain.

Since that branch is actually BoringSSL, you can add this repo as a remote for your BoringSSL checkout and then fetch that branch into it.

If we need to make more changes to generate-tests.cc, we can create PRs against the `boringssl-generate_tests` branch.

briansmith:
Closing this; see the previous comment.

PR #1885, opened 2024-01-09
Require __BYTE_ORDER__ to be defined for bi-endian target architectures.

briansmith:


briansmith:
cc @uweigand @erichte-ibm @pkubaj. The basic idea here is to ensure that __BYTE_ORDER__ is defined properly since @uweigand's contribution to crypto/internal.h requires __BYTE_ORDER__ to be defined correctly, and to avoid using any other indication of (big-)endianness.

ecnelises (commented):
maybe off-topic but isn't `mipsel` little-endian?

include/ring-core/target.h
```diff
@@ -45,16 +58,11 @@
...
-#define OPENSSL_64_BIT
-#define OPENSSL_RISCV64
 #elif defined(__s390x__)
 #define OPENSSL_64_BIT
 #define OPENSSL_S390X
-#elif defined(__wasm__)
-#define OPENSSL_32_BIT
 #else
 // Note BoringSSL only supports standard 32-bit and 64-bit two's-complement,
 // little-endian architectures. Functions will not produce the correct answer
```
DavidHorton5339, line 68: Please update comment to reflect that some big-endian targets are supported.

DavidHorton5339 (approved):
LGTM - though please see suggestion about comment.

DavidHorton5339:
> maybe off-topic but isn't `mipsel` little-endian?

MIPS is bi-endian, so although mipsel is the little-endian variant, OPENSSL_MIPS is insufficient on it own to uniquely identify the target elsewhere in the code. That's why __BYTE_ORDER__ is needed too, and the purpose of this PR is to make sure that is set for bi-endian target architectures where this uncertainty exists.

Perhaps it would be clearer to prepend 'Require __BYTE_ORDER__ to be defined for bi-endian target architectures.' to the comment block starting on line 45, and to replace 'big-endianness' with 'endianness'.

include/ring-core/target.h
briansmith: I just removed the comment.

briansmith:
> replace 'big-endianness' with 'endianness'.

I implemented this suggestion.

briansmith:
Thanks for reviewing this. I really appreciate it.

PR #1892, opened 2024-01-11
bigint: Improve temporary array alignment.

briansmith:
Replace pointer-to-usize `as` conversions with use of `align_offset`. This avoids any dependency on the way pointers are encoded.

Previously when the buffer was already aligned the way we needed it, we'd treat the buffer as being misaligned by `ALIGNMENT` bytes. This wasn't problematic but it wasn't optimal. By using `align_offset`, we'll recognize it as being misaligned by 0 bytes, i.e. already aligned.

Previously we were allocating `ALIGNMENT` extra Limbs to have slack for alignment where we could have allocated `ALIGNMMENT / LIMB_BYTES` extra limbs. Now, since we make the minimum alignment adjustment necessary, we can allocate 1 less limb than that.

Be explicit that the type of the buffer elements is `Limb`. Use `core::mem::{align_of}::size_of::<Limb>()` instead of `LIMB_BYTES`.

### DIFF
--- a/src/arithmetic/bigint.rs
+++ b/src/arithmetic/bigint.rs
@@ -491,7 +491,9 @@ pub fn elem_exp_consttime<M>(
     exponent: &PrivateExponent,
     m: &Modulus<M>,
 ) -> Result<Elem<M, Unencoded>, error::Unspecified> {
-    use crate::{cpu, limb::LIMB_BYTES};
+    use crate::cpu;
+    use alloc::vec::Vec;
+    use core::mem;
 
     // Pretty much all the math here requires CPU feature detection to have
     // been done. `cpu_features` isn't threaded through all the internal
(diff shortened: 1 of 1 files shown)


briansmith:
@joshlf Are you around? This isn't dealing with `unsafe` exactly but it seems like you'd be the best person to review it.

src/arithmetic/bigint.rs
```diff
@@ -512,13 +514,19 @@ pub fn elem_exp_consttime<M>(
...
+    const _ALIGNMENT_IS_MULTIPLE_OF_LIMB_ALIGNMENT: () =
+        assert!(ALIGNMENT % mem::align_of::<Limb>() == 0);
+    const EXTRA_FOR_ALIGNMENT: usize = (ALIGNMENT / mem::size_of::<Limb>()) - 1;
+    let mut table: Vec<Limb> = vec![0; ((TABLE_ENTRIES + 3) * num_limbs) + EXTRA_FOR_ALIGNMENT];
     let (table, state) = {
-        let misalignment = (table.as_ptr() as usize) % ALIGNMENT;
-        let table = &mut table[((ALIGNMENT - misalignment) / LIMB_BYTES)..];
-        assert_eq!((table.as_ptr() as usize) % ALIGNMENT, 0);
+        let align_offset = table.as_ptr().align_offset(ALIGNMENT);
+        let table = &mut table[align_offset..];
```
joshlf, line 529: Assuming the "obvious" behavior of `align_offset`, this is fine, but technically if we want to be conservative about reading this part of [the documentation](https://doc.rust-lang.org/std/primitive.pointer.html#method.align_offset):

> If it is not possible to align the pointer, the implementation returns `usize::MAX`. It is permissible for the implementation to always return `usize::MAX`. Only your algorithm’s performance can depend on getting a usable offset here, not its correctness.

...then this isn't sufficient, and we need to check for `usize::MAX`. Let me know if that's something you care about.

joshlf (changes requested):

src/arithmetic/bigint.rs
briansmith: > ...then this isn't sufficient, and we need to check for usize::MAX. Let me know if that's something you care about.

Well, I don't think we should rely on behavior that isn't guaranteed. In this case, it seems like there could be a more useful guarantee, and it would be good to change libcore so that such a stronger guarantee is made. But I doubt that will happen soon.

This was intended to be an intermediate step towards stack-allocating this array. I think I'm just going to skip this step and go directly to stack allocating.

Thanks for your careful review. I really appreciate it!

joshlf: Of course!

briansmith:
This was redone a better way in #2299

Issue #1827, opened 2023-11-27
Stop using OpenSSL-license-licensed code.


tisonkun:
Currently, the [LICENSE](https://github.com/briansmith/ring/blob/6c29bf61cd0e9d478deeec11e922e8513d5d0f3c/LICENSE) file is large and complicated, while can be vaguely regarded as a combination of MIT, ISC, and OpenSSL’s licenses.

Although [the latest OpenSSL license is Apache License 2.0](https://www.openssl.org/source/apache-license-2.0.txt), in this repo, we use a few of sentences refer to the BSD-style licenses with advertising clauses:

https://github.com/briansmith/ring/blob/6c29bf61cd0e9d478deeec11e922e8513d5d0f3c/LICENSE#L66-L83

https://github.com/briansmith/ring/blob/6c29bf61cd0e9d478deeec11e922e8513d5d0f3c/LICENSE#L137-L145

It causes concerns in downstream for using this software in a mindset like so-called "permissive OSS license" or "weak copyleft license": https://lists.apache.org/thread/ptwdv18z4wd9r11nmdwj7wgzwvm3b8l2

@briansmith do you have more background of the license content, or how generally a downstream user use it?

The background of this questions is from https://www.apache.org/legal/resolved.html#category-x that "BSD-4-Clause/BSD-4-Clause (University of California-Specific)" can introduce burden for users to convey this software - they're, be required, to include extra acknowledgement for certern actions. And while if we can either:

1. use an Apache License 2.0 base, or;
2. be clear that what part of this crate is not covered by these licenses.

briansmith:
I believe it is already clear which parts of the crate are covered by which license as each source file has a license header at the top. The code in question comes from BoringSSL and they didn't do the relicensing. Unfortunately they (BoringSSL) put some important bits in OpenSSL-license-licensed header files. Our goal is to replace all the OpenSSL-license-licensed code as we move completely away from C to Rust. Once that is done there won't be any issue regarding this anymore.

See also #902 which tracks making the licensing use SPDX.

tisonkun:
@briansmith Thanks for your reply. I have one more question:

> it is already clear which parts of the crate are covered by which license

For a takeaway, are those files OpenSSL-license-licensed included in any downstream software that use ring? Or it's optional if certain feature unused.

I did a check and it seems the fileset is:

* [ ] crypto/constant_time_test.c
* [ ] crypto/cpu_intel.c
* [ ] crypto/internal.h
[...]

PR #1893, opened 2024-01-11
Refactor fe_isnonzero to fe_isnonzero_vartime

tisonkun:
This refers to #1827.

An intuitive solution according to @briansmith's comment. But I know little about crypto so perhaps some invariants are broken.


Cargo.toml
```diff
@@ -94,7 +94,6 @@ include = [
     "crypto/limbs/limbs.c",
     "crypto/limbs/limbs.h",
     "crypto/limbs/limbs.inl",
-    "crypto/mem.c",
```
briansmith, line 97: We can't remove crypto/mem.c until #1899 is implemented.

Cargo.toml
```diff
@@ -113,7 +112,6 @@ include = [
     "include/ring-core/asm_base.h",
     "include/ring-core/base.h",
     "include/ring-core/check.h",
-    "include/ring-core/mem.h",
```
briansmith, line 116: We should be able to remove mem.h with this PR though, if we remove the include from mem.c.

build.rs
```diff
@@ -43,7 +43,6 @@ const RING_SRCS: &[(&[&str], &str)] = &[
     (&[], "crypto/fipsmodule/ec/gfp_p384.c"),
     (&[], "crypto/fipsmodule/ec/p256.c"),
     (&[], "crypto/limbs/limbs.c"),
-    (&[], "crypto/mem.c"),
```
briansmith, line 46: Ditto. We need to retain it for now.

build.rs
```diff
@@ -885,7 +884,6 @@ fn prefix_all_symbols(pp: char, prefix_prefix: &str, prefix: &str) -> String {
     ];
 
     static SYMBOLS_TO_PREFIX: &[&str] = &[
-        "CRYPTO_memcmp",
```
briansmith, line 888: Ditto.

crypto/curve25519/curve25519.c
```diff
@@ -397,14 +395,18 @@ static void fe_invert(fe *out, const fe *z) {
...
-static int fe_isnonzero(const fe_loose *f) {
+static int fe_isnonzero_vartime(const fe_loose *f) {
   fe tight;
   fe_carry(&tight, f);
   uint8_t s[32];
   fe_tobytes(s, &tight);
 
-  static const uint8_t zero[32] = {0};
-  return CRYPTO_memcmp(s, zero, sizeof(zero)) != 0;
+  for (unsigned i = 0; i < 32; i++) {
```
briansmith, line 404: We use `size_t` for array indexes. I think we should use `sizeof(s)` instead of `32`.

crypto/mem.c
```diff
@@ -1,70 +0,0 @@
...
- * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
- * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
- * SUCH DAMAGE.
- *
- * The licence and distribution terms for any publically available version or
- * derivative of this code cannot be changed.  i.e. this code cannot simply be
- * copied and put under another distribution licence
- * [including the GNU Public Licence.] */
-
-#include <ring-core/mem.h>
```
briansmith, line 57: We need to restore this file but remove this `#include`.

src/constant_time.rs
```diff
@@ -14,27 +14,20 @@
...
 /// `a` and `b`.
 pub fn verify_slices_are_equal(a: &[u8], b: &[u8]) -> Result<(), error::Unspecified> {
-    if a.len() != b.len() {
-        return Err(error::Unspecified);
-    }
-    let result = unsafe { CRYPTO_memcmp(a.as_ptr(), b.as_ptr(), a.len()) };
-    match result {
-        0 => Ok(()),
-        _ => Err(error::Unspecified),
+    if a != b {
```
briansmith, line 24: We cannot do this because we need a constant-time comparison. See #1899.

briansmith (changes requested):
Thanks for tackling this.

tisonkun:
Thanks for your review @briansmith!

Do you think it's better to exclude changes in `src/constant_time.rs` and move forward this PR, or we can wait for https://github.com/briansmith/ring/pull/1899 merged?

briansmith:
Yes, I think we should limit this to the curve25519.c change and mem.h deletion.

tisonkun:
Updated. PTAL @briansmith.

crypto/curve25519/curve25519.c
```diff
@@ -403,8 +401,13 @@ static int fe_isnonzero(const fe_loose *f) {
   uint8_t s[32];
   fe_tobytes(s, &tight);
 
-  static const uint8_t zero[32] = {0};
```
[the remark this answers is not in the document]
tisonkun: Thanks for your information! Updated.

tisonkun:
@briansmith ping as a reminder :D

tisonkun:
@briansmith I suppose the codecov regression is false positive..

tisonkun:
@briansmith is there other blocker to this PR?

tisonkun:
As #1827 resolved, this seems unrelated.

PR #1859, opened 2023-12-08
Added MIPS (big-endian, 32-bit) support

DavidHorton5339:
Now that ring has generic fallback code, we can compile it for new targets by allowing them through target.h checks.
I'd like to add MIPS support. This works well (all unit tests pass, and our application works on the target).
Unfortunately the Rust compiler recently got broken for MIPS, so version 1.71 or earlier is required.

PR #1894, opened 2024-01-11
Added MIPS 32-bit big-endian

DavidHorton5339:
Now that ring has generic fallback code, we can compile it for new targets by allowing them through target.h checks.
I'd like to add MIPS support. This works well (all unit tests pass, and our application works on the target).
Unfortunately the Rust compiler recently got broken for MIPS, so version 1.71 or earlier is required.
Same as https://github.com/briansmith/ring/pull/1859, but without the unsquashable multiple commits over > 1 branch.

### DIFF
--- a/include/ring-core/target.h
+++ b/include/ring-core/target.h
@@ -43,6 +43,9 @@
 #elif defined(__MIPSEL__) && defined(__LP64__)
 #define OPENSSL_64_BIT
 #define OPENSSL_MIPS64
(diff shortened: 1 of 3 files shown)


briansmith:
My understanding is that this target basically wouldn't work in Rust 1.72 or later without a self-built toolchain? Is that right?

briansmith:
@DavidHorton5339 I am closing this in favor of #1910, which rebases this PR on top of the latest changes to main and adds this and mipsel to the CI test matrix.

PR #1898, opened 2024-01-11
internal: Add generalized "noinline" attribute support.

briansmith:
There are some functions that we never want to be (cross-language) inlined, and we expect to add more. Make it easier to add more, and make it easier to extend this capability to more C compilers.

### DIFF
--- a/crypto/internal.h
+++ b/crypto/internal.h
@@ -132,6 +132,14 @@
 #include <stdalign.h>
 #endif
 
+#if defined(__clang__) || defined(__GNUC__)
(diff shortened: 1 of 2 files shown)


briansmith:
The original submission of this PR used standard attirbute syntax `[[clang::noinline]]` when compiling with clang. But, apparently clang doesn't like mixing `__attribute__((...))` syntax with the standard attribute syntax, which was being done in the ADX code. So, I change RING_NOINLINE to use `__attribute__((...))` syntax for clang too.

PR #1900, opened 2024-01-12
clippy: fix warning on aarch64

vkrasnov:
Fix `warning: unneeded `return` statement` on aarch65


briansmith:
The code is written this way so that if we change the `#[cfg(...)]` logic, it will still do the right thing. Accordingly, normally we'd address this instead by putting a `#[allow(needless_return)]` on the function. In this case, it seems like we need to deal with some situations where we shouldn't be using NEON even on AAarch64, so I've submitted #1901 as an alternative solution that addresses this while stepping in that direction.

briansmith:
Closing since this was subsumed by PR #1901.

PR #1901, opened 2024-01-12
aes-gcm: Don't assume NEON is available on AArch64.

briansmith:
Some of the assembly probably assumes NEON is available on Aarch64, so don't remove the assertion in the feature detection logic. But move closer to supporting NEON-less Aarch64 configurations, e.g. within an OS kernel or when NEON is disabled for testing purposes, or when we have a target that for whatever reason doesn't statically enable NEON for AAarch64 and also doesn't support dynamic feature detection.


briansmith:
@vkrasnov Could you please take a look at this? I think this will subsume #1900.

PR #1905, opened 2024-01-13
CI: Use macos-13-large instead of macos-13-xlarge for aarch64-apple-darwin jobs.

briansmith:
Running these jobs cost $111.04 in December.


briansmith:
Sadly, only the xlarge runners are ARM64. the large runners are x86-64. So we can't do this.

PR #1908, opened 2024-01-13
CI: Use llvm-tools-preview instead of external llvm-tools for symbol prefix testing

briansmith:
See individual commit messages for details. This makes the script portable to all hosts and allows us to expand the symbol prefix checking to all tested targets. It also allows us to avoid incompatibilities between Apple nm and the toolchain's output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -290,10 +290,16 @@ jobs:
       #       Unknown attribute kind (528)
       #       (Producer: 'LLVM12.0.0-rust-1.54.0-nightly'
       #       Reader: 'LLVM APPLE_1_1200.0.32.29_0')
+
+      - if: ${{ matrix.target != 'aarch64-apple-ios' &&
+          !contains(matrix.host_os, 'windows') &&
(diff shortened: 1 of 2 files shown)


briansmith:
https://github.com/briansmith/ring/actions/runs/7515538763/job/20459551889 is a run that intentionally breaks symbol prefixing, to test this.

briansmith:
It looks like the Windows builds are using PowerShell to run the shell script, and "passing" even though they aren't doing any work. :( I'll need to update this to force the use of mingw bash for windows.

briansmith:
> It looks like the Windows builds are using PowerShell to run the shell script, and "passing" even though they aren't doing any work. :( I'll need to update this to force the use of mingw bash for windows.

Well, after that, it seems more work is needed for Windows. Will split the Windows work off into a separate issue/PR.

briansmith:
https://github.com/briansmith/ring/actions/runs/7516068245 is the new test run to make sure CI fails when an unprefixed symbol exists.

briansmith:
https://github.com/briansmith/ring/actions/runs/7516194196 was the final test run that shows that the builds break when there is an unprefixed symbol.

PR #1918, opened 2024-01-16
Build: Add compile-time checks regarding ABI assumptions.

briansmith:


briansmith:
@pixlwave PTAL at this, to understand the issues with arm64_32 in particular.

pixlwave:
Ok yep, this helps explain more about the build system/configuration to me 👍

PR #1914, opened 2024-01-15
Add support for watchOS.

pixlwave:
Hoping to make a follow-up PR to #1689 to add watchOS support, although in the current state this doesn't yet work for all of the targets:

- `aarch64-apple-watchos` is the only target to compile happily, although I don't believe there's any hardware actually using this yet 🙈
- `armv7k-apple-watchos` builds although there are 4 warnings about .o files with no symbols which I'm not sure if that is to be expected or not?
- `arm64_32-apple-watchos` fails to build, due to a couple of `implicit conversion loses integer precision` errors. I think this makes sense as arm64_32 is considered to be aarch64 by Rust, even if the pointer size is 32-bit but I have no idea where to fix this.
- `aarch64-apple-watchos-sim` and `x86_64-apple-watchos-sim` both fail to build due to conflicting clang arguments `-mwatchsimulator-version-min=2.0` and `--target={aarch64|x86_64}-apple-watchos5.0-simulator`, although again, I'm failing to see where these 2 parameters come from 🤔

Any pointers towards potential answers would be greatly appreciated 🙂

**Build output**

<details>
<summary>aarch64-apple-watchos ✅</summary>
<pre><code>ring git:(watchos) ❯ cargo +nightly build -Zbuild-std --target "aarch64-apple-watchos"
   Compiling ring v0.17.7 (/Users/runner/Fork/ring)
warning: field `0` is never read
  --> src/arithmetic/n0.rs:19:15
   |19
   | pub struct N0([Limb; 2]);
   |            -- ^^^^^^^^^
   |            |
   |            field in this struct
   |
   = note: `N0` has a derived impl for the trait `Clone`, but this is intentionally ignored during dead code analysis
   = note: `#[warn(dead_code)]` on by default
help: consider changing the field to be of unit type to suppress this warning while preserving the field numbering, or remove the field
   |
19 | pub struct N0(());
   |               ~~

warning: field `0` is never read
  --> src/ec/curve25519/scalar.rs:18:19
   |
18 | pub struct Scalar([u8; SCALAR_LEN]);
   |            ------ ^^^^^^^^^^^^^^^^
   |            |
   |            field in this struct
   |
help: consider changing the field to be of unit type to suppress this warning while preserving the field numbering, or remove the field
   |
18 | pub struct Scalar(());
   |                   ~~

warning: `ring` (lib) generated 2 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 1.90s</code></pre>
</details>

<details>
[...]

PR #1919, opened 2024-01-16
Build: DRY conditions for allowing use of x25519_neon assembly.

briansmith:
Expand the conditions in which the x25519_neon function will be avoided. This is a temporary measure until we can properly condition it on the target ABI.

### DIFF
--- a/src/ec/curve25519/x25519.rs
+++ b/src/ec/curve25519/x25519.rs
@@ -63,7 +63,7 @@ fn x25519_public_from_private(
     let private_key: &[u8; SCALAR_LEN] = private_key.bytes_less_safe().try_into()?;
     let private_key = ops::MaskedScalar::from_bytes_masked(*private_key);
 
-    #[cfg(all(not(target_os = "ios"), target_arch = "arm"))]
+    #[cfg(ring_x25519_neon)]
     {
         if cpu::arm::NEON.available(cpu_features) {
             static MONTGOMERY_BASE_POINT: [u8; 32] = [
@@ -109,7 +109,7 @@ fn x25519_ecdh(
         point: &ops::EncodedPoint,
         #[allow(unused_variables)] cpu_features: cpu::Features,
     ) {
-        #[cfg(all(not(target_os = "ios"), target_arch = "arm"))]
+        #[cfg(ring_x25519_neon)]
         {
             if cpu::arm::NEON.available(cpu_features) {
                 return x25519_neon(out, scalar, point);
@@ -158,8 +158,10 @@ fn x25519_ecdh(
     Ok(())
 }
 
-#[cfg(all(not(target_os = "ios"), target_arch = "arm"))]
+#[cfg(ring_x25519_neon)]
 fn x25519_neon(out: &mut ops::EncodedPoint, scalar: &ops::MaskedScalar, point: &ops::EncodedPoint) {
+    // x25519_NEON isn't compatible with Apple's ABI.
+    const _NOT_APPLE_COMPATIBLE: () = assert!(!cfg!(target_vendor = "apple"));
     prefixed_extern! {
         fn x25519_NEON(
             out: &mut ops::EncodedPoint,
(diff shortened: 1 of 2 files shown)


briansmith:
@pixlwave @lcruz99 @BlackHoleFox PTAL at this, since you've been interested in this topic previously.

src/ec/curve25519/x25519.rs
```diff
@@ -158,8 +158,10 @@ fn x25519_ecdh(
     Ok(())
 }
 
-#[cfg(all(not(target_os = "ios"), target_arch = "arm"))]
+#[cfg(ring_x25519_neon)]
 fn x25519_neon(out: &mut ops::EncodedPoint, scalar: &ops::MaskedScalar, point: &ops::EncodedPoint) {
+    const _NOT_APPLE_COMPATIBLE: () = assert!(!cfg!(target_vendor = "apple"));
```
BlackHoleFox, line 163: This doesn't make sense out of the gate, at least to me. Was the idea to followup and allow some Apple targets to use this? Even with that dropping M{1, 2, 3} in the middle comes off weird and puts main in a odd state (see below about the feasibility of `target_abi` gating). 

However I noticed this does fix a possible bug where 32-bit watchOS may try and use this incorrectly.

> This is a temporary measure until we can properly condition it on the target ABI.

I don't really think the ABI is needed, target arch and OS get the whole way to what you seem to want? And then `#[cfg(target_abi = "")]` [isn't even stable yet](https://github.com/rust-lang/rust/pull/119590) and probably won't be for a while so relying on it for a new release is foggy. This I'm less sure of, since I don't have context, but this plan might also stumble over the simulator targets that have `target_abi="sim"` without a workaround.

The only Apple device's that don't support NEON are 32-bit iOS (`armv7s)` and actually-32-bit watchOS (`armv7k`). 64-bit-but-half-pointers (`arm64_32`) and the other 64-bit watchOS target (`aarch64`) have it available. Those are the two targets that need excluded vs many safe-to-assume combos for Apple's platforms.

With the specifics covered, does _ring_'s policy forbid turning this on for any target with `cfg(target_feature = "neon")`? It seems like the simplest approach (if you're not worried about CPU bugs) and the assembly can be tested on another aarch64 device for correctness (register restoring, etc).

briansmith: The problem is that x25519_NEON's assembly language code doesn't work in Apple's OS's because it isn't ABI-compatible with them, I think because it uses a register that the OS reserves.

This only affects 32-bit targets (target_arch = "arm"), not 64-bit or arm64-32.

briansmith:
@BlackHoleFox Thanks for taking a look. I updated the PR to address the build.rs nit you mentioned, and to clarify the issue with x25519_NEON. PTAL.

pixlwave (approved):
Seems reasonable to me as a first step based on the discussions in #1914 👍

(Although full disclosure: I'm still relatively new to Rust, so learning about things like build.rs as I go).

src/ec/curve25519/x25519.rs
pixlwave: > The problem is that x25519_NEON's assembly language code doesn't work in Apple's OS's because it isn't ABI-compatible with them, I think because it uses a register that the OS reserves.
> 
> This only affects 32-bit targets (target_arch = "arm"), not 64-bit or arm64-32.

Not sure about conventions, but this seems like potentially useful insight to sum up in a comment in the code.

briansmith: I added a comment calling out the ABI-incompatibility, but without the speculation about the register usage, as I didn't find a definitive reason.

briansmith: RE: target_abi not being stable, and target_abi = sim, those are all good points. target_abi not being stable is one of the main reasons we have to condition on the target vendor.

briansmith:
I'm going to close this without merging it. I don't like the denylist-like approach. I think we only need to support x25519_neon for, at most, Linux and Android, so we should switch to an allowlist-based approach that lists them specifically. Further, we should change build.rs so that x25519_neon is not compiled at all on non-Linux/Android targets.

BlackHoleFox:
Seems reasonable, thanks for the backnforth.

Billyum50 (approved):

PR #1924, opened 2024-01-17
build.rs: Use named constants instead of string literals more consistently

briansmith:
This doesn't try to fix every case.


briansmith:
@BlackHoleFox PTAL. I wrote this based on your comments in the other PR.

BlackHoleFox:
Late getting back to this but nice, I think its a good improvement for maintainability :)

PR #1927, opened 2024-01-19
generate_tests.cc: Remove ecc-*-point-mul[-base] test generation.

briansmith:
The generation of these tests was replaced by crypto/fipsmodule/ec/make_ec_scalar_base_mult_tests.go in *ring*'s source tree).

The twin-mul test generation was left since there's no replacement in the *ring* source tree, though it seems we never have used its output. Before using them we should probably try to adapt the Go generator instead; if that succeeds, we should remove `GeneratePointMulTwinTests` too.

### DIFF
--- a/tool/generate_tests.cc
+++ b/tool/generate_tests.cc
@@ -990,86 +990,6 @@ static bool GeneratePointMulTest(const InterestingPoints &points,
 }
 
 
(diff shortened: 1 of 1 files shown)


briansmith:
@vkrasnov FYI.

PR #1928, opened 2024-01-19
ec: `go run crypto/fipsmodule/ec/make_ec_scalar_base_mult_tests.go`

briansmith:
Update the point multiplication test vectors to match exactly the output of the (new) generation script.


briansmith:
@vkrasnov FYI.

PR #1934, opened 2024-01-24
bench: Report throughput in AEAD benchmarks

vkrasnov:
This makes the results directly comparable to BoringSSL and OpenSSL

### DIFF
--- a/bench/aead.rs
+++ b/bench/aead.rs
@@ -70,11 +70,14 @@ impl aead::NonceSequence for NonceSequence {
 }
 
 fn seal_in_place_separate_tag(c: &mut Criterion) {
+    let mut group = c.benchmark_group("aead");
+
     let rng = SystemRandom::new();
 
     for &(alg_name, algorithm) in ALGORITHMS {
         for record_len in RECORD_LENGTHS {
-            c.bench_with_input(
+            group.throughput(criterion::Throughput::BytesDecimal(*record_len as _));
+            group.bench_with_input(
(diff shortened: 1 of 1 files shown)


bench/aead.rs
```diff
@@ -70,11 +70,14 @@ impl aead::NonceSequence for NonceSequence {
...
 
 fn seal_in_place_separate_tag(c: &mut Criterion) {
+    let mut group = c.benchmark_group("aead");
+
     let rng = SystemRandom::new();
 
     for &(alg_name, algorithm) in ALGORITHMS {
         for record_len in RECORD_LENGTHS {
-            c.bench_with_input(
+            group.throughput(criterion::Throughput::BytesDecimal(*record_len as _));
```
briansmith, line 79: It seems odd to set the throughput parameter on the group and not the individual benchmark, when the input length varies by benchmark, not by group. But this does seem to be what the criterion.rs documentation says to do and also I read the criterion.rs source code to verify that it does seem to do the right thing.

It also seems strange to measure in decimal units (1,000) instead of 1,024, but indeed that seems to be what BoringSSL does:
```
          static_cast<double>(bytes_per_call * num_calls) /
              static_cast<double>(us));
```
where us is the number of milliseconds.

briansmith (approved):

bench/aead.rs
vkrasnov: Indeed

PR #1937, opened 2024-01-27
Improve conditional compilation around PerlAsm

nspin:
`build.rs` determines whether the target platform is supported by PerlAsm using both `target_arch` and `target_os`. Instances of conditional compilation in both `src/` and `crypto/` were using just `target_arch` to determine whether PerlAsm symbols are present, resulting in link-time build failures for certain targets, including, for example, `aarch64-unknown-none`.

This commit fixes those instances of conditional compilation to align with the build script.

I've made the most conservative possible modifications to these `cfg` conditions, so some of them may contain redundancy. In particular, those that were  `any(target_arch = "aarch64", target_arch = "arm", target_arch = "x86_64", target_arch = "x86")` are now `all(have_perlasm, any(target_arch = "aarch64", target_arch = "arm", target_arch = "x86_64", target_arch = "x86"))`, but it may be the case that some of these ought to just be `have_perlasm`.

This PR resolves https://github.com/briansmith/ring/issues/1793. As discussed in that issue, the conditions in `build.rs` for PerlAsm support may be overly conservative, but expanding the set of targets which benefit from PerlAsm will be trickier. This PR is just a bug fix that enables platforms such as `aarch64-unknown-none` to use the fallbacks rather than failing to build at link-time.


build.rs
```diff
@@ -433,7 +435,9 @@ fn build_c_code(
 
     generate_prefix_symbols_asm_headers(out_dir, ring_core_prefix).unwrap();
 
-    let (asm_srcs, obj_srcs) = if let Some(asm_target) = asm_target {
+    let (asm_srcs, obj_srcs) = if let Some(asm_target) = find_asm_target(target) {
+        println!("cargo:rustc-cfg=have_perlasm");
```
briansmith, line 439: I suggest we name it just `perlasm`.

I am worried about the risk for users of non-Cargo build systems. I suggest when we do not set `perlasm`, we set `no_perlasm`. Then we should add a check like this to lib.rs to ensure that the such users have addressed this:
```
const PERLASM_CONFIGURED: () = assert!((cfg!(perlasm) && !cfg!(no_perlasm)) || (!cfg!(perlasm) && cfg!(no_perlasm)));
```

briansmith (changes requested):
Thank you for doing this. I am sorry about all the changes that were merged recently that created conflicts for this PR. 

I think this is a good step towards what we need.

src/lib.rs
```diff
@@ -142,3 +142,6 @@ mod sealed {
     // ```
     pub trait Sealed {}
 }
+
+const PERLASM_CONFIGURED: () =
+    assert!((cfg!(perlasm) && !cfg!(no_perlasm)) || (!cfg!(perlasm) && cfg!(no_perlasm)));
```
nspin, line 147: I couldn't find a more appropriate place for this assertion. What do you think about its placement?

build.rs
nspin: Good point. I've implemented this.

src/lib.rs
nspin: Oops, I just noticed that you specifically suggested we add this assertion to `lib.rs`. Marking as resolved.

crypto/fipsmodule/ec/p256_shared.h
```diff
@@ -24,7 +24,7 @@
 #include "../bn/internal.h"
 
 #if !defined(OPENSSL_NO_ASM) && \
-    (defined(OPENSSL_X86_64) || defined(OPENSSL_AARCH64)) && \
+    defined(RING_HAVE_PERLASM) && \
```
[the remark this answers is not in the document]
nspin: Done.

src/lib.rs
```diff
@@ -142,3 +142,6 @@ mod sealed {
     // ```
     pub trait Sealed {}
 }
+
+const PERLASM_CONFIGURED: () =
```
briansmith, line 146: This must be named `_PERLASM_CONFIGURED` to pass the `clippy` job.

build.rs
```diff
@@ -419,6 +423,8 @@ fn build_c_code(
     core_name_and_version: &str,
 ) {
     let (asm_srcs, obj_srcs) = if let Some(asm_target) = asm_target {
+        println!("cargo:rustc-cfg=perlasm");
```
briansmith, line 426: It seems like the `cargo:rustc-cfg` directives need to be emitted from `generate_sources_and_preassemble` instead so that the `package` job (mk/package.sh) succeeds.

src/lib.rs
briansmith: This assertion fails when running the `package` job.

briansmith (changes requested):
The `deny` job failure is fixed in PR #2012, so just ignore that.

When building without perlasm, there will be a lot of dead code. We should probably do something like:
`#![cfg_attr(all(no_perlasm, any(target_arch = "aarch64", ....)),allow(dead_code)]` in lib.rs to deal with this, at least temporarily.

We should also add a build-only job to CI for target `aarch64-unknown-none` or similar that ensures the build succeeds.

src/lib.rs
nspin: > This must be named `_PERLASM_CONFIGURED` to pass the `clippy` job.

Done.

nspin: > This assertion fails when running the `package` job.

See below.

build.rs
nspin: I believe that wouldn't cover the case where `RING_PREGENERATE_ASM != 1 && !is_git`. I've moved these `println!()` statements into `main()` itself, which covers all cases.

nspin:
@briansmith I just now rebased again. Are there any more changes you'd like me to make to this PR?

PR #1946, opened 2024-02-16
Address new clippy lint.

briansmith:


briansmith:
The coverage job errors were addressed in PR #1947.

PR #1947, opened 2024-02-16
CI: Use nightly-2024-02-13 for code coverage testing.

briansmith:
More recent versions of the toolchain cause the tests to crash with a segmentation fault when built/run with profiling.


briansmith:
The clippy failure is not due to this change; it will be fixed in PR #1946.

PR #1949, opened 2024-02-16
AEAD: Move input length checks to to individual algorithm implementations

briansmith:
Use `usize`-based math instead of `u64`-based math, addressing the longstanding TODO. Then move the length check into the individual algorithm implementations. Especially in the case of AES-GCM, this makes it clearer that there will be no overflows in the handling of the input lengths. (There weren't any, but this makes that clearer.) See the individual commit messages for details.

### DIFF
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -143,7 +143,7 @@ pub struct Algorithm {
         aad: Aad<&[u8]>,
         in_out: &mut [u8],
         cpu_features: cpu::Features,
-    ) -> Tag,
+    ) -> Result<Tag, error::Unspecified>,
     open: fn(
         key: &KeyInner,
         nonce: Nonce,
@@ -151,12 +151,10 @@ pub struct Algorithm {
         in_out: &mut [u8],
         src: RangeFrom<usize>,
         cpu_features: cpu::Features,
-    ) -> Tag,
+    ) -> Result<Tag, error::Unspecified>,
 
     key_len: usize,
     id: AlgorithmID,
-
-    max_input_len: usize,
 }
 
 const fn max_input_len(block_len: usize, overhead_blocks_per_nonce: usize) -> usize {
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -21,7 +21,10 @@
 //! [AEAD]: https://eprint.iacr.org/2000/025.pdf
 //! [`crypto.cipher.AEAD`]: https://golang.org/pkg/crypto/cipher/#AEAD
 
-use crate::{cpu, error, hkdf, polyfill};
+use crate::{
+    cpu, error, hkdf,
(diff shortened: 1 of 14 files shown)


src/aead.rs
```diff
@@ -140,29 +143,26 @@ pub struct Algorithm {
...
     id: AlgorithmID,
-
-    /// Use `max_input_len!()` to initialize this.
-    // TODO: Make this `usize`.
-    max_input_len: u64,
 }
 
-const fn max_input_len(block_len: usize, overhead_blocks_per_nonce: usize) -> u64 {
+const fn max_input_len(block_len: usize, overhead_blocks_per_nonce: usize) -> usize {
     // Each of our AEADs use a 32-bit block counter so the maximum is the
```
briansmith, line 161: Codecov complains that this is not covered by tests, but it is only used in const contexts, so that is OK.

src/aead/chacha20_poly1305.rs
```diff
@@ -51,12 +55,20 @@ fn chacha20_poly1305_seal(
...
-) -> Tag {
+) -> Result<Tag, error::Unspecified> {
     let chacha20_key = match key {
         aead::KeyInner::ChaCha20Poly1305(key) => key,
         _ => unreachable!(),
     };
 
+    if in_out.len() > MAX_IN_OUT_LEN {
+        return Err(error::Unspecified);
+    }
```
briansmith, line 66: This code path only occurs with a (practically) impossibly-long input, so we can't test it.

src/aead/chacha20_poly1305.rs
```diff
@@ -131,12 +143,24 @@ fn chacha20_poly1305_open(
...
         _ => unreachable!(),
     };
 
+    let unprefixed_len = in_out
+        .len()
+        .checked_sub(src.start)
+        .ok_or(error::Unspecified)?;
+    if unprefixed_len > MAX_IN_OUT_LEN {
+        return Err(error::Unspecified);
+    }
```
briansmith, line 158: This code path only occurs with a (practically) impossibly-long input, so we can't test it.

src/aead/gcm.rs
```diff
@@ -84,16 +89,33 @@ impl Key {
...
-    pub(crate) fn new(key: &Key, aad: Aad<&[u8]>, cpu_features: cpu::Features) -> Self {
+    pub(crate) fn new(
+        key: &Key,
+        aad: Aad<&[u8]>,
+        in_out_len: usize,
+        cpu_features: cpu::Features,
+    ) -> Result<Self, error::Unspecified> {
+        if in_out_len > aes_gcm::MAX_IN_OUT_LEN {
+            return Err(error::Unspecified);
+        }
```
briansmith, line 106: This code path only occurs with a (practically) impossibly-long input, so we can't test it.

PR #1955, opened 2024-02-18
rsa/der: Make it clearer that ASN.1 serialization is panic-free and lossless.

briansmith:
Eliminate `as` conversions and an unreachable. See the individual commits for details.


briansmith:
codecov rightly reports that we're not testing the case where we need to serialize the high byte of the length. To do this, we need to add a test that constructs a 4096-bit RrsaKeyPair.

briansmith:
Replaced by #1959.

PR #1960, opened 2024-02-20
Enable sha512 instruction extension on Apple silicon

BusyJay:
Tested on M1 and time consumed by sha512 is reduced to half.
```
digest::oneshot::sha512/0
                        time:   [93.800 ns 94.427 ns 95.059 ns]
                        change: [-55.464% -55.238% -55.020%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/src/cpu/arm.rs
+++ b/src/cpu/arm.rs
@@ -172,13 +172,19 @@ fn detect_features() -> u32 {
     features
 }
 
+#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+fn detect_features() -> u32 {
+    1 << 6 // sha512
+}
+
 #[cfg(all(
     any(target_arch = "aarch64", target_arch = "arm"),
     not(any(
         target_os = "android",
         target_os = "fuchsia",
         all(target_os = "linux", not(target_env = "uclibc")),
-        target_os = "windows"
+        target_os = "windows",
+        target_os = "macos"
     ))
 ))]
 fn detect_features() -> u32 {
--- a/src/cpu/arm.rs
+++ b/src/cpu/arm.rs
@@ -96,6 +96,9 @@ fn detect_features() -> u32 {
         const HWCAP_PMULL: c_ulong = 1 << 1 + OFFSET;
         const HWCAP_SHA2: c_ulong = 1 << 3 + OFFSET;
 
+        #[cfg(target_arch = "aarch64")]
+        const HWCAP_SHA512: c_ulong = 1 << 21;
+
         if caps & HWCAP_AES == HWCAP_AES {
             features |= AES.mask;
         }
@@ -104,6 +107,11 @@ fn detect_features() -> u32 {
         }
         if caps & HWCAP_SHA2 == HWCAP_SHA2 {
             features |= SHA256.mask;
+        }
+
+        #[cfg(target_arch = "aarch64")]
+        if caps & HWCAP_SHA512 == HWCAP_SHA512 {
+            features |= SHA512.mask;
         }
     }
 
@@ -172,9 +180,36 @@ fn detect_features() -> u32 {
     features
 }
 
-#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+#[cfg(all(target_os = "ios", target_arch = "aarch64"))]
 fn detect_features() -> u32 {
-    1 << 6 // sha512
+    let mut features = ARMCAP_STATIC;
+
+    extern "C" {
+        fn sysctlbyname(
+            name: *const u8,
+            oldp: *mut core::ffi::c_void,
+            oldlenp: *mut crate::c::size_t,
+            newp: *mut core::ffi::c_void,
+            newlen: crate::c::size_t,
+        ) -> core::ffi::c_int;
+    }
+
+    let mut value: core::ffi::c_int = 0;
+    let mut len = core::mem::size_of_val(&value);
+    let rc = unsafe {
+        sysctlbyname(
+            "hw.optional.armv8_2_sha512\0".as_ptr(),
+            &mut value as *mut _ as _,
+            &mut len,
+            core::ptr::null_mut(),
+            0,
+        )
+    };
+    if rc == 0 && len == core::mem::size_of_val(&value) && value != 0 {
+        features |= SHA512.mask;
+    }
+
+    features
 }
 
 #[cfg(all(
@@ -184,7 +219,7 @@ fn detect_features() -> u32 {
         target_os = "fuchsia",
         all(target_os = "linux", not(target_env = "uclibc")),
         target_os = "windows",
-        target_os = "macos"
+        target_os = "ios"
     ))
 ))]
 fn detect_features() -> u32 {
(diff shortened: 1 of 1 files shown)


src/cpu/arm.rs
```diff
@@ -172,13 +172,19 @@ fn detect_features() -> u32 {
     features
 }
 
+#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+fn detect_features() -> u32 {
+    1 << 6 // sha512
+}
```
briansmith, line 178: above in the `features!` macro invocation, we have:
```
    // Keep in sync with `ARMV8_SHA256`.
    "sha2" => SHA256 {
        mask: 1 << 4,
    },
```
Apparently in C and in Rust, the "sha3" Aarch64 CPU feature flag is overloaded for both SHA-3 and SHA-512 CPU features. So you should add an entry:
```
    // Keep in sync with `ARMV8_SHA512`.
    // "sha3" is overloaded for both SHA-3 and SHA512.
    "sha3" => SHA512 {
        mask: 1 << 6,
    },
```

Then you will not need to add a `detect_features()` because it will be statically detected on aarch64-apple-darwin:
```
$ rustc --print=cfg --target aarch64-apple-darwin | grep sha
target_feature="sha2"
target_feature="sha3"
# MSRV
$ rustc +1.61.0 --print=cfg --target aarch64-apple-darwin | grep sha
target_feature="sha2"
target_feature="sha3"
```

Also below, add an assertion analogous to const `_AARCH64_APPLE_TARGETS_EXPECTED_FEATURES` called `_AARCH64_APPLE_DARWIN_TARGETS_EXPECTED_FEATURES` that asserts that the "sha3" feature is enabled statically for aarch64-apple-darwin.

If you want to do the same for aarch64-apple-ios, then we have to use dynamic feature detection because aarch64-apple-ios doesn't enable it statically:
```
$ rustc +1.61.0 --print=cfg --target aarch64-apple-ios | grep sha
target_feature="sha2"
```
The BoringSSL implementation of this at https://boringssl.googlesource.com/boringssl/+/1e15682f1a4bb64c48b84884976a2b5c4201e878%5E%21/ demonstrates how to do that dynamic feature detection for iOS and also for other targets.

briansmith (changes requested):

src/cpu/arm.rs
```diff
@@ -172,13 +180,46 @@ fn detect_features() -> u32 {
...
+    let mut features = ARMCAP_STATIC;
+
+    extern "C" {
+        fn sysctlbyname(
+            name: *const u8,
+            oldp: *mut core::ffi::c_void,
+            oldlenp: *mut crate::c::size_t,
+            newp: *mut core::ffi::c_void,
+            newlen: crate::c::size_t,
+        ) -> core::ffi::c_int;
```
briansmith, line 194: `core::ffi::c_int` is too new for *ring* to use because of the MSRV, so you have to use `c::int`. (`c_void` is fine to use.)

I think we should just use the libc crate to call this, if that works. We have a goal to avoid libc dependencies but it seems like this function is provided by the OS's libc so we can't get around libc anyway, just like with `getauxval` on Linux/Android. And hopefully using the libc crate to call it will handle any/all linking-related aspects.

src/cpu/arm.rs
```diff
@@ -172,13 +180,46 @@ fn detect_features() -> u32 {
...
+    }
+
+    let mut value: core::ffi::c_int = 0;
+    let mut len = core::mem::size_of_val(&value);
+    let rc = unsafe {
+        sysctlbyname(
+            "hw.optional.armv8_2_sha512\0".as_ptr(),
+            &mut value as *mut _ as _,
```
briansmith, line 202: We don't use `as _` style casts but the clippy job in CI doesn't run on this target. Instead you can add this polyfill to `ring::polyfill`:

```
// TODO(MSRV 1.76): Replace with `core::ptr::from_mut`.
#[inline(always)]
pub fn ptr_from_mut<T: ?Sized>(r: &mut T) -> *mut T {
    <*mut T>::from(r)
}
```
and then use `ptr_from_mut(&mut x).cast::<c_void>()`.

src/cpu/arm.rs
```diff
@@ -172,13 +180,46 @@ fn detect_features() -> u32 {
     features
 }
 
+#[cfg(all(target_os = "ios", target_arch = "aarch64"))]
+fn detect_features() -> u32 {
+    let mut features = ARMCAP_STATIC;
+
+    extern "C" {
+        fn sysctlbyname(
+            name: *const u8,
```
briansmith, line 189: This should be `c_char`.

src/cpu/arm.rs
```diff
@@ -172,13 +180,46 @@ fn detect_features() -> u32 {
...
+            newp: *mut core::ffi::c_void,
+            newlen: crate::c::size_t,
+        ) -> core::ffi::c_int;
+    }
+
+    let mut value: core::ffi::c_int = 0;
+    let mut len = core::mem::size_of_val(&value);
+    let rc = unsafe {
+        sysctlbyname(
+            "hw.optional.armv8_2_sha512\0".as_ptr(),
```
briansmith, line 201: You'll need a `.cast<c_char>()` here.

src/cpu/arm.rs
```diff
@@ -172,13 +180,46 @@ fn detect_features() -> u32 {
     features
 }
 
+#[cfg(all(target_os = "ios", target_arch = "aarch64"))]
+fn detect_features() -> u32 {
+    let mut features = ARMCAP_STATIC;
```
briansmith, line 185: Let use this:
```
use core::{mem::size_of_val, ffi:{...}, ...};
use c::{...};
```
to avoid having to qualify everything.

src/cpu/arm.rs
```diff
@@ -172,13 +180,46 @@ fn detect_features() -> u32 {
...
+    let rc = unsafe {
+        sysctlbyname(
+            "hw.optional.armv8_2_sha512\0".as_ptr(),
+            &mut value as *mut _ as _,
+            &mut len,
+            core::ptr::null_mut(),
+            0,
+        )
+    };
+    if rc == 0 && len == core::mem::size_of_val(&value) && value != 0 {
```
briansmith, line 208: It would be better for us to have wrapper for `sysctlbyname` that looks something like this:

```
fn detect_feature(null_terminated_name: &[c_char]) -> bool {
    debug_assert_eq!(null_terminated_name.last(), Some(0));
    let mut value = 0;
    let mut len = core::mem::size_of_val(&value);
    ... yada yada ...
    if rc != 0 {
      return false;
   }
   if len != size_of_val(&value) {
     return false; // unreachable
   }
   value != 0
}
```
This will allow code coverage measurement to show us in more detail if we're triggering the error/impossible condition(s).

briansmith (changes requested):
Thanks for these improvements. This is looking pretty good to me.

I would appreciate it if you could, when you make the requested changes, squash everything into a single commit.

I can see that the code coverage measurement is not going to like this as we don't have CI for non-macos Apple targets. So as a temporary measure, we'll need to add a second commit that filters out `SHA512.mask` from `ARMCAP_STATIC` if the target is an apple target, so that we ignore the statically-indicated presence of the feature and then fall back to the dynamic detection logic on macosx too (all apple targets). This way we can meaningfully say we are covering this code path to some extent.

briansmith:
@BusyJay I have made the tweaks I suggested above in PR #1978. PTAL.

briansmith:
This was merged as part of PR #1978. Thanks for submitting this!

PR #1962, opened 2024-02-22
build.rs can be run from arbitrary directories

illicitonion:
Currently it hard-codes an assumption that the build script is run from the $CARGO_MANIFEST_DIR. While this is true for cargo, and typically is true for other build systems, some toolchains may have other requirements.

1e1021d05bdd41cb02bbc43615c9dd7ecd35f853 started encoding that this assumption may not always hold, this commit just extends that other places the assumption is currently made.

This was tested by adding a call to
`std::env::set_current_dir("/tmp").unwrap()` at the top of the build script's `main` function, and has been used to build both the pregenerated version and the non-pregenerated version of the build script.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -19,6 +19,7 @@
 // another for the concrete logging implementation). Instead we use `eprintln!`
 // to log everything to stderr.
 
+use std::process::Stdio;
 use std::{
     ffi::OsString,
     fs::{self, DirEntry},
@@ -616,6 +617,7 @@ fn run_command_with_args(command_name: &Path, args: &[String]) {
 
 fn run_command(mut cmd: Command) {
     eprintln!("running {:?}", cmd);
+    cmd.stderr(Stdio::inherit());
     let status = cmd.status().unwrap_or_else(|e| {
         panic!("failed to execute [{:?}]: {}", cmd, e);
     });
--- a/build.rs
+++ b/build.rs
@@ -276,19 +276,25 @@ fn read_env_var(name: &'static str) -> Option<OsString> {
 }
 
 fn main() {
+    // Avoid assuming the working directory is the same is the $CARGO_MANIFEST_DIR so that toolchains
+    // which may assume other working directories can still build this code.
+    let c_root_dir = PathBuf::from(
+        std::env::var_os("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR should always be set"),
+    );
+
     const RING_PREGENERATE_ASM: &str = "RING_PREGENERATE_ASM";
     match read_env_var(RING_PREGENERATE_ASM).as_deref() {
         Some(s) if s == "1" => {
-            pregenerate_asm_main();
+            pregenerate_asm_main(&c_root_dir);
         }
-        None => ring_build_rs_main(),
+        None => ring_build_rs_main(&c_root_dir),
         _ => {
             panic!("${} has an invalid value", RING_PREGENERATE_ASM);
         }
     }
 }
 
-fn ring_build_rs_main() {
+fn ring_build_rs_main(c_root_dir: &Path) {
     use std::env;
 
     let out_dir = env::var_os("OUT_DIR").unwrap();
@@ -298,7 +304,7 @@ fn ring_build_rs_main() {
     let os = env::var("CARGO_CFG_TARGET_OS").unwrap();
     let env = env::var("CARGO_CFG_TARGET_ENV").unwrap();
 
-    let is_git = std::fs::metadata(".git").is_ok();
+    let is_git = std::fs::metadata(c_root_dir.join(".git")).is_ok();
 
     // Published builds are always built in release mode.
     let is_debug = is_git && env::var("DEBUG").unwrap() != "false";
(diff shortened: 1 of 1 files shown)


briansmith:
Thanks. The cargo docs at https://doc.rust-lang.org/cargo/reference/build-scripts.html do say "In addition to environment variables, the build script’s current directory is the source directory of the build script’s package." It seems to me that an alternate runner of build.rs (like Bazel) should change the current working directory accordingly before executing the build script.

briansmith:
I am already working on adapting build.rs so that it can be run in a 2nd mode where it will generate a BUILD.gn for GN, which is not too different from Bazel. I would very much like to extend that work so that it could be run on a 3rd mode to generate a Bazel Build file too. Then users of GN and Bazel would not need to run the build.rs at all, as some users of *ring* are not allowed to run build.rs scripts. Is this an alternative you'd be interested in helping with?

build.rs
```diff
@@ -759,7 +759,7 @@ fn get_command(var: &'static str, default: &str) -> OsString {
 // environment variables that affect the build.
 fn emit_rerun_if_changed() {
     for path in &["crypto", "include", "third_party/fiat"] {
-        walk_dir(&PathBuf::from(path), &|entry| {
+        walk_dir(&absolute_path_from_manifest_dir_relative_path(path), &|entry| {
```
briansmith, line 762: I don't think this needs to be modified since any interpreter of the `cargo:rerun-if-changed` output would need to be able to handle relative paths. I would expect that Bazel wouldn't process the `cargo:rerun-if-changed` output anyway.

build.rs
```diff
@@ -1015,3 +1015,9 @@ fn prefix_all_symbols(pp: char, prefix_prefix: &str, prefix: &str) -> String {
 
     out
 }
+
+// Avoid assuming the working directory is the same is the $CARGO_MANIFEST_DIR so that toolchains
+// which may assume other working directories can still build this code.
+fn absolute_path_from_manifest_dir_relative_path<P: AsRef<Path>>(path: P) -> PathBuf {
```
briansmith, line 1021: The docs for `Path::join` say that it handles absolute paths fine so I suggest we drop the "_from_relative_path" part of the name.

I suggest that we avoid having low-level functions in build.rs call this function, because it makes all the logic dependent on everything being compatible with Cargo w.r.t. the environment variables set. Instead, I suggest that at the top level, we compute `c_root_dir: PathBuf` and thread that through the functions that need it; those functions would then `join` paths to `c_root_dir` as necessary. We need to move to a design that is more like this anyway as we're working on generators for BUILD.gn and Bazel BUILD files that reuse some of this logic; they likely won't read any of these environment variables.

build.rs
```diff
@@ -692,9 +692,9 @@ fn perlasm_src_dsts(out_dir: &Path, asm_target: &AsmTarget) -> Vec<(PathBuf, Pat
     {
         // Appease the borrow checker.
         let mut maybe_synthesize = |concrete, synthesized| {
-            let concrete_path = PathBuf::from(concrete);
+            let concrete_path = absolute_path_from_manifest_dir_relative_path(concrete);
             if srcs.contains(&concrete_path) {
-                let synthesized_path = PathBuf::from(synthesized);
+                let synthesized_path = absolute_path_from_manifest_dir_relative_path(synthesized);
```
briansmith, line 697: I think we don't need to make these absolute here. Instead, we should make the paths absolute just before we pass them to cc-rs and/or `perlasm()`.

illicitonion: Makes sense - done

build.rs
```diff
@@ -472,7 +477,7 @@ fn build_c_code(
         })
         .collect::<Vec<_>>();
 
-    let test_srcs = RING_TEST_SRCS.iter().map(PathBuf::from).collect::<Vec<_>>();
+    let test_srcs = RING_TEST_SRCS.iter().map(|path| PathBuf::from(path)).collect::<Vec<_>>();
```
briansmith, line 480: I'm not sure why this change is made. I think it's not needed? The `clippy` job fails because of this.

Also, the `rustfmt` job fails, so I guess you need a `cargo fmt`.

build.rs
```diff
@@ -734,10 +740,10 @@ fn asm_path(out_dir: &Path, src: &Path, asm_target: &AsmTarget) -> PathBuf {
     out_dir.join(dst_filename)
 }
 
-fn perlasm(perl_exe: &Path, src_dst: &[(PathBuf, PathBuf)], asm_target: &AsmTarget) {
+fn perlasm(perl_exe: &Path, src_dst: &[(PathBuf, PathBuf)], asm_target: &AsmTarget, c_root_dir: &Path) {
     for (src, dst) in src_dst {
         let mut args = vec![
-            src.to_string_lossy().into_owned(),
+            c_root_dir.join(src).to_string_lossy().into_owned(),
```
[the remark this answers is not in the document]
illicitonion: Yeah, looks like the forward slash replacement was needed

build.rs
```diff
@@ -585,18 +605,19 @@ fn nasm(file: &Path, arch: &str, include_dir: &Path, out_dir: &Path) {
         .arg(include_dir)
         .arg("-Xgnu")
         .arg("-gcv8")
-        .arg(file);
+        .arg(c_root_dir.join(file));
```
briansmith, line 608: There are still two relative paths here, in `Command::new` and the `include/`, so I think this probably won't work if the current working directory isn't what Cargo would set, unless you are building from a packaged release of ring that skips the `generate_sources_and_preassemble` step. 

If you actually are running nasm in Bazel through build.rs, I am guessing you need a solution for the `Command::new` independently of the current working directory issue. Probably we should introduce a cargo cfg directive for providing the path to nasm.

briansmith (approved):

PR #1978, opened 2024-03-01
Enable SHA512 acceleration on Aarch4 Android, Linux, and Apple targets.

briansmith:


briansmith:
@BusyJay I added aarch64-apple-darwin code coverage measurement to CI in PR #1980 and #1979, and then rebased this on top of that. That way we can measure the code coverage of your changes (as amended by me) accurately.

src/cpu/arm.rs
```diff
@@ -172,13 +180,59 @@ fn detect_features() -> u32 {
...
+        let mut len = mem::size_of_val(&value);
+        let value_ptr = polyfill::ptr::from_mut(&mut value).cast::<c_void>();
+        let rc = unsafe { libc::sysctlbyname(name, value_ptr, &mut len, core::ptr::null_mut(), 0) };
+        // All the conditions are separated so we can observe them in code coverage.
+        if rc != 0 {
+            return false;
+        }
+        debug_assert_eq!(len, mem::size_of_val(&value));
+        if len != mem::size_of_val(&value) {
+            return false;
```
briansmith, line 207: Hmm...why is codecov telling me that this line is being hit?

src/cpu/arm.rs
```diff
@@ -172,13 +180,59 @@ fn detect_features() -> u32 {
...
+        use core::mem;
+        use libc::{c_int, c_void};
+
+        let mut value: c_int = 0;
+        let mut len = mem::size_of_val(&value);
+        let value_ptr = polyfill::ptr::from_mut(&mut value).cast::<c_void>();
+        let rc = unsafe { libc::sysctlbyname(name, value_ptr, &mut len, core::ptr::null_mut(), 0) };
+        // All the conditions are separated so we can observe them in code coverage.
+        if rc != 0 {
+            return false;
```
briansmith, line 203: I don't understand why codecov is saying this is being hit.

src/cpu/arm.rs
```diff
@@ -50,61 +50,62 @@ mod abi_assumptions {
...
 fn detect_features() -> u32 {
-    use libc::c_ulong;
-
-    // XXX: The `libc` crate doesn't provide `libc::getauxval` consistently
-    // across all Android/Linux targets, e.g. musl.
-    extern "C" {
-        fn getauxval(type_: c_ulong) -> c_ulong;
+    cfg_if::cfg_if! {
+        if #[cfg(target_arch = "aarch64")] {
+            use libc::{getauxval,AT_HWCAP, HWCAP_AES, HWCAP_PMULL, HWCAP_SHA2, HWCAP_SHA512};
```
briansmith, line 55: @BusyJay When I rebased your change on top of PR #1982, this was the only non-trivial merge conflict. Instead of using your definition of `HWCAP_SHA512`, I used the one from libc.

Cargo.toml
```diff
@@ -162,7 +162,7 @@ untrusted = { version = "0.9" }
 [target.'cfg(any(target_arch = "aarch64", target_arch = "arm", target_arch = "x86",target_arch = "x86_64"))'.dependencies]
 spin = { version = "0.9.8", default-features = false, features = ["once"] }
 
-[target.'cfg(all(any(target_os = "android", target_os = "linux"), any(target_arch = "aarch64", target_arch = "arm")))'.dependencies]
+[target.'cfg(all(any(target_os = "android", target_os = "linux", any(target_os = "ios", target_os = "macos", target_os = "tvos")), any(target_arch = "aarch64", target_arch = "arm")))'.dependencies]
```
BusyJay, line 165: Why repeat `any` inside `any`?

BusyJay (approved):

Cargo.toml
briansmith: Mostly just to keep the apple stuff together. It is kind of frustrating that these are all very similar but need to be spelled out separately everywhere, and this way it is a little easier to (re-)review that it is done correctly whenever we extend it, e.g. to add watchos.

PR #1981, opened 2024-03-02
Split 32-bit and 64-bit ARM feature detection on Linux.

briansmith:


briansmith:
This cannot be done until we have code coverage measurement for 32-bit ARM.

PR #1985, opened 2024-03-03
cpu: Clarify `cfg(...)` logic by removing redundant checks.

briansmith:
Make it clearer when reading `arm.rs` and `intel.rs` tthat these modules are only compiled for the relevant target architectures. Remove redundant conditional logic that was confusing regarding this.

### DIFF
--- a/src/cpu/arm.rs
+++ b/src/cpu/arm.rs
@@ -12,12 +12,8 @@
 // OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
-#![cfg_attr(
-    not(any(target_arch = "aarch64", target_arch = "arm")),
-    allow(dead_code)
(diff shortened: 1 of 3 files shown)


src/cpu/arm.rs
```diff
@@ -286,24 +281,8 @@ impl Feature {
...
-        {
-            // SAFETY: See `OPENSSL_armcap_P`'s safety documentation.
-            if self.mask == self.mask & unsafe { OPENSSL_armcap_P } {
-                return true;
-            }
-        }
-
-        false
+        // SAFETY: See `OPENSSL_armcap_P`'s safety documentation.
+        self.mask == self.mask & unsafe { OPENSSL_armcap_P }
```
briansmith, line 285: This function isn't currently used by aarch64 targets for anything other than NEON; all the other checks of `OPENSSL_armcap_P` are done in assembly code for AArch64.

This line covered by 32-bit ARM targets in the `test` matrix but it isn't covered in the `coverage` matrix because 32-bit ARM targets don't support source-based code coverage measurement yet.

PR #1986, opened 2024-03-04
aarch64-apple-*: Improve `unsafe` usage.

briansmith:
Make `detect_feature` a safe function by haivng it do a (clearly superfluous) check on the well-formedness of its input.


briansmith:
FYI @BusyJay. In the midst of getting it to work with MSRV 1.61, I got the priorities mixed up regarding `unsafe`. This should work exactly the same

PR #1987, opened 2024-03-08
Changed toolchain to stable for coverage

### LINKED ISSUE #1902: CI: Use stable Rust for source-based code coverage measurement
Rust 1.76 is scheduled to be released on Feburary 8th, 2024. Assuming it continues to include the change in https://github.com/rust-lang/rust/pull/118100, we should switch all the `coverage` jobs in ci.yml to use Rust Stable instead of Rust Nightly.

Raghav-Bell:
It fixes #1902

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -542,7 +542,7 @@ jobs:
 
         # Coverage collection is Nightly-only
         rust_channel:
-          - nightly
+         - stable
 
         # TODO: targets
         include:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -542,7 +542,7 @@ jobs:
 
         # Coverage collection is Nightly-only
         rust_channel:
-         - beta
+          - stable
 
         # TODO: targets
         include:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -542,7 +542,7 @@ jobs:
 
         # Coverage collection is Nightly-only
         rust_channel:
-         - stable
+         - beta
 
         # TODO: targets
         include:


briansmith:
From the CI run:
> ```
> qemu: uncaught target signal 11 (Segmentation fault) - core dumped
> /home/runner/work/ring/ring/mk/runner: line 21:  6499 Segmentation fault      (core dumped) $*
> error: test failed, to rerun pass `-p ring --lib`
> ```
This kind of crash seems to happen when the clang version doesn't match the LLVM version that Rust uses. It seems like we upgraded clang in CI to what Nightly Rust uses. In order to move forward with this change, we may either need to downgrade to the earlier version of clang that matches stable Rust's LLVM version, or wait until the next stable Rust is released.

You might try using channel `beta` to see if the beta channel is using the new version of LLVM to get an idea of how long we'd need to wait.

Raghav-Bell:
image

@briansmith You are right
`stable` needs [LLVM $`\geq`$ 16](https://github.com/rust-lang/rust/pull/117947/)
`beta` & `nightly` are compatible with [LLVM ==18](https://github.com/rust-lang/rust/pull/120055).
I will push it to `beta` for now, later we can move it to stable. Thanks

briansmith:
@Raghav-Bell Do you want to try again now?

briansmith:
OK, this seems to be working now on stable. Could you please do the following?:
1. Rebase this on top of the latest main. Your PR branch is 64 commits behind the main branch, so it is hard to compare its effect on code coverage measurement as it is. If you rebase, then we should get a useful report from codecov.io.
2. Squash the three commits into one.
3. Improve the commit message. Something like this:

> CI: Use stable Rust toolchain for code coverage.
>
> Use the stable Rust toolchain for code coverage in CI. This should make coverage measurements more stable day-to-day and also make it easier for people to replicate coverage measurements on locally.

briansmith:
Since PR #2056 was merged, we'll need to wait for Rust 1.80 (IIUC) to become stable, or else add a wokaround that forces the use of nightly for the armv7 target.

briansmith:
Thanks for the PR. I definitely did want to switch to stable Rust for code coverage. This would have been a good improvement. Unfortunately, the branch coverage option hasn't been stablized, IIUC. So I'm going to close this. Thanks again!

PR #1992, opened 2024-03-13
digest internals: Refactor to clarify safety

briansmith:
Refactor `ring::digest` to minimize the use of `unsafe`.


briansmith:
The codecov.io checks are calse alarms. Two of them are in `const fn` that are only to be used at compile time, and the others are in `unreachable!()` calls.

PR #1993, opened 2024-03-13
style: address shellcheck warnings

reubenmiller:
Addressing the warnings raised by [shellcheck](https://www.shellcheck.net/).

Using shellcheck (slightly) improves the style consistency and reduces the chance of subtle errors.

There are still two shellcheck warnings for missing shebangs (in `mk/package.sh` and `mk/publish.sh`, however looking at the scripts it seems these might be for windows environments running MinGW, where shebangs might be incompatible.

### DIFF
--- a/mk/cargo.sh
+++ b/mk/cargo.sh
@@ -43,7 +43,7 @@ if [ -n "${ANDROID_NDK_ROOT-}" ]; then
   android_tools=${ANDROID_NDK_ROOT}/toolchains/llvm/prebuilt/linux-x86_64/bin
 fi
 
-for arg in $*; do
+for arg in "$@"; do
   case $arg in
     --target=*)
       target=${arg#*=}
--- a/mk/cargo.sh
+++ b/mk/cargo.sh
@@ -255,9 +255,9 @@ if [ -n "${RING_COVERAGE-}" ]; then
 
   while read -r executable; do
     basename=$(basename "$executable")
-    "${llvm_root}"/llvm-profdata merge -sparse "$coverage_dir/$basename.profraw" -o "$coverage_dir/$basename.profdata"
+    "${llvm_root}/llvm-profdata" merge -sparse "$coverage_dir/$basename.profraw" -o "$coverage_dir/$basename.profdata"
     mkdir -p "$coverage_dir"/reports
-    "${llvm_root}"/llvm-cov export \
+    "${llvm_root}/llvm-cov" export \
       --instr-profile "$coverage_dir/$basename.profdata" \
       --format lcov \
       "$executable" \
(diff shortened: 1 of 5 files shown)


mk/cargo.sh
```diff
@@ -251,16 +251,16 @@ cargo "$@"
 if [ -n "${RING_COVERAGE-}" ]; then
   # Keep in sync with check-symbol-prefixes.sh.
   # Use the host target-libdir, not the target target-libdir.
-  llvm_root="$(rustc +${toolchain} --print target-libdir)/../bin"
+  llvm_root="$(rustc +"${toolchain}" --print target-libdir)/../bin"
```
briansmith, line 254: This looks weird. I wonder if there's a better way of doing it. Maybe removing the "outermost"-looking double quotes, similar to the `target_upper=...` line above?

mk/cargo.sh
```diff
@@ -251,16 +251,16 @@ cargo "$@"
...
 
-  while read executable; do
+  while read -r executable; do
     basename=$(basename "$executable")
-    ${llvm_root}/llvm-profdata merge -sparse "$coverage_dir/$basename.profraw" -o "$coverage_dir/$basename.profdata"
+    "${llvm_root}"/llvm-profdata merge -sparse "$coverage_dir/$basename.profraw" -o "$coverage_dir/$basename.profdata"
```
briansmith, line 258: It looks weird to be inconsistent with having the double quotes only around the variable usage in `"${llvm_root}"/llvm-profdata` but then less tight in `"$coverage_dir/$basename.profraw"` and `"$coverage_dir/$basename.profdata"`. Maybe I'm missing something, but some more consistency would be nice.

briansmith (commented):
Thanks. It would be great if somebody who is more familiar with shell scripting could review this. There are some things that seems strange to me. I'll point out a couple.

mk/cargo.sh
reubenmiller: Having outher quotes is fine because $() starings a new context where the quoting is processed independently, so no escaping is required. However for consistency, I'll remove the outer quotes as it has not effect on the variable assignment.

reubenmiller: Resolved by 9a18e6fdb9312615608f6be0f366f25e8b4367ec

mk/cargo.sh
reubenmiller: Resolved by 31dc1e6eb4badf6ff3f09310d729d3784fc60f39.

Though I was thinking about normalizing the usage of `${example}` vs `$example` using the following rules:
* Use `$example` when the variable is not being without additional characters or variable modifiers (e.g. `${example:-default_value}`
* Use `${example}` otherwise, e.g. `+${example}`, `my/path/${example}` to make it more (visually) obvious that a string is being built from mixing strings with variables.

@briansmith Are you ok with this?

briansmith:
Thanks for the PR. Probably your changes make a really big improvement, but I'm so ignorant about shell scripting that I was scared of overlooking something so I kept putting it off. I really appreciate the effort. Now I'm not accepting PRs so I'm going to close this.

PR #2001, opened 2024-03-20
gcm: add missing #[allow(clippy::assertions_on_constants)] for aarch64

goffrie:
`cargo clippy` was complaining when targeting aarch64.


briansmith:
Thank you!

PR #2003, opened 2024-03-22
AES-GCM internals: Make the Block type an alias for [u8; BLOCK_LEN]

briansmith:


src/aead/shift.rs
```diff
@@ -28,14 +29,14 @@ where
             transform(input)
         };
         let output = <&mut [u8; BLOCK_LEN]>::try_from(&mut in_out[i..][..BLOCK_LEN]).unwrap();
-        *output = *block.as_ref();
+        *output = block;
```
briansmith, line 32: Codecov is right to complain about the lack of test coverage for this function in the `coverage` job in CI. That's something to be addressed separately.

PR #2010, opened 2024-03-31
Add __XTENSA__ to target.h

coder0xff:


coder0xff:
Hi. Would you trigger the workflow please?

coder0xff:
@briansmith Is it safe to assume that the ci / deny failure is unrelated?

briansmith:
Please:
1. Make a mk/install-build-tools-unsupported.sh that is in the same form of mk/install-build-tools.sh, which adds support for installing the xtensa toolchain on an Ubuntu system, starting from the state where qemu-xtensa is not installed.
2. Make a mk/cargo-unsupported.sh that mirrors mk/cargo.sh which uses the xtensa toolchain to run cargo commands in qemu-xtensa.
3. Update .github/workflows/ci.yml to add the xtensa target to the test matrix.

briansmith:
Instead of continuing to grow this allowlist, I'd rather eliminate it. Please help with implementing the strategy in https://github.com/briansmith/ring/pull/2042#issuecomment-2120898907 if you can.

Issue #1886, opened 2024-01-09
Vendor rust-aes/soft implementation


briansmith:
The code is at https://github.com/RustCrypto/block-ciphers/tree/master/aes/src/soft.

* The code should be placed in `src/rust_crypto/aes/soft/` in the *ring* sources.
* The first commit should add the two source files, unmodified. They will not successfully build. Include the full commit hash from the upstream repository in the commit message.
* Remove `hazmat` and other code that will never be needed. Retain the decryption code; even though it will be dead code; the future AES-CBC code will use it.
* Add src/rust_crypto/aes/soft/block.rs to define the `Block` type in a way that does not use the `cipher` crate.
* Replace types `FixsliceBlocks` and `BatchBlocks` with types that do not use the `cipher` crate.
* Adapt `ring::aead::aes` to use the newly-vendored code instead of aes_nohw.c. Remove aes_nohw.c, aes/internal.h, and all references to the files from build.rs and Cargo.toml.

briansmith:
The change that implements this should be done in a single PR. I will use `diff` against the latest commit of the `aes` crate sources on GitHub to verify that the first commit is exactly identical to what's in the upstream source. Then I will review each commit in the PR as though it were its own PR. So basically the submitter needs to be prepared to do a lot of rebasing through the review process if there are many rounds of review.

PR #2024, opened 2024-04-10
vendor rust-crypto aes/soft implementation

rohhamh:
I've tried following the instructions at https://github.com/briansmith/ring/issues/1886#issue-2072949071

I ran the tests for fixslice64 and they all seem to pass.
However I'm very reluctant about some of the changes made here,
like increasing rd_key's size to 240 and having u32-u64 conversion functions
or using self.inner.rounds to differentiate between the target functions
or sending less than four actual blocks for encryption to the vendored code.

------

This should fail for fixslice32, I would like it if I could have your opinion on how you think it is best to handle fixslice32 as it takes u32 parameters.

The code in ctr32_encrypt_within is basically my attempt at replicating what aes_nohw_ctr32_encrypt_blocks in aes_nohw.c does, I imagine it's accurate as the tests did pass.

There's also the [contribution guide](https://github.com/briansmith/ring/tree/681402342bfe69f72238e50b75deb5edeff741af#contributing).
I'm not sure if the statement should be added to every commit message or one would suffice,
or how the license would apply to the vendored code.

And finally, this is basically the first time I've done anything in rust or cryptography, so I imagine there would be many ways to improve the code. I'd be very happy to know about them.


src/rust_crypto/aes/soft/fixslice32.rs
```diff
@@ -0,0 +1,1371 @@
...
+            add_round_constant_bit(&mut rkeys[rk_off..(rk_off + 8)], rcon);
+        } else {
+            add_round_constant_bit(&mut rkeys[rk_off..(rk_off + 8)], rcon - 8);
+            add_round_constant_bit(&mut rkeys[rk_off..(rk_off + 8)], rcon - 7);
+            add_round_constant_bit(&mut rkeys[rk_off..(rk_off + 8)], rcon - 5);
+            add_round_constant_bit(&mut rkeys[rk_off..(rk_off + 8)], rcon - 4);
+        }
+
+        xor_columns(&mut rkeys, rk_off, 8, ror_distance(1, 3));
+    }
```
briansmith, line 62: It looks like maybe we could split this into two functions so that the above logic initializes an `AES_KEY` structure as it exists today, which would be stored in the `aes::Key`. Then whenever we do an encryption/decryption operation, we would first execute the below logic to expand the round keys into a temporary buffer that contains the fixedslice representation. Of course, this would add some overhead to each encryption batch. Then we'd need to measure to see if it would still be faster than the non-SIMD bitsliced implementation.

WDYT?

src/rust_crypto/aes/soft/fixslice32.rs
```diff
@@ -0,0 +1,1371 @@
...
+    for i in 1..11 {
+        sub_bytes_nots(&mut rkeys[(i * 8)..(i * 8 + 8)]);
+    }
+
+    rkeys
+}
+
+/// Fully bitsliced AES-192 key schedule to match the fully-fixsliced representation.
+#[allow(dead_code)]
+pub(crate) fn aes192_key_schedule(key: &[u8; 24]) -> FixsliceKeys192 {
```
briansmith, line 91: I think we can add a commit to just delete the aes192-specific code.

src/rust_crypto/aes/soft/fixslice32.rs
```diff
@@ -0,0 +1,1371 @@
...
+    }
+
+    // Adjust to match fixslicing format
+    #[cfg(aes_compact)]
+    {
+        for i in (8..88).step_by(16) {
+            inv_shift_rows_1(&mut rkeys[i..(i + 8)]);
+        }
+    }
+    #[cfg(not(aes_compact))]
```
briansmith, line 71: And/or we should consider deleting all the `not(aes_compact)` code paths and reduce the key size accordingly.

src/rust_crypto/aes/soft/fixslice32.rs
```diff
@@ -0,0 +1,1371 @@
...
+
+    add_round_key(&mut state, &rkeys[80..]);
+    inv_sub_bytes(&mut state);
+
+    #[cfg(not(aes_compact))]
+    {
+        inv_shift_rows_2(&mut state);
+    }
+
+    let mut rk_off = 72;
```
briansmith, line 260: This makes me think that many of the round keys in `rkeys` are only needed for decryption and not encryption? If so, we could reduce the size of `rkeys` by half by just supporting only encryption?

briansmith (commented):
Thank you very much for doing this. I am sorry to admit that I overlooked that the Rust Crypto key representation was so much larger when I originally suggested this plan. I think it's important for us to find a way to minimize (ideally eliminate) the size overhead, at least for targets that are likely to have 128-bit SIMD and/or hardware instructions available.

src/rust_crypto/aes/soft/fixslice32.rs
rohhamh: Is it so we won't change the AES_KEY structure of today's and/or have smaller changes to the current codebase?

I'm afraid I'm not well-versed enough in the subject to have a definitive answer, normally, I'd say that if we can avoid the overhead and have the expansion happen only once then we could do just that.
But I imagine it wouldn't matter much if the difference in performance is negligible when compared to what rust-crypto provides at best (and is still faster than what's currently available).

Do you think this could be the best approach?

src/rust_crypto/aes/soft/fixslice32.rs
rohhamh: We might have to check more accurately, but I think that's not the case as the following is part of the code of encryption
```rust
    let mut rk_off = 8;
    loop {
        sub_bytes(&mut state);
        mix_columns_1(&mut state);
        add_round_key(&mut state, &rkeys[rk_off..(rk_off + 8)]);
        rk_off += 8;

        #[cfg(aes_compact)]
        {
            shift_rows_2(&mut state);
        }

        if rk_off == 80 {
            break;
        }
```

rohhamh:
Is the PR not relevant anymore? @briansmith

briansmith:
@rohhamh Thank you for working on this. In https://github.com/RustCrypto/block-ciphers/issues/475 and https://github.com/RustCrypto/block-ciphers/issues/191 I attempted to transfer the knowledge from what we learned here to the RustCrypto team.

rohhamh:
I wish you good luck during your break
@briansmith

PR #2032, opened 2024-04-16
Add Apple visionOS support

QuentinPerez:
## Motivation

Add the support of Apple VisionOS  [121419](https://github.com/rust-lang/rust/pull/121419)

## Solution
Add target_os = "visionos"

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -161,7 +161,7 @@ untrusted = { version = "0.9" }
 [target.'cfg(any(target_arch = "aarch64", target_arch = "arm", target_arch = "x86",target_arch = "x86_64"))'.dependencies]
 spin = { version = "0.9.8", default-features = false, features = ["once"] }
 
-[target.'cfg(all(any(target_os = "android", target_os = "linux", any(target_os = "ios", target_os = "macos", target_os = "tvos")), any(target_arch = "aarch64", target_arch = "arm")))'.dependencies]
+[target.'cfg(all(any(target_os = "android", target_os = "linux", target_vendor = "apple"), any(target_arch = "aarch64", target_arch = "arm")))'.dependencies]
 libc = { version = "0.2.148", default-features = false }
 
 [target.'cfg(all(target_arch = "aarch64", target_os = "windows"))'.dependencies]
@@ -174,7 +174,7 @@ wasm-bindgen-test = { version = "0.3.37", default-features = false }
 libc = { version = "0.2.148", default-features = false }
 
 [build-dependencies]
-cc = { version = "1.0.83", default-features = false }
+cc = { version = "1.0.94", default-features = false }
 
 [features]
 # These features are documented in the top-level module's documentation.
--- a/build.rs
+++ b/build.rs
@@ -254,7 +254,7 @@ const NASM: &str = "nasm";
 
 /// Operating systems that have the same ABI as macOS on every architecture
 /// mentioned in `ASM_TARGETS`.
-const MACOS_ABI: &[&str] = &["ios", MACOS, "tvos"];
+const MACOS_ABI: &[&str] = &["ios", MACOS, "tvos", "visionos"];
 
 const MACOS: &str = "macos";
 const WINDOWS: &str = "windows";
--- a/src/cpu/arm.rs
+++ b/src/cpu/arm.rs
@@ -41,8 +41,7 @@ mod abi_assumptions {
 // detection.
 
 cfg_if::cfg_if! {
-    if #[cfg(all(target_arch = "aarch64",
-                 any(target_os = "ios", target_os = "macos", target_os = "tvos")))] {
+    if #[cfg(all(target_arch = "aarch64", target_vendor = "apple"))] {
         mod darwin;
         use darwin as detect;
     } else if #[cfg(all(target_arch = "aarch64", target_os = "fuchsia"))] {
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -131,16 +131,14 @@ impl crate::sealed::Sealed for SystemRandom {}
     target_os = "haiku",
     target_os = "hermit",
     target_os = "illumos",
-    target_os = "ios",
     target_os = "linux",
-    target_os = "macos",
     target_os = "netbsd",
     target_os = "openbsd",
     target_os = "redox",
     target_os = "solaris",
-    target_os = "tvos",
     target_os = "vita",
     target_os = "windows",
+    target_vendor = "apple",
     all(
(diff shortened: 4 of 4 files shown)


Cargo.toml
```diff
@@ -162,7 +162,7 @@ untrusted = { version = "0.9" }
 [target.'cfg(any(target_arch = "aarch64", target_arch = "arm", target_arch = "x86",target_arch = "x86_64"))'.dependencies]
 spin = { version = "0.9.8", default-features = false, features = ["once"] }
 
-[target.'cfg(all(any(target_os = "android", target_os = "linux", any(target_os = "ios", target_os = "macos", target_os = "tvos")), any(target_arch = "aarch64", target_arch = "arm")))'.dependencies]
+[target.'cfg(all(any(target_os = "android", target_os = "linux", any(target_os = "ios", target_os = "visionos", target_os = "macos", target_os = "tvos")), any(target_arch = "aarch64", target_arch = "arm")))'.dependencies]
```
briansmith, line 165: Please alphabetize this.

build.rs
```diff
@@ -197,6 +197,11 @@ const ASM_TARGETS: &[AsmTarget] = &[
         arch: AARCH64,
         perlasm_format: "ios64",
     },
+    AsmTarget {
+        oss: MACOS_ABI,
+        arch: AARCH64,
+        perlasm_format: "visionos64",
```
briansmith, line 203: How did you decide on this? Does it not work when we use `perlasm_format: "ios64"`?

In any case, it doesn't make sense for `oss: MACOS_ABI` to be here AND for there to be a different perlasm format. Either we should have an `AsmTarget` that has `oss: "visionos"` or we should add `"visionos"` to MACOS_ABI so that the above AsmTarget is used. RIght?

crypto/perlasm/arm-xlate.pl
```diff
@@ -30,7 +30,7 @@
     else			{ ""; }
 };
 my $hidden = sub {
-    if ($flavour =~ /ios/)	{ ".private_extern\t".join(',',@_); }
+    if ($flavour =~ /ios|visionos/)	{ ".private_extern\t".join(',',@_); }
```
briansmith, line 33: If we really can't use `perlasm_format: "ios64"` then we need to document near the top of this script why visionos needs to be supported separately, but tvos and other variants don't. We should also try to upstream these changes to BoringSSL, if they are really needed, i.e. if we can't just use `"ios64"`.

src/cpu/arm.rs
```diff
@@ -170,7 +170,12 @@ fn detect_features() -> u32 {
 
 #[cfg(all(
     target_arch = "aarch64",
-    any(target_os = "ios", target_os = "macos", target_os = "tvos")
+    any(
+        target_os = "ios",
+        target_os = "visionos",
```
briansmith, line 175: Please alphabetize this and other similar lists.

src/ec/curve25519/x25519.rs
```diff
@@ -63,7 +63,10 @@ fn x25519_public_from_private(
     let private_key: &[u8; SCALAR_LEN] = private_key.bytes_less_safe().try_into()?;
     let private_key = ops::MaskedScalar::from_bytes_masked(*private_key);
 
-    #[cfg(all(not(target_os = "ios"), target_arch = "arm"))]
+    #[cfg(all(
+        not(any(target_os = "ios", target_os = "visionos")),
+        target_arch = "arm"
```
briansmith, line 68: I don't think we need this change. nobody is going to be trying to build for 32-bit ARM VisionOS, right?

briansmith (changes requested):
Thanks for this. Have you tested this on visionOS hardware and in the simulator?

QuentinPerez:
Thank you for the review. It's clearer now how the ring works. The change is much simpler! And yes, I tested it; the app is even on the App Store.

sax:
Hello! I'm curious if this same change would be required to add support for `watchos`.

briansmith:
@QuentinPerez could you please rebase this on top of the latest changes on the main branch. The recent changes should result in it being even simpler now.

briansmith:
> Hello! I'm curious if this same change would be required to add support for `watchos`.

There are several watchos targets, some of which are easier to support than others. PR #1914 started to add support for watchOS, but it got closed before it was finished. I suggest you pick where that left off.

Cargo.toml
```diff
@@ -175,7 +175,7 @@ wasm-bindgen-test = { version = "0.3.37", default-features = false }
 libc = { version = "0.2.148", default-features = false }
 
 [build-dependencies]
-cc = { version = "1.0.83", default-features = false }
+cc = { version = "1.0.94", default-features = false }
```
briansmith, line 178: What is the oldest cc-rs that is known to work for this? Because of issues like https://github.com/rust-lang/cc-rs/issues/984 I'm weary of forcing users to update to too-new version of cc-rs right now, if we can avoid it.

sax: VisionOS support was added in [1.0.93](https://github.com/rust-lang/cc-rs/releases/tag/1.0.93), but [1.0.94](https://github.com/rust-lang/cc-rs/releases/tag/1.0.94) includes a fix to how it integrates with `xcrun` that may be necessary. I don't see anything in the changelog after 1.0.94 related to visionos.

briansmith: Thanks. I think we need to go through the release notes of cc-rs 1.0.94 and higher, and recent issues in that crate, to verify that requiring such a new release is not going to be problematic, given recent instability regarding Apple targets in that crate.

eugenehp: so what would be the best course of action here? seems like https://github.com/rust-lang/cc-rs/issues/984 is still a blocker.

original PR was merged on April 12, and release followed on April 13
https://github.com/rust-lang/cc-rs/releases/tag/1.0.93

briansmith: Maybe we could leave the cc-rs dependency as-is but add a comment above it saying cc-rs 1.0.94 or later is required for visionos support, but we're waiting to increase our dependency until https://github.com/rust-lang/cc-rs/issues/984 and maybe other issues are addressed, and also add a note to that same effect to BUILDING.md. Then we can merge this.

eugenehp: yeah, a conditional dependency in `Cargo.toml` sounds like a feasible option.

@QuentinPerez wanna do it?

briansmith:
Closing in favor of #2112 and #2113. Please see those PRs for important details. I would appreciate testing of them and feedback, especially related to performance and code size.

Issue #1858, opened 2023-12-08
Illegal instruction when running on Raspberry Pi when building with `RUSTFLAGS` containing `-C target_cpu=cortex-a72`


jupyterkat:
NOTE: this is compiled with `-target-cpu=native` and `--release`, and target triple `stable-aarch64-unknown-linux-gnu`. But it fails with default `target-cpu` anyways

backtrace:
```
Thread 1 "musicalcat" received signal SIGILL, Illegal instruction.
ring_core_0_17_7_aes_hw_set_encrypt_key ()
    at /home/katbot/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ring-0.17.7/pregenerated/aesv8-armx-linux64.S:122
122             aese    v6.16b,v0.16b
(gdb) bt full
#0  ring_core_0_17_7_aes_hw_set_encrypt_key ()
    at /home/katbot/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ring-0.17.7/pregenerated/aesv8-armx-linux64.S:122
No locals.
#1  0x0000005555f71bec in ring::aead::aes_gcm::init_256 ()
No symbol table info available.
#2  0x0000005555f6f72c in <ring::aead::unbound_key::UnboundKey as core::convert::From<ring::hkdf::Okm<&ring::aead::Algorithm>>>::from ()
No symbol table info available.
#3  0x0000005555f4e71c in rustls::tls13::key_schedule::KeySchedule::derive_decrypter::he91ce5cdc238f2b2 ()
No symbol table info available.
#4  0x0000005555f4db68 in rustls::tls13::key_schedule::KeyScheduleHandshakeStart::derive_client_handshake_secrets ()
No symbol table info available.
#5  0x0000005555f56a08 in rustls::client::tls13::handle_server_hello ()
No symbol table info available.
#6  0x0000005555f404a0 in <rustls::client::hs::ExpectServerHello as rustls::common_state::State<rustls::client::client_conn::ClientConnectionData>>::handle ()
No symbol table info available.
#7  0x0000005555f409f0 in <rustls::client::hs::ExpectServerHelloOrHelloRetryRequest as rustls::common_state::State<rustls::client::client_conn::ClientConnectionData>>::handle ()
No symbol table info available.
#8  0x0000005555eb7950 in rustls::common_state::CommonState::process_main_protocol ()
```
lscpu:
```
Architecture:                    aarch64
CPU op-mode(s):                  32-bit, 64-bit
[...]

PR #2046, opened 2024-05-03
Implement detection if CPU doesn't support hardware AES crypto.

Arne91:
Pull request as solution of issue #1858 Illegal instruction on arm-a72

### DIFF
--- a/src/aead/aes.rs
+++ b/src/aead/aes.rs
@@ -439,6 +439,68 @@ pub enum Implementation {
     NOHW,
 }
 
+#[cfg(all(target_arch = "aarch64", target_os = "linux"))]
+fn check_aes_support(buffer: &[u8], len: isize) -> bool {
+    let flags_str = b"flags\t\t: ";
+    let features_str = b"Features\t: ";
+    let aes_str = b"aes";
+
+    let data = &buffer[..len as usize];
+
+    if let Some(pos) = data
+        .windows(flags_str.len())
+        .position(|window| window == flags_str)
+    {
+        let flags_section = &data[pos + flags_str.len()..];
+        if let Some(end) = flags_section.iter().position(|&b| b == b'\n') {
+            return flags_section[..end]
+                .windows(aes_str.len())
+                .any(|window| window == aes_str);
+        }
+    } else if let Some(pos) = data
+        .windows(features_str.len())
+        .position(|window| window == features_str)
+    {
+        let features_section = &data[pos + features_str.len()..];
+        if let Some(end) = features_section.iter().position(|&b| b == b'\n') {
+            return features_section[..end]
+                .windows(aes_str.len())
+                .any(|window| window == aes_str);
+        }
+    }
+
+    false
+}
+
+/// check "/proc/cpuinfo" for AES flag
+#[cfg(all(target_arch = "aarch64", target_os = "linux"))]
+pub fn aes_support() -> Result<bool, String> {
+    let path = "/proc/cpuinfo\0";
+    let fd = unsafe { libc::open(path.as_ptr() as *const _, libc::O_RDONLY) };
+
+    if fd < 0 {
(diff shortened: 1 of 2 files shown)


Arne91:
Introduces a new check to ensure that the target CPU actually supports AES hardware cryptography. Previously, the Rust compiler might assume that all aarch64 CPUs support AES hardware acceleration, which is not always the case. For instance, devices like the Raspberry Pi 4, despite being aarch64, do not have this capability.

The lack of an explicit check for AES support lead to a runtime errors with "illegal instruction" when cryptographic operations are assumed to be hardware-accelerated but are not. With this update, the code now explicitly verifies the availability of AES hardware cryptography before attempting to use it, enhancing reliability and performance on diverse architectures.

briansmith:
Are you saying that `getauxval` indicates that AES instructions are available, even when they aren't?

briansmith:
> . Previously, the Rust compiler might assume that all aarch64 CPUs support AES hardware acceleration, which is not always the case. 

Are you sating that the compiler always evaluates `cfg!(target_feature = "aes")` to `true` for aarch64-unknown-linux-gnu? Or some other linux target? If it is a different Linux target, what is the target? I am not seeing that using these commands:

```
$ rustc +1.61.0 --print cfg --target=aarch64-unknown-linux-gnu | grep feature
target_feature="neon"
target_feature="pmuv3"
```

```
$ rustc --version
rustc 1.78.0 (9b00956e5 2024-04-29)
$ rustc --print cfg --target=aarch64-unknown-linux-gnu | grep feature
target_feature="neon"
```

src/aead/aes.rs
```diff
@@ -449,7 +511,22 @@ fn detect_implementation(cpu_features: cpu::Features) -> Implementation {
     )))]
     let _cpu_features = cpu_features;
 
-    #[cfg(target_arch = "aarch64")]
+    #[cfg(all(target_arch = "aarch64", target_os = "linux"))]
+    {
+        match aes_support() {
+            Ok(aes_support) => {
+                if cpu::arm::AES.available(cpu_features) && aes_support {
```
briansmith, line 518: We cannot open a file and read from it every time we try to encrypt/decrypt something.

Instead, change the implementation of `cpu::arm` so that `cpu::arm::AES.available(cpu_features)` returns false. This would either require changing the implementation of `detect_features()` if `getauxval` is the problem, or the way `ARMCAP_STATIC` is initialized, if `cfg!(target_feature = "aes")` is the problem.

But, first, we need a clearer understanding of what is the actual cause of the problem. I feel like this PR is just wallpapering over the issue.

Arne91:
> > . Previously, the Rust compiler might assume that all aarch64 CPUs support AES hardware acceleration, which is not always the case.
> 
> Are you sating that the compiler always evaluates `cfg!(target_feature = "aes")` to `true` for aarch64-unknown-linux-gnu? Or some other linux target? If it is a different Linux target, what is the target? I am not seeing that using these commands:
> 
> ```
> $ rustc +1.61.0 --print cfg --target=aarch64-unknown-linux-gnu | grep feature
> target_feature="neon"
> target_feature="pmuv3"
> ```
> 
> ```
> $ rustc --version
> rustc 1.78.0 (9b00956e5 2024-04-29)
> $ rustc --print cfg --target=aarch64-unknown-linux-gnu | grep feature
> target_feature="neon"
> ```

Thanks for your reply.

Yes, when I build with target cpu set to cortex-a72, I get the "aes" feature. Unfortunately I am not at my workstation (rapsberry pi 4 and my work laptop) at the moment. On Monday I will give you more details on what the problem is.

briansmith:
> Yes, when I build with target cpu set to cortex-a72

Thank you. I understand now exactly what the issue is; see https://github.com/briansmith/ring/issues/1858#issuecomment-2093675988. I am going to close this PR because I don't think this PR is on the right track.

PR #2052, opened 2024-05-06
AES: Use `NonZero_size_t` for ctr32_encrypt functions.

briansmith:
It turns out BSAES isn't the only one that required a non-empty in/out; at least the x86-64 implementation of `aes_hw_ctr32_encrypt_blocks` requires a non-empty in/out as well. Just assume all of them require it.

As the code currently exists, this is a no-op because `ctr32_encrypt_within` is never called with an empty in/out, but this would prevent memory safety issues if/when that changes.

### DIFF
--- a/src/aead/aes.rs
+++ b/src/aead/aes.rs
@@ -18,7 +18,7 @@ use crate::{
     c, constant_time, cpu, error,
     polyfill::{self, slice},
 };
-use core::ops::RangeFrom;
+use core::{num::NonZeroUsize, ops::RangeFrom};
 
 #[derive(Clone)]
 pub(super) struct Key {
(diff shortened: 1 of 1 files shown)


src/aead/aes.rs
```diff
@@ -136,10 +136,16 @@ unsafe fn ctr32_encrypt_blocks(
     let (input, leftover) = slice::as_chunks(&in_out[src]);
     debug_assert_eq!(leftover.len(), 0);
 
-    let blocks = input.len();
+    let blocks = match NonZeroUsize::new(input.len()) {
+        Some(blocks) => blocks,
+        None => {
+            return;
```
briansmith, line 142: This is covered by tests, but only when NEON is enabled on 32-bit ARM, when BSAES falls back to VPAES. Unfortunately, we don't have code coverage measurement for 32-bit ARM since the Rust toolchain doesn't enable the necessary infrastructure for it for these targets.

PR #2054, opened 2024-05-08
Add "unknown" to LINUX_ABI

exFalso:
This PR adds `unknown` to the LINUX_ABI array in build.rs. This is specifically to support `--target=x86_64-fortanix-unknown-sgx`, this target currently fails to link altogether because of the missing object files. With `unknown` added, everything works as expected.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -248,6 +248,7 @@ const LINUX_ABI: &[&str] = &[
     "linux",
     "redox",
     "solaris",
+    "unknown",
 ];
 
 const WIN32N: &str = "win32n";


exFalso:
Also, a comment on build.rs: the `is_git` branching and the `pregenerated/` folder flow makes it difficult to work with the project. It essentially means that the git repo does not reflect the sources that are on crates.io, which also means that the crate cannot be referred to as a cargo git dependency. cargo vendoring also breaks, and in general build tooling (like nix) that assumes the repo contains all sources required for the build stop working.

Suggestion: include the `pregenerated/` folder in the repository itself

briansmith:
Obviously every "unknown" OS isn't guaranteed to have the same ABI as Linux. A more precise change to build.rs is needed to support SGX.

exFalso:
Just noting: in SGX **there is no operating system**, it's all straight x86 code.

Also, it turned out that even with the project linking, at runtime the C code faulted, so we ended up completely replacing the ring dependency.

briansmith:
> Also, it turned out that even with the project linking, at runtime the C code faulted, so we ended up completely replacing the ring dependency.

That's because SGX requires its own implementation of the CPU feature detection in `ring::cpu::intel`. See the discussion in #775.

exFalso:
Makes sense, fair enough, thank you for the context!

PR #2060, opened 2024-05-14
x86: Remove 32-bit SIMD AES implementation.

briansmith:
The C code is never built with `OPENSSL_SSE2` so this has been dead code. Don't put any effort into getting it to work as there seems to be very little interest in 32-bit x86 performance.


briansmith:
Closing in favor of doing the opposite in PR #2062.

PR #2062, opened 2024-05-14
AES: Enable SSE2-based fallback implementation.

briansmith:
Address some warnings from Clang/GCC in the SSE2 code that are disabled by default in BoringSSL but not in *ring*.


briansmith:
Closing this in favor of #2063, which does the opposite, and which was merged.

PR #2067, opened 2024-05-19
aarch64-apple-*: Use a CStr polyfill in feature detection.

briansmith:
Make the memory safety clearer, and make it the check happen at compile time instead of runtime.

### DIFF
--- a/src/polyfill/cstr.rs
+++ b/src/polyfill/cstr.rs
@@ -1,0 +1,93 @@
+// Copyright 2024 Brian Smith.
+//
+// Permission to use, copy, modify, and/or distribute this software for any
+// purpose with or without fee is hereby granted, provided that the above
+// copyright notice and this permission notice appear in all copies.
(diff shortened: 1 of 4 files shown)


src/polyfill/cstr.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        // It is safe to cast a `*const u8` to a `const c_char` as they are the
+        // same size and alignment.
+        self.0.as_ptr().cast()
+    }
+
+    // SAFETY: Same as `CStr::from_bytes_with_nul_unchecked`.
+    const unsafe fn from_bytes_with_nul_unchecked(value: &'static [u8]) -> Self {
+        Self(value)
+    }
```
briansmith, line 42: Codecov is right that this isn't covered, as it is intended to only be executed at compile time. Ditto for the rest of the code in this `cstr` submodule.

PR #2068, opened 2024-05-19
CPU: Test the SHA-512 detection logic for macOS.

briansmith:


src/cpu/arm/darwin.rs
```diff
@@ -97,3 +97,23 @@ pub fn detect_features() -> u32 {
...
+        const _SHA512_NOT_STATICALLY_DETECTED: () = assert!((ARMCAP_STATIC & SHA512.mask) == 0);
+
+        if cfg!(target_os = "macos") {
+            // All aarch64-apple-darwin targets have SHA3 enabled statically...
+            assert!(cfg!(target_feature = "sha3"));
+
+            // ...so we should detect it.
+            assert!(SHA512.available(cpu::features()));
+        }
+    }
```
briansmith, line 118: Assuming codecov is complaining that the hidden `else` branch isn't covered, that's true. We'll need to add testing for the other apple targets in the near future.

PR #2070, opened 2024-05-21
Rewrite aes_nohw.c in Rust

briansmith:


briansmith:
I updated this PR with the following changes:
* Rebased on the current main branch.
* Merged aes_nohw.rs into fallback.rs. (In retrospect, I'm not sure this was the best choice, but it's not worse enough to undo.)
* Rewrote `ctr32_encrypt_within` to be a more literal transliteration of the original C implementation.
* `ctr32_encrypt_within` is now implemented in terms of `Overlapping`.

Issue #1512, opened 2022-09-02
Fails to build on SPARC64


Richard-Rogalski:
Both ring 0.16.20 and the latest master branch fail to build on SPARC64. I'll include what I believe to be the relevant output.

```
error: failed to run custom build command for `ring v0.17.0-not-released-yet (/jdk/conduit/ring)`

Caused by:

  process didn't exit successfully: `/jdk/conduit/ring/target/debug/build/ring-12602144b6b3c18f/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=RING_PREGENERATE_ASM
  cargo:rustc-env=RING_CORE_PREFIX=ring_core_0_17_0_not_released_yet_
  OPT_LEVEL = Some("0")
  TARGET = Some("sparc64-unknown-linux-gnu")
  HOST = Some("sparc64-unknown-linux-gnu")
  CC_sparc64-unknown-linux-gnu = None
  CC_sparc64_unknown_linux_gnu = None
  HOST_CC = None
  CC = None
  CFLAGS_sparc64-unknown-linux-gnu = None
  CFLAGS_sparc64_unknown_linux_gnu = None
  HOST_CFLAGS = None
  CFLAGS = None
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some("true")
  CARGO_CFG_TARGET_FEATURE = None

  --- stderr

[...]

PR #2077, opened 2024-05-23
Introduce SPARC64 support

Richard-Rogalski:
Adds the minimum information to build on sparc64 / sparcv9. Tests pass 100%.

Finally a sore in my conscience gone. Much thanks to the devs from Gentoo and Debian who work very hard to keep these machines in the shape they're in.

For end users: remember to `CFLAGS="-mcpu=ultrasparc" CXXFLAGS="-mcpu=ultrasparc" RUSTFLAGS="-C target-cpu=ultrasparc"`. If you want to build for a higher target, the C/CXX target can be found in `man gcc`, where for rust you can find it with `rustc --print target-cpus`.

Closes: #1512

### DIFF
--- a/include/ring-core/target.h
+++ b/include/ring-core/target.h
@@ -60,6 +60,8 @@
 #define OPENSSL_32_BIT
 #elif defined(__s390x__)
 #define OPENSSL_64_BIT
+#elif defined(__sparc_v9__) && defined(__LP64__)
+#define OPENSSL_64_BIT
 #else
 #error "Unknown target CPU"
 #endif
--- a/mk/cargo.sh
+++ b/mk/cargo.sh
@@ -30,6 +30,7 @@ qemu_powerpc64="qemu-ppc64 -L /usr/powerpc64-linux-gnu"
 qemu_powerpc64le="qemu-ppc64le -L /usr/powerpc64le-linux-gnu"
 qemu_riscv64="qemu-riscv64 -L /usr/riscv64-linux-gnu"
 qemu_s390x="qemu-s390x -L /usr/s390x-linux-gnu"
+qemu_sparc64="qemu-sparc64 -L /usr/sparc64-linux-gnu"
 
 # Avoid putting the Android tools in `$PATH` because there are tools in this
 # directory like `clang` that would conflict with the same-named tools that may
@@ -169,6 +170,11 @@ case $target in
     export CFLAGS_s390x_unknown_linux_gnu="--sysroot=/usr/s390x-linux-gnu -march=zEC12"
     export CARGO_TARGET_S390X_UNKNOWN_LINUX_GNU_LINKER=s390x-linux-gnu-gcc
     export CARGO_TARGET_S390X_UNKNOWN_LINUX_GNU_RUNNER="$qemu_s390x"
+    ;;
+  sparc64-unknown-linux-gnu)
+    export CFLAGS_sparc64_unknown_linux_gnu="--sysroot=/usr/sparc64-linux-gnu"
+    export CARGO_TARGET_SPARC64_UNKNOWN_LINUX_GNU_LINKER=sparc64-linux-gnu-gcc
+    export CARGO_TARGET_SPARC64_UNKNOWN_LINUX_GNU_RUNNER="$qemu_sparc64"
     ;;
   x86_64-unknown-linux-musl)
     use_clang=1
--- a/mk/install-build-tools.sh
+++ b/mk/install-build-tools.sh
@@ -165,6 +165,12 @@ s390x-unknown-linux-gnu)
     gcc-s390x-linux-gnu \
     libc6-dev-s390x-cross
   ;;
+sparc64-unknown-linux-gnu)
+  install_packages \
+    qemu-user \
+    gcc-sparc64-linux-gnu \
+    libc6-dev-sparc64-cross
+  ;;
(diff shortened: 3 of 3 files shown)


include/ring-core/target.h
```diff
@@ -60,6 +60,8 @@
 #define OPENSSL_32_BIT
 #elif defined(__s390x__)
 #define OPENSSL_64_BIT
+#elif defined(__sparc_v9__)
```
briansmith, line 63: I understand that `__sparc_v9__` is supposed to imply 64-bit. But some versions of LLVM/clang have had bugs where they defined `__sparc_v9__` even for 32-bit mode.
So I think we also need to check that `__LP64__` is defined too, just in case the user is using an old compiler. See https://reviews.llvm.org/D98574.

mk/cargo.sh
```diff
@@ -170,6 +171,13 @@ case $target in
     export CARGO_TARGET_S390X_UNKNOWN_LINUX_GNU_LINKER=s390x-linux-gnu-gcc
     export CARGO_TARGET_S390X_UNKNOWN_LINUX_GNU_RUNNER="$qemu_s390x"
     ;;
+  sparc64-unknown-linux-gnu)
+    # Ultrasparc is the sparcv9 ISA, without this -mcpu compilers can
+    # default to lower ISA versions.
+    export CFLAGS_sparc64_unknown_linux_gnu="--sysroot=/usr/sparc64-linux-gnu -mcpu=ultrasparc"
```
briansmith, line 177: Well...I am a bit unsure about this `-mcpu-ultrasparc`. If this is something that everybody targeting sparc64 should do, then this should be done in cc-rs instead of here.

briansmith (changes requested):
I am nervous about https://github.com/rust-lang/rust/issues/115336 since we do use `#[repr(transparent)]` structures in *ring*. Especially in conjunction with the `-mcpu=` issues, where I guess could affect the ABI compatibility.

Do all the tests pass even without the `-mcpu=` bits

mk/cargo.sh
glaubitz: The `-mcpu=ultrasparc` is definitely not necessary. sparc64 defaults to SPARCv9 which is what we want here.

See: https://gcc.gnu.org/onlinedocs/gcc/SPARC-Options.html

Richard-Rogalski:
Okay. It builds and still passes all tests, regardless of what {C, CXX, RUST}FLAGS are set. It is exclusively for 64 bit sparc.

@glaubitz : On gentoo, we don't package rust for our 32-bit userland profile. (Upstream doesn't provide binary toolchains for sparc at all, and it's a hassle enough as it is for 64bit). Does debian support(or have plans to continue support for) rust programs on 32bit sparc? If so, I can do another PR in the future for sparcv9 32-bit.

Sparcv8 and lower, even I am not insane enough for :b and neither are any linux distros

Richard-Rogalski:
Actually, I just saw the few duplicate issues and PRs, and [this](https://github.com/briansmith/ring/pull/2042#issuecomment-2120898907) comment. To me, this seems like a much better approach-- the only thing I can see this whitelist doing is determining 32/64 bit, which there are definitely better ways to do than listing every possible architecture ever. Can I do a new PR implementing this instead?

glaubitz:
> Okay. It builds and still passes all tests, regardless of what {C, CXX, RUST}FLAGS are set. It is exclusively for 64 bit sparc.
> 
> @glaubitz : On gentoo, we don't package rust for our 32-bit userland profile. (Upstream doesn't provide binary toolchains for sparc at all, and it's a hassle enough as it is for 64bit).

Well, upstream doesn't provide binary toolchains for sparc64 either. However, Debian's SPARC port is fully 64-bit these days.

However, I usually add 32-bit SPARC support when I add 64-bit SPARC support since there is still a commercially supported 32-bit SPARC platform, namely Leon, which is used by ESA, NASA and other high-reliability industry products.

> Does debian support(or have plans to continue support for) rust programs on 32bit sparc? If so, I can do another PR in the future for sparcv9 32-bit.
> 
> Sparcv8 and lower, even I am not insane enough for :b and neither are any linux distros

32-bit SPARC defaults to SPARCv9 these days, except for Leon which is SPARCv7 with support for atomic operations.

glaubitz:
> Actually, I just saw the few duplicate issues and PRs, and [this](https://github.com/briansmith/ring/pull/2042#issuecomment-2120898907) comment. To me, this seems like a much better approach-- the only thing I can see this whitelist doing is determining 32/64 bit, which there are definitely better ways to do than listing every possible architecture ever. Can I do a new PR implementing this instead?

Well, how about we merge this commit first so that the package finally builds on sparc64. The improved version can still be implemented later, can't it?

briansmith (approved):

briansmith:
> Actually, I just saw the few duplicate issues and PRs, and [this](https://github.com/briansmith/ring/pull/2042#issuecomment-2120898907) comment. To me, this seems like a much better approach-- the only thing I can see this whitelist doing is determining 32/64 bit, which there are definitely better ways to do than listing every possible architecture ever. Can I do a new PR implementing this instead?

First, I just merged this, so that we get the cargo.sh and install-build-tools.sh changes. Those are the most difficult for me to produce myself.

For generalizing target.h away from the allowlist approach, please see PR #2082.

PR #2078, opened 2024-05-23
CI: Don't pass `-march` to GCC for s390x.

briansmith:
Use the defaults.

### DIFF
--- a/mk/cargo.sh
+++ b/mk/cargo.sh
@@ -166,7 +166,7 @@ case $target in
     # XXX: Using -march=zEC12 to work around a z13 instruction bug in
     # QEMU 8.0.2 and earlier that causes `test_constant_time` to fail
     # (https://lists.gnu.org/archive/html/qemu-devel/2023-05/msg06965.html).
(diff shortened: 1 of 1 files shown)


briansmith:
@uweigand CI seems to pass with this change. Is there any reason we shouldn't do it?

uweigand:
> @uweigand CI seems to pass with this change. Is there any reason we shouldn't do it?

The CI still seems to be using the stock Jammy qemu (6.2), which still has the bug https://lists.gnu.org/archive/html/qemu-devel/2023-05/msg06965.html.   This means if the generated code contains the z13 instruction LOCFHR, qemu will mistranslate it and execution will fail.

Now whether or not the generated code does contain the instruction of course depend both on the source code and the compiler level, so any change to any of that might cause the instruction to appear or disappear.  So even if in the *current* state, the instruction is not used and CI passes, any random change might cause the instruction to appear again, which would make the CI fail again.

So I'd prefer to leave the -march=zEC12 in by default, until such time as the CI server moves to use a recent-enough qemu build that no longer has this bug.

PR #2079, opened 2024-05-24
Add support for 64-bit big-endian MIPS

matoro:
This seems to have partial support in the mk scripts, but missing from target.h or CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -219,6 +219,11 @@ jobs:
           - target: i686-unknown-linux-musl
             host_os: ubuntu-22.04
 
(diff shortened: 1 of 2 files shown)


briansmith:
Thanks! Please try #2082 and see if that works for you. I think we don't need to add MIPSEB to the GitHub build matrix since we can't keep growing it to support every target arch.

matoro:
> Thanks! Please try #2082 and see if that works for you. I think we don't need to add MIPSEB to the GitHub build matrix since we can't keep growing it to support every target arch.

That certainly seems the better way to go, I'll close this in favor of that.

Issue #2053, opened 2024-05-08
Add functional store/restore to digest::Context


anatse:
For some cases, such as compute hash sum for very big data passed through the HTTP streams, it is necessary to store and restore digest context. Our case: we are have storage which provides HTTP API to download with resume, using ranges. And this storage provides SHA_XXX hash sum for whole provided data, not for part. Therefore it need to check integrity for all data at the end of stream. In most cases it is not possible to complete the operation in one session, in cause to network errors. And compute whole hash sum using ring impossible because Context cannot be restored from pervious session. We also cannot see whole data at a time because we are acts just as a proxy and target system also not provides as a hash sum.

briansmith:
I don't have a problem with adding this, but I don't have cycles to do the work on it, so somebody would need to contribute it.

* The format of the internal state should be opaque to the API. The API should export/import a slice of bytes. This means we need to pick an endianness for the bytes. I suggest little endian, since that's less work for most targets.
* We shouldn't try to serialize `digest::Algorithm` or ``digest::AlgorithmId`. Instead the user should pass in the `&digest::Algorithm`. The user will need to be responsible for keeping track of what algorithm goes to which bytes.
* The APIs should be named "....less_safe()`.
* We should have tests that exercise every algorithm, in the same style as the existing digest tests.

anatse:
PR - https://github.com/briansmith/ring/pull/2081

PR #2081, opened 2024-05-27
2053: Add store/restore functionality to digest::Context

anatse:
Add **store/restore** functionality to digest::Context (https://github.com/briansmith/ring/issues/2053)


Cargo.toml
```diff
@@ -157,6 +157,7 @@ name = "ring"
 cfg-if = { version = "1.0.0", default-features = false }
 getrandom = { version = "0.2.10" }
 untrusted = { version = "0.9" }
+serde = { version = "1.0.203", optional = true, features = ["derive"] }
```
briansmith, line 160: Hi, I don't want a serde dependency. We need a serde-free API.

anatse: Hi, serde removed

anatse:
up

PR #2097, opened 2024-06-12
Define native word size as `constant_time::Word`.

briansmith:


src/limb.rs
```diff
@@ -21,34 +21,20 @@
 use crate::{c, error, polyfill::ArrayFlatMap};
 
 #[cfg(any(test, feature = "alloc"))]
-use crate::bits;
+use crate::{bits, constant_time, polyfill::usize_from_u32};
```
MarijnS95, line 24: Hey! When building without `"alloc"` feature (common when using `default-features = false`) the below `pub const LIMB_BITS: usize = usize_from_u32(Limb::BITS);` tries to use this `feature = "alloc"`-guarded import and fails to compile, is that intended?

MarijnS95: Pushed a fix in https://github.com/briansmith/ring/pull/2214 :+1:

PR #2102, opened 2024-06-14
aes: Split FFI helpers into their own submodule.

briansmith:
Make it clearer which types are used in the FFI and start separating the (unsafe) code for dealing with the C/assembly implementations from the (mostly safe) rest of the code.

`git difftool HEAD^1:src/aead/aes.rs src/aead/aes/ffi.rs`.

### DIFF
--- a/src/aead/aes.rs
+++ b/src/aead/aes.rs
@@ -1,4 +1,4 @@
-// Copyright 2018 Brian Smith.
+// Copyright 2018-2024 Brian Smith.
 //
 // Permission to use, copy, modify, and/or distribute this software for any
 // purpose with or without fee is hereby granted, provided that the above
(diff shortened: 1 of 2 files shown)


src/aead/aes/ffi.rs
```diff
@@ -0,0 +1,211 @@
...
+            KeyBytes::AES_128(bytes) => (&bytes[..], BitLength::from_bits(128)),
+            KeyBytes::AES_256(bytes) => (&bytes[..], BitLength::from_bits(256)),
+        };
+
+        // Unusually, in this case zero means success and non-zero means failure.
+        if 0 == unsafe { f(bytes.as_ptr(), key_bits, &mut key) } {
+            debug_assert_ne!(key.rounds, 0); // Sanity check initialization.
+            Ok(key)
+        } else {
+            Err(error::Unspecified)
```
briansmith, line 56: It is expected that this isn't reached because we've met all the preconditions to ensure the function won't fail.

src/aead/aes.rs
```diff
@@ -192,15 +57,15 @@ impl Key {
...
             },
 
             // SAFETY: `aes_nohw_set_encrypt_key` satisfies the `set_encrypt_key!`
             // contract.
             Implementation::NOHW => unsafe {
-                set_encrypt_key!(aes_nohw_set_encrypt_key, bytes, &mut key, cpu_features)?;
+                set_encrypt_key!(aes_nohw_set_encrypt_key, bytes, cpu_features)
             },
-        };
+        }?;
```
briansmith, line 68: This isn't reachable because the branch where the called function returns an error isn't reachable. No different tha before. We'll deal with this in a future change.

PR #2103, opened 2024-06-14
gcm: Split FFI helpers into a separate submodule.

briansmith:
Make it clearer which types are used in the FFI and start separating the (unsafe) code that's used only for the FFI from the (safe) code used in the Rust implementation.

`git difftool HEAD^1:src/aead/gcm.rs src/aead/gcm/ffi.rs`

### DIFF
--- a/src/aead/gcm/ffi.rs
+++ b/src/aead/gcm/ffi.rs
@@ -1,0 +1,186 @@
+// Copyright 2018 Brian Smith.
+//
+// Permission to use, copy, modify, and/or distribute this software for any
+// purpose with or without fee is hereby granted, provided that the above
+// copyright notice and this permission notice appear in all copies.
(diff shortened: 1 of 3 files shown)


src/aead/gcm/ffi.rs
```diff
@@ -0,0 +1,186 @@
...
+    ) {
+        use crate::polyfill::slice;
+        use core::num::NonZeroUsize;
+
+        let input = slice::flatten(input);
+
+        let input_len = match NonZeroUsize::new(input.len()) {
+            Some(len) => len,
+            None => {
+                return;
```
briansmith, line 128: This is currently unreachable (before and after this commit), but it might not always be unreachable, so we'll leave it as-is. In the future we should write a test.

PR #2112, opened 2024-06-20
Add *-apple-{visionos,watchos} support; CI: build tvOS,visionOS,watchOS.

briansmith:


briansmith:
Note that this might cause some large tables to be added to the build for P-256 and X25519 that weren't included before, which could significantly bloat the size of the app. This might be especially concerning for watchOS applications. We may want to merge BoringSSL's https://boringssl.googlesource.com/boringssl/+/20c9406971b39d214d4d6997f3a6e3ec772c440a to address this.

eugenehp:
Looks good, just be careful with the `target_vendor` part https://github.com/rust-lang/rust/issues/100343

PR #2113, opened 2024-06-20
{aarch64,x86_64}-apple-*: Use assembly language implementations.

briansmith:
Change visionOS and watchOS to use the assembly implementations like iOS, macOS, and tvOS were already.

Use the same C compiler options for all these targets.


briansmith:
This had to be folded into #2112.

PR #2116, opened 2024-06-21
arm: Switch from denylisting Apple targets to allowlisting Android/Linux.

briansmith:
The allowlist approach is safer.


briansmith:
@eugenehp Thanks for your comment in the other PR about the `target_vendor` usage. It caused me to reconsider some of the use of it, and this PR removes those uses. PTAL.

I think the rest of the `target_vendor = "apple"` usage is benign if redundant. I guess if/when it causes trouble we can reconsider it.

Issue #2088, opened 2024-06-06
Unknown target cpu : xtensa-esp32s3-espidf


lorenzo-pirchio:
Hi,
I'm working on a project on esp32s3, I need to use rustls but I have problems with the ring crate since xtensa is not supported .
I'm using the esp32-idf-template (https://github.com/esp-rs/esp-idf-template), and I changed the depedencies adding the ones I needed.
As suggested by @briansmith, I am using the latest commit on the main branch, checking if there is this change (https://github.com/briansmith/ring/commit/ed5b2a86cccfe30e670b33536fc7e494fb5e9e19), but It still doesn't work.
To be sure that I was using the latest commit, I downloaded the repo and added it as a local dependency.

*This is my cargo.toml*:

`[package]
name = "s3_rust"
version = "0.1.0"
edition = "2021"
resolver = "2"
rust-version = "1.71"

[[bin]]
name = "s3_rust"
harness = false # do not use the built in cargo test harness -> resolve rust-analyzer errors

[profile.release]
opt-level = "s"

[profile.dev]
debug = true    # Symbols are nice and they don't increase the size on Flash
opt-level = "z"

[features]
default = ["std", "embassy", "esp-idf-svc/native"]

pio = ["esp-idf-svc/pio"]
std = ["alloc", "esp-idf-svc/binstart", "esp-idf-svc/std"]
alloc = ["esp-idf-svc/alloc"]
nightly = ["esp-idf-svc/nightly"]
experimental = ["esp-idf-svc/experimental"]
embassy = ["esp-idf-svc/embassy-sync", "esp-idf-svc/critical-section", "esp-idf-svc/embassy-time-driver"]

[dependencies]
log = { version = "0.4", default-features = false }
esp-idf-svc = { version = "0.48", default-features = false }

[target.'cfg(target_os="espidf")'.dependencies]
ring = {path="C:/ring-main/ring-main", features =[ "less-safe-getrandom-espidf"]  }
rustls = { version = "0.23.7", default-features = false, features = ["ring", "logging", "std", "tls12"] }

[build-dependencies]
embuild = "0.31.3"
`

This is my config.toml:
`[build]
target = "xtensa-esp32s3-espidf"

[target.xtensa-esp32s3-espidf]
linker = "ldproxy"
runner = "espflash flash --monitor" # Select this runner for espflash v2.x.x
[...]

PR #2124, opened 2024-07-01
Add XTENSA architecture to target.h

juliankrieger:
In accordance to #2088


briansmith:
Thanks for doing this. This looks good to me but the commit message is not good. Please improve the commit message to something like "Fix build for Xtensa targets when using older versions of GCC."

juliankrieger:
Ah, sorry. Consider it done.

juliankrieger:
Done.

PR #2130, opened 2024-07-29
Split out wasm32_unknown_unknown_js feature

Protryon:
This PR:
* Adds a feature `wasm32_unknown_unknown`, which is identical to `wasm32_unknown_unknown_js`, except it doesn't enable the `getrandom` crate features.
* This allows WASM-only environments to implement their own context-specific rand functionality and register it with `getrandom`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -195,7 +195,8 @@ std = ["alloc"]
 unstable-testing-arm-no-hw = []
 unstable-testing-arm-no-neon = []
 test_logging = []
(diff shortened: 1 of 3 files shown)


briansmith:
One of the getrandom crate's maintainers has proposed something related to this in https://github.com/rust-random/getrandom/issues/498. I think we should resolve that issue and then revisit this once the getrandom issue is resolved.

briansmith:
Thanks for the PR. I'm no longer accepting PRs for this project, but I appreciate your effort.

PR #2147, opened 2024-09-24
Remove needless as_bytes() call

samueltardieu:
`String::len()` already returns the number of bytes in the string.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -204,7 +204,7 @@ impl TestCase {
         let result = if s.starts_with('\"') {
             // The value is a quoted UTF-8 string.
 
-            let mut bytes = Vec::with_capacity(s.as_bytes().len() - 2);
+            let mut bytes = Vec::with_capacity(s.len() - 2);
             let mut s = s.as_bytes().iter().skip(1);
             loop {
                 let b = match s.next() {


briansmith:
Thanks for the PR!

> `String::len()` already returns the number of bytes in the string.

This is true but I would argue that the way the code is currently written is clearer and perhaps {`str`/`String`}::`len()` shouldn't exist. Since the current code passes Clippy I am inclined to leave it as-is and I hope Clippy doesn't start warning about this pattern.

WDYT?

samueltardieu:
I totally agree that {`String`, `str`}`::len()` should not exist in the first place. But they do, and I think using `.as_bytes().len()` adds to the confusion, not at the place where it is used, but everywhere else: it may let people think that this is necessary, and that `len()` alone would return the number of codepoints.

This is why I prefer to use the function directly and have a clippy lint (which can be disabled of course, at a local or project level) to warn people that this is unnecessary because precisely `len()` is weirdly defined on strings.

briansmith:
I think this is just a matter of taste that's not worth working on so I'm going to close it. I appreciate the effort though.

briansmith:
See PR #2265.

PR #2152, opened 2024-10-01
Derive SecureRandom for SystemRandom on Fuchsia

erickt:
The `getrandom::getrandom` on Fuchsia uses `zx_cprng_draw` to get cryptographically secure random data, which can be seen [here](https://github.com/rust-random/getrandom/blob/master/src/fuchsia.rs).


briansmith:
Thanks!

PR #2159, opened 2024-11-14
ci: add support for loongarch64-unknown-linux-{gnu,musl}

heiher:


.github/workflows/ci.yml
```diff
@@ -578,6 +592,8 @@ jobs:
           - aarch64-unknown-linux-gnu      # Has assembly
           - arm-unknown-linux-gnueabi      # Has assembly
           - armv7-unknown-linux-gnueabihf  # Has assembly
+          - loongarch64-unknown-linux-gnu  # Has assembly
+          - loongarch64-unknown-linux-musl # Has assembly
```
briansmith, line 596: There is no assembly in our code for these targets.

briansmith (changes requested):
Thanks. I don't want to add these to the CI matrix because they are just going to be running the same code that is already tested in other targets and the matrix is already too huge. However, I would like to accept the changes to cargo.sh and install-build-tools.sh. Could you revert the changes to ci.yml and leave the rest? Thanks!

heiher:
> Thanks. I don't want to add these to the CI matrix because they are just going to be running the same code that is already tested in other targets and the matrix is already too huge. However, I would like to accept the changes to cargo.sh and install-build-tools.sh. Could you revert the changes to ci.yml and leave the rest? Thanks!

Thanks for your review. It's done.

briansmith (approved):

PR #2163, opened 2024-12-05
cpu: Silence warnings about unnecessary `unsafe`.

briansmith:


briansmith:
The remaining CI failures will be dealt with in (a) separate PR(s).

PR #2164, opened 2024-12-05
aead: Support stacked borrows model using a new `InOut` type.

briansmith:
Notably, `InOut::input_output_len` constructs the `input` pointer from the `output` pointer in a way that safely avoids any concerns about the order of borrowing the (now implicit) input slice and output slice, and in particular whether any such borrowing invalidates any pointers derived from those slices.

Practically, this helps people who are using Miri in its default stacked borrows mode (as opposed to the tree borrows mode) verify the memory safety of our code.

### DIFF
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -1,4 +1,4 @@
-// Copyright 2015-2021 Brian Smith.
+// Copyright 2015-2024 Brian Smith.
 //
 // Permission to use, copy, modify, and/or distribute this software for any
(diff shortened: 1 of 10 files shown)


briansmith:
@jasonyu1996 PTAL at this, to see if it completely addresses your concern.

jasonyu1996:
> @jasonyu1996 PTAL at this, to see if it completely addresses your concern.

The change looks good to me.

PR #2167, opened 2024-12-05
CI: Update cargo-semver-checks to be compatible with newest toolchain.

briansmith:


briansmith:
The wasm32-browser test failures were dealt with in PR #2166.

PR #2216, opened 2025-01-10
build: Only replace `cl.exe` with `clang-cl` for ARM64 Windows builds

### LINKED ISSUE #2215: Conflicting `C(XX)FLAGS` when cross-compiling to `aarch64-pc-windows-msvc`
First of all thanks for landing #1339 in some crates.io release! Would be great if this repository had `git` tags to follow along where commits are reachable :sweat_smile:

## Setup

When cross-compiling (currently from a Linux setup) to Windows, it's common to use https://jake-shadle.github.io/xwin/ / https://github.com/Jake-Shadle/xwin to set up headers and libraries.

When running this in my home directory (`/home/marijn`):

```console
$ xwin --arch x86_64,aarch64 --sdk-version 10.0.22621 splat
```

I add the following to my `~/.cargo/config.toml` to support cross-compiling C(++) sources in our Rust tree, and ultimately linking a binary:

```toml
[target.aarch64-pc-windows-msvc]
linker = "lld-link"
rustflags = [
    "-Lnative=/home/marijn/.xwin-cache/splat/crt/lib/aarch64",
    "-Lnative=/home/marijn/.xwin-cache/splat/sdk/lib/um/aarch64",
    "-Lnative=/home/marijn/.xwin-cache/splat/sdk/lib/ucrt/aarch64",
]

[env]
CC_aarch64-pc-windows-msvc = "clang-cl"
CXX_aarch64-pc-windows-msvc = "clang-cl"
AR_aarch64-pc-windows-msvc = "llvm-lib"

CFLAGS_aarch64-pc-windows-msvc = "-Wno-unused-command-line-argument -fuse-ld=lld-link -vctoolsdir/home/marijn/.xwin-cache/splat/crt -winsdkdir/home/marijn/.xwin-cache/splat/sdk"
CXXFLAGS_aarch64-pc-windows-msvc = "-Wno-unused-command-line-argument -fuse-ld=lld-link -vctoolsdir/home/marijn/.xwin-cache/splat/crt -winsdkdir/home/marijn/.xwin-cache/splat/sdk"
```

## Problem statement

Unfortunately the Windows ARM64 support PR (#1339) overwrites the compiler with `clang`:

https://github.com/briansmith/ring/blob/d2e401ff0b5bf0d056e553075e16458b0d8a882b/build.rs#L549-L556

This code doesn't modify the flags, and leaves what my environment has in `CFLAGS_aarch64-pc-windows-msvc` in place. Without using the `clang-cl` driver, `clang` doesn't understand the above commands and compiling `ring` errors out with a bunch of errors like:

```
warning: ring@0.17.8: clang: error: unknown argument: '-vctoolsdir/home/marijn/.xwin-cache/splat/crt'
warning: ring@0.17.8: clang: error: unknown argument: '-winsdkdir/home/marijn/.xwin-cache/splat/sdk'
warning: ring@0.17.8: ToolExecError: Command "clang" "-O3" "--target=aarch64-pc-windows-msvc" "-ffunction-sections" "-fdata-sections" "-g" "-fno-omit-frame-pointer" "--target=aarch64-pc-windows-msvc" "-I" "ring/include" "-I" "myproject/target/aarch64-pc-windows-msvc/debug/build/ring-81f4360e84b9e6a3/out" "-Wno-unused-command-line-argument" "-fuse-ld=lld-link" "-vctoolsdir/home/marijn/.xwin-cache/splat/crt" "-winsdkdir/home/marijn/.xwin-cache/splat/sdk" "-fvisibility=hidden" "-std=c1x" "-Wall" "-Wbad-function-cast" "-Wcast-align" "-Wcast-qual" "-Wconversion" "-Wmissing-field-initializers" "-Wmissing-include-dirs" "-Wnested-externs" "-Wredundant-decls" "-Wshadow" "-Wsign-compare" "-Wsign-conversion" "-Wstrict-prototypes" "-Wundef" "-Wuninitialized" "-g3" "-Werror" "-o" "myproject/target/aarch64-pc-windows-msvc/debug/build/ring-81f4360e84b9e6a3/out/8551bf18c762c50d-curve25519.o" "-c" "ring/crypto/curve25519/curve25519.c" with args clang did not execute successfully (status code exit status: 1).
```

## Proposed solution

As mentioned at https://github.com/briansmith/ring/pull/1339#issuecomment-899678208 `clang-cl` works, albeit currently (on LLVM `18.1.8`) with lots of warnings of the form:
```
error: mixing declarations and code is incompatible with standards before C99 [-Werror,-Wdeclaration-after-statement]
error: unsafe buffer access [-Werror,-Wunsafe-buffer-usage]
```

I'd be happy to PR a change to _not_ overwrite the compiler when it is `clang-cl` like, next to the existing `... && !compiler.is_clang_like()`. `cc-rs` [already tracks if it's `clang_cl`](https://github.com/rust-lang/cc-rs/blob/5e17879f98b7186edde10b26993a8627d95b4eee/src/tool.rs#L446) but doesn't currently expose that publicly, so we'll have to add that first.

Thanks!

MarijnS95:
Fixes #2215
Closes https://github.com/briansmith/ring/pull/2699
Closes https://github.com/briansmith/ring/pull/2727
Closes: https://github.com/briansmith/ring/pull/2747

> [!CAUTION]
> This change depends on https://github.com/rust-lang/cc-rs/pull/1357.

When cross-compiling to Windows from Linux or similar, it's common to use the `clang-cl` driver from the LLVM toolchain which supports parsing `cl.exe`-like arguments.

Ring however doesn't seem to compile for ARM64 Windows using `cl.exe`, and contains a `// FIXME`-style workaround to use `clang` to compile its C files instead.

The command-line interface for `clang` isn't always compatible with that of `cl.exe` and `clang-cl`.  There didn't seem to be any trouble with this yet, but when cross-compiling from Linux it's common to explicitly provide "sysroots" via `-vctoolsdir` and `-winsdkdir` in `CFLAGS`. In such a setup this workaround in `ring` would pass those arguments to `clang`, resulting in "unknown argument" errors.

`cc-rs` can tell us exactly what compiler it found, and we can use this information to decide how to fill in this workaround.  If the user was already compiling their code with `clang-cl`, nothing has to be replaced. In the end, all this entails is changing the workaround to not compile with `clang`, but with `clang-cl` instead.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -181,9 +181,30 @@ fn cpp_flags(compiler: &cc::Tool) -> &'static [&'static str] {
             "-Wuninitialized",
         ];
         NON_MSVC_FLAGS
+    } else if compiler.is_like_clang_cl() {
+        static MSVC_FLAGS: &[&str] = &[
+            "/Gy",      // Enable function-level linking.
+            "/std:c11", // Standard
+            "/Zc:wchar_t",
+            "/Zc:forScope",
+            "/Zc:inline",
+            // Default warnings.
+            "/W4", // /Wall defaults to -Weverything (https://github.com/llvm/llvm-project/commit/f9b08a382cc1e0669805991849ad69efbd4703e8), use original W4 instead
+            // Warnings.
+            "/wd4127", // C4127: conditional expression is constant
+            "/wd4464", // C4464: relative include path contains '..'
+            "/wd4514", // C4514: <name>: unreferenced inline function has be
+            "/wd4710", // C4710: function not inlined
+            "/wd4711", // C4711: function 'function' selected for inline expansion
+            "/wd4820", // C4820: <struct>: <n> bytes padding added after <name>
+            "/wd5045", /* C5045: Compiler will insert Spectre mitigation for memory load if
+                        * /Qspectre switch specified */
+        ];
+        MSVC_FLAGS
     } else {
         static MSVC_FLAGS: &[&str] = &[
-            "/Gy", // Enable function-level linking.
+            "/Gy",      // Enable function-level linking.
+            "/std:c11", // Standard
             "/Zc:wchar_t",
             "/Zc:forScope",
             "/Zc:inline",


build.rs
```diff
@@ -547,9 +547,13 @@ fn obj_path(out_dir: &Path, src: &Path) -> PathBuf {
...
 fn configure_cc(c: &mut cc::Build, target: &Target, c_root_dir: &Path, include_dir: &Path) {
     let compiler = c.get_compiler();
-    // FIXME: On Windows AArch64 we currently must use Clang to compile C code
-    let compiler = if target.os == WINDOWS && target.arch == AARCH64 && !compiler.is_like_clang() {
-        let _ = c.compiler("clang");
+    // FIXME: On Windows AArch64 we currently must use clang or clang-cl to compile C code
+    let compiler = if target.os == WINDOWS
+        && target.arch == AARCH64
+        && !compiler.is_like_clang()
+        && !compiler.is_like_clang_cl()
```
briansmith, line 554: Which is more commonly used, clang-cl or clang? Is there advantage to putting one check in front of the other?

briansmith (changes requested):
In `ci.yml`, we have (twice):
```
      - if: ${{ matrix.target == 'aarch64-pc-windows-msvc' }}
        run: |
          echo "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\Llvm\x64\bin" >> $GITHUB_PATH
```
When using clang-cl, is this still necessary, or is clang-cl already in the path in GitHub actions?

If clang-cl is more common and clang-cl is already in the path, then we should remove this from ci.yml.

We should also update BUILDING.md "For Windows ARM64 targets..." to describe the updated situation.

build.rs
```diff
@@ -547,9 +547,13 @@ fn obj_path(out_dir: &Path, src: &Path) -> PathBuf {
...
+    {
+        c.compiler("clang");
```
briansmith, line 556: Should we start defaulting to clang-cl instead?

briansmith:
In the commit message:
> build: Don't overwrite clang-cl with clang for ARM64 Windows builds

here and elsewhere you use the word "overwrite" when I think "override" is the better word.

> When `ring` however overwrites this compiler with `clang`, 

I suggest instead we say "When this is done,".

> existing user arguments
in i.e. `CFLAGS` are not compatible resulting in various "unknown
argument" errors such as for `-vctoolsdir` and `-winsdkdir`.

Most likely, we need to modify our compiler flag handling in build.rs to pass the correct flags to clang-cl.

> As pointed out in the original ARM64 Windows support PR, compiling
`ring` with `clang-cl` works equally well as `clang`: allow that by not
overwriting the compiler with `clang` if it's already `clang-cl`, to not
have to sort out incompatible arguments.

IIRC, when I tried it, clang-cl couldn't compile the assembly source files, which is why we hard-code clang.

We need to modify ci.yml to add the clang-cl-based configuration.

I also think that, assuming it works "equally well," we should default to the clang-cl-based build (instead of clang) when we detect the C compiler is MSVC and the target is aarch64.

build.rs
MarijnS95: We're unlikely to keep this function at all, in favour of a `compiler.family()` and manual `match` or `matches!()` against `cc::ToolFamily`. Given that GNU cannot compile to MSVC at all (and that is caught inside `cc-rs`, though I don't know if this `ring` workaround is supposed to apply to `aarch64-pc-windows-gnu` as well), we might also replace these two lines with `&& compiler.is_like_msvc()` and document that `cl.exe` is unable to compile these files (which I understand to be the original comment)?

build.rs
MarijnS95: Perhaps. If users are more commonly compiling for Windows MSVC than Windows GNU, arbitrarily replacing their requested compiler with `clang-cl` whose arguments are compatible with `cl.exe` seems a lot less error-prone than giving them the `clang` interface. Any `cl.exe`-specific argument in their `CFLAGS` (my issue: #2215) and this fails to compile.

Either way, overwriting compilers - which might have been carefully set up via `CC` environment flags etc - seems very painful. As said below, having `ring` compile natively with `cl.exe` would be awesome. At least it's isolated to ARM64 Windows though.

build.rs
briansmith: It does seem like the logic should be "if the compiler is MSVC, then use clang-cl instead."

MarijnS95:
> In the commit message:
> 
> > build: Don't overwrite clang-cl with clang for ARM64 Windows builds
> 
> here and elsewhere you use the word "overwrite" when I think "override" is the better word.

Agreed. It doesn't overwrite the compiler, but it does "overwrite" or "overrule" a `CC` environment flag or similar in the users' environment.

> > When `ring` however overwrites this compiler with `clang`,
> 
> I suggest instead we say "When this is done,".

NAK. I didn't introduce the idea of "overwrite the compiler with `clang`" yet before this part of the paragraph, so `this` would refer to nothing.

> > existing user arguments
> > in i.e. `CFLAGS` are not compatible resulting in various "unknown
> > argument" errors such as for `-vctoolsdir` and `-winsdkdir`.
> 
> Most likely, we need to modify our compiler flag handling in build.rs to pass the correct flags to clang-cl.

As above, if we assume that user flags are more likely compatible with `cl.exe` than with `clang`, using `clang-cl` is the better option than picking `clang`.

I do think we can be more strict here. We're going to expose the `.family()` flag in `cc-rs`, so I'm thinking this:

```rust
match compiler.family() {
    // The workaround
    ToolFamily::Msvc { clang_cl: false } => c.compiler("clang-cl"),
    ToolFamily::Msvc { clang_cl: true } | ToolFamily::Clang { .. } => {},
    ToolFamily::Gnu { .. } => todo!("Test this"),
}
```

> > As pointed out in the original ARM64 Windows support PR, compiling
> > `ring` with `clang-cl` works equally well as `clang`: allow that by not
> > overwriting the compiler with `clang` if it's already `clang-cl`, to not
> > have to sort out incompatible arguments.
> 
> IIRC, when I tried it, clang-cl couldn't compile the assembly source files, which is why we hard-code clang.

Curious what that entails. For me it failed only because of warnings forced as errors in the `git` tree:
https://github.com/briansmith/ring/blob/d2e401ff0b5bf0d056e553075e16458b0d8a882b/build.rs#L314

> We need to modify ci.yml to add the clang-cl-based configuration.
> 
> I also think that, assuming it works "equally well," we should default to the clang-cl-based build (instead of clang) when we detect the C compiler is MSVC and the target is aarch64.

Sure let's always use `clang-cl`.

briansmith:
> Curious what that entails. For me it failed only because of warnings forced as errors in the git tree:

In the issue, you mentioned:
> ```
> error: mixing declarations and code is incompatible with standards before C99 [-Werror,-Wdeclaration-after-statement]
> ```

Of course, we require C99 at least, so we need to change build.rs to tell clang-cl to use the right version of C. You can see here we have:

```
        static NON_MSVC_FLAGS: &[&str] = &[
            "-fvisibility=hidden",
            "-std=c1x", // GCC 4.6 requires "c1x" instead of "c11"
            "-Wall",
            "-Wbad-function-cast",
            "-Wcast-align",
            "-Wcast-qual",
            "-Wconversion",
            "-Wmissing-field-initializers",
            "-Wmissing-include-dirs",
            "-Wnested-externs",
            "-Wredundant-decls",
            "-Wshadow",
            "-Wsign-compare",
            "-Wsign-conversion",
            "-Wstrict-prototypes",
            "-Wundef",
            "-Wuninitialized",
        ];
```
So, when the compiler is clang-cl, we need to pass the clang-cl equivalent of `-std=c11` .


In the issue, you also mentioned:
> ```
> error: unsafe buffer access [-Werror,-Wunsafe-buffer-usage]
> ```
Obviously, this is pretty concerning and I'd like to prioritize addressing it.

The reason we have all of these warnings enabled in the .git build is precisely to ensure compatibility with the compilers we support.

If we're adding support for a new compiler, as we are here with clang-cl, then we need to have CI test this configuration. Which means doing the `ci.yml` changes and getting the build to pass in CI in the `.git` configuration where all the warnings are enabled.

briansmith:
> NAK. I didn't introduce the idea of "overwrite the compiler with clang" yet before this part of the paragraph, so this would refer to nothing.

It is introduced by the subject line "Don't override clang-cl with clang for ARM64 Windows builds."

build.rs
briansmith: We don't do anything to support -gnu windows targets, especially for aarch64, so you don't need to worry about them.

build.rs
briansmith: At the time, cl.exe didn't have `uint128_t` and probably there were other issues too. Maybe BoringSSL actually made some changes upstream to avoid requiring uint128_t for the ECC code? If so, then we can probably start supporting cl.exe.  Definitely, I would love it, and I think there's already an issue for it.

BUILDING.md
```diff
@@ -22,9 +22,11 @@ edition of Visual Studio, like Community, Standard, or Enterprise). The
...
 
+### (Cross-)compiling to Windows ARM64
+
 For Windows ARM64 targets (aarch64-pc-windows-msvc), the Visual Studio Build
 Tools “VS 2022 C++ ARM64 build tools” and "clang" components must be installed.
-Add Microsoft's provided version of `clang` to `%PATH%`, which will allow the
+Add Microsoft's provided version of `clang-cl` to `%PATH%`, which will allow the
 build to work in GitHub Actions without installing anything:
 ```
 $env:Path += ";C:\Program Files (x86)\Microsoft Visual Studio\2022\Enterprise\VC\Tools\Llvm\x64\bin"
```
briansmith, line 32: Is this where clang-cl lives?

BUILDING.md
```diff
@@ -22,9 +22,11 @@ edition of Visual Studio, like Community, Standard, or Enterprise). The
 “Desktop development with C++” workflow must be installed. Visual Studio
 2022 Version 17.5 is supported; earlier versions of Visual Studio may work.
 
+### (Cross-)compiling to Windows ARM64
+
 For Windows ARM64 targets (aarch64-pc-windows-msvc), the Visual Studio Build
 Tools “VS 2022 C++ ARM64 build tools” and "clang" components must be installed.
```
briansmith, line 28: Is clang-cl in the "clang" component or in a different one?

build.rs
```diff
@@ -547,10 +548,18 @@ fn obj_path(out_dir: &Path, src: &Path) -> PathBuf {
 
 fn configure_cc(c: &mut cc::Build, target: &Target, c_root_dir: &Path, include_dir: &Path) {
     let compiler = c.get_compiler();
-    // FIXME: On Windows AArch64 we currently must use Clang to compile C code
-    let compiler = if target.os == WINDOWS && target.arch == AARCH64 && !compiler.is_like_clang() {
-        let _ = c.compiler("clang");
-        c.get_compiler()
+    // FIXME: On Windows AArch64 we currently must use LLVM (clang-cl) to compile C code
```
briansmith, line 551: I think we should say "clang or clang-cl" instead of "clang-cl" here. I think we should add a comment "If the compiler is MSVC, switch to clang-cl so that `CFLAGS`, etc., are more likely to be compatible."

MarijnS95:
I've rewritten the patch and description:

1. We now use `clang-cl` in place of `clang`, because:
   1. The error I got when cross-compiling (passing `cl.exe` arguments to `clang` instead of `clang-cl`);
   2. If the user is already compiling with `clang`, nothing needs to be changed;
   3. Compiling with GNU is currently not supported (but is a valid target);
2. This means the `clang-cl` path is now natively tested in CI;
3. Added `/std:c11` to the MSVC arguments, but it doesn't seem to bite yet...

Note that this means we have the following targets to add to CI:

- [x] (Cross-)compiling from a Windows machine to ARM64 Windows using the default `cl.exe` (triggers the `// FIXME` to replace it with `clang-cl`);
- [ ] (Cross-)compiling from a Windows machine to ARM64 Windows using LLVM / `clang`;
- [ ] Cross-compiling from a Linux machine to ARM64 Windows with LLVM (my local target for testing);
- [ ] :grey_question: Compiling natively on an ARM64 Windows machine using the default `cl.exe`;
- [ ] :grey_question: Compiling natively on an ARM64 Windows machine using LLVM / `clang`.

build.rs
MarijnS95: I had that at first, but considered that the weight is mostly on the file being compiled via an `LLVM` toolchain. Whether the compiler/argument "driver" is `clang` or `clang-cl` doesn't matter, but you're right that I could comment this better in the `match` arms.

build.rs
MarijnS95: Cool - I'll try to cross-compile with `cl.exe` from my Windows X64 machine to find out and document what specifically is holding it back if at all.

At the same time I'm inquiring to get access to an ARM64 Windows machine soon, to test what happens when natively compiling on that machine.

build.rs
```diff
@@ -547,10 +548,24 @@ fn obj_path(out_dir: &Path, src: &Path) -> PathBuf {
...
-    // FIXME: On Windows AArch64 we currently must use Clang to compile C code
-    let compiler = if target.os == WINDOWS && target.arch == AARCH64 && !compiler.is_like_clang() {
-        let _ = c.compiler("clang");
-        c.get_compiler()
+    // FIXME: On Windows AArch64, ring C code cannot be compiled using cl.exe, but must be compiled
+    // using the LLVM toolchain.
+    let compiler = if target.os == WINDOWS && target.arch == AARCH64 {
+        match compiler.family() {
+            // If the user is already using clang or clang-cl, nothing needs to be replaced
+            cc::ToolFamily::Clang { .. } | cc::ToolFamily::Msvc { clang_cl: true, .. } => compiler,
```
briansmith, line 556: I suggest we deleted this case. Either we detect non-clang-cl MSVC and then switch it to clang-cl because we know cl.exe doesn't work, or we do nothing.

build.rs
```diff
@@ -547,10 +548,24 @@ fn obj_path(out_dir: &Path, src: &Path) -> PathBuf {
...
+            // with flags that are already in place (i.e. custom CFLAGS that the user provided to cc).
+            cc::ToolFamily::Msvc {
+                clang_cl: false, ..
+            } => {
+                // FIXME: This requires clang-cl to be available in PATH, regardless of any explicit
+                // or custom path that the user might have provided to cc via the CC flag.
+                c.compiler("clang-cl");
+                c.get_compiler()
+            }
+            t => panic!("Unsupported tool `{t:?}` for ARM64 Windows compilation"),
```
briansmith, line 567: Similarly, I think we just keep going with whatever the compiler is, and see what happens.

build.rs
MarijnS95: Not sure I understand, this already does nothing. Unless you simply mean that when replacing the below panic with `_ => compiler,`, this is a duplicate match arm and would intrinsically be removed anyway?

MarijnS95:
> In `ci.yml`, we have (twice):
> 
> ```
>       - if: ${{ matrix.target == 'aarch64-pc-windows-msvc' }}
>         run: |
>           echo "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\Llvm\x64\bin" >> $GITHUB_PATH
> ```
> 
> When using clang-cl, is this still necessary, or is clang-cl already in the path in GitHub actions?

Going all the way back to this, I'm confused by the CI setup:

For the `test:` job, there's a massive `target:` list. However, no `host_os` is set (mandatory for `runs-on`), so I assume only the explicit pairs in `include:` are actually running and this list is redundant?

For the "twice" case you mentioned, in `test-features:` there's no `aarch64-pc-windows-msvc` present in the matrix at all so `if: ${{ matrix.target == 'aarch64-pc-windows-msvc' }}` will likely never hit?

MarijnS95:
We don't yet exercise this codepath in the CI (and `cc-rs` seems to default-detect the presence of `clang` and uses that as-is when [natively compiling on Windows](https://github.com/MarijnS95/ring/actions/runs/13807138626/job/38620252134).

In this PR I added `/std:c11` and it's clearly visible in the CLI args for `clang-cl`, yet it keeps failing with a bunch of warnings indicating that `c11` support wasn't enabled?

```
  cargo:warning=ToolExecError: command did not execute successfully (status code exit status: 1): LC_ALL="C" "clang-cl" "-nologo" "-MD" "-O2" "-Z7" "-Brepro" "--target=aarch64-pc-windows-msvc" "-I" "ring/include" "-I" "target/aarch64-pc-windows-msvc/debug/build/ring-5d0b5febe883c645/out" "-WX" "/Gy" "/Zc:wchar_t" "/Zc:forScope" "/Zc:inline" "/Wall" "/wd4127" "/wd4464" "/wd4514" "/wd4710" "/wd4711" "/wd4820" "/wd5045" "/std:c11" "-fuse-ld=lld-link" "-vctoolsdir/home/marijn/.xwin-cache/splat/crt" "-winsdkdir/home/marijn/.xwin-cache/splat/sdk" "-o" "target/aarch64-pc-windows-msvc/debug/build/ring-5d0b5febe883c645/out/356905f97a05e3f6-sha256-armv8-win64.o" "-c" "--" "target/aarch64-pc-windows-msvc/debug/build/ring-5d0b5febe883c645/out/sha256-armv8-win64.S"
  cargo:warning=ring/crypto/fipsmodule/ec/p256-nistz.c(231,3): error: '_Alignas' is incompatible with C standards before C11 [-Werror,-Wpre-c11-compat]
  cargo:warning=  231 |   alignas(64) P256_POINT table[16];
  cargo:warning=      |   ^
...
```

The better solution is / might still be to prefer `cc` to compile with `clang` (which should ignore user `CFLAGS` when `CC` is likewise "ignored") and not have a `clang-cl` path to supersede a user-set `CFLAGS`+`CC` for `cl.exe`.

MarijnS95:
Regarding availability of `clang-cl` again, I was experimenting with our Debian bookworm runner image (Linux), and it doesn't provide `clang-cl` from neither the `llvm-15` nor the `clang-15` package, you need to install [`clang-tools-15`](https://packages.debian.org/bookworm/amd64/clang-tools-15/filelist) to get access to that.

Before however, our `clang-cl` was a simple symlink to `/usr/lib/llvm-15/bin/clang` (and `llvm-bin` was a symlink to `llvm-ar`, since these binaries likely look at `argv[0]` to understand in what mode they're supposed to be running). Not sure why it's a binary blob in that package, maybe it's a copy of `clang` for no reason.

build.rs
dragonmux: We would point out that uint128_t is not standardised at all either in [C++](https://en.cppreference.com/w/cpp/types/integer.html) or [C](https://en.cppreference.com/w/c/types/integer.html) so sin-binning the compiler for not supporting something that isn't even a part of the language standard seems misguided as this indicates broken code needing fixed.

MarijnS95:
I gleaned the idea to apply the `-Wno-*` flags **that I already had in this PR** exclusively to `clang-cl` from https://github.com/briansmith/ring/pull/2727 (so that we don't break existing `cl.exe` invocations), but that PR is otherwise identical to what I already had here.

dragonmux (changes requested):
Your changes for which warnings to disable goes far too far - silencing all warnings being produced indiscriminately is not the way. The two extra warnings this PR now silences should really be errors from a Rust perspective because they are the compiler telling you about legitimate bugs. Using reserved identifiers the compiler may itself be using, accessing buffers in a way that it is finding via static analysis likely leads to a buffer overflow (this is what unsafe buffer usage implies).

build.rs
MarijnS95: Thanks @mati865; yes the logic in this PR was written to not change anything if the used compiler is already LLVM-based.

cyqsimon:
Can we have this (or one of its alternatives #2699 #2727 #2747) merged please? Thanks to all in advance.

PR #2217, opened 2025-01-10
CI: Fix build by dealing with wasm32-wasi -> wasip1 transition.

briansmith:


briansmith:
It seems like we need to also update wasmtime. :sigh:

briansmith:
> It seems like we need to also update wasmtime. :sigh:

Nope, the problem was a bug in the update to mk/cargo.sh, which I've fixed.

PR #2230, opened 2025-01-16
chacha20_poly1305 internals: DRY per-packet initialization.

briansmith:
Create a new function `begin` that is used by all the opening and sealing operations in both `chacha20_poly1305` and `chacha20_poly1305_openssh`. Inline `derive_poly1305_key` into it.


briansmith:
Hmm....something is going on with the code coverage measurement.

PR #2253, opened 2025-01-20
arithmetic internals: Restore old signature of bn_mul_mont.

briansmith:
In BoringSSL, it returns `int`. Have it do that again, to make the next bit of refactoring easier, where we replace `bn_mul_mont` with a Rust function that dispatches to a bunch of other functions, all of which also are declared as having the same semantics.

The good thing is that this restores the assembly code to match BoringSSL.

The bad thing is that this some bits of dead code to handle zero return values that we'll never see.

### DIFF
--- a/crypto/fipsmodule/bn/internal.h
+++ b/crypto/fipsmodule/bn/internal.h
@@ -165,23 +165,16 @@ typedef crypto_word_t BN_ULONG;
 #error "Must define either OPENSSL_32_BIT or OPENSSL_64_BIT"
 #endif
 
-
-// |num| must be at least 4, at least on x86.
-//
-// In other forks, |bn_mul_mont| returns an |int| indicating whether it
-// actually did the multiplication. All our implementations always do the
-// multiplication, and forcing callers to deal with the possibility of it
-// failing just leads to further problems.
-//
-// In other forks, |bn_mod_mul|'s `num` argument has type |int| but it is
-// implicitly treated as a |size_t|; when |int| is smaller than |size_t|
-// then the |movq 48(%rsp),%r9| done by x86_64-xlate.pl implicitly does the
-// conversion.
+// See `bn_mul_mont_ffi` and `_MAX_LIMBS_ADDRESSES_MEMORY_SAFETY_ISSUES`.
 OPENSSL_STATIC_ASSERT(sizeof(int) == sizeof(size_t) ||
                       (sizeof(int) == 4 && sizeof(size_t) == 8),
                       "int and size_t ABI mismatch");
-void bn_mul_mont(BN_ULONG *rp, const BN_ULONG *ap, const BN_ULONG *bp,
-                 const BN_ULONG *np, const BN_ULONG *n0, size_t num);
+// |num| must be at least 4, at least on x86.
+// BoringSSL documents the return value as being 1 or 0, but some
+// implementations (e.g. 32-bit ARM) return different values on
+// success.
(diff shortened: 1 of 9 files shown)


crypto/fipsmodule/bn/internal.h
```diff
@@ -165,23 +165,16 @@ typedef crypto_word_t BN_ULONG;
...
+// See `bn_mul_mont_ffi` and `_MAX_LIMBS_ADDRESSES_MEMORY_SAFETY_ISSUES`.
 OPENSSL_STATIC_ASSERT(sizeof(int) == sizeof(size_t) ||
                       (sizeof(int) == 4 && sizeof(size_t) == 8),
                       "int and size_t ABI mismatch");
-void bn_mul_mont(BN_ULONG *rp, const BN_ULONG *ap, const BN_ULONG *bp,
-                 const BN_ULONG *np, const BN_ULONG *n0, size_t num);
+// |num| must be at least 4, at least on x86.
+// BoringSSL documents the return value as being 1 or 0, but some
+// implementations (e.g. 32-bit ARM) return different values on
+// success.
```
briansmith, line 175: In very recent versions of BoringSSL, where `bn_mul_mont` is written in C, it actually does return 1 or 0 as defined in BoringSSL's version of bn/internal.h.  However, before that refactoring was done, it was documented to return 1 or 0, but it would sometimes return non-zero values.

In particular, 32-bit ARM bn_mul_mont_nohw looks like it returns a pointer, which will be non-NULL, instead of 1 or 0. Looks like the NEON version does the same.

I think it would be better to change BoringSSL to do what *ring* is currently doing, namely just making all the `bn_mul_mont_*` functions that `bn_mul_mont` calls return `void`. Especially since there's no great way to handle them failing when we've met the preconditions.

WDYT @davidben?

briansmith: In particular, we need to go through all the 64-bit implementations and ensure that they aren't returning a pointer that gets interpreted as a 32-bit int, because it is possible for `(int)p == 0` when `p != 0` for a pointer p, when `sizeof(int) < sizeof(void*)`.

briansmith: Also, maybe we could remove the extra instruction in each implementation that truncates `num` to 32-bits while we're at it.

briansmith: It seems like 32-bit ARM is the only unusual case.

I did verify that the aarch64 version does `mov x0,#1` in each of the three functions that implement `bn_mul_mont`.

Also, I verified in x86_64-mont.pl that there is a `mov \$1,%rax` just prior to the epilogue of each function.

In x86-mont.pl, the last instruction of the function is `&mov	("eax",1);`

(I verified that all the functions in x86_4-mont5.pl, which aren't actually `bn_mul_mont` have a void return type. )

briansmith:
It is just as much work to ensure that every implementation correctly sets the return value as it is to change things to safely let them just assume that `len` meets their prerequisites. I have implemented the latter approach in PR #2254.

crypto/fipsmodule/bn/internal.h
davidben: Hmm? Not sure I'm following the pointer business. Here's 32-bit Arm.
https://boringssl.googlesource.com/boringssl/+/refs/heads/master/crypto/fipsmodule/bn/asm/armv4-mont.pl#262

PR #2254, opened 2025-01-20
arithmetic internals: Further clarify bn_mul_mont ABI.

briansmith:
The checks are superceded by the checks in the Rust code, so replace the checks with documenting the preconditions so we don't need to worry about the functions failing anymore.

Document some additional caveats discovered.


briansmith:
This was done in BoringSSL b918bad0f02fdd3ed6eff8704d2419209b18fa94.

PR #2269, opened 2025-01-22
Make it clearer that we're not passing empty slices to C code.

briansmith:
See each individual commit.


briansmith:
I reduced this to a single commit after merging #2272, which subsumed the poly1305 case. I will deal with other cases separately.

PR #2270, opened 2025-01-22
poly1305 internals: Remove unneeded I-U-F buffering.

briansmith:
`CRYPTO_poly1305_update` was designed to buffer any partial blocks, only passing full blocks to poly1305_update. Then
`CRYPTO_poly1305_finish` would pass in the final partial block.

`chacha20_poly1305`'s poly1305_update_padded_16 was working around that logic to ensure that a partial block never got buffered. Then it was doing extra work to pad the last block with zeros.

These two mechanisms were basically cancelling each other out. Instead, just avoid all the work.

This removes some non-trivial buffer management from C.


briansmith:
This requires some other changes first.

PR #2281, opened 2025-01-24
arithmetic/limbs: Rewrite limbs_equal_limbs_consttime in Rust.

briansmith:
The C variant is still used by the P-384 C code, so leave it.

In ec, combine the `Elem` equality functions into a single implementation that abstracts over the encoding.

This eliminates another call from Rust to a pointer-taking function written in C.


briansmith:
Unfortunately this creates a number of bits of unreachable code that lower the code coverage metrics. But they are all indeed unreachable.

PR #2295, opened 2025-01-26
Merge BoringSSL through d2e3212de29bac1ceed33ca8ab8bbff3f41a2459

briansmith:


crypto/fipsmodule/bn/asm/x86-mont.pl
```diff
@@ -288,8 +282,8 @@
 	&jle	(&label("outer"));
 
 	&emms	();				# done with mmx bank
-
-}	# The non-SSE2 code was removed. 
+	&jmp	(&label("common_tail"));
```
briansmith, line 285: I made this change just so the DIFF between *ring* and BoringSSL would be minimal.

PR #2299, opened 2025-01-27
arithmetic: Avoid heap & simplify alignment logic in `elem_exp_consttime`.

briansmith:
Avoid allocating on the heap. Let the compiler do the alignment instead of manually aligning the start of the table.


briansmith:
Updated substantially.

PR #2320, opened 2025-02-02
Merge BoringSSL through afaccd42b9804b0aaa1754e85ca2b4961cd95a27

briansmith:


crypto/internal.h
```diff
@@ -327,11 +327,22 @@ static inline void constant_time_conditional_memcpy(void *dst, const void *src,
...
+  // gcc 13.2.0 doesn't automatically vectorize this loop regardless of barrier
+  typedef aliasing_uint8_t v32u8 __attribute__((vector_size(32), aligned(1), may_alias));
+  size_t n_vec = n&~(size_t)31;
+  v32u8 masks = ((aliasing_uint8_t)mask-(v32u8){}); // broadcast
+  for (size_t i = 0; i < n_vec; i += 32) {
+    *(v32u8*)&out[i] ^= masks & *(v32u8 const*)&in[i];
+  }
+  out += n_vec;
+  n -= n_vec;
+#endif
```
briansmith, line 345: This is tested in CI but not in the code coverage measurement jobs, because we use clang, to allow code coverage for C code, ironically.

crypto/cpu_intel.c
```diff
@@ -166,19 +180,67 @@ void OPENSSL_cpuid_setup(void) {
...
+  // to have any downclocking problem when zmm registers are used.
+  if (is_intel && family == 6 &&
+      (model == 85 ||    // Skylake, Cascade Lake, Cooper Lake (server)
+       model == 106 ||   // Ice Lake (server)
+       model == 108 ||   // Ice Lake (micro server)
+       model == 125 ||   // Ice Lake (client)
+       model == 126 ||   // Ice Lake (mobile)
+       model == 140 ||   // Tiger Lake (mobile)
+       model == 141)) {  // Tiger Lake (client)
+    extended_features[0] |= 1u << 14;
```
briansmith, line 241: Nothing in *ring* is making use of AVX512 so for now this is dead code. We will update code coverage measurement when we add AVX512

PR #2321, opened 2025-02-02
digest: Enable AVX SHA-{384,512/256,512} on Zen+ AMD CPUs.

briansmith:
See BoringSSL fcef13a49852397a0d39c00be8d7bc2ba1ab6fb9.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -689,6 +689,29 @@ jobs:
             host_os: ubuntu-24.04
 
           - target: x86_64-unknown-linux-gnu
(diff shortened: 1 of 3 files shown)


briansmith:
It seems like the purpose of the "Is intel CPU" check is really to exclude older pre-Zen AMD CPUs, so perhaps we should just do that instead. Then we don't have to worry about reducing performance on those (admittedly very old) CPUs but also we can avoid any risk of (timing leak) issues with the microcode that implemented the SHRD/SHLD on those CPUs.

PR #2329, opened 2025-02-05
cpu/intel: Further reduce the C<->Rust interface for CPU feature detection.

briansmith:


src/cpu/intel.rs
```diff
@@ -35,139 +36,190 @@ mod abi_assumptions {
...
-        #[cfg(any( $( target_arch = $arch ),+ ))]
-        const $NAME: Feature = Feature {
-            word: $word,
-            mask: 1 << $bit,
-        };
+    let mut caps = 0;
 
-        #[cfg(any( $( target_arch = $arch ),+ ))]
-        impl_get_feature! { $NAME => $Name }
+    if check(cpuid, 0, 24) {
```
briansmith, line 92: It is expected, presently, that this condition is always true because we only support platforms that have SSE2 enabled by default, and there are not SSE2-enabled CPUs without FXSR.

PR #2331, opened 2025-02-05
Remove spin dependency on x86 and x86_64 targets.

briansmith:
Here's a representative example of how this changes the generated assembly in callers:

```
@@ -36,14 +36,11 @@
 	lea r12, [rcx - 4]
3343
 		Acquire => intrinsics::atomic_load_acquire(dst),
-	mov rax, qword ptr [rip + ring::cpu::intel::featureflags::FEATURES@GOTPCREL]
-	movzx eax, byte ptr [rax + 4]
-		match self.status.load(Ordering::Acquire) {
-	cmp al, 2
-		if let Some(value) = self.get() {
-	jne .LBB85_2
+	mov rax, qword ptr [rip + ring::cpu::intel::featureflags::FEATURES]
+		match NonZeroUsize::new(val) {
+	test rax, rax
+	je .LBB85_2
 	movabs rax, 274877906881
 		if input.len() > MAX_IN_OUT_LEN {
@@ -104,10 +101,9 @@
 		if in_out.len() >= SSE_MIN_LEN {
 	cmp r12, 129
 	jb .LBB85_8
-		*features
-	mov rax, qword ptr [rip + ring::cpu::intel::featureflags::FEATURES@GOTPCREL]
-	mov eax, dword ptr [rax]
3343
+		Acquire => intrinsics::atomic_load_acquire(dst),
+	mov rax, qword ptr [rip + ring::cpu::intel::featureflags::FEATURES]
 		if (self.values() & MASK) == MASK {
 	test eax, 256
@@ -216,9 +212,9 @@
 .LBB85_2:
 	.cfi_def_cfa rbp, 16
 	mov r13, r8
-		self.try_call_once_slow(f)
-	call spin::once::Once<T,R>::try_call_once_slow
+		None => self.init(f),
+	call ring::polyfill::once_cell::race::OnceNonZeroUsize::init
 	mov r8, r13
 	movabs rax, 274877906881
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -160,7 +160,7 @@ cfg-if = { version = "1.0.0", default-features = false }
 getrandom = { version = "0.2.10" }
 untrusted = { version = "0.9" }
 
-[target.'cfg(any(all(target_arch = "aarch64", target_endian = "little"), all(target_arch = "arm", target_endian = "little"), target_arch = "x86",target_arch = "x86_64"))'.dependencies]
+[target.'cfg(any(all(target_arch = "aarch64", target_endian = "little"), all(target_arch = "arm", target_endian = "little")))'.dependencies]
 spin = { version = "0.9.8", default-features = false, features = ["once"] }
 
 [target.'cfg(all(any(all(target_arch = "aarch64", target_endian = "little"), all(target_arch = "arm", target_endian = "little")), any(target_os = "android", target_os = "linux")))'.dependencies]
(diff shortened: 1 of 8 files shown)


briansmith:
This was rebased on top of #2338 and #2337 to eliminate some tiny regressions in CPU feature dispatching performance due to reloads.

The new version incorporates https://github.com/matklad/once_cell/pull/273. PR #2338 seems to obviate the need for adding `get_unchecked` but I did start the process of working w/ upstream on that at https://github.com/matklad/once_cell/pull/274. I also filed a documentation clarification at https://github.com/rust-lang/rust/issues/136669 to support the once_cell change.

PR #2333, opened 2025-02-06
cpu/arm: Remove old CPU feature detection API; unify with Intel.

briansmith:
After d39ac0fd8ab26dfe910088841ddc5243df0d9b40 we no longer have the need to be ABI-compatible with the `OPENSSL_armcap_P`-based feature detection.

Remove the tests for that ABI compatibility.

Don't add static CPU feature detection for Intel yet.

Use the same feature flag mechanism as Intel by moving the implementation from intel.rs to cpu.rs and expanding it with the static feature detection capability from arm.rs.

### DIFF
--- a/src/cpu/arm/fuchsia.rs
+++ b/src/cpu/arm/fuchsia.rs
@@ -12,7 +12,7 @@
 // OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
-use super::{AES, ARMCAP_STATIC, NEON, PMULL, SHA256};
+use super::{Aes, Neon, PMull, Sha256, CAPS_STATIC};
 
 pub const FORCE_DYNAMIC_DETECTION: u32 = 0;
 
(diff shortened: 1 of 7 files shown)


src/cpu/arm/fuchsia.rs
```diff
@@ -12,7 +12,7 @@
 // OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
-use super::{AES, ARMCAP_STATIC, NEON, PMULL, SHA256};
```
briansmith, line 15: @erickt We don't have Fuchsia in our CI yet so I'm not sure if these changes are enough to avoid breaking the build. PTAL if you can.

briansmith:
Net -55 Rust LoC is going to hurt our GitHub language metrics!

src/cpu/arm/fuchsia.rs
erickt: Thanks for the ping! I'll test it out and get back to you.

briansmith: PR #2336 is the next step in this which has similar hazards of missing `use`s and whatnot.

briansmith: OK, thanks!

PR #2346, opened 2025-02-09
cpu::intel: Do CPU feature detection in Rust.

briansmith:


briansmith:
It looks like Rust 1.63..=1.65 may contain bugs in `__cpuid`: https://github.com/rust-lang/rust/pull/101861 so we may need to vendor a newer version of `__cpuid` or do another workaround.

briansmith:
> It looks like Rust 1.63..=1.65 may contain bugs in __cpuid: https://github.com/rust-lang/rust/pull/101861 so we may need to vendor a newer version of __cpuid or do another workaround.

We should check what the latest versions of stdarch do and what libstd is using. Maybe we can look into workarounds used by other projects to use these safely.

briansmith:
This has two commits; the first is PR #2352. When PR #2352 gets merged, I will rebase this.

PR #2363, opened 2025-02-11
x86: Reduce harm when patched to remove the SSE2 requirement.

briansmith:


src/arithmetic/montgomery.rs
```diff
@@ -152,14 +152,21 @@ pub(super) fn limbs_mul_mont(
...
+            if let Some(cpu) = cpu.get_feature() {
+                bn_mul_mont_ffi!(in_out, n, n0, cpu, unsafe {
+                    (MIN_LIMBS, MOD_FALLBACK, Sse2) => bn_mul_mont
+                })
+            } else {
+                // This isn't really an FFI call; it's defined below.
+                unsafe {
+                    super::ffi::bn_mul_mont_ffi::<(), {MIN_LIMBS}, 1>(in_out, n, n0, (),
+                    bn_mul_mont_fallback)
+                }
```
briansmith, line 168: Codecov is right to complain that we don't test this branch in CI since we can't reach it without patching. This is intentional.

PR #2376, opened 2025-02-12
digest: Align state on 16-byte boundary.

briansmith:


briansmith:
The benchmarks for this are noisy so this might not be a win.

PR #2379, opened 2025-02-13
Update target.h to support MIPS64

humaita-github:
target.h was missing the definitions for MIPS64, cf. https://github.com/briansmith/ring/issues/1866


include/ring-core/target.h
```diff
@@ -49,6 +49,8 @@
 // Versions of GCC before 10.0 didn't define `__ILP32__` for all 32-bit targets.
 #elif defined(__MIPSEL__) || defined(__MIPSEB__) || defined(__PPC__) || defined(__powerpc__) || defined(__csky__) || defined(__XTENSA__)
 #define OPENSSL_32_BIT
+#elif defined(__MIPSEB__) && defined(__LP64__)
```
briansmith, line 52: We already have `#elif defined(__LP64__)` , which handles all pretty much all 64-bit targets. Why do we need this MIPS-specific check?

humaita-github:
PR not needed, as discussed in the previous post

PR #2381, opened 2025-02-13
Merge BoringSSL through e43169a421099d99d1f9c78409704a3128cb22e8

briansmith:


briansmith:
These changes seem pretty OK to merge but it is weird to merge dead code and it may be a while before we can merge more recent commits from BoringSSL as described in issue #2382.

briansmith:
I guess we'll merge it anyway because we need to get the AUTHORS file, to make sure people are being acknowledged well, and because the license headers in future merges may refer to it.

PR #2387, opened 2025-02-16
Fuchsia AArch64: Detect SHA-512 CPU feature & clean up existing detection.

briansmith:
See each individual commit message for details.


briansmith:
@erickt I noticed `cargo +nightly check --target=aarch64-unknown-fuchsia -Zbuild-std` does work though that doesn't help with verifying this change. At least I verified that previous changes don't break the build.

PR #2402, opened 2025-02-18
Merge BoringSSL 33d1049: Switch the license to Apache 2.0, matching OpenSSL upstream

briansmith:


LICENSE
```diff
@@ -22,183 +18,7 @@ including in particular all the Rust code:
...
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
-
-The above copyright notice and this permission notice shall be included in all
-copies or substantial portions of the Software.
+See LICENSE-BoringSSL for code that was sourced from BoringSSL.
```
briansmith, line 21: I don't really know the best way to deal with the multi-license situation insofar as a LICENSE file goes. Suggestions appreciated.

Cargo.toml
```diff
@@ -5,7 +5,7 @@ categories = ["cryptography", "no-std"]
 description = "Safe, fast, small crypto using Rust."
 edition = "2021"
 keywords = ["crypto", "cryptography", "rand", "ECC", "RSA"]
-license-file = "LICENSE"
+license = "APACHE AND (APACHE OR MIT) AND ISC"
```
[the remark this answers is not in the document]
briansmith: Thanks. I updated things based on what you wrote above. PTAL. Your help is much appreciated.

LICENSE
briansmith: Thanks. I think the scope of this PR should be limited to making sure the LICENSE file isn't wrong, making sure we include all the license files in the crate that are for the crate contents, having the machine-readable license for the crate contents, and getting rid of the use of the old licenses. I think documenting the licensing of things that don't end up in the crate should be done in another PR.

PR #2430, opened 2025-03-03
TODO(MSRV): RSA: deprecate accidentally-exposed API.

briansmith:
The `encode` method was exposed long, long ago accidentally. Start the process of fixing that.


briansmith:
Replaced with PR #2466.

PR #2439, opened 2025-03-04
cpu::intel: Do CPU feature detection in Rust.

briansmith:
It's easiest to review this PR commit-by-commit.


src/cpu/intel.rs
```diff
@@ -98,47 +94,127 @@ pub(super) mod featureflags {
...
+        // Expected to be unreachable on any environment we currently
+        // support.
+        leaf1_edx = 0;
+        leaf1_ecx = 0;
+        #[cfg(target_arch = "x86_64")]
+        {
+            extended_features_ecx = 0;
+            extended_features_ebx = 0;
+        }
+    }
```
briansmith, line 169: As codecov points out, and the comment notes, these lines are unreachable in *unpatched* versions of *ring*. (Some people do patch it in ways that make these lines reachable.)

PR #2447, opened 2025-03-05
aes: Avoid unwanted overflow check when using u32::MAX as the counter.

briansmith:
The problem occurs:
* release mode with `RUSTFLAGS="-C overflow-checks"`
* release mode with `overflow-checks = true` in the Cargo.toml profile.
* debug mode.

Thanks to Mike (GitHub user MikeRomaniuk).


briansmith:
@cpu @ctz @djc PTAL. I intend to do a release with this fix today.

ctz (approved):
I think this should be a single AES block, so it is a surprise that the input is being treated as a counter? Though the change itself looks alright.

briansmith:
> I think this should be a single AES block[1](#user-content-fn-1-98dcae082e03d94627ad1e6ac71ba2ec), so it is a surprise that the input is being treated as a counter? Though the change itself looks alright.

Sometimes we use this implementation of block encryption via ctr encryption because it is faster and/or the same speed but less code.

```
fn encrypt_block_using_encrypt_iv_xor_block(key: &impl EncryptBlock, block: Block) -> Block {
    key.encrypt_iv_xor_block(Iv(block), ZERO_BLOCK)
}
```

briansmith:
PR #2448 is the PR for the 0.17.12 release. Note that it is a minimal diff from the 0.17.11 release.

djc:
@briansmith is there a denial of service issue here that deserves an advisory?

briansmith:
The advisory is at https://github.com/rustsec/advisory-db/pull/2240. Additional tests at https://github.com/briansmith/ring/pull/2455 verify the hypothesis that the AES-GCM API is affected in that one edge case. Interestingly, only 64-bit, not 32-bit is affected, because usize::MAX is too small on smaller targets.

briansmith:
Yes, I think that's true. I was thinking that even in the best case scenario, it would still happen with non-negligible likelihood.

djc:
> I don't know for sure, but I was thinking a malicious peer can just arrange `sample` to end with 0xffff_ffff? AIUI it comes directly from parts of the received packet.

Yes, that matches my understanding.

briansmith:
> @briansmith Do we have any CVE-ID assigned or requested for this issue?

I see that Github has https://github.com/advisories/GHSA-4p46-pwfr-66x6 and dependabot has started sending out notices to people. My understanding is that every GitHub advisory will get a CVE assigned automatically? But, IDK for sure.

I noticed that GitHub's advisory doesn't show up in the "advisories" at https://github.com/briansmith/ring/security/advisories as of this moment, which I find curious. I would expect that if they're sending out Dependabot alerts then they'd also add their advisory to the project's own advisory list. Anway, we'll see what happens.

PR #2448, opened 2025-03-05
0.17.12

briansmith:


briansmith:
Note that the target branch for this is the newly-created release-branch/0.17.12 branch that is based on release-branch/0.17.11 which is what the 0.17.11 release was done from.

briansmith:
On the main branch, there were changes to increase MSRV and to potentially increase the binutils version required, so I decided to do this off a branch. Now the changes scheduled for 0.17.12 will be 0.17.13.

briansmith:
The clippy failure is something we fixed after 0.17.11 and it's benign.

PR #2456, opened 2025-03-06
limb.rs: specify ambiguous into type

RensHijdra:
I experienced a compile error for ring as a dependency of jsonwebtoken.
A `len.into()` is ambigous because of an impl definition of the serde_json crate; see below:

```rs
error[E0283]: type annotations needed
  --> /home/rens/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ring-0.17.12/src/limb.rs:66:23
   |
66 |     if a.len() != len.into() {
   |                --     ^^^^
   |                |
   |                type must be known at this point
   |
   = note: multiple `impl`s satisfying `usize: PartialEq<_>` found in the following crates: `core`, `serde_json`:
           - impl PartialEq for usize;
           - impl PartialEq<serde_json::value::Value> for usize;
help: try using a fully qualified path to specify the expected types
   |
66 -     if a.len() != len.into() {
66 +     if a.len() != <NonZero<usize> as Into<T>>::into(len) {
   |
```

My commit fixes this issue in the way the error message suggests.
Thanks!

### DIFF
--- a/src/limb.rs
+++ b/src/limb.rs
@@ -25,7 +25,7 @@ use crate::{
     polyfill::{sliceutil, usize_from_u32, ArrayFlatMap},
 };
 use core::{iter, num::NonZeroUsize};
-
+use core::num::NonZero;
 #[cfg(any(test, feature = "alloc"))]
 use crate::bits;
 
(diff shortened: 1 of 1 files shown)


src/limb.rs
```diff
@@ -60,7 +60,7 @@ fn limbs_less_than_limbs(a: &[Limb], b: &[Limb]) -> Result<LimbMask, LenMismatch
     // optimizer.
     // XXX: Questionable whether `LenMismatchError` is appropriate.
     let len = NonZeroUsize::new(b.len()).ok_or_else(|| LenMismatchError::new(a.len()))?;
-    if a.len() != len.into() {
+    if a.len() != <NonZero<usize> as Into<usize>>::into(len) {
```
briansmith, line 63: Thank you for the PR.

Instead, we should use `len.get()` instead of `len.into()`. Your change won't work with old versions of Rust we support because they don't have `NonZero<T>`.

briansmith:
I believe this issue is specific to whatever fork you have, right? Because `serde_json` is relevant to this project as it is here.

briansmith:
PTAL at PR #2457.

RensHijdra:
Yes, this issue only occurs when there is another impl definition of `PartialEq<T> for usize` in some crate, which happened to be serde_json in this case.
I wasn't that familiar with NonZero and its effect in older Rust versions, thanks for the fix!

PR #2467, opened 2025-03-09
Deprecate `ring::test`.

briansmith:


briansmith:
@riquito I expect that, for each public *ring* API, we'll probably need to define a function that takes the output of the `SecureRandom` as a byte slice (or similar) with a name "_less_safe_for_testing_only" or similar and perhaps guarded by a feature flag or similar.

For your own APIs that are parameterized like *ring* APIs, I guess you would do similar.

riquito:
Thank you for the reply, it made me realize that `hmac::Key` can be created from a slice of bytes.
If it helps any future reader, I solved my issue by replacing the above function with

```
fn setup_fake_random_key() -> hmac::Key {
    hmac::Key::new(hmac::HMAC_SHA256, &[])
}
```

PR #2470, opened 2025-03-10
Avoid using YMM registers in `gcm_ghash_vpclmulqdq_avx2_1`.

briansmith:
[Adapted by Brian Smith to *ring*'s prior modifications.]

gcm_ghash_vpclmulqdq_avx2() executes vzeroupper only when len >= 32, as it was supposed to use only xmm registers for len == 16.  But actually it was writing to two ymm registers unconditionally for $BSWAP_MASK and $GFPOLY.  Therefore, there could be a slow-down in later code using legacy SSE instructions, in the (probably rare) case where gcm_ghash_vpclmulqdq_avx2() was called with len=16 *and* wasn't followed by a function that does vzeroupper, e.g. aes_gcm_enc_update_vaes_avx2().

(The Windows xmm register restore epilogue of
gcm_ghash_vpclmulqdq_avx2() itself does use legacy SSE instructions, so probably was slowed down by this, but that's just 8 instructions.)

Fix this by updating gcm_ghash_vpclmulqdq_avx2() to correctly use only xmm registers when len=16.  This makes it match the similar code in aes-gcm-avx10-x86_64.pl, which does do it correctly, more closely.

Also, make both functions just execute vzeroupper unconditionally, so that it won't be missed again.  It's actually only 1 cycle on the CPUs this code runs on, and it no longer seems worth executing conditionally.

### DIFF
--- a/crypto/fipsmodule/aes/asm/aes-gcm-avx10-x86_64.pl
+++ b/crypto/fipsmodule/aes/asm/aes-gcm-avx10-x86_64.pl
@@ -786,9 +786,6 @@ sub _ghash_update {
     jae             .Laad_loop_1x$local_label_suffix
 
 .Laad_large_done$local_label_suffix:
-    # Issue the vzeroupper that is needed after using ymm or zmm registers.
-    # Do it here instead of at the end, to minimize overhead for small AADLEN.
-    vzeroupper
 
     # GHASH the remaining data 16 bytes at a time, using xmm registers only.
 .Laad_blockbyblock$local_label_suffix:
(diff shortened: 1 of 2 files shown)


briansmith:
@arielb1 @jlizen This may conflict with the thing you're doing.

arielb1:
@briansmith 

My PR works fine on top of the x/xmm branch

briansmith:
Oh yeah, regarding whether it is worthwhile to remove the vzeroupper: I benchmarked with and without it, and benchmarks showed no difference, so I left it, to match upstream.

PR #2471, opened 2025-03-10
hand-assemble instructions not supported in old binutils

### LINKED ISSUE #2463: Work around incompatibility with ancient GNU as (GNU binutils)
We have multiple bug reports where the latest release of *ring* doesn't build when using old versions of GNU binutils:

* https://github.com/briansmith/ring/issues/2462
* https://github.com/briansmith/ring/issues/2461

Luckily, this isn't urgent as *ring* 0.17.12 is available for people to use, but it could become urgent at any time if we have a security fix because we only do security releases on top of the latest version.

There are multiple workarounds available:
* Upgrade GNU binutils to a later version (2.38 seems sufficient; maybe some earlier versions work?)
* Use `CC=clang` or `TARGET_CC=clang` or similar, to use clang and llvm-as instead of GNU `as`.

I think somebody needs to investigate exactly why GNU as 2.29 rejects the instruction, to see if we can find an easy workaround. I don't have access to GNU as older than 2.38.

Perhaps there is an easy workaround, such as adding an `.arch` directive to `aes-gcm-avx2-x86_64.pl` that would make it work.

Or, there could be more tedious workarounds. In OpenSSL, in the past at least, they have sometimes used `.byte` directives to directly encode instructions. For example, for `mulx` in x86_64-mont.pl we have:

```
.byte	0xc4,0x62,0xfb,0xf6,0xa1,0x10,0x00,0x00,0x00	# mulx	2*8($nptr),%rax,%r12
```

The difficulty here, of course, is that the registers being accessed are encoded into the instruction, so this is kind of a last resort workaround.

I don't intend to implement such a workaround myself, but others might be interested in doing so.

arielb1:
This turned out to be quite ugly, but it fixes #2463, making ring work again on AL2.

The generated .o files are slightly different since there is no debuginfo, and I don't know of an easy way to add it back when using .byte instructions. Checked with objdump that there are no other differences in the .o files before and after this PR - might be worth having someone else check.


arielb1:
Maybe a global regex-replace would be nicer than using assemble - not sure.

briansmith:
> Maybe a global regex-replace would be nicer than using assemble - not sure.

Yes, I think that approach is what we'll need to do, because otherwise we're going to have lots of merge conflicts when merging changes from upstream BoringSSL.

Here is an example of how something similar is done in sha512-x86-64.pl:

```
sub sha256op38 {
    my $instr = shift;
    my %opcodelet = (
		"sha256rnds2" => 0xcb,
  		"sha256msg1"  => 0xcc,
		"sha256msg2"  => 0xcd	);

    if (defined($opcodelet{$instr}) && @_[0] =~ /%xmm([0-7]),\s*%xmm([0-7])/) {
      my @opcode=(0x0f,0x38);
	push @opcode,$opcodelet{$instr};
	push @opcode,0xc0|($1&7)|(($2&7)<<3);		# ModR/M
	return ".byte\t".join(',',@opcode);
    } else {
	return $instr."\t".@_[0];
    }
}

foreach (split("\n",$code)) {
	s/\`([^\`]*)\`/eval $1/geo;

	s/\b(sha256[^\s]*)\s+(.*)/sha256op38($1,$2)/geo;

	print $_,"\n";
}
```

Basically what happens is that the perlasm stuff happens to generate `$code` as a big string, and then this logic steps through it line-by-line and replaces each instruction.

arielb1:
Are you fine with a big list of instructions, or do you think I need to write code that does the encoding?

crypto/fipsmodule/aes/asm/aes-gcm-avx2-x86_64.pl
```diff
@@ -958,6 +958,66 @@ sub _aes_gcm_update {
 $code .= _aes_gcm_update 0;
 $code .= _end_func;
 
-print $code;
+sub filter_and_print {
+        my %asmMap = (
```
briansmith, line 962: The indention here looks a little off. I think it should look like this:
```
sub filter_and_print {
    my %asmMap = (
        'vaesenc %ymm2, %ymm12, %ymm12' => '.byte 0xc4,0x62,0x1d,0xdc,0xe2',
        ....
    )
    ....
```

briansmith (commented):
> Are you fine with a big list of instructions, or do you think I need to write code that does the encoding?

I think the big list of instructions is fine because the worst that will happen is that an unsupported combination of instruction/operands will be output verbatim, which will just break the build with older tools again, and then presumably somebody will make a new PR to fix this.

arielb1:
Fixed indentation. I don't know why coverage doesn't like me

crypto/fipsmodule/aes/asm/aes-gcm-avx2-x86_64.pl
arielb1: Fixed

arielb1:
> I think the big list of instructions is fine because the worst that will happen is that an unsupported combination of instruction/operands will be output verbatim, which will just break the build with older tools again, and then presumably somebody will make a new PR to fix this.

I agree, this looks less scary than having an encoder run wild

briansmith:
> I don't know why coverage doesn't like me

The coverage check likes to report that it is failing between the time the first coverage job finishes until all the coverage jobs have been run and codecov.io has updated the results.

briansmith:
> Checked with objdump that there are no other differences in the .o files before and after this PR - might be worth having someone else check.

The tests pass, so it's probably right. But, I will try with Intel XED (https://github.com/intelxed/xed) as I've used that in the past and found it helpful for this kind of thing.

crypto/fipsmodule/aes/asm/aes-gcm-avx2-x86_64.pl
```diff
@@ -958,6 +958,68 @@ sub _aes_gcm_update {
...
+        'vpclmulqdq      $0x10, %ymm5, %ymm4, %ymm2' => '.byte 0xc4,0xe3,0x5d,0x44,0xd5,0x10',
+        'vpclmulqdq      $0x10, %ymm7, %ymm2, %ymm2' => '.byte 0xc4,0xe3,0x6d,0x44,0xd7,0x10',
+        'vpclmulqdq      $0x10, %ymm8, %ymm2, %ymm2' => '.byte 0xc4,0xc3,0x6d,0x44,0xd0,0x10',
+        'vpclmulqdq      $0x11, %ymm2, %ymm12, %ymm4' => '.byte 0xc4,0xe3,0x1d,0x44,0xe2,0x11',
+        'vpclmulqdq      $0x11, %ymm2, %ymm12, %ymm7' => '.byte 0xc4,0xe3,0x1d,0x44,0xfa,0x11',
+        'vpclmulqdq      $0x11, %ymm3, %ymm13, %ymm4' => '.byte 0xc4,0xe3,0x15,0x44,0xe3,0x11',
+        'vpclmulqdq      $0x11, %ymm4, %ymm3, %ymm1' => '.byte 0xc4,0xe3,0x65,0x44,0xcc,0x11',
+        'vpclmulqdq      $0x11, %ymm4, %ymm3, %ymm2' => '.byte 0xc4,0xe3,0x65,0x44,0xd4,0x11',
+        'vpclmulqdq      $0x11, %ymm5, %ymm3, %ymm4' => '.byte 0xc4,0xe3,0x65,0x44,0xe5,0x11',
+        'vpclmulqdq      $0x11, %ymm5, %ymm4, %ymm3' => '.byte 0xc4,0xe3,0x5d,0x44,0xdd,0x11',
```
briansmith, line 1008: This is going to be extremely brittle.

There are, I guess, about 4*16*16*16 = 16,384 possible variants of just vpclmulqdq, so we don't want to try to enumerate every possible one.  Nor is anybody excited about writing an instruction encoder.

crypto/fipsmodule/aes/asm/aes-gcm-avx2-x86_64.pl
```diff
@@ -958,6 +958,68 @@ sub _aes_gcm_update {
...
+    );
+    for my $line (split("\n",$code)) {
+        my $trimmed;
+        $trimmed = $line;
+        $trimmed =~ s/^\s+//;
+        $trimmed =~ s/\s+(#.*)?$//;
+        if (exists $asmMap{$trimmed}) {
+            $line = $asmMap{$trimmed};
+        }
```
briansmith, line 1017: I think we should have an `else` here that does 'else if it is a vaesenc/vaesenclast/vpclmul instruction and the line mentioned "ymm", exit with an error`. This way we don't silently break this workaround.

We should have a comment about how to update this when it fails, something like:
```perl
      # run crypto/fipsmodule/aes/asm/aes-gcm-avx2-encoding.whatever and paste its output above.
```

Then in `crypto/fipsmodule/aes/asm/aes-gcm-avx2-encoding.(sh or py)` we would have the script that disassembles the .o file, filters the output to the relevant lines, and then transforms each line into the form used above, sorting the same way you sorted them, so we can copy/paste it into this function.

briansmith (changes requested):
First, thank you very much for doing this!

We will very soon land the avx512 version which will have the analogous issue. Are you planning to submit the analogous change for the avx512 version? Otherwise, this workaround would probably be short-lived.

arielb1:
> We will very soon land the avx512 version which will have the analogous issue. Are you planning to submit the analogous change for the avx512 version? Otherwise, this workaround would probably be short-lived.

I will definitely do this workaround for avx512 as well.

crypto/fipsmodule/aes/asm/aes-gcm-avx2-x86_64.pl
arielb1: On the other hand,
1. I believe that the code that calls `vpclmulqdq` changes fairly slowly, most patches don't touch it.
2. If it happens it's a fairly easy fix

If we ever have a problem with it changing too frequently, we can of course try to do something smarter.

briansmith:
```
$ find target/x86_64-unknown-linux-gnu/debug -name "*aes-gcm-avx2*.o" -exec objdump -S {} \; | grep -E "(vaesenc|vpclmul).*ymm" | grep c4 | cut -f2,3 | sort | uniq
```
^ This isn't quite right but only because I suck at shell scripting and I ran out of time. Unfortunately it generates output something like:
```
c4 e3 65 44 d5 10       vpclmullqhqdq %ymm5,%ymm3,%ymm2
```
which we'd need to transform into:
```
'vpclmulqdq $0x10, %ymm5, %ymm3, %ymm2' => '.byte 0xc4,0xe3,0x65,0x44,0xd5,0x10',
```

arielb1:
Yeah. Do you think it's important to do that?

briansmith:
> Yeah. Do you think it's important to do that?

Not sure what you mean. I think it's good to have a script that at least mostly automates dealing with future merges, even if there's a manual copy-paste step. Presumably if we do such a script for this AVX2 implementation, it will help automate the creation of the PR for the avx512 version too, as we'll be able to tweak it very quickly to adapt to avx512.

Sorry my shell scripting is terrible; I'm sure you probably have it already solved.

arielb1:
I mean, do you want to block this PR on me writing that script?

arielb1:
added that script

crypto/fipsmodule/aes/asm/aes-gcm-avx2-x86_64.pl
arielb1: added that script

crypto/fipsmodule/aes/asm/aes-gcm-avx2-x86_64.pl
```diff
@@ -958,6 +958,73 @@ sub _aes_gcm_update {
...
+        my $trimmed;
+        $trimmed = $line;
+        $trimmed =~ s/^\s+//;
+        $trimmed =~ s/\s+(#.*)?$//;
+        if (exists $asmMap{$trimmed}) {
+            $line = $asmMap{$trimmed};
+        } else {
+            if($trimmed =~ /(vpclmulqdq|vaes).*%ymm/) {
+                die ("found instruction not supported under old binutils, please update asmMap with the results of running\n" .
+                     'find target -name "*aes-gcm-avx2*.o" -exec python3 crypto/fipsmodule/aes/asm/make-avx-map-for-old-binutils.py \{\} \; | sort | uniq');
```
briansmith, line 1020: I suggest `find target -name "*aes-gcm-avx2*.o" -exec python3 crypto/fipsmodule/aes/asm/make-avx-map-for-old-binutils.py \{\} \; | LC_ALL=C sort | uniq`

Without `LC_ALL=C` my system sorts the lines in a different order (aesenc after aesenclast). With `L_ALL=C` I get the same output as what's in the source.

briansmith:
Thanks. This looks good to me, modulo the `LC_ALL=C` bit. I ran your script and pasted in the output (with `LC_ALL=C`) and verified that result was a no-op.

When we do the AVX-512 version, I think we should move `filter_and_print` to x86_64-xlate.pl so that it can be shared between all modules that will use these instructions without having to copy/paste it. You can do this now or we can do it during the avx512 work.

Could you please squash this?

arielb1:
> When we do the AVX-512 version, I think we should move filter_and_print to x86_64-xlate.pl so that it can be shared between all modules that will use these instructions without having to copy/paste it

How would you do that?

briansmith:
> > When we do the AVX-512 version, I think we should move filter_and_print to x86_64-xlate.pl so that it can be shared between all modules that will use these instructions without having to copy/paste it
> 
> How would you do that?

See this in the aes-gcm-avx2-x86_64.pl file:
``` 
$0 =~ m/(.*[\/\\])[^\/\\]+$/;
my $dir = $1;
my $xlate;
( $xlate = "${dir}x86_64-xlate.pl" and -f $xlate )
  or ( $xlate = "${dir}../../../perlasm/x86_64-xlate.pl" and -f $xlate )
  or die "can't locate x86_64-xlate.pl";

open OUT, "| \"$^X\" \"$xlate\" $flavour \"$output\"";
*STDOUT = *OUT;
```

My understanding is that whatever processing we put in `x86_64-xlate.pl` will get applied to all x86_64 files. You can see there is a function in `x86_64-xlate.pl` called `process_line` that seems to do the kind of line-by-line rewriting that we are doing here, for a variety of reasons.

arielb1:
Done. Put it in a separate file to make things cleaner.

crypto/fipsmodule/aes/asm/aes-gcm-avx2-x86_64.pl
```diff
@@ -57,7 +57,12 @@
   or ( $xlate = "${dir}../../../perlasm/x86_64-xlate.pl" and -f $xlate )
   or die "can't locate x86_64-xlate.pl";
 
-open OUT, "| \"$^X\" \"$xlate\" $flavour \"$output\"";
+my $xlate_binutils;
+( $xlate_binutils = "${dir}xlate-old-binutils.pl" and -f $xlate_binutils )
+  or ( $xlate_binutils = "${dir}../../../perlasm/xlate-old-binutils.pl" and -f $xlate_binutils )
+  or die "can't locate xlate-old-binutils.pl";
```
briansmith, line 63: I like your thinking here but this is likely to break some build systems I am aware of but am not allowed to explain to you because it adds a new file dependency. My suggestion is that we change back to the inline approach from the previous version and then develop a more general approach in the avx512 version where we'll have more time. I will do a release this morning if we can get this going. I will comment in the avx512 issue about the more general approach.

Thanks!

arielb1: k

arielb1:
Done. I'll rather brute force copy it over to AVX512 rather than play with the other file

arielb1:
In any case it's done for AVX-2

briansmith:
Thank you very much for contributing this, @arielb1!

briansmith:
I verified that the output of `objdump -d` on the object files, before and after this change, for windows and linux (but not macOS or other OS) is identical other than the symbol prefixing of "0_17_13" -> "0_17_14".

crypto/fipsmodule/aes/asm/aes-gcm-avx2-x86_64.pl
```diff
@@ -958,6 +958,73 @@ sub _aes_gcm_update {
 $code .= _aes_gcm_update 0;
 $code .= _end_func;
 
-print $code;
+sub filter_and_print {
+    # This function replaces AVX2 assembly instructions with their assembled forms,
```
ebiggers, line 962: Technically these are VAES and VPCLMULQDQ instructions, not AVX2.

briansmith: @ebiggers Thanks. I really appreciate the attention to detail. I clarified this in PR #2480.

PR #2476, opened 2025-03-12
have old binutils support on avx512 as well and improve it

arielb1:
Currently there are no ymm/zmm in the AVX512 code, but there will be.

I tested that when you fill out the map correctly, this code will work for AVX-512 as well, being able to run on AL2 and result in the same objdump if you use make-avx-map-for-old-binutils to generate an asmMap.

Made the perl code less sensitive to whitespace differences, I still think it's pretty understandable.

This currently does not change anything, not even the .S output [and therefore does not change the .o output or its objdump].


briansmith:
OK, I will review this probably next week. In the meantime, could you replace the "AVX 1.0" references in the PR title/summary and in the patch itself (if any) with "AVX10"?

Also, could you do a diff between aes-gcm-avx2-x86_64.pl an aes-gcm-avx10-x86_64.pl to ensure that the perl code you added to both matches as closely as possible (no superfluous differences)?

briansmith:
Also, thank you!

arielb1:
> Also, could you do a diff between aes-gcm-avx2-x86_64.pl an aes-gcm-avx10-x86_64.pl to ensure that the perl code you added to both matches as closely as possible (no superfluous differences)?

I copy pasted it

briansmith:
@arielb1 Could you please update this PR after the latest changes to `aes-gcm-avx512-x86_64.pl` (renamed from `*-avx10-*`)? Sorry about the hassle.

arielb1:
Updated.

crypto/fipsmodule/aes/asm/aes-gcm-avx512-x86_64.pl
```diff
@@ -1268,6 +1268,42 @@ sub _aes_gcm_update {
...
+            my $trimmed = $+{instruction} . " " . $+{args};
+            my $prespace = $+{prespace};
+            my $postspace = $+{postspace};
+            if (exists $asmMap{$trimmed}) {
+                $line = ${prespace} . $asmMap{$trimmed} . ${postspace};
+            } else {
+                if($trimmed =~ /(vpclmulqdq|vaes).*%[yz]mm/) {
+                    die ("found instruction not supported under old binutils, please update asmMap with the results of running\n" .
+                         'find target -name "*aes-gcm-avx512*.o" -exec python3 crypto/fipsmodule/aes/asm/make-avx-map-for-old-binutils.py \{\} \; | LC_ALL=C sort | uniq');
+                }
```
briansmith, line 1299: I have found that this works very poorly when hacking on this file. Here's what happens:
* I do something that triggers this check, so the build fails.
* I run the suggested command, but it will only ever find object files that were successfully built, and those successfully-built object files don't use the new forms of the instructions!
* I have to comment out this check.

Also, the `find target` has another downside: It will find every object file that's ever been built. Basically, we'll end up with the union of all forms of all these instructions that have ever been needed unless/until `cargo clean` is done. 

There are ways to hack around it but it is ugly.

I think we should move this check to CI, in a separate script. For example, in the package job, we could look at the files generated into `pregenerated` to find lines that look like VAES/VPCLMULQDQ in the .S files. We could do this in Python instead of Perl. This would solve 

The other thing is that this is done for nasm output but we really don't need to do it for nasm. There are important reasons related to Windows ABI stuff where it important that we keep the nasm output readable. So we should avoid doing this rewriting when the output format is nasm. Note that looking only at .S files is the right thing to do, if we're avoiding rewriting the nasm output.

briansmith (changes requested):

briansmith (approved):

briansmith:
Thanks for doing this! I really appreciate the effort you put into this.

PR #2482, opened 2025-03-14
cpu: Replace use of `OnceNonZeroUsize::get()` with `get_unchecked()`.

briansmith:


briansmith:
TODO:
- [ ] Check the difference in the output assembly and update the commit message with it.
- [ ] Check if this obviates the need for the `Values` construct and remove `Values` if so.

briansmith:
Done in #2499

PR #2496, opened 2025-03-24
Don't rebuild when various Cargo-set environment variables like `CARGO_MANIFEST_DIR` change.

### LINKED ISSUE #2454: ring's build script shouldn't output rerun-if-env-changed for Cargo-related env vars
(Originally filed as https://github.com/ctz/ring/issues/10, and ported here.)

Hi there, and thank you so much for maintaining/remaintaining ring!

Currently, ring's build script outputs a number of rerun-if-env-changed instructions ([full output](https://gist.github.com/sunshowers/d02774a6412bb8d588b77431de687b39)):

```
cargo:rerun-if-env-changed=CARGO_MANIFEST_DIR
cargo:rerun-if-env-changed=CARGO_PKG_NAME
cargo:rerun-if-env-changed=CARGO_PKG_VERSION_MAJOR
cargo:rerun-if-env-changed=CARGO_PKG_VERSION_MINOR
cargo:rerun-if-env-changed=CARGO_PKG_VERSION_PATCH
cargo:rerun-if-env-changed=CARGO_PKG_VERSION_PRE
cargo:rerun-if-env-changed=CARGO_MANIFEST_LINKS
cargo:rerun-if-env-changed=RING_PREGENERATE_ASM
cargo:rerun-if-env-changed=OUT_DIR
cargo:rerun-if-env-changed=CARGO_CFG_TARGET_ARCH
cargo:rerun-if-env-changed=CARGO_CFG_TARGET_OS
cargo:rerun-if-env-changed=CARGO_CFG_TARGET_ENV
cargo:rerun-if-env-changed=CARGO_CFG_TARGET_ENDIAN
cargo:rerun-if-env-changed=CC_x86_64-unknown-linux-gnu
cargo:rerun-if-env-changed=CC_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=HOST_CC
cargo:rerun-if-env-changed=CC
cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
cargo:rerun-if-env-changed=CFLAGS
cargo:rerun-if-env-changed=HOST_CFLAGS
cargo:rerun-if-env-changed=CFLAGS_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=CFLAGS_x86_64-unknown-linux-gnu
cargo:rerun-if-env-changed=CC_x86_64-unknown-linux-gnu
cargo:rerun-if-env-changed=CC_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=HOST_CC
cargo:rerun-if-env-changed=CC
cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
cargo:rerun-if-env-changed=CFLAGS
cargo:rerun-if-env-changed=HOST_CFLAGS
cargo:rerun-if-env-changed=CFLAGS_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=CFLAGS_x86_64-unknown-linux-gnu
```

This causes a number of extra recompiles if a ring build is invoked via a Cargo xtask or other wrapper. For more details, see https://github.com/oxidecomputer/omicron/pull/7740.

While many of these are appropriate, the ones prefixed by CARGO_ should probably not be output, I believe. Cargo's documentation says in [this section](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed):

> Note that the environment variables here are intended for global environment variables like CC and such, it is not possible to use this for environment variables like TARGET that [Cargo sets for build scripts](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts). The environment variables in use are those received by cargo invocations, not those received by the executable of the build script.

I think an easy fix here is to ensure that no environment variables prefixed by CARGO_ are produced. That would correctly filter out the current list, and also likely be correct if and when the build script changes.

janlugt:
`CARGO_MANIFEST_DIR` should be allowed to change without triggering a rebuild. Related to #2454.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -30,18 +30,95 @@ use std::{
 mod env {
     use std::ffi::OsString;
 
-    /// Read an environment variable and tell Cargo that we depend on it.
-    ///
-    /// The name is static since we intend to only read a static set of environment
-    /// variables.
-    pub fn var_os(name: &'static str) -> Option<OsString> {
-        println!("cargo:rerun-if-env-changed={}", name);
-        std::env::var_os(name)
+    enum EnvVarTy {
+        RerunIfChanged,
+        SetByCargo,
     }
 
-    pub fn var(name: &'static str) -> Option<String> {
-        var_os(name).and_then(|value| value.into_string().ok())
+    pub struct EnvVar {
+        pub name: &'static str,
+        ty: EnvVarTy,
     }
+
+    /// Read an environment variable and optionally tell Cargo that we depend on it.
+    ///
+    /// The env var is static since we intend to only read a static set of environment
+    /// variables.
+    pub fn var_os(env_var: &'static EnvVar) -> Option<OsString> {
+        match env_var.ty {
+            EnvVarTy::RerunIfChanged => {
+                println!("cargo:rerun-if-env-changed={}", env_var.name);
+            }
+            EnvVarTy::SetByCargo => {}
+        }
+        std::env::var_os(env_var.name)
+    }
+
+    pub fn var(env_var: &'static EnvVar) -> Option<String> {
+        var_os(env_var).and_then(|value| value.into_string().ok())
+    }
+
+    // In alphabetical order
+    pub const CARGO_CFG_TARGET_ARCH: EnvVar = EnvVar {
+        name: "CARGO_CFG_TARGET_ARCH",
+        ty: EnvVarTy::RerunIfChanged,
+    };
+    pub const CARGO_CFG_TARGET_ENDIAN: EnvVar = EnvVar {
+        name: "CARGO_CFG_TARGET_ENDIAN",
+        ty: EnvVarTy::RerunIfChanged,
+    };
+    pub const CARGO_CFG_TARGET_ENV: EnvVar = EnvVar {
+        name: "CARGO_CFG_TARGET_ENV",
+        ty: EnvVarTy::RerunIfChanged,
+    };
+    pub const CARGO_CFG_TARGET_OS: EnvVar = EnvVar {
+        name: "CARGO_CFG_TARGET_OS",
+        ty: EnvVarTy::RerunIfChanged,
+    };
+    pub const CARGO_MANIFEST_DIR: EnvVar = EnvVar {
+        name: "CARGO_MANIFEST_DIR",
+        ty: EnvVarTy::SetByCargo,
+    };
+    pub const CARGO_MANIFEST_LINKS: EnvVar = EnvVar {
+        name: "CARGO_MANIFEST_LINKS",
+        ty: EnvVarTy::RerunIfChanged,
+    };
+    pub const CARGO_PKG_NAME: EnvVar = EnvVar {
+        name: "CARGO_PKG_NAME",
+        ty: EnvVarTy::SetByCargo,
+    };
+    pub const CARGO_PKG_VERSION_MAJOR: EnvVar = EnvVar {
+        name: "CARGO_PKG_VERSION_MAJOR",
+        ty: EnvVarTy::SetByCargo,
+    };
+    pub const CARGO_PKG_VERSION_MINOR: EnvVar = EnvVar {
+        name: "CARGO_PKG_VERSION_MINOR",
+        ty: EnvVarTy::SetByCargo,
+    };
+    pub const CARGO_PKG_VERSION_PATCH: EnvVar = EnvVar {
+        name: "CARGO_PKG_VERSION_PATCH",
+        ty: EnvVarTy::SetByCargo,
+    };
+    pub const CARGO_PKG_VERSION_PRE: EnvVar = EnvVar {
+        name: "CARGO_PKG_VERSION_PRE",
+        ty: EnvVarTy::SetByCargo,
+    };
+    pub const DEBUG: EnvVar = EnvVar {
+        name: "DEBUG",
+        ty: EnvVarTy::RerunIfChanged,
+    };
+    pub const OUT_DIR: EnvVar = EnvVar {
+        name: "OUT_DIR",
+        ty: EnvVarTy::SetByCargo,
+    };
+    pub const PERL_EXECUTABLE: EnvVar = EnvVar {
+        name: "PERL_EXECUTABLE",
+        ty: EnvVarTy::RerunIfChanged,
+    };
+    pub const RING_PREGENERATE_ASM: EnvVar = EnvVar {
+        name: "RING_PREGENERATE_ASM",
+        ty: EnvVarTy::RerunIfChanged,
+    };
 }
 
 const X86: &str = "x86";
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -269,7 +269,8 @@ fn main() {
     // Avoid assuming the working directory is the same is the $CARGO_MANIFEST_DIR so that toolchains
     // which may assume other working directories can still build this code.
     let c_root_dir = PathBuf::from(
-        env::var_os("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR should always be set"),
+        // Don't emit `cargo:rerun-if-env-changed` for $CARGO_MANIFEST_DIR.
+        std::env::var_os("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR should always be set"),
```
briansmith, line 273: I suggest that instead of doing this, we change the implementation of the local `env::var_os`/`env::var` to take an `&'static EnvVar` instead of a `&str` for the environment variable name:

```rust
struct EnvVar {
    name: &'static str,
    ty: EnvVarType
}
enum EnvVarType {
    RerunIfChanged,
    CargoSetEnvVar,
}
const CARGO_MANIFEST_DIR: EnvVarName = EnvVarName {
    name: "CARGO_MANIFEST_DIR",
    ty: CargoSetEnvVar,
};
...
```

Otherwise, we'd have to repeat the pattern you have above for every Cargo-set environment variable, apparently.

briansmith (changes requested):

build.rs
janlugt: Thanks for the review! I've updated the PR, I hope I've understood your comment correctly.

build.rs
```diff
@@ -30,18 +30,91 @@ use std::{
...
+        std::env::var_os(env_var.name)
     }
 
-    pub fn var(name: &'static str) -> Option<String> {
-        var_os(name).and_then(|value| value.into_string().ok())
+    pub fn var(env_var: &'static EnvVar) -> Option<String> {
+        var_os(env_var).and_then(|value| value.into_string().ok())
     }
+
+    pub const CARGO_MANIFEST_DIR: EnvVar = EnvVar {
```
briansmith, line 58: I guess we need to decide some kind of order for these. I suggest alphabetical order. Could you alphabetize them and then add a comment above the first one `// In alphabetical order` or similar?

build.rs
```diff
@@ -30,18 +30,91 @@ use std::{
...
+    /// Read an environment variable and optionally tell Cargo that we depend on it.
     ///
-    /// The name is static since we intend to only read a static set of environment
+    /// The env var is static since we intend to only read a static set of environment
     /// variables.
-    pub fn var_os(name: &'static str) -> Option<OsString> {
-        println!("cargo:rerun-if-env-changed={}", name);
-        std::env::var_os(name)
+    pub fn var_os(env_var: &'static EnvVar) -> Option<OsString> {
+        if let EnvVarTy::SetByCargo = env_var.ty {
```
briansmith, line 48: 1. In this codebase we use `match` instead of `if let` for matching enums other than `Option`.
2. Isn't this the opposite of what you want? I think if the env var. is set by cargo, then we DON'T want to print "cargo::rerun-if-env-changed=...", right?

briansmith (changes requested):
Thanks for doing this!

briansmith:
Also, if/when you update this, please squash the changes into a single commit (I use use `git rebase -i main` for this.)

build.rs
janlugt: Done.

build.rs
janlugt: 1. Fixed.
2. Thanks for catching that! It was the first thing I did this morning, should've waited until I finished my coffee ;-)

build.rs
```diff
@@ -269,51 +346,50 @@ fn main() {
...
-    const RING_PREGENERATE_ASM: &str = "RING_PREGENERATE_ASM";
-    match env::var_os(RING_PREGENERATE_ASM).as_deref() {
+    match env::var_os(&env::RING_PREGENERATE_ASM).as_deref() {
         Some(s) if s == "1" => {
             pregenerate_asm_main(&c_root_dir, &core_name_and_version);
         }
         None => ring_build_rs_main(&c_root_dir, &core_name_and_version),
         _ => {
-            panic!("${} has an invalid value", RING_PREGENERATE_ASM);
+            panic!("RING_PREGENERATE_ASM has an invalid value",);
```
briansmith, line 374: NIT: I don't think it makes sense to keep the trailing comma.

But, I think you can instead use `RING_PREGENERATE_ASM.name` as the parameter instead, to DRY this.

briansmith:
NOTE: It is OK tha tthe codecov job is failing. It just doesn't like that the coverage is very slightly lower than it used to be, due to other changes.

build.rs
janlugt: Fixed.

mcronce:
@briansmith asked me to test with this PR and I had similar results as reported in the issue thread - I had to switch `CARGO_PKG_NAME`, `CARGO_PKG_VERSION_MAJOR`, `CARGO_PKG_VERSION_MINOR`, `CARGO_PKG_VERSION_PATCH`, `CARGO_PKG_VERSION_PRE`, and `OUT_DIR` to be `EnvVarTy::SetByCargo` to avoid a rebuild

```diff
diff --git a/build.rs b/build.rs
index e26b0409b..579d9dbdb 100644
--- a/build.rs
+++ b/build.rs
@@ -85,23 +85,23 @@ mod env {
     };
     pub const CARGO_PKG_NAME: EnvVar = EnvVar {
         name: "CARGO_PKG_NAME",
-        ty: EnvVarTy::RerunIfChanged,
+        ty: EnvVarTy::SetByCargo,
     };
     pub const CARGO_PKG_VERSION_MAJOR: EnvVar = EnvVar {
         name: "CARGO_PKG_VERSION_MAJOR",
-        ty: EnvVarTy::RerunIfChanged,
+        ty: EnvVarTy::SetByCargo,
     };
     pub const CARGO_PKG_VERSION_MINOR: EnvVar = EnvVar {
         name: "CARGO_PKG_VERSION_MINOR",
-        ty: EnvVarTy::RerunIfChanged,
+        ty: EnvVarTy::SetByCargo,
     };
     pub const CARGO_PKG_VERSION_PATCH: EnvVar = EnvVar {
         name: "CARGO_PKG_VERSION_PATCH",
-        ty: EnvVarTy::RerunIfChanged,
+        ty: EnvVarTy::SetByCargo,
     };
     pub const CARGO_PKG_VERSION_PRE: EnvVar = EnvVar {
         name: "CARGO_PKG_VERSION_PRE",
-        ty: EnvVarTy::RerunIfChanged,
+        ty: EnvVarTy::SetByCargo,
     };
     pub const DEBUG: EnvVar = EnvVar {
         name: "DEBUG",
@@ -109,7 +109,7 @@ mod env {
     };
     pub const OUT_DIR: EnvVar = EnvVar {
         name: "OUT_DIR",
-        ty: EnvVarTy::RerunIfChanged,
+        ty: EnvVarTy::SetByCargo,
     };
     pub const PERL_EXECUTABLE: EnvVar = EnvVar {
         name: "PERL_EXECUTABLE",
```

janlugt:
I've incorporated your changes into the PR.

janlugt:
Can this be merged now that @jhillyerd tested this for the MSRV?

briansmith (approved):

briansmith:
I used a merge commit to merge this so that I could tweak the commit message that shows up on the main branch, instead of going through a new review cycle to update the commit message.

I created follow-up PR #2506 to DRY the names so that we can be more certain that there were no typos that would cause the variable names to be different than expected.

briansmith:
Please test the follow-up in PR #2507 that marks the remaining Cargo-set environment variables as set by Cargo.

georglauterbach:
@briansmith thank you very much! Do you have an estimate of when these changes will be available via a new release?

PR #2508, opened 2025-04-10
CI: Add `overflow-checks = true` to release and bench profiles.

briansmith:
I hope that this allows code coverage measurement to find cases where there is a potential overflow panic, as the branch should show up as uncovered.


briansmith:
> I hope that this allows code coverage measurement to find cases where there is a potential overflow panic, as the branch should show up as uncovered.

Nope, it doesn't affect it.

PR #2518, opened 2025-04-13
cpu: intel: correct assumptions for x32 abi

jpalus:
Fixes build for rust's Tier 2 platform `x86_64-unknown-linux-gnux32` ([x32 ABI](https://en.wikipedia.org/wiki/X32_ABI))

### DIFF
--- a/src/cpu/intel.rs
+++ b/src/cpu/intel.rs
@@ -25,9 +25,9 @@ mod abi_assumptions {
     const _ASSUMES_SSE2: () =
         assert!(cfg!(target_feature = "sse") && cfg!(target_feature = "sse2"));
 
-    #[cfg(target_arch = "x86_64")]
+    #[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
     const _ASSUMED_POINTER_SIZE: usize = 8;
-    #[cfg(target_arch = "x86")]
+    #[cfg(any(target_arch = "x86", all(target_arch = "x86_64", target_pointer_width = "32")))]
     const _ASSUMED_POINTER_SIZE: usize = 4;
     const _ASSUMED_USIZE_SIZE: () = assert!(size_of::<usize>() == _ASSUMED_POINTER_SIZE);
     const _ASSUMED_REF_SIZE: () = assert!(size_of::<&'static u8>() == _ASSUMED_POINTER_SIZE);


src/cpu/intel.rs
```diff
@@ -25,9 +25,9 @@ mod abi_assumptions {
     const _ASSUMES_SSE2: () =
         assert!(cfg!(target_feature = "sse") && cfg!(target_feature = "sse2"));
 
-    #[cfg(target_arch = "x86_64")]
+    #[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
     const _ASSUMED_POINTER_SIZE: usize = 8;
-    #[cfg(target_arch = "x86")]
+    #[cfg(any(target_arch = "x86", all(target_arch = "x86_64", target_pointer_width = "32")))]
```
briansmith, line 30: These assertions are here because we don't know if our code works with the x32 ABI.

At a minimum, we'd also need to duplicate all the x86-64 `coverage` tests in ci.yml for this new ABI, in addition to doing this.

This is because we're unsure whether *all* the assembly code correctly supports 32-bit `size_t` and 32-bit pointers.

jpalus:
> The last I heard, nobody is using the x32 ABI in practice. Do we really need to support it?

We do do maintain x32 port at PLD Linux, so does Debian so it would be nice if `ring` built fine for this target. I'm not really familiar with the project, I'm only here because of `maturin` build failure (through `rustls` dependency). If there is any assembly involved I don't think it's worth spending time on supporting x32 in there. Is there some generic code path not involving assembly that could be taken for x32?

briansmith:
Either way, to support this target we'd need the coverage tests to be expanded to run for the x32 target too. I think that's the first step here. Especially, I am curious if GitHub Actions even supports an operating system that can run x32 binaries. You can try it out by copy/pasting all the x86_64-* stuff in the `coverage` job in ci.yml to create duplicate jobs for the gnux32 target.

src/cpu/intel.rs
briansmith, line 30: I suggest we also require target_os = "linux" for x32 support. For example, we don't want to have to deal with the combination of Windows and x32 or FreeBSD and x32, etc.

briansmith:
> If there is any assembly involved I don't think it's worth spending time on supporting x32 in there. 

I didn't look at the details of the x32 ABI. If the ABI is defined such that pointers and size_t values are passed in 64-registers, zero-extended, then it can work fine without changing the assembly. My understanding is that x32 does zero-extend pointers passed in registers. If it doesn't zero-extend size_t values passed in registers then we'd need to audit our assembly code to verify that it always assumes zeroes the high half of every size/length parameter declared as `usize`/`size_t`/`NonZero_size_t` and/or that the size/length parameter is typed `u64`.

> Is there some generic code path not involving assembly that could be taken for x32?

I don't think it is a good strategy for take the generic code path for x32 ABI. One thing we could do is change the CPU feature detection logic so that for the x32 ABI, no CPU features are detected. This will limit the amount of assembly code that would be used for the x32 ABI. But we'd still need to review `bn_mul_mont` and the SHA-256/SHA-512 assembly code (at least) that is used even when no CPU features are detected.

briansmith:
I took a step towards this in PR #2560.

PR #2521, opened 2025-04-29
Use `&Digest` in ECDSA to avoid Copy in `sign_digest`'s `100`-iteration loop

dare3path:
The primary goal is to prevent implicit copies of
`Digest` (which implements `Copy`) in `sign_digest`’s 100-iteration loop,
which could copy ~72 bytes per iteration (up to 7,200 bytes total, but this is worse case scenario, generally this loop iterates only once as per the correction in the [comment](https://github.com/briansmith/ring/pull/2521#issuecomment-2843669286) below - thanks). So this only saves 64 bytes instead of what I initially thought (6400 bytes) due to my superficially looking at the loop.

Modified `digest_scalar`, `sign_digest`, and `verification` to pass `digest::Digest` by reference (`&Digest`) instead of by value, preventing implicit copies via `Digest`'s `Copy` trait in `sign_digest`'s `100`-iteration loop.

<del>Using `&h` in `digest_scalar(n, &h)` ensures clarity by explicitly indicating
reference passing</del>(undone this in [3rd](https://github.com/briansmith/ring/pull/2521/commits/8a1e442e58a9d44421be6386b5adbf991da50d29) commit). There is no runtime performance impact, as Rust’s reference
coercion optimizes `&h` (or ~~even multiple `&` operators like `&&&&&&&&h`~~) to a
single `&Digest` pointer at compile time, producing identical machine code to
passing `h` directly. This was verified via Rust’s type system ~~but NOT verified via assembly
inspection - I just trust that Grok 3 isn't wrong about this as it would make sense to be so.~~ **EDIT:** `&h` is optimized (as [seen](https://godbolt.org/z/hG4oabhj1) in the assembly), but multiple `&` operators (e.g., `&&h` or `&&&&&&&&h`) trigger runtime dereferencing, which I missed. Fortunately it doesn't affect current PR then.

Passes:
- [x] `cargo test`
- [x] `cargo test --all-targets`
- [x] `cargo test --all-features`
- [x] `cargo test --all-targets --all-features`


briansmith:
The loop will only execute one iteration generally. Are you saying that in the unusual case where more than one iteration is needed, there is some memcpy that this PR eliminates?

dare3path:
You're right, I haven't actually looked at the loop at all(to realize it's generally only executing once), the PR is just a side-effect of me implementing `ZeroizeOnDrop`(manually, like [here](https://github.com/dare3path/ring/tree/zeroize_on_drop)), locally for my own needs, and therefore Digest having implemented `Drop` and thus not being `Copy` anymore, I've had to use a reference at step 5 there (at least) and superficially looking at the loop I thought it would be nice for this to be a ref even when it is `Copy`. My bad, thank you for pointing this out and apologies for wasting your time with this. For just one iteration this is PR is obviously not necessary.

dare3path:
oh sorry, to answer your question, originally since `h` is passed by value at step 5, and it being `Copy`, it means it gets copied once, in the case when the loop executes only once. Using `&` would mean it only gets 8 bytes copied instead of the whole thing that `h` is (72 bytes apparently, I could be wrong).

src/ec/suite_b/ecdsa/signing.rs
```diff
@@ -201,14 +201,14 @@ impl EcdsaKeyPair {
...
 
     /// Returns the signature of message digest `h` using a "random" nonce
     /// generated by `rng`.
     fn sign_digest(
         &self,
-        h: digest::Digest,
+        h: &digest::Digest,
         rng: &dyn rand::SecureRandom,
         cpu: cpu::Features,
     ) -> Result<signature::Signature, error::Unspecified> {
```
briansmith, line 214: Why not just `let h = &*h;` here? Wouldn't this have the intended effect?

dare3path: do you mean do that `let h=&h;` in order to keep the arg without the `&` ?

src/ec/suite_b/ecdsa/verification.rs
```diff
@@ -65,7 +65,7 @@ impl signature::VerificationAlgorithm for EcdsaVerificationAlgorithm {
             // NSA Guide Step 3: "Convert the bit string H to an integer e as
             // described in Appendix B.2."
             let n = &self.ops.scalar_ops.scalar_modulus(cpu);
-            digest_scalar(n, h)
+            digest_scalar(n, &h)
```
dare3path, line 68: here, we can use either `h` or `&h` to get the same `asm`, if the arg is `h: &digest::Digest,` - but i kept it `&h` since it's no cost, in order to make it obvious that that `digest_scalar` is taking a reference there.

dare3path: oh sorry, wrong line, not here, where was it...

dare3path: at it was:
```rust

            // Step 5.
            let e = digest_scalar(n, h);
```
which is above here: https://github.com/briansmith/ring/pull/2521/files/8a1e442e58a9d44421be6386b5adbf991da50d29#diff-ab4f51b624f3e5d349af84abdcb5a848d975b71ce35bf85db234ba0e211bd23cR268
but I undid that and used `h` instead, I forgot.

src/ec/suite_b/ecdsa/signing.rs
briansmith: > But yeah if you want to keep their sig, sure, but since Digest is Copy, it won't move the value into the function, it will just pass a copy of it (was it 72 bytes instead of 8 when it's a ref) to the sign_digest.

OK, thanks. I guess then this PR does make sense. If. you want to re-open it, I'll have it run through CI and then I can merge it.

dare3path: sure, if you wish.

briansmith:
Thank you!

dare3path:
Glad it's useful, thank you!

PR #2522, opened 2025-04-30
build.rs: Check symbol prefixes in archive.

briansmith:
Check symbol prefixes on every target.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -215,61 +215,51 @@ const ASM_TARGETS: &[AsmTarget] = &[
         oss: LINUX_ABI,
         arch: AARCH64,
         perlasm_format: "linux64",
-        extern_prefix: "",
     },
     AsmTarget {
         oss: LINUX_ABI,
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -1082,3 +1125,44 @@ fn prefix_all_symbols(pp: char, prefix_prefix: &str, prefix: &str) -> String {
...
+    } else if symbol.name().starts_with(b"__covrec_") {
+        // LLVM code coverage adds these symbols.
+        // TODO: Will these be unique?
+        eprintln!(
+            "{path_str}: symbol not prefixed as expected but temporarily allowed: {name_approx}"
+        );
+        true
+    } else if target.os == WINDOWS
+        && (symbol.name().starts_with(b"__xmm@") || symbol.name().starts_with(b"__real@"))
+    {
```
briansmith, line 1156: Maybe we should also check `target.env == MSVC`.

PR #2529, opened 2025-05-20
CI: Build aarch64-pc-windows-msvc on ARM Windows 11, and run tests.

briansmith:


briansmith:
I guess we should try to add a coverage test for this too.

PR #2530, opened 2025-05-20
Remove `lto=true` in Cargo.toml profiles.

briansmith:
Rust 1.87 seems to miscompile the tests and/or *ring* itself when LTO is enabled. See https://github.com/rust-lang/rust/issues/141306.


briansmith:
There are multiple compatibility-breaking changes in Rust 1.87 and/or nightly. This addresses just one of them; the other CI failures are due to other compatibility-breaking changes that will be dealt with in separate PRs.

PR #2531, opened 2025-05-20
CI: Only measure coverage for tests.

briansmith:
Don't measure coverage for doctests, bins, benchmarks, etc.

This works around a failure to build doctests in the coverage jobs which is probably caused by a compatibility-breaking change in Rust's new doctest-xcompile feature and/or its doctest bundling feature.


briansmith:
The test-features failure will be fixed by PR #2532.

PR #2532, opened 2025-05-20
CI: Don't build/run doctests in test-features jobs.

briansmith:
The doctests are built/run in separate test-doc jobs so we don't need to run them in test-features.

This will speed up CI slightly.

This will also work around some incompatibility introduced in recent Rust 1.87 and/or Nightly releases where linking the doctests fails when cross-compiling. (Rust recently started building doctests when cross-compiling where previously it ignored them during cross- compilation.)


briansmith:
The `coverage` job failures were addressed in PR #2531.

PR #2537, opened 2025-05-22
aes: Make internal AES key creation infallible.

briansmith:
Once we've constructed a `KeyBytes` we know the input is valid.


briansmith:
Failed "(aarch64-pc-windows-msvc, --release, 1.66.0)" test, which caused a bunch of other `test` jobs to be skipped, is a known issue that will be dealt with separately.

PR #2546, opened 2025-05-29
aes fallback: Don't inline `sub_bytes` or `shift_rows`.

briansmith:
Help the optimizer optimize a bit. When this was in C, Clang wasn't Inlining them either (on x86 and arm, at least).

### DIFF
--- a/src/aead/aes/fallback.rs
+++ b/src/aead/aes/fallback.rs
@@ -316,6 +316,7 @@ impl Batch {
 
 // AES round steps.
 impl Batch {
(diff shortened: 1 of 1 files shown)


briansmith:
With this change, the armv7 no-neon coverage job finished in 10m30s, whereas previously the job ran in 11m9s, and the last run before the C->Rust conversion was 10m0s; the i686-linux coreduo-v1 coverage job finished in 30m 26s with this change, where previously it was 30m 25s.

PR #2570, opened 2025-06-11
cpu/aarch64/windows: Use libstd instead of windows_sys.

briansmith:
Eliminate the windows-sys dependency on Aarch64. We believe there is no use for `no_std` support for this target.


briansmith:
Does `IsProcessorFeaturePresent` report AES features being available in an enclave?

I wonder if there is some way we could run the tests in an enclave in CI?

complexspaces:
I don't have an enclave dev environment set up at this moment, sorry, but I can get back to you later on checking if `IsProcessorFeaturePresent` functions correctly in the enclave.

> I wonder if there is some way we could run the tests in an enclave in CI?

I don't believe it'd be possible today for several reasons, at least on GitHub-hosted CI. It requires VBS features to be enabled (which means the Windows host for the test code to run on needs to be able to do virtualization) and it needs a bleeding edge Windows 11 release. Everything else seems workable long-term but I think the nested virtualization would be an issue.

PR #2580, opened 2025-06-16
Fix Comments and Minor Code Consistency in ECDSA and P256 ASM

leopardracer:
**Description:**  
- Corrected a comment in the ECDSA verification test to accurately describe the modulus operation.
- Improved code consistency in the P256 x86-64 assembly by updating a comment and ensuring clarity.


briansmith:
Thanks for the contribution. It is unfortunate that these typos are here. It would be better to fix them upstream (BoringSSL at least, if not in OpenSSL) and then we can merge the changes from upstream. That way it will be easier to deal with any licensing issues that we inherit from upstream in the future.

PR #2589, opened 2025-06-27
Use `let <pat> = ... else ...` to clarify some bits of code

briansmith:


briansmith:
The problem here is that every `let ... else` that has an unreachable `else` now results in a partial code coverage measurement for the `let ... else` line, which can be confusing if the expression in the `let` is non-trivial:
image

briansmith:
> The problem here is that every `let ... else` that has an unreachable `else` now results in a partial code coverage measurement for the `let ... else` line, which can be confusing if the expression in the `let` is non-trivial:
> image

Reported as https://github.com/rust-lang/rust/issues/143149.

PR #2593, opened 2025-07-02
digest/hmac: Make `Context` types a bit smaller.

briansmith:
Reduce the size of each `Context` type by 8 bytes.

Also, eliminate a redundant store of `num_pending` that was found as a side effect.


src/digest.rs
```diff
@@ -185,71 +187,69 @@ impl FinishError {
...
                 }
             };
-            sliceutil::overwrite_at_start(buffer_to_fill, data);
-            match slice::split_at_checked(data, buffer_to_fill.len()) {
-                Some((just_copied, to_digest)) => {
-                    debug_assert_eq!(buffer_to_fill.len(), just_copied.len());
-                    debug_assert_eq!(self.num_pending + just_copied.len(), block_len);
-                    let leftover = self.block.update(buffer, cpu_features);
-                    debug_assert_eq!(leftover.len(), 0);
-                    self.num_pending = 0;
```
briansmith, line 237: This is the redundant store of `self.num_pending` that was eliminated.

PR #2673, opened 2025-08-03
Windows arm64: enable building with MSVC

dennisameling:
Closes https://github.com/briansmith/ring/issues/2674
Closes https://github.com/briansmith/ring/issues/2215

Currently, a [workaround](https://github.com/briansmith/ring/blob/1f96c99015fefd7c03db68ae2416660e35ab1e63/build.rs#L596) is in place to always build with Visual Studio's `clang.exe` instead of MSVC.

This is causing [problems](https://github.com/SeleniumHQ/selenium/actions/runs/16576636789/job/46883267219?pr=16046#step:18:291) in some cases, because the `aarch64-pc-windows-msvc` target typically [only uses MSVC-provided tooling](https://doc.rust-lang.org/nightly/rustc/platform-support/windows-msvc.html#host-tooling) and has no hard requirement on `clang.exe` at all. It gets even trickier when it tries to pass MSVC flags like `/nologo`, `/DCOMPILER_MSVC`, etc. to `clang.exe`, which it doesn't support. Sure, `clang-cl.exe` _might_ work, but that can come with its own set of challenges.

I learned that the `ring` library basically has [two ways of using it](https://github.com/briansmith/ring/blob/1f96c99015fefd7c03db68ae2416660e35ab1e63/BUILDING.md):
1. **Using packaged builds**, where object files are pre-compiled using Perl, NASM, etc. - the user doesn't need to have these tools installed locally.
2. **Building from Git**: the mentioned tools (Perl, NASM, etc.) need to run on the user's machine.

I was thinking:

> What if we can use Visual Studio's `clang.exe` to pre-compile the object files, and then let users build the rest of the project with MSVC, so that it cleanly follows the `aarch64-pc-windows-msvc` target structure again?

This PR implements exactly that. With some minor changes, I was able to get things to compile using MSVC, using pre-compiled object files generated with Visual Studio's `clang.exe`.

CC @awakecoding @Alovchin91

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -24,27 +24,14 @@ version used by default in GitHub Actions is the one that is primarily
 supported.
 
 For Windows ARM64 targets (aarch64-pc-windows-msvc), the Visual Studio Build
-Tools “VS 2022 C++ ARM64 build tools” and "clang" components must be installed.
-Add Microsoft's provided version of `clang` to `%PATH%`, which will allow the
-build to work in GitHub Actions without installing anything:
-```
-$env:Path += ";C:\Program Files (x86)\Microsoft Visual Studio\2022\Enterprise\VC\Tools\Llvm\x64\bin"
-```
-If you (locally) have “Build Tools for Visual Studio 2022” instead, use:
-```
-$env:Path += ";C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\Llvm\x64\bin"
-```
-Alternatively, if the host machine is already a Windows ARM64 then use:
-```
-$env:Path += ";C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\ARM64\bin"
-```
+Tools “VS 2022 C++ ARM64 build tools” must be installed.
 
 Packaged Builds
 ---------------
 
 The *ring* crate released on crates.io needs fewer tools than are required
 for building *ring* from Git. Perl isn't required because the output of
-the Perl build steps is packaged into the crate. For Windows x86 and x86-64,
+the Perl build steps is packaged into the crate. For Windows x86, x86-64 and arm64,
 the packaged crate contains precompiled object files for the assembly
 language modules so no macro assembler. 
 
@@ -63,12 +50,11 @@ apply when building from crates.io:
   is used as the assembler. The version to use, how to download it, and where
   to put it, is all documented in [mk/install-build-tools.ps1](mk/install-build-tools.ps1).
 
+* For Windows arm64 only, Visual Studio's `clang.exe` (which can be installed as a component in Visual Studio 2019 and later) is used as the assembler. Installation instructions can be found [here](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170).
+
 For Windows hosts, it is easiest to work in a Git Bash prompt as this will
 make Perl available automatically. You should not need to mess with
-VCVARSALL/etc since cc-rs will find MSVC. When targeting
-aarch64-pc-windows-msvc, just add the Visual-Studio-supplied LLVM directory
-to the PATH by transliterating the relevant PowerShell commands into Bash
-syntax; this copy of clang will be able to find all the headers, etc.
+VCVARSALL/etc since cc-rs will find MSVC and Clang.
 
 
 Additional Features that are Useful for Development


Alovchin91:
FYI, the original idea was to use _Clang-cl_ with the MSVC toolchain for 2 main reasons:

1. Back then, the Nasm assembler wasn't available for Windows Arm64, but Clang did provide asm support, so we used it.

2. MSVC didn't have `int128_t` support (apparently, it's still the case 🤷‍♂️), but Clang had.

I'm not sure what specifically has changed in Ring 0.17 (a lot, from what I know), so maybe today it is easier to build it with MSVC. But the original idea was to use _Clang-cl_.

crypto/limbs/limbs.inl
```diff
@@ -60,6 +60,13 @@ static inline Carry limb_adc(Limb *r, Limb a, Limb b, Carry carry_in) {
   Carry ret;
 #if defined(RING_CORE_ADDCARRY_INTRINSIC)
   ret = RING_CORE_ADDCARRY_INTRINSIC(carry_in, a, b, r);
+#elif defined(OPENSSL_AARCH64) && defined(_MSC_VER) && !defined(__clang__)
+  /* Manual 64-bit addition for MSVC ARM64 (no __uint128_t) */
+  Limb sum = a + carry_in;
+  Carry carry1 = (sum < a) ? 1 : 0;  /* carry from a + carry_in */
+  *r = sum + b;
+  Carry carry2 = (*r < sum) ? 1 : 0;  /* carry from sum + b */
+  ret = carry1 | carry2;
```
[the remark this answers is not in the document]
briansmith: > used AI 

I cannot accept PRs of unknown provenance.

build.rs
```diff
@@ -669,6 +670,45 @@ fn nasm(file: &Path, arch: &str, include_dir: &Path, out_dir: &Path, c_root_dir:
     run_command(c);
 }
 
+fn clang_win(file: &Path, arch: &str, include_dir: &Path, out_dir: &Path) {
```
briansmith, line 673: I think instead of this, it would be better to create two separate libraries: one for assembly code, and one for C code. Then configure the `cc::Build` for the assembly code library to force the use of clang as the C compiler for Windows AArch64. This way, we maximize the use of cc-rs's infrastructure, e.g. handling compiler options and whatnot.

crypto/fipsmodule/bn/internal.h
```diff
@@ -18,11 +18,15 @@
...
 
-#if defined(OPENSSL_X86_64) && defined(_MSC_VER) && !defined(__clang__)
+#if defined(_MSC_VER) && defined(OPENSSL_64_BIT) && !defined(__clang__)
 #pragma warning(push, 3)
 #include <intrin.h>
 #pragma warning(pop)
+#if defined(OPENSSL_X86_64)
 #pragma intrinsic(_umul128)
+#elif defined(OPENSSL_AARCH64)
+#pragma intrinsic(__umulh)
```
briansmith, line 28: I'm not sure why BoringSSL avoids doing this. My guess is they aren't testing AArch64 Windows builds using MSVC, but instead forcing the use of clang.

MarijnS95:
> FYI, the original idea was to use _Clang-cl_ with the MSVC toolchain for 2 main reasons:

@Alovchin91 just noting that it doesn't currently use `clang-cl.exe` but plain `clang.exe`, overwritten in such a way in `cc-rs` that it may be receiving `CFLAGS` that the user initially assumed / set up for `cl.exe`: #2215. That was proposed to be changed to `clang-cl.exe` in #2216, but it's all a bit hacky in case the user really has really intentionally set `CC=clang`.

Alovchin91:
@MarijnS95 Right. I can hardly remember much of our discussion (you can find it [here](https://github.com/briansmith/ring/pull/1339) and maybe something also in #1167), but I _guess_ _maybe_ back in the v0.16 days the C code was compiled separately, so it didn't really matter whether we used `clang` or `clang-cl`. It seems now it matters 😅

I'm not sure if it's a good idea to move forward with MSVC since it apparently still lacks 128-bit support. But as a drop-in replacement, `clang-cl` should be used of course.

Is it really a problem though that the user could explicitly set `CC` to something random? 🤔  They could do it with MSVC, too.

MarijnS95:
@Alovchin91 reading through those links you seem to mostly mention `clang-cl` yet the end result uses `clang`.

> the C code was compiled separately, so it didn't really matter whether we used `clang` or `clang-cl`

I don't understand exactly what you mean here with "separately". Since the compiler is replaced inside `cc-rs`, that crate has still taken into account the users' `CFLAGS` from the environment which the user might have purposefully set up for `cl.exe`. `clang-cl.exe` understands those while `clang.exe` does not.

That is also the reason to check what `CC` was already set to inside `cc-rs`. If the user purposefully chose `clang.exe` and set their `CFLAGS` to be compatible with that (and `ring` is currently compiling fine for them on Windows ARM64!), an unconditional switch to `clang-cl.exe` would break as their arguments are not compatible with the `cl.exe`-like interface.

PR #2688, opened 2025-08-10
Add qnx support

AkhilTThomas:
This PR adds support for the QNX target, As mentioned in #2429 since there is no CI support the test was run on both x86-64 and aarch64 qemu and the reports are attached.
[test_output-aarch64.txt](https://github.com/user-attachments/files/21707438/test_output-aarch64.txt)
[test_output-x86-64.txt](https://github.com/user-attachments/files/21707439/test_output-x86-64.txt)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -207,7 +207,7 @@ fn cpp_flags(compiler: &cc::Tool) -> &'static [&'static str] {
 const ASM_TARGETS: &[AsmTarget] = &[
     AsmTarget {
         oss: &[
-            ANDROID, FREEBSD, FUCHSIA, ILLUMOS, LINUX, NETBSD, OPENBSD, REDOX,
+            ANDROID, FREEBSD, FUCHSIA, ILLUMOS, LINUX, NETBSD, NTO, OPENBSD, REDOX,
         ],
         arch: AARCH64,
         perlasm_format: "linux64",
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -207,7 +207,7 @@ fn cpp_flags(compiler: &cc::Tool) -> &'static [&'static str] {
 const ASM_TARGETS: &[AsmTarget] = &[
     AsmTarget {
         oss: &[
-            ANDROID, FREEBSD, FUCHSIA, ILLUMOS, LINUX, NETBSD, OPENBSD, REDOX,
+            ANDROID, FREEBSD, FUCHSIA, ILLUMOS, LINUX, NETBSD, OPENBSD, REDOX, NTO,
```
briansmith, line 210: Please put these in sorted order.

build.rs
```diff
@@ -225,7 +225,7 @@ const ASM_TARGETS: &[AsmTarget] = &[
     AsmTarget {
         oss: &[
             ANDROID, DRAGONFLY, FREEBSD, FUCHSIA, HAIKU, HURD, ILLUMOS, LINUX, NETBSD, OPENBSD,
-            REDOX, SOLARIS,
+            REDOX, SOLARIS, NTO,
```
briansmith, line 228: Ditto.

build.rs
```diff
@@ -287,6 +287,7 @@ const NETBSD: &str = "netbsd";
 const OPENBSD: &str = "openbsd";
 const REDOX: &str = "redox";
 const SOLARIS: &str = "solaris";
+const NTO: &str = "nto";
```
briansmith, line 290: Ditto.

briansmith (changes requested):
Thanks. This seems reasonable to me. Please make the requested tweaks.

build.rs
AkhilTThomas: Done

briansmith:
Thanks for doing this, and thanks to @szclsya for the initial PR.

Issue #2341, opened 2025-02-08
update getrandom


oligamiq:
please update getrandom.

The version of getrandom has been updated. Until now, if you want to compile for wasm,
```toml
getrandom = { version = "0.2", features = ["std", "js"] }
```

```toml
getrandom = { version = "0.3.1", features = ["std", "wasm_js"] }
```

I had written code like this, but it causes problems because the version of getrandom on which this library depends is out of date.

Related issues, etc.
https://github.com/briansmith/ring/pull/2218
https://github.com/briansmith/ring/pull/2130
https://github.com/briansmith/ring/issues/2187

https://rust-random.github.io/book/update-0.9.html
https://github.com/rust-random/rand/blob/master/CHANGELOG.md

briansmith:
It will be a while before I can review the changes between the last time I looked at getrandom and the getrandom 0.3 release. Don't hold your breath.

oligamiq:
https://github.com/rust-random/getrandom/blob/master/CHANGELOG.md

The changes that may be relevant are
・Rename getrandom and getrandom_uninit functions to fill and fill_uninit respectively 
・js features to wasm_js

oligamiq:
https://github.com/briansmith/ring/pull/2342

briansmith:
Thanks. AFAICT getrandom 0.3 changes the way configuration is done in a way that is difficult for *ring* to adjust to, or at least I am not sure how to adjust *ring* to use the new configuration mechanism.

Plus, I review every commit in getrandom before I upgrade to a new version. At one point last year I was reviewing getrandom commits in real time, but I haven't reviewed them for a while. So I have about a year of getrandom commits to review 1-by-1 before I can do the update.

oligamiq:
Is it for security?
Good luck!

briansmith:
My current perspective is that it is annoying to others to be depending on getrandom 0.2 when getrandom 0.3 is available, so it is worth somebody trying to resolve things. I don't want to break backward compatibility with 0.17 in the next few months, so such a change would need to be backward compatible. I am open to hearing creative solutions.

oligamiq:
How about switching using features?

In Cargo.toml, you can write it like this:

```toml
_getrandom_alias_3 = { version = "0.3.1", package = "getrandom", features = [
    "std",
    "wasm_js",
] }
_getrandom_alias_2 = { version = "0.2", package = "getrandom", features = [
    "std",
    "js",
] }
```
[...]

PR #2700, opened 2025-09-08
Update getrandom

hanyuone:
This PR implements the "first PR" outlined [here](https://github.com/briansmith/ring/issues/2341#issuecomment-2940827952) in #2341:

> First PR: Vendor the getrandom wasm_js.rs backend along with wrappers in lib.rs into rand/getrandom/{lib.rs,backends.rs,wasm_js.rs}. Compile and use the vendored implementation only if the target is wasm32 and wasm32_unknown_unknown_js is enabled. Use getrandom 0.3 otherwise.

I needed to add `error.rs` and `utils.rs` as well, since the `Error` struct and some utility functions were required by the other files.

Currently, running `ring` on `wasm32-unknown-unknown` requires the `getrandom_backend` config variable to be set in order for `getrandom` to compile properly, but making the ergonomics smoother is outside the scope of PR.

### DIFF
--- a/src/rand/getrandom/error.rs
+++ b/src/rand/getrandom/error.rs
@@ -1,6 +1,3 @@
-#[cfg(feature = "std")]
-extern crate std;
-
 use cfg_if::cfg_if;
 use core::fmt;
 
--- a/src/rand/getrandom/error.rs
+++ b/src/rand/getrandom/error.rs
@@ -1,0 +1,213 @@
+#[cfg(feature = "std")]
+extern crate std;
+
+use cfg_if::cfg_if;
+use core::fmt;
+
+// This private alias mirrors `std::io::RawOsError`:
+// https://doc.rust-lang.org/std/io/type.RawOsError.html)
+cfg_if::cfg_if!(
+    if #[cfg(target_os = "uefi")] {
+        // See the UEFI spec for more information:
+        // https://uefi.org/specs/UEFI/2.10/Apx_D_Status_Codes.html
+        type RawOsError = usize;
+        type NonZeroRawOsError = core::num::NonZeroUsize;
+        const UEFI_ERROR_FLAG: RawOsError = 1 << (RawOsError::BITS - 1);
+    } else {
+        type RawOsError = i32;
+        type NonZeroRawOsError = core::num::NonZeroI32;
+    }
+);
+
+/// A small and `no_std` compatible error type
+///
+/// The [`Error::raw_os_error()`] will indicate if the error is from the OS, and
+/// if so, which error code the OS gave the application. If such an error is
+/// encountered, please consult with your system documentation.
+///
+/// *If this crate's `"std"` Cargo feature is enabled*, then:
+/// - [`getrandom::Error`][Error] implements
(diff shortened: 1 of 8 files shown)


src/rand/getrandom/error.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    /// Internal errors can be in the range of 2^16..2^17
+    const INTERNAL_START: RawOsError = 1 << 16;
+    /// Custom errors can be in the range of 2^17..(2^17 + 2^16)
+    const CUSTOM_START: RawOsError = 1 << 17;
+
+    /// Creates a new instance of an `Error` from a negative error code.
+    #[cfg(not(target_os = "uefi"))]
+    #[allow(dead_code)]
+    pub(super) fn from_neg_error_code(code: RawOsError) -> Self {
```
briansmith, line 52: It would be great to get rid of the dead code in this submodule that is for supporting non-warm targets.

src/rand/getrandom/error.rs
```diff
@@ -0,0 +1,210 @@
...
+            } else if cfg!(not(target_os = "solid_asp3")) {
+                code.checked_neg()
+            } else {
+                Some(code)
+            }
+        }
+    }
+
+    /// Creates a new instance of an `Error` from a particular custom error code.
+    pub const fn new_custom(n: u16) -> Error {
```
briansmith, line 124: AFAICT, this is the only part that the wasm_js.rs uses.

src/rand/getrandom/error.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    /// Creates a new instance of an `Error` from a particular internal error code.
+    pub(crate) const fn new_internal(n: u16) -> Error {
+        // SAFETY: code > 0 as INTERNAL_START > 0 and adding `n` won't overflow `RawOsError`.
+        let code = Error::INTERNAL_START + (n as RawOsError);
+        Error(unsafe { NonZeroRawOsError::new_unchecked(code) })
+    }
+
+    fn internal_desc(&self) -> Option<&'static str> {
```
briansmith, line 137: We can also get rid of this and all the `Debug` and `Display` support as we throw all that away anyway.

briansmith (commented):
Thanks for taking this on. Please post a new version with these changes and squash them all into one commit with a good commit message. Then I will review the updated version by diffing it with upstream. Also, please use an upstream commit that corresponds to a crates.io release of `getrandom`.

hanyuone:
Sorry for getting back to this late - all the points should be addressed, and I'm using the upstream commit corresponding to v0.3.3 on crates.io. Once you've confirmed everything is good, I'll squash everything down to one commit.

With `ring`, are we certain that we'll only ever be using `getrandom_backend="wasm_js"`? If so, then I can get rid of the `getrandom_msan` optional code in `lib` as well.

With v0.3.4, the `getrandom_backend` config flag is now default - would that change our approach to updating `getrandom`?

hanyuone:
@briansmith With #2756, should this PR be closed?

PR #2824, opened 2026-04-27
Remove windows-sys

Jake-Shadle:
This is an alternative to #2570 that still removes the windows-sys dependency but inlines the bindings (1 function, 1 constant) instead of relying on std.

### DIFF
--- a/src/cpu/aarch64/windows.rs
+++ b/src/cpu/aarch64/windows.rs
@@ -13,9 +13,7 @@
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 use super::{Aes, CAPS_STATIC, Neon, PMull, Sha256};
-use windows_sys::Win32::System::Threading::{
-    IsProcessorFeaturePresent, PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE,
-};
+use crate::polyfill::windows_sys;
 
 pub const FORCE_DYNAMIC_DETECTION: u32 = 0;
 
@@ -25,7 +23,9 @@ pub fn detect_features() -> u32 {
 
     let mut features = 0;
 
-    let result = unsafe { IsProcessorFeaturePresent(PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE) };
+    let result = unsafe {
+        windows_sys::IsProcessorFeaturePresent(windows_sys::PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE)
+    };
 
     if result != 0 {
         // These are all covered by one call in Windows
--- a/src/polyfill/mod.rs
+++ b/src/polyfill/mod.rs
@@ -91,6 +91,9 @@ mod test;
 pub(crate) mod uninit;
 mod uninit_slice;
 
+#[cfg(all(windows, target_arch = "aarch64"))]
+pub(crate) mod windows_sys;
+
 pub use self::{array_flat_map::ArrayFlatMap, array_split_map::ArraySplitMap, notsend::NotSend};
 
 #[allow(unused_imports)]
--- a/src/polyfill/windows_sys.rs
+++ b/src/polyfill/windows_sys.rs
@@ -1,0 +1,11 @@
+pub type DWORD = u32;
+pub type BOOL = i32;
+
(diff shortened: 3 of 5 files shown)


src/cpu/aarch64/windows.rs
```diff
@@ -13,11 +13,16 @@
 // CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 use super::{Aes, CAPS_STATIC, Neon, PMull, Sha256};
-use windows_sys::Win32::System::Threading::{
-    IsProcessorFeaturePresent, PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE,
-};
 
 pub const FORCE_DYNAMIC_DETECTION: u32 = 0;
+/// This Arm processor implements the Arm v8 extra cryptographic instructions (for example, AES, SHA1 and SHA2).
+const PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE: u32 = 30;
```
briansmith: Please move all of these to `polyfill/windows_sys.rs`. Not sure what the license is supposed to be, but it should be documented in that file.

Please define and use the `DWORD` and `BOOL` type aliases.

src/cpu/aarch64/windows.rs
```diff
@@ -13,11 +13,16 @@
...
+
+#[link(name = "kernel32")]
+extern "system" {
```
briansmith: What's the advantage/disadvantage of using windows-link vs. doing this manually? I'm not sure why windows-link exists.

Jake-Shadle: There are some slight differences between x86 and other arches, in some cases, since this is only aarch64 I didn't see the point.

src/cpu/aarch64/windows.rs
Jake-Shadle: I've moved them to that file as you asked.

konstin:
This change would be great to have, as we're seeing a lot of duplication with the windows crates.

src/polyfill/mod.rs
```diff
@@ -93,6 +93,9 @@ mod test;
 mod uninit_slice;
 mod uninit_slice_cursor;
 
+#[cfg(target_arch = "aarch64")]
```
briansmith: This needs `all(target_os = "windows", ...)' otherwise all aarch64 builds for other OS's will fail.

src/polyfill/windows_sys.rs
```diff
@@ -0,0 +1,25 @@
...
+// purpose with or without fee is hereby granted, provided that the above
+// copyright notice and this permission notice appear in all copies.
+//
+// THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+// WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
+// SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+// OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
```
briansmith: The license  and copyright should be the license/copyright of whatever you copied/reduced the bindings from.

briansmith (changes requested):
Thanks.

Besides addressing the other two comments, please squash these into one commit.

The commit message should be something like "Vendor `windows_sys::IsProcessorFeaturePresent` from `windows_sys <whatever the version you copied it from>`."

Jake-Shadle:
I just manually wrote this binding from the data in the linked API doc, so I'm not sure what license you want me to put in the header, the windows API doesn't really have a license AFAIK, just a terms of use <https://learn.microsoft.com/en-us/legal/microsoft-apis/terms-of-use>.

briansmith:
@Jake-Shadle PTAL at #2865, which uses windows-bindgen to automate the generation. Closing this in favor of that. Thanks for your contribution.

PR #2857, opened 2026-06-08
aead: Remove unnecessary clone

FranciscoTGouveia:
If I am not overlooking anything, I believe this clone is unnecessary and can thus be deleted.

CC @nunoplopes


briansmith:
Thanks! CI failures are unrelated.

PR #2871, opened 2026-07-17
Change error types to use `core::error::Error`

jschillem:
With [`core::error::Error` being stabilized in Rust 1.81](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0/), the error types no longer need to depend on `std` for this trait.

Removing this also renders the `std` unnecessary for any library consumers as the only usage of the `std` is now for test logging. Though it may make sense to leave the `std` flag to avoid breakages.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,9 +37,6 @@
 //!         on esp-idf despite the likelihood that RNG is not secure.
 //!         This feature only works with <code>os = espidf</code> targets.
 //!         See <a href="https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/random.html">
-//! <tr><td><code>std</code>
-//!     <td>Enable features that use libstd, in particular
-//!         <code>std::error::Error</code> integration. Implies `alloc`.
 //! <tr><td><code>wasm32_unknown_unknown_js</code>
 //!     <td>When this feature is enabled, for the wasm32-unknown-unknown target,
 //!         Web APIs will be used to implement features like `ring::rand` that
--- a/tests/error_tests.rs
+++ b/tests/error_tests.rs
@@ -1,5 +1,9 @@
 #![allow(missing_docs)]
 
+/// `compile_time_assert_core_error_error::<T>();` fails to compile if `T`
+/// doesn't implement `core::error::Error`.
+const fn compile_time_assert_core_error_error<T: core::error::Error>() {}
+
 #[cfg(feature = "std")]
 #[test]
 fn error_impl_std_error_error_test() {
@@ -9,4 +13,12 @@ fn error_impl_std_error_error_test() {
 
     test::compile_time_assert_std_error_error::<error::Unspecified>();
     test::compile_time_assert_std_error_error::<error::KeyRejected>();
+}
+
+#[test]
+fn error_impl_core_error_error_test() {
+    use ring::error;
+
+    compile_time_assert_core_error_error::<error::Unspecified>();
+    compile_time_assert_core_error_error::<error::KeyRejected>();
 }
--- a/src/error/key_rejected.rs
+++ b/src/error/key_rejected.rs
@@ -14,9 +14,6 @@
 
 //! Error reporting.
 
-#[cfg(feature = "std")]
-extern crate std;
-
 /// An error parsing or validating a key.
 ///
 /// The `Display` implementation will return a string that will help you better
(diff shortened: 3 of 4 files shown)


src/lib.rs
```diff
@@ -37,9 +37,6 @@
 //!         on esp-idf despite the likelihood that RNG is not secure.
 //!         This feature only works with <code>os = espidf</code> targets.
 //!         See <a href="https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/random.html">
-//! <tr><td><code>std</code>
```
briansmith, line 40: The `std` feature must be preserved for backward compatibility, and it must continue to imply `alloc`.

src/testutil.rs
```diff
@@ -148,10 +148,9 @@ pub const fn compile_time_assert_send<T: Send>() {}
 /// implement `Sync`.
 pub const fn compile_time_assert_sync<T: Sync>() {}
 
-/// `compile_time_assert_std_error_error::<T>();` fails to compile if `T`
-/// doesn't implement `std::error::Error`.
-#[cfg(feature = "std")]
-pub const fn compile_time_assert_std_error_error<T: std::error::Error>() {}
```
briansmith, line 154: We should keep this, and also add `compile_time_assert_core_error_error`

tests/error_tests.rs
```diff
@@ -1,12 +1,11 @@
 #![allow(missing_docs)]
 
-#[cfg(feature = "std")]
 #[test]
-fn error_impl_std_error_error_test() {
+fn error_impl_core_error_error_test() {
```
briansmith: we should keep the test that verifies `std::error::Error`, and also add the `core::error::Error` variant.

Cargo.toml
```diff
@@ -192,7 +192,6 @@ dev_urandom_fallback = []
 less-safe-getrandom-custom-or-rdrand = []
 less-safe-getrandom-espidf = []
 slow_tests = []
-std = ["alloc"]
```
briansmith, line 195: We need to keep this for backward compatibility.

src/deprecated_test.rs
```diff
@@ -31,8 +31,7 @@ pub use crate::testutil::{
     compile_time_assert_send, compile_time_assert_sync, from_hex, run,
 };
 
-#[cfg(feature = "std")]
-pub use crate::testutil::compile_time_assert_std_error_error;
+pub use crate::testutil::compile_time_assert_core_error_error;
```
briansmith: We have to keep exporting `compile_time_assert_std_error_error` because it is a public API.

Conversely, we shouldn't expose `compile_time_assert_core_error_error` from here, as this module is for backward compatibility.

briansmith (changes requested):
Thanks!

src/lib.rs
jschillem: Do you think there is still reason to  "advertise" the `std` feature in the crate docs however? The `std` error is a re-export of the `core` error so  the feature is really only kept for backwards compatibility and implying `alloc`.

At the very least I think saying that it is necessary for the trait impl is not needed.

src/deprecated_test.rs
jschillem: Added as a separate function that is not re-exported.

Cargo.toml
jschillem: Reverted.

tests/error_tests.rs
jschillem: Added specific core tests within `src/error/mod.rs` (due to `testutil` not being public)

src/testutil.rs
jschillem: Added as separate function.

src/error/mod.rs
```diff
@@ -56,3 +56,18 @@ cold_exhaustive_error! {
...
+    pub const fn compile_time_assert_core_error_error<T: core::error::Error>() {}
+
+    #[test]
+    fn error_impl_core_error_error_test() {
+        use crate::error;
+
+        compile_time_assert_core_error_error::<error::Unspecified>();
+        compile_time_assert_core_error_error::<error::KeyRejected>();
+    }
+}
```
briansmith: Thanks for adding these. These should go in error_tests.rs since they are testing public API.

src/lib.rs
briansmith: Makes sense.

briansmith (changes requested):

briansmith:
In addition to the change mentioned in the review, please squash the commits into a single commit, and update the commit message to remove  "and removing the `std` feature".

src/error/mod.rs
jschillem: Moved them to `error_tests.rs`, the utility to actually test that they impl cannot be in `testutil.rs` like the `std::error::Error` one though without being re-exported since `testutil` is a private module.

briansmith (approved):
Thank you!

Issue #2527, opened 2025-05-15
macOS/darwin compilation fails due to `MIN_STATIC_FEATURES` assertions


zachschuermann:
Our [github actions macOS job] is failing due to failing compile time assertions in `ring`:
```
error[E0080]: evaluation of constant value failed
  --> /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ring-0.17.14/src/cpu/arm/darwin.rs:44:5
   |
44 |     assert!((CAPS_STATIC & MIN_STATIC_FEATURES) == MIN_STATIC_FEATURES);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation panicked: assertion failed: (CAPS_STATIC & MIN_STATIC_FEATURES) == MIN_STATIC_FEATURES

error[E0080]: evaluation of constant value failed
  --> /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ring-0.17.14/src/cpu/arm/darwin.rs:51:5
   |
51 |     assert!(CAPS_STATIC == MIN_STATIC_FEATURES);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation panicked: assertion failed: CAPS_STATIC == MIN_STATIC_FEATURES

For more information about this error, try `rustc --explain E0080`.
error: could not compile `ring` (lib) due to 2 previous errors
```

Reference to those [assertions] just to make investigation easier:
```rust
// MSRV: Enforce 1.61.0 onaarch64-apple-*, in particular) prior to. Earlier
// versions of Rust before did not report the AAarch64 CPU features correctly
// for these targets. Cargo.toml specifies `rust-version` but versions before
// Rust 1.56 don't know about it.
#[allow(clippy::assertions_on_constants)]
const _AARCH64_APPLE_TARGETS_EXPECTED_FEATURES: () =
    assert!((CAPS_STATIC & MIN_STATIC_FEATURES) == MIN_STATIC_FEATURES);

// Ensure we don't accidentally allow features statically beyond
// `MIN_STATIC_FEATURES` so that dynamic detection is done uniformly for
// all of these targets.
#[allow(clippy::assertions_on_constants)]
const _AARCH64_APPLE_DARWIN_TARGETS_EXPECTED_FEATURES: () =
[...]

PR #2872, opened 2026-08-01
_AARCH64_APPLE_DARWIN_TARGETS_EXPECTED_FEATURES failure

mahogny:
This fix is for the error below

I believe the error has been reported before as
https://github.com/briansmith/ring/issues/2527

...and I confirm that this happens only when doing native cpu builds. The checks in the current code don't line up

The fix is a "dynamic" test of features. It's a smaller subset than this patch

Included is a related CI-test you can choose to ignore

## Error

error[E0080]: evaluation panicked: assertion failed: (CAPS_STATIC & MIN_STATIC_FEATURES) == MIN_STATIC_FEATURES
  --> /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ring-0.17.14/src/cpu/arm/darwin.rs:44:5
   |
44 |     assert!((CAPS_STATIC & MIN_STATIC_FEATURES) == MIN_STATIC_FEATURES);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation of `cpu::arm::darwin::_AARCH64_APPLE_TARGETS_EXPECTED_FEATURES` failed here

error[E0080]: evaluation panicked: assertion failed: CAPS_STATIC == MIN_STATIC_FEATURES
  --> /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ring-0.17.14/src/cpu/arm/darwin.rs:51:5
   |
51 |     assert!(CAPS_STATIC == MIN_STATIC_FEATURES);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation of `cpu::arm::darwin::_AARCH64_APPLE_DARWIN_TARGETS_EXPECTED_FEATURES` failed here

For more information about this error, try `rustc --explain E0080`.
error: could not compile `ring` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.

### DIFF
--- a/src/cpu/aarch64/darwin.rs
+++ b/src/cpu/aarch64/darwin.rs
@@ -17,16 +17,39 @@ use super::{Aes, CAPS_STATIC, Neon, PMull, Sha256};
 use {super::Sha512, core::ffi::CStr};
 
 // ```
-// $ rustc +1.61.0 --print cfg --target=aarch64-apple-ios | grep -E "neon|aes|sha|pmull"
+// $ rustc --print cfg --target=aarch64-apple-ios | grep -E "neon|aes|sha|pmull"
 // target_feature="aes"
 // target_feature="neon"
 // target_feature="sha2"
-// $ rustc +1.61.0 --print cfg --target=aarch64-apple-darwin | grep -E "neon|aes|sha|pmull"
+// $ rustc --print cfg --target=aarch64-apple-darwin | grep -E "neon|aes|sha|pmull"
 // target_feature="aes"
 // target_feature="neon"
 // target_feature="sha2"
 // target_feature="sha3"
 // ```
+//
+// Every aarch64-apple-* device has AES, PMULL and SHA-256, and by default the
+// compiler tells us so. However, we cannot *rely* on being told, because
+// `-Ctarget-cpu` can leave the target feature set *smaller* than the target's
+// default:
+//
+// ```
+// $ rustc --print cfg --target=aarch64-apple-darwin -Ctarget-cpu=generic | grep -E "neon|aes|sha"
+// target_feature="neon"
+// ```
+//
+// On Apple Silicon, `-Ctarget-cpu=native` makes LLVM pick an older CPU than
+// the default one (`cyclone`, or `generic` on hosts whose CPU LLVM does not
+// recognize, e.g. virtualized CI runners); see
+// https://github.com/rust-lang/rust/issues/93889. Users also disable target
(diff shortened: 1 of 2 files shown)


src/cpu/aarch64/darwin.rs
```diff
@@ -17,17 +17,40 @@ use super::{Aes, CAPS_STATIC, Neon, PMull, Sha256};
...
 // target_feature="sha3"
 // ```
 //
+// Every aarch64-apple-* device has AES, PMULL and SHA-256, and by default the
+// compiler tells us so. However, we cannot *rely* on being told, because
+// `-Ctarget-cpu` can leave the target feature set *smaller* than the target's
+// default:
+//
+// ```
+// $ rustc --print cfg --target=aarch64-apple-darwin -Ctarget-cpu=generic | grep -E "neon|aes|sha"
```
briansmith, line 37: We don't intend to support `-Ctarget-cpu=generic` on apple hardware.

src/cpu/aarch64/darwin.rs
```diff
@@ -17,17 +17,40 @@ use super::{Aes, CAPS_STATIC, Neon, PMull, Sha256};
...
+// target_feature="neon"
+// ```
+//
+// On Apple Silicon, `-Ctarget-cpu=native` makes LLVM pick an older CPU than
+// the default one (`cyclone`, or `generic` on hosts whose CPU LLVM does not
+// recognize, e.g. virtualized CI runners); see
+// https://github.com/rust-lang/rust/issues/93889. Users also disable target
+// features explicitly. Formerly such a configuration was a hard compile error
```
briansmith, line 45: If users want to do `-Ctarget-cpu=native` or `-Ctarget-cpu=generic` or similar, then they should also enable the target features we expect ("aes", etc.). I think we should document this in BUILDING.md and avoid making these code changes.

mahogny:
In my mind, target-cpu and feature sets are two completely different beast. Target CPU is a performance choice (esp when we perform benchmarking to see what instructions are crucial, or what sections need hand tuning); Features is a choice of how much code bloat we are ok with pulling in (more important in wasm context, but there's also compile time aspects)

I think it will create a lot of confusion if these two aspects are confounded

Issue #1999, opened 2024-03-18
assert!(cfg!(target_feature = "sse") && cfg!(target_feature = "sse2")) panic at build on i386, i586, FreeBSD, or many target_os=none x86 targets.


girgen:
Hi!

I get this error when building an app using ring-0.17.8

```
error[E0080]: evaluation of constant value failed
 --> /wrkdirs/usr/ports/www/sqlpage/work/SQLpage-0.20.0/cargo-crates/ring-0.17.8/src/cpu/[intel.rs:28](http://intel.rs:28/):9
  |
28 |         assert!(cfg!(target_feature = "sse") && cfg!(target_feature = "sse2"));
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'assertion failed: cfg!(target_feature = "sse") && cfg!(target_feature = "sse2")', /wrkdirs/usr/ports/www/sqlpage/work/SQLpage-0.20.0/cargo-crates/ring-0.17.8/src/cpu/[intel.rs:28](http://intel.rs:28/):9
  |
  = note: this error originates in the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0080`.
error: could not compile `ring` (lib) due to 1 previous error
```

I tried to find info using the three issue links in the code, but I don't have enough experience with rust, to be honest, I'm just packaging an application. It seems to me that the problem is with the assert call itself, so perhaps this is actually a bug?

Removing the check will make the code build.

Do you need more information, just specify what is needed. Build log attached.

[sqlpage-0.20.0.log](https://github.com/briansmith/ring/files/14635026/sqlpage-0.20.0.log)

briansmith:
What are the steps to reproduce? `cargo build --target=???`?

girgen:
See the attached log above ☝️ 

Looks like
```
--target i686-unknown-freebsd
```

briansmith:
What do you get when you do this?
```
$ rustc --print=cfg --target i686-unknown-freebsd | grep sse
debug_assertions
target_feature="sse"
target_feature="sse2"
```
For me, rustc indicates that sse and sse2 are there. 

I see in your build log that there are warnings that `CARGO_CFG_TARGET_FEATURE` not set. THat might be a clue.

[...]

PR #2878, opened 2026-09-17
build: Fixes for non-SSE x86

willnode:
This fixes compilation for non-SSE x86 after commenting out assertions.

This is the cherrypick of what I send to fix redox i586 build after rebase: https://gitlab.redox-os.org/redox-os/ring/-/merge_requests/5

The change in `x86.rs` is because `leaf1_edx` is already exist in the function

The change in `prefixed.rs` is because in `src/arithmetic/limbs/fallback/mont.rs` the macro is used for non-SSE2 x86

```
 if #[cfg(not(any(
            all(target_arch = "aarch64", target_endian = "little"),
            all(target_arch = "arm", target_endian = "little"),
            all(target_arch = "x86", target_feature = "sse2"),
            target_arch = "x86_64")))] {
        // this is called if all(target_arch = "x86", not(target_feature = "sse2"))
        prefixed_export! {
            #[cfg_attr(target_arch = "x86", cold)]
            #[cfg_attr(target_arch = "x86", inline(never))]
            unsafe extern "C" fn bn_mul_mont_fallback(
```


willnode:
Related issue #1999 i opened this also because more change was needed to fix compilation for targets without SSE

